PR #1, opened 2018-11-12
Bring over files from old repo

kalilsn:
Just copying files directly from the old repo separately from other changes to come so they're not buried.


kalilsn:
Other than the Jenkinsfile (which was reviewed), this should just be copying files directly from vets-contrib.

PR #11, opened 2018-12-03
Update argonaut docs URL

edpaget:
Closes department-of-veterans-affairs/vets-contrib#705

Depends on https://github.com/department-of-veterans-affairs/devops/pull/3710


edpaget:
Thanks for the catch got that updated.

PR #20, opened 2018-12-14
Add visual regression testing

kalilsn:
Resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/716

Adds visual regression testing to our CI: on every PR, Jenkins will generate screenshots of the site on a few pages, and then diff those against the ones stored in the repo. If there are changes, it will fail. If they're expected, the developer can then update the stored screenshots.

These screenshots are just full page grabs without any other interaction, but if we have particular components that need to be tested, it might be worth adding some interaction (say, clicking on the mobile menu, or expanding some documentation). These tests are relatively slow, so I don't think we want to go too deep into that, but we aren't actually running any tests at the moment (😬) other than `npm audit` for security.

These tests have to be run in docker, but the command is kind of a mouthful. I'm open to adding a Makefile for shorthand if that seems like a good idea.


README.md
```diff
@@ -28,10 +28,30 @@ This will start the app on port 3001 as well as the webpack dev server to automa
...
+```
+If the tests don't pass, an image showing the diff will be generated in `src/__image_snapshots__/__diff_output__` 
+
+If you need to update your snapshots (after seeing a failing diff), run with the `-u` flag:
+```
+docker run --rm -itv "$PWD:/application" -v "/application/node_modules" developer-portal npm run test:visual -- -u
+```
+
+It's important to run these tests inside the docker container, or else the tests will report false negatives due to font differences.
+
```
dvogel: Thank you for the clear instructions!

src/visualRegressionTest.ts
```diff
@@ -0,0 +1,60 @@
+import { Page } from 'puppeteer';
+
+import { mockSwagger } from './e2eHelpers';
+
+jest.setTimeout(100000);
```
dvogel: I'm guessing this is a global watchdog timeout to prevent the job from hanging indefinitely. I assume this is milliseconds and and thus 1m40s? Maybe it's worth making this a calculated value based on the length of `paths` so  we can't forget to increase this as we add pages we'd like to snapshot?

kalilsn:
test

test

src/visualRegressionTest.ts
kalilsn: It's actually per test, so I don't think it should be calculated per path. AFAIK it's just how long jest will wait for a promise to resolve before quitting, and their defaults cause a lot of issues when doing something as slow as chromium integration tests or waiting for webpack to compile.

This number was just taken from the existing e2e test setup, so I could do some experimentation to figure out a better value, but I haven't had any scenarios where it actually waits this long.

kalilsn:
Previously the swagger screenshots were actually broken because I messed up the test for mocking those requests.

PR #24, opened 2018-12-21
Add LGY docs to benefits section. (#780)

dvogel:
This adds the LGY docs. Those docs aren't ready for primetime yet, so I won't merge this yet. Wanted to put it out for review to get the ball rolling once the upstream docs are ready though.


src/containers/Explore.tsx
```diff
@@ -56,6 +56,11 @@ class Explore extends React.Component<IExploreProps, { }> {
                     <SwaggerDocs url={`${process.env.REACT_APP_VETSGOV_SWAGGER_API}/services/appeals/docs/v0/api`} />
                 );
                 break;
+            case 'loan_guarantees':
+                docs = (
+                    <SwaggerDocs url={`${process.env.REACT_APP_VETSGOV_SWAGGER_API}/services/loan_guarantees/docs/v0/api`} />
```
kalilsn: Right now our readme suggests that locally your `REACT_APP_VETSGOV_SWAGGER_API` should be `localhost:3000` where you have vets-api running. Since this service doesn't go through vets-api, that won't ever work. Solving that problem might be beyond the scope of this PR, but it seems worth noting.

Looks like for argonaut we've solved that problem by just pointing directly at staging-api.va.gov, which doesn't seem ideal either.

edpaget: Due to how argonaut is deployed their docs are only served from the staging kong environment (though things may have changed while I'm out!). I think we should suggest developing with `REACT_APP_VETSGOV_SWAGGER_API` set to `dev-api.va.gov` since we've change the CORS policy to allow all origins, and note that if people need to test API docs they're writting locally to change it to `localhost:3000`.

dvogel: I just pushed an a change to the README that makes it clear either approach is appropriate, depending on what your goal is.

src/App.tsx
```diff
@@ -16,21 +17,43 @@ history.listen(location => {
...
+  hosted_apis: {
+    appeals: isHostedApiEnabled('appeals', true),
+    argonaut: isHostedApiEnabled('argonaut', true),
+    benefits: isHostedApiEnabled('benefits', true),
+    claims: isHostedApiEnabled('claims', true),
+    facilities: isHostedApiEnabled('facilities', true),
+    loan_guarantees: isHostedApiEnabled('loan_guarantees', false)
+  }
+};
+
```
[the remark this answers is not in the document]
kalilsn: My strong opinion: I like this approach!

PR #26, opened 2018-12-27
Reorganize css to only allow variable/mixin importing to all modules

edpaget:
Resolves department-of-veterans-affairs/vets-contrib#711


edpaget:
Oh sick that regression test was awesome, @kalilsn!

edpaget:
@juliaelman 

1. No, I wouldn't work unfortunately. What makes this fix work is changing the `base.scss` file to only export Sass variables and no CSS. We're using the variables set by uswds in some of the component scss files, so there needs to be a central importable file with all of them. 
1. I think we might want to wait until we have a documented way to organize the repo. I personally lean towards keeping scss and image files in the same folder as their associated component since webpack manages where the files actually live (and in some cases can inline small images for performance).

juliaelman:
@edpaget agree on the point about having images + CSS files in one place.

Also, the images listed in number 3 are one's that are not used on the site AFAIK. Going to go ahead and approve the PR because that should not block the merge.

edpaget:
Ah got it I'll delete the other images.

PR #28, opened 2019-01-02
Fixes from npm audit (to unlock #780)

dvogel:


dvogel:
Just pushed a change to use the `critical` settings for `audit-level`. The `high` setting is too loose with the definition of "vulnerability" if failing to verify the origin in a dev server is considered a serious vulnerability.

PR #33, opened 2019-01-04
Expose disability_rating and service_history api docs

kalilsn:
Resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/554


src/App.tsx
```diff
@@ -32,8 +32,10 @@ const flags = {
     argonaut: isHostedApiEnabled('argonaut', true),
     benefits: isHostedApiEnabled('benefits', true),
     claims: isHostedApiEnabled('claims', true),
+    disability_rating: isHostedApiEnabled('disability_rating', true),
     facilities: isHostedApiEnabled('facilities', true),
-    loan_guarantees: isHostedApiEnabled('loan_guarantees', false)
+    loan_guarantees: isHostedApiEnabled('loan_guarantees', false),
+    service_history: isHostedApiEnabled('service_history', true)
```
edpaget:
```suggestion
    service_history: isHostedApiEnabled('service_history', true),
```

I think we might want to add to our linter, but its easier for having to redo alpha sorts latter if we leave trailing commas.

PR #36, opened 2019-01-07
Ep add unit tests

edpaget:


edpaget:
Whoops guess my local linter wasn't running against my test files.

src/actions/index.spec.ts
```diff
@@ -0,0 +1,238 @@
...
+    await actions.submitForm()(dispatch, getState, undefined);
+    expect(dispatch).toBeCalledWith({type: constants.SUBMIT_APPLICATION_BEGIN});
+    expect(dispatch).toBeCalledWith({
+      type: constants.SUBMIT_APPLICATION_ERROR,
+      status: 'Max Retries Exceeded. Last Status: Oops',
+    });
+  });
+
+  it('retries the correct number of times the fetch errors', async () => {
+    fetch.mockReject(new Error('Oops'));
```
dvogel: If the test below is simulating an error response from the remote server, what type of error is this test simulating? Is this duplicative of the test below? Is this simulating an error with the input to `fetch` that causes it to never send the request? I'm guessing not since retry seems inappropriate there.

nit: The test name seems to be missing a word. Maybe "_when_ the fetch _receives_ an error"?

src/actions/index.spec.ts
```diff
@@ -0,0 +1,238 @@
...
+});
+
+describe('submitFormError', () => {
+  it('should return the correct type', () => {
+    expect(actions.submitFormError('test-token')).toEqual({
+      type: constants.SUBMIT_APPLICATION_ERROR,
+      status: 'test-token'
+    });
+  });
+});
```
dvogel: The code in `submitFormError` and `submitFormSuccess` is so simple that this test and the one above it feel almost tautological. It would be pretty hard to change this method in a way that violated this test without it first being caught by the type checker since the return type of the method uses literal types:

```
export interface ISubmitFormError extends Action {
  type: constants.SUBMIT_APPLICATION_ERROR;
  status: string;
}
```

```
export const submitFormError : ActionCreator<ISubmitFormError> = (status : string) => {
  return {
    status,
    type: constants.SUBMIT_APPLICATION_ERROR,
  };
}
```

The same is true for the toggle tests below. I'm fine reducing the specificity of the typing and verifying this behavior via tests. I'm also fine omitting tests for these actions since they are so specifically typed. Doing both feels a tad superfluous.

src/reducers/index.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+          benefits: true
+        })
+      }));
+    expect(application(newApp, { type: constants.TOGGLE_BENEFITS_CHECKED }))
+      .toEqual(expect.objectContaining({
+        apis: expect.objectContaining({
+          benefits: false
+        })
+      }));
+  });
```
dvogel: I think this is the appropriate level to test these toggles. This test covers multiple potential mistakes we could make. It also tests against regressions introduced by upgraded dependencies.

dvogel:
Lots of tests! Thanks for adding these. I think this will really help cover one of our blind spots while developing the portal.

src/actions/index.spec.ts
dvogel: I see. Can you change the test title or the `Oops` message to say "network error" or something similar?

dvogel: Nevermind. You already did that :1st_place_medal:

PR #37, opened 2019-01-08
Add internal only API section

dvogel:
This adds a new section for "internal only" APIs to house the Vet360 address validation API. The copy is made up by me on the fly. I'll fix that tomorrow, before merging this, after discussing it with Jason.


dvogel:
> Visual regression testing failed. Review these diffs and then update the snapshots.
> 
> [renders-explore-benefits-docs-appeals-properly-1-diff.png](https://s3-us-gov-west-1.amazonaws.com/developer-portal-screenshots/5cbb16495dd5876e05c24ef54cb776a794c30ddd/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-appeals-properly-1-diff.png)
> [renders-explore-benefits-docs-benefits-properly-1-diff.png](https://s3-us-gov-west-1.amazonaws.com/developer-portal-screenshots/5cbb16495dd5876e05c24ef54cb776a794c30ddd/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-benefits-properly-1-diff.png)
> [renders-explore-properly-1-diff.png](https://s3-us-gov-west-1.amazonaws.com/developer-portal-screenshots/5cbb16495dd5876e05c24ef54cb776a794c30ddd/visual-regression-test-ts-visual-regression-test-renders-explore-properly-1-diff.png)

I believe these are intended. I'll review them more closely tomorrow and commit updated standards if appropriate.

src/App.tsx
```diff
@@ -68,7 +69,7 @@ class App extends React.Component {
         <Route path="/apply" component={Apply} />
         <Route path="/beta" component={BetaPage} />
         <Route path="/beta-success" component={BetaSuccess} />
-        <Route path="/explore/:apiCategory?" component={ExploreDocs} />
+        <Route path="/explore/:apiCategoryKey?" component={ExploreDocs} />
```
dvogel: The `Key` suffix is just to differentiate this string from the references to `IApiCategory` objects elsewhere.

src/containers/ExploreDocs.tsx
```diff
@@ -90,26 +93,36 @@ const apiDefs : IApiCategories = {
...
             {
-                name: 'Disability Rating',
-                shortDescription: "Get a veteran's disability rating",
-                urlFragment: 'disability_rating',
-            },
-            {
-                name: 'Service History',
-                shortDescription: "Get a veteran's service history",
-                urlFragment: 'service_history',
-            },
```
dvogel: I'm not sure how these got dropped. I'll put them back :)

tslint.json
```diff
@@ -13,7 +13,13 @@
...
         "singleline": "never"
       },
-      "severity": "default"
+      "severity": "warning"
+    },
+    "no-console": {
+      "severity": "error"
+    },
+    "prefer-const": {
+      "severity": "error"
```
dvogel: These seemed to be the defaults. Since I often change them to "warning" while actively developing, I added them here. I'm personally OK if we want to just permanently change these to "warning".

src/containers/ExploreDocs.tsx
```diff
@@ -90,26 +93,36 @@ const apiDefs : IApiCategories = {
...
 };
 
+const apiCategoryOrder: string[] = [
+    'benefits',
+    'facilities',
+    'health',
+    'verification',
+    'internal',
+];
+
```
dvogel: The tslint rule for requiring alphabatized keys in object literals requires a list to specify this order. I have a half-done refactor to move the API data out of this file so that it can be used by multiple components. When that happens this `apiCategoryOrder` list will remain here, so the split is better IMO anyway.

src/containers/ExploreDocs.tsx
```diff
@@ -215,106 +235,109 @@ class ApiPage extends React.Component<RouteComponentProps<IApiNameParam>, {scrol
...
-                const { history, location: { pathname } } = this.props;
-                if (pathname !== path) {
-                    history.push(path);
+    // Determines which section is being viewed. Ensures that the window location is set to the URL
+    // for that section. The left nav rendering uses the window location to determine which nav item
+    // to visually focus.
+    private updatePathToReflectView() {
+        const distanceRecords: Array<[number, number, string]> = [];
+
+        Object.keys(this.sectionRefs).forEach((apiCategoryKey, idx) => {
```
dvogel: This is an example of where I think the `prefer-const` tslint rule goes too far. Using `flatMap` for this would probably be overkill. Declaring the array reference `const` while changing it's contents below feels like a lie. LIES I TELL YOU! :smile:

juliaelman:
@dvogel a few questions:

1. Where will Developers (and maybe Partners) be coming from in order to find this information?
1. What VA Internal Only APIs are we currently including on the Developer Portal?

Below is a screenshot of what I've got running locally, for reference.

screen shot 2019-01-08 at 10 19 30 am

dvogel:
Good question @juliaelman . The first API being hosted in that section is the address validation API (from Vet360). It isn't appearing in that screenshot because it's disabled unless `.env` includes:

```
REACT_APP_ADDRESS_VALIDATION_API_ENABLED=true
```

The immediate goal here is to make this documentation available on the portal for the Vet360 team to send potential consumers to directly. They have a few other teams asking how they can consume the API and Vet360 would like to get them on-boarded through us soon.

juliaelman:
@dvogel great - thank you!

Based on the [first task results from the Developer Portal IA Treejack Summary](https://github.com/department-of-veterans-affairs/vets-contrib/blob/master/lighthouse/products/developer-portal/research/ia/va-dev-portal-ia-treejack-summary.md), Developers respond better when information corresponds with a particular API. As you can see in these results, we had a 0% success rate with OAuth living on it's own.

With that, it might make more sense to place this under the "Veteran Verification API" and add some sort of signifier that it is "VA Internal Only".

@jwoodman5 would love to hear your thoughts. Happy to help do this design work too!

dvogel:
@juliaelman I think those are good suggestions. We discussed a similar aspect of this in sprint planning. In that discussion we were concerned that developers outside the VA would find this API and feel confused. We were trying to leverage the idea behind that 0% success rate in order to hide it from most developers and make it easy enough for someone in the know to navigate back to in the event they were linked directly to the docs and left the page to view something else on the portal but needed to get back.

I think your suggestion of making the API discoverable but finding a different way to clearly indicate that this API is only for internal VA consumers would be useful too. Probably much better in the long term. If this pattern continues with more internal APIs, we would essentially have to replicate the full API categorization in the nav a second time under Internal-Only.

For the goal of this sprint, which is getting the Vet360 documentation on the portal at all, I am cool with either approach.

src/containers/ExploreDocs.tsx
```diff
@@ -106,10 +122,25 @@ const apiDefs : IApiCategories = {
         buttonText: "Stay Informed",
         name: "Veteran Verification",
         overview: VerificationOverview,
-        shortDescription: "Empowering Veterans to take control of their data and put it to work.",
+        shortDescription: "Coming soon! Empowering Veterans to take control of their data and put it to work.",
```
kalilsn: I removed this when we added the first APIs in this category but I wasn't really sure if it was the right choice. I also got rid of [the notice on the homepage](https://github.com/department-of-veterans-affairs/developer-portal/pull/33/files#diff-11945690eecf779e3b81b16f40cf777fL60), so do you want to add that back too?

dvogel: Oops. I didn't mean to revert any of your changes! I'll remove the "Coming soon!" bit again.

src/containers/ExploreDocs.tsx
kalilsn: I don't really have strong feelings about the display order being implicit in the object or declared in this array, but in general I don't think we should feel bad about just disabling linting rules (either permanently or via a comment) if they're making us write more complicated code

jwoodman5:
@juliaelman @dvogel Great suggestions and comments. Julia - Are you suggesting the address validation API documentation lives within the menu similar to Veteran Verification > VA Internal Only > Address Validation API?  And basically, based on the results of the Treejack test we could potentially have a "VA Internal Only" section within each main API category?

tslint.json
kalilsn: I don't think we should _permanently_ set anything to warning so that we don't clutter the linter output (especially on Jenkins). The accessibility ones are set to warning with the hope that we will fix those issues soon, and then bump them up back up to errors.

I'm all for leaving them in here so it's easy to manually change them when you're developing, but I think we should leave trailing comma as an error (default is error right now) or just disable it if we don't like it!

jwoodman5:
@dvogel I think that is a fair suggestion and to your main point, getting the docs up is main priority. I would second the suggestion of nesting the address validation API docs under Veteran Verification if that is what it relates to. And for the wording, I like the last example you have - "*VA Internal-Only APIs* Some APIs are made available by the VA are only authorized for internal use by VA teams. This particular API is not available for public use." 

We could even add a link and something like, "We do have a variety of APIs which are available for public use, which can be found _here_." And return them to the dev portal homepage. @juliaelman Thoughts on that?

src/containers/ExploreDocs.tsx
edpaget: Yeah I support disabling rules when it feels natural to do so, just including a comment about why you're disabling would be helpful for future devs.

tslint.json
edpaget: I'm open to changing the `console` lint rule to a warning. I'm heavy user of `println` debugging, so I use console statements all the time and have to disable with a comment, which has meant my `console` statements have made it into the master branch in the past.

PR #38, opened 2019-01-08
Add Vet360 address_validation API docs. (#906)

dvogel:
This PR is a follow-up from [PR #37](https://github.com/department-of-veterans-affairs/developer-portal/pull/37). This PR moves the `address_validation` API under the `verification` section and adds a tag to the API name to indicate that it's for internal VA use only.


tslint.json
```diff
@@ -13,7 +13,13 @@
         "multiline": "always", 
         "singleline": "never"
       },
-      "severity": "default"
+      "severity": "warning"
```
kalilsn: I think this should be error (that was the default before) still, or we should just remove it. Not gonna argue about it if you feel strongly, but it seemed like this might have just been a mistake.

dvogel:
@jwoodman5 hopefully these screenshots help. Moving the address_validation API under the verification section means we don't have to define nearly as much explanatory text. It's just he blurb inside the link bucket.

Here's the nav, same as it was:
address_validation_pr_nav

Here's the view when you click the verification nav link:
address_validation_pr_links

src/containers/ExploreDocs.tsx
```diff
@@ -171,10 +197,19 @@ function ApiSection({ apiCategory, sectionRef } : { apiCategory : string, sectio
...
+class ApiPage extends React.Component<RouteComponentProps<IApiNameParam>, {}> {
     private sectionRefs : { [key: string]: React.RefObject<HTMLElement> } = {}
+    // Our componentDidUpdate method will be called whenever any of these happen:
+    //   * NavLink clicked.
+    //   * The initial route is rendered.
+    //   * The after-scroll event changes the navigation stack and the page re-renders.
+    // The skipAutoScroll flag is set when the user is manually scrolling in order to avoid the
+    // navigation stack modifications from scrolling the viewport to the top of the newly-current
+    // section.
+    private skipAutoScroll: boolean;
```
kalilsn: Really appreciate this being written out here! And the rewrite in general looks great!

tslint.json
dvogel: Sounds good. When I first ran the linter it found a bunch of pre-existing violations so I assumed it must have been set to warning of disabled altogether previously. Switching it to error now doesn't find any violations so we're back in good standing :)

dvogel:
@jwoodman5 @juliaelman do either you have preferences re: @kalilsn's suggestion for adding an indication that address verification is internal only to the left nav?

This looks OK when it's selected:
selected

It looks a little strange before it's selected though:
unselected

jwoodman5:
I like the idea as it provide notice one step earlier, but not sure how I feel about how it looks in deselected state. And I don't have any ideas at the moment. I will defer to @juliaelman on that.

juliaelman:
@dvogel I am having trouble getting my local env working. Please update the "VA Internal" copy on the side nav to:

1. Left align with "Address Validation"
1. Remove parenthesis.
1. s Internal VA use/Internal VA use only.
1. Use `<small>` around the copy to differentiate it from the other side nav copy.
1. Change the line height on that copy so that it is closer to the "Address Validation"

@jwoodman5 Let's plan on chatting next steps for the overall IA of the Developer Portal in our meeting tomorrow. I think we're moving a bit fast and making assumptions on how our users will leverage the side nav. I'd also like to know where you see the post Treejack test work going so that we can create an optimal user experience so that Developers can easily find what they need.

PR #40, opened 2019-01-11
Run accessibility tests

kalilsn:
Resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/908


Jenkinsfile
```diff
@@ -127,7 +127,21 @@ node('vetsgov-general-purpose') {
...
+    try {
+      dockerImage.inside(args) {
+        sh 'cd /application && npm run-script test:accessibility:ci'
+      }
+    } catch (error) {
+      notify()
+      dir(pwd()) {
+        step([$class: 'JUnitResultArchiver', testResults: 'test-report.xml'])
       }
       throw error
```
edpaget: I think these shouldn't throw an error right since we want them to report, but not fail the builds?

kalilsn: That's true, but I don't think we can even achieve that at the moment. If you archive a junit report with test failures it marks the build unstable (which github reads as a failure) even if no error is thrown. I've just disabled the tests with `xit` for now so that we'll be completely ready to actually run them once they're fixable.

PR #45, opened 2019-01-18
Change LGY doc path (#1019)

dvogel:
This mirrors a change we made to kong in dev.


src/containers/Explore.tsx
```diff
@@ -56,7 +56,7 @@ class Explore extends React.Component<IExploreProps, { }> {
             case 'loan_guarantees':
                 docs = (
                     <Flag name="hosted_apis.loan_guarantees">
```
kalilsn: Let's make the flag name match too

dvogel: Good catch!

dvogel: Just pushed changes to rename the feature flag too.

PR #50, opened 2019-01-25
Move overview section to the vba_documents swagger file.

dvogel:
Since this language is so specific to the benefits intake API it's moving to the swagger file in vets-api. Here's the corresponding PR from vets-api: https://github.com/department-of-veterans-affairs/vets-api/pull/2760


src/content/apiDocs/benefitsOverview.mdx
```diff
@@ -8,12 +8,6 @@ Our goal is to make easier and more secure for VSOs and other organizations to i
 
  VA is moving towards an Application Programming Interface (API) first driven digital enterprise, which will establish the next generation open management platform for Veterans and accelerate transformation in VA's core functions, such as Health, Benefits, Burial and Memorials. This platform will be a system for designing, developing, publishing, operating, monitoring, analyzing, and optimizing VA’s API ecosystem. We are in the alpha stage of the project, wherein we provide one API that enables parties external to VA to submit VBA forms and supporting documentation via PDF on behalf of Veterans.
```
kalilsn: Does this still make sense here? Seems like this is general information about the whole program

kalilsn: I guess I'd never really read these all the way through, but there's a lot of duplicated information in the Health and Benefits overview sections about environments, path to production, etc. I think those should be removed if they're not specific to those APIs, or if they duplicate info that's in other places on the site (so they don't get out of sync). Maybe that's work for a future PR though.

dvogel: I agree that this info isn't specific to this API and it should move. The text I moved to the swagger docs differs from this text in that I don't know where this should move go. As we deliver the navigation re-org, we'll have to go over all of the copy to make sure it still makes sense.

PR #52, opened 2019-01-28
Add OAuth sample application

edpaget:
Resolves department-of-veterans-affairs/vets-contrib#463


edpaget:
Working on a corresponding update with this code to vets-api-clients.

juliaelman:
@dvogel agreed. Something I've been thinking about is how the auth content is worked into the overall site. AFAIK auth is used in different ways for both Health + VV APIs. 

These thoughts feed directly into some next steps on research around our docs overall. There are lots of questions around our current documentations performance and how quickly Developers are able to get to that "aha" moment.

@edpaget I wonder if it would be helpful to add a link to this from https://github.com/department-of-veterans-affairs/vets-api-clients/ 🤔

PR #53, opened 2019-01-28
Don't block master deploys based on npm audit failures.

dvogel:
This should prevent us from getting to the situation we found ourselves in a few weeks ago where pre-merge checks passed but post-merge checks of the same code failed due to intermittent `npm audit` errors. Those errors were due ot operational issues experienced by the npm team. However, the remote-database architecture of npm's audit mechanism causes newly discovered security vulnerabilities to exhibit the same issues in our CI workflow.


config/env.js
```diff
@@ -15,15 +15,24 @@ if (!NODE_ENV) {
...
+
   // Don't include `.env.local` for `test` environment
   // since normally you expect tests to produce the same
   // results for everyone
   NODE_ENV !== 'test' && `${paths.dotenv}.local`,
+
+  // e.g. .env.test
+  BUILD_ENV !== null && `${paths.dotenv}.${BUILD_ENV}`,
+  `${paths.dotenv}.${NODE_ENV}`,
+
```
dvogel: I don't think these changes should have any practical effect since `.env.local` already override `.env` and `.env.dev.local` already overrode `.env.dev`. Having them specified in the order documented by the linked gem just made it much easier for me to visually verify that.

Jenkinsfile
```diff
@@ -120,11 +122,19 @@ node('vetsgov-general-purpose') {
...
-      throw error
+      if (!onDeployableBranch()) {
+        // Only PR branches should be blocked by npm audit failures. If a
+        // package we already rely on is newly discovered to be vulnerable
+        // deploys of master it not clear that the security problem is worse
+        // than the problem motivating the deploy. Hopefully we will learn
+        // about these new audit errors while preparing our next PR.
+        notify()
+        throw error
+      }
```
dvogel: This is the heart of the change for the ticket.

Jenkinsfile
```diff
@@ -120,11 +122,19 @@ node('vetsgov-general-purpose') {
   stage('Security') {
     try {
       dockerImage.inside(args) {
-        sh "cd /application && npm config set audit-level critical && npm audit"
+        // sh "cd /application && npm config set audit-level critical && npm audit"
+        sh "exit 1"
```
dvogel: This is just to exercise the block below on Jenkins. The [job failed](http://jenkins.vetsgov-internal/job/testing/job/developer-portal/job/dv-994-npm-audit-ci/4/) and the slack notification didn't occur so we should be good. I'll return this to it's original state before merging.

Jenkinsfile
```diff
@@ -36,9 +40,7 @@ def buildDetails = { vars ->
 }
 
 def notify = { ->
-  if (env.BRANCH_NAME == devBranch ||
-      env.BRANCH_NAME == stagingBranch ||
-      env.BRANCH_NAME == prodBranch) {
+  if (onDeployableBranch()) {
```
dvogel: Everything above here should be simple consolidation based on De Morgan's law ... assuming I remembered it correctly :smile:

README.md
```diff
@@ -11,7 +11,7 @@ First install the dependencies:
 npm install
 ```
 
-Next, create a `.env` file with the following contents (don't worry about what these mean yet):
+Next, create a `.env.local` file with the following contents (don't worry about what these mean yet):
```
dvogel: Since `.env` is used by the test suite but `.env.local` is not, this should give people an easier time with the visual regression tests. Specifically, they won't accidentally commit visual artifacts that depend on their local feature flagging.

kalilsn:
I might be wrong, but I think that ticket was originally about our security audit step in the devops repo because it was holding up other teams.

Jenkinsfile
dvogel: The second sentence is meant to address a concern about whether newly-vulnerable packages will go unnoticed if our deploys aren't giving us that feedback anymore. That seems like a concern for an app that isn't being actively deployed but since we're pretty active in this repo, there's plenty of PR runs to give us that feedback. Any idea how to make that clearer? Just remove it?

README.md
kalilsn: This makes sense! I also think we could also improve this with a `.dockerignore` file to exclude `.env.*.local` files (or whichever one it is that people have been using)

Jenkinsfile
dvogel: Is this clearer?

```
        // Only PR branches should be blocked by npm audit failures. Packages
        // that receive new vulnerability notices after master already depends
        // on them should be handled separately from our post-merge
        // auto-deploys.
```

dvogel:
> I might be wrong, but I think that ticket was originally about our security audit step in the devops repo because it was holding up other teams.

Thanks for pointing that out. The ticket was a little vague. I first experienced this problem here so I tried to fix it here first. I think you're probably right since the title of the ticket specifically mentions the npm-audit-ci package, which is depended on by the devops Jenkinsfile. I'll follow up with another PR for that.

Jenkinsfile
kalilsn: Perfect!

dvogel:
re: the devops jenkinsfile changes, we've decided to defer breaking out the devops jenkinsfile entries for the lambdas until we get the new backend up and running.

PR #57, opened 2019-02-07
disable Swagger's "Try it out" button by default

nfasulo:
Resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/998.

An API can have the "Try it out" button enabled by adding an environmental variable (e.g. `REACT_APP_LOAN_GUARANTY_TRY_IT_OUT_ENABLED=true`).


src/components/swaggerPlugins/TryItOut.jsx
```diff
@@ -0,0 +1,21 @@
+export class TryItOut {
+
+  static toggle() {
+
+    return {
+      spec: {
+        wrapSelectors: {
+          allowTryItOutFor: () => () => {
+            const currAPI = window.location.pathname.split('/').pop().toUpperCase();
+            const envValue = process.env[`REACT_APP_${currAPI}_TRY_IT_OUT_ENABLED`];
```
dvogel: I have two thoughts here that intersect:

* Could we make the `Explore` container pass in a `disableTryItOut` prop in order to disentangle this from `window`? Seems like `SwaggeDocs.renderSwaggerUI` could customize the list of plugins used based on that prop.
* Disabling this is mainly for us versus the public so we could probably get away with a global setting instead of a per-API setting. That would eliminate the need to reach into `window` for the API name.

nfasulo: This is my first experience with React so I would be happy to go with just having a global setting to disable this for all APIs for the time being. Though I can start looking into how to modify the `Explore` container to pass in a `disableTryItOut` prop if that seems valuable.

nfasulo:
> Visual regression testing failed. Review these diffs and then update the snapshots.
> 
> [renders-explore-benefits-docs-appeals-properly-1-diff.png](https://s3-us-gov-west-1.amazonaws.com/developer-portal-screenshots/8afcee56679fd6095e5f5ed53ed9acb2cda86905/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-appeals-properly-1-diff.png)
> [renders-explore-benefits-docs-benefits-properly-1-diff.png](https://s3-us-gov-west-1.amazonaws.com/developer-portal-screenshots/8afcee56679fd6095e5f5ed53ed9acb2cda86905/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-benefits-properly-1-diff.png)

@kalilsn I forgot to run the visual regression tests before pushing changes. Am I ok to update the snapshots to show that the Authorize button is no longer visible?

kalilsn:
@nfasulo Yeah, that's the intended process! If anything, having them fail once is kind of nice because reviewers can see the diff a little more easily.

src/components/swaggerPlugins/TryItOut.jsx
kalilsn: I think if you change from disabling try it out on a per-api basis, you don't need to pass a `disableTryItOut` prop at all. You can just read from a single env var in `src/components/swaggerPlugins/index.ts` to determine if those plugins are used or not.

PR #58, opened 2019-02-07
Add OAuth redirect URL to sign-up form.

dvogel:
This adds the redirect URL field to the apply form, which is saved by the changes in this PR: https://github.com/department-of-veterans-affairs/devops/pull/4107


src/containers/ApplyForm.tsx
```diff
@@ -134,6 +130,18 @@ class ApplyForm extends React.Component<IApplyProps> {
                   <label htmlFor="facilities">VA Facilities API</label>
                 </div>
 
+                <h4> OAuth APIs:</h4>
```
kalilsn:
```suggestion
                <h4>OAuth APIs:</h4>
```

src/actions/index.ts
```diff
@@ -181,11 +187,21 @@ export const submitFormError : ActionCreator<ISubmitFormError> = (status : strin
...
   }
+}
+
+export const validateEmail = (newValue: IErrorableInput) => {
+  validateByPattern(newValue, EMAIL_REGEX, "Must be a valid email address.");
+  return newValue;
+}
+
+export const validateOAuthRedirectURI = (newValue: IErrorableInput) => {
+  const partialUrlPattern = /^http[s]?:[/][/][^/:?]+(:[0-9]+)?([/][^?#]*)?$/;
```
edpaget: I think this is allowing URLs with fragments (like `?test=foo`), which okra will reject. 

```suggestion
try {
  const uri = new URL(newValue.value);
  if (uri.search !== "") {
    newValue.validation = "Cannot include a query fragment";
  }
} catch (error) {
  // This happens when the URL is malformed, for instance does not include a protocol
  newValue.validation = "Must be an http or https URL.";
}
```

dvogel: The regex rejects query strings. It does accept fragments. I'm not tied to that regex but the URL class doesn't provide direct access to the fragment.

edpaget: Hmm I tried it on the linked preview paged and seemed to be able to submit with a query string.

src/containers/ApplyForm.tsx
```diff
@@ -182,6 +192,22 @@ class ApplyForm extends React.Component<IApplyProps> {
     );
   }
 
+  private renderOAuthFields() {
+    if (this.props.inputs.apis.health || this.props.inputs.apis.verification) {
+      const oAuthRedirectURI = this.props.inputs.oAuthRedirectURI;
+      return (
+          <ErrorableTextInput
+            errorMessage={null}
```
edpaget: This should be `this.props.updateOAuthRedirectURI.validation`

src/containers/ApplyForm.tsx
```diff
@@ -195,11 +221,29 @@ class ApplyForm extends React.Component<IApplyProps> {
...
+    return incompleteFields.length === 0;
+  }
+
   private readyToSubmit() {
-    const { inputs: { apis, email, firstName, lastName, organization, termsOfService } } = this.props;
-    return !!email.value && !!firstName.value && !!lastName.value && !!organization.value &&
-           ( apis.verification || apis.health || apis.benefits || apis.facilities ) &&
-           termsOfService;
+    const { inputs: { oAuthRedirectURI, termsOfService } } = this.props;
+    const redirectURIComplete = (!this.anyOAuthApisSelected() || !!oAuthRedirectURI.value);
```
edpaget: Should also check: `!oAuthRedirectURI.validation`

src/actions/index.ts
edpaget: Ah got it the issue was elsewhere.

src/containers/ApplyForm.tsx
dvogel: Good catch! I put too much trust in my validation test and didn't verify that the submission check was accurate.

dvogel: Fixed.

src/containers/ApplyForm.tsx
dvogel: Fixed.

PR #60, opened 2019-02-11
Properly save termsOfService.

dvogel:


src/actions/index.ts
```diff
@@ -121,11 +121,12 @@ const fetchWithRetry = async (fetchFn : () => Promise<Response>) : Promise<Respo
 function buildApplicationBody({ application }: IRootState) {
   const applicationBody : any = {};
   applicationBody.apis = apisToList(application.inputs.apis);
-  ['description', 'email', 'firstName', 'lastName', 'oAuthRedirectURI', 'organization', 'termsOfService'].forEach((property) => {
+  ['description', 'email', 'firstName', 'lastName', 'oAuthRedirectURI', 'organization'].forEach((property) => {
     if (application.inputs[property]) {
       applicationBody[property] = application.inputs[property].value;
     }
   });
+  applicationBody.termsOfService = application.inputs.termsOfService;
```
dvogel: This is the heart of the fix. The `termsOfService` field is a plain boolean and not an Errorable input.

src/store.ts
```diff
@@ -51,9 +57,9 @@ const store = createStore(
     ),
 )
 
-store.subscribe(() => {
+store.subscribe(debounce(() => {
   saveApplicationState(store.getState());
-});
+}));
```
dvogel: This is a slight perf improvement that was annoying me while debugging.

src/store.ts
```diff
@@ -29,7 +30,12 @@ function loadApplicationState(): { application: IApplication } {
 
 function saveApplicationState(state: IRootState) {
   try {
-    const serializedState = JSON.stringify({application: state.application})
+    const stateToSerialize = {
+      application: {
+        inputs: state.application.inputs,
+      },
+    };
+    const serializedState = JSON.stringify(stateToSerialize);
```
kalilsn: This changes it to stop storing any properties other than `application.inputs` in session storage, right? Was this also for performance?

dvogel: It was to avoid saving the `sending` flag. If the form submission fails to return, then the user would be stuck, unable to re-submit the form. I think this is a bigger risk for developers than real users, but in general we never want to save UI state.

kalilsn: Makes sense! I was just slightly concerned about this biting someone later on if they were expecting a piece of state to be persisted and it wasn't.

PR #63, opened 2019-02-12
Move terms of service from left nav to footer.

dvogel:


package.json
```diff
@@ -100,7 +100,7 @@
     "deploy": "gh-pages -d build",
     "lint": "./node_modules/.bin/tslint --config tslint.json --project tsconfig.json",
     "lint:ci": "./node_modules/.bin/tslint --config tslint.json --project tsconfig.json -t junit > lint-results.xml",
-    "test": "node scripts/test.js",
+    "test": "echo You should run test:unit, test:e2e, test:accessibility, or test:visual",
```
dvogel: Since this is broken by default, let's not have any new folks fall into the trap :smile:

src/App.tsx
```diff
@@ -62,6 +63,11 @@ class App extends React.Component {
     return (
       <PageContent {...props} >
         <Route exact={true} path={["/", "/index.html"]} component={Home} />
+
+        {/* Legacy routes that we want to maintain: */}
+        <Route path="/explore/terms-of-service" render={props1 => <Redirect to="/terms-of-service" />} />
+
+        {/* Current routes: */}
```
dvogel: I'm hoping we can make the portal as easy to understand on it's own as possible. To that end, I figured it would be best to handle these links at the top-level of our routes instead of in the revproxy rewrite rules, which are hard to test locally.

tslint.json
```diff
@@ -72,6 +72,9 @@
     },
     "react-a11y-titles": {
       "severity": "warning"
+    },
+    "jsx-no-lambda": {
+      "severity": "off"
```
dvogel: This preset seems far too restrictive given [this guidance from Dan Abramhov](https://github.com/facebook/create-react-app/issues/4801#issuecomment-409553780). The alternative approach, which seems to be instantiating the `<Redirect>` element separate from the instantiation of the `<Route>` element seems worse to me.

src/App.e2e.ts
```diff
@@ -0,0 +1,20 @@
...
+
+describe('terms of service', () => {
+  // This is tested in App to test coverage broadly. We care that the user has
+  // access to this link whereever they are.
+  it('is displayed in the footer', async () => {
+
+    for (const path of ['/', '/apply', '/explore']) {
+      await page.goto(`${puppeteerHost}${path}`, { waitUntil: 'networkidle0' });
+      const tosLinkCount = await page.evaluate(() => {
+        const pathExpr = "//footer//a[contains(., 'Terms of Service')]";
```
kalilsn: I'm curious about your choice of xpath over a css selector here!

dvogel: I wanted to select the element based on the link text and AFAIK the query selector API does not support that. Does it?

kalilsn: Yep, that makes sense now. Didn't really think it through! I think you could still filter the results of the query based on the node's `textContent`, but I don't think there's any _reason_ to do that :)

PR #65, opened 2019-02-15
Dev portal changes for the banner and Veterans Crisis Line

mjlumetta:
Includes changes to the banner with the official government site notice and the Veterans Crisis Line button/modal. 

### Desktop

- Check that the official site banner expands and contracts and that it looks as expected on desktop.

expandedgovbannerdesktop

- Check that the Veterans Crisis Line button looks as expected and opens the correctly styled modal on desktop.

opencrisislinemodaldesktop

- General: try out responsive styling.

### Mobile

- Check that the official site banner expands and contracts and that it looks as expected on mobile.

expandedgovbannermobile

- Check that the Veterans Crisis Line button looks as expected and opens the correctly styled modal on mobile.

opencrisislinemodalmobile


src/components/Banner.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+		const accordionWrapper = shallow(<Banner />).find('.usa-accordion-content');
+		expect(accordionWrapper.prop('aria-hidden')).toBe('true');
+	});
+
+	it('should show the site guidance accordion when accordionVisible is true', () => {
+		const wrapper = shallow(<Banner />);
+		wrapper.setState({ accordionVisible: true });
+		expect(wrapper.find('.usa-accordion-content').prop('aria-hidden')).toBe('false');
+	});
+});
```
dvogel: This repo has developed a slightly annoying inconsistency of 2 vs 4 space indents. Now that it's clear that all teams will be working on the portal, I'll start a convo in #dev-portal-chat to settle the the 2 vs 4 question and we can install a tslint rule for it. In the meantime, let's not add 8 space indents too :smile:

src/components/VeteransCrisisLine.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <p>
+            	Get more resources at &nbsp;
+            	<a href="https://www.veteranscrisisline.net/">VeteransCrisisLine.net</a>.
+    				</p>
+          </div>
+        </div>
+      </div>
+    );
+	}
+}
```
dvogel: Is this borrowed from va.gov? If not, would they benefit from us pushing this component upstream to the formation design system repo?

dvogel:
It looks like some styling changes bled over into the top nav bar:

screenshot from 2019-02-18 10-01-41

This is what it looks like in production:

screenshot from 2019-02-18 10-01-33

src/components/VeteransCrisisLine.tsx
mjlumetta: umm it's a close parallel. it may already exist in that repo in some form, possibly, so i can check. but i think the answer is yes in general if it's not there and it's possible with the exist design system repo

kalilsn:
One more question that I forgot to add there 😓 (also probably for @juliaelman )

When I encountered the issue with being unable to close the modal because the button was off screen, I tried pressing `esc` and clicking in the background area, which seem like common patterns for closing a modal window. Is that something we can consider adding here or are there UX reasons to avoid that?

juliaelman:
@kalilsn lengthy comments are welcmoe 😄 

> Our active state is different (grey dotted outline vs solid yellow outline)

Nice catch. That should be updated to the yellow outline represented in the [Figma prototype.](https://www.figma.com/proto/SGQZyWGE44uYM4lI1twVsw/IA-Updates-Jan-2019?node-id=26%3A428&viewport=211%2C193%2C0.229344&scaling=min-zoom)

>  When the va.gov modal opens, the "Call 1-800-273-8255 and press 1" item is focused

We should match what is on VA.gov

* The links have a grey background on hover

> To clarify, do you mean the `Here's how you know` link? If so, you are correct that the hover state should stay the same as the default state.

> Our close button is different. This one is actually part of the screenshot in the ticket, so I'm assuming we should make that change.

Correct.

> When I encountered the issue with being unable to close the modal because the button was off screen, I tried pressing `esc` and clicking in the background area, which seem like common patterns for closing a modal window. Is that something we can consider adding here or are there UX reasons to avoid that?

The VCL modal should function exactly like what is on VA.gov, which does include `esc` as a way too close the modal.

src/components/Banner.tsx
```diff
@@ -1,73 +1,121 @@
 import * as React from 'react';
 
 import AlertBox from '@department-of-veterans-affairs/formation/AlertBox';
+import { VeteransCrisisLine } from './VeteransCrisisLine';
 
 import './Banner.scss';
 
-import closeButton from '../../node_modules/uswds/src/img/close.svg';
 import flagIcon from '../../node_modules/uswds/src/img/favicons/favicon-40.png';
+import rightArrow from '../assets/arrow-right-white.png';
```
kalilsn: Can we replace this with an SVG? They've got an svg in the va.gov implementation that you can probably grab. It's pretty blurry on my screen as is

src/components/VeteransCrisisLine.tsx
```diff
@@ -0,0 +1,99 @@
+import * as React from 'react';
+
+import './VeteransCrisisLine.scss';
+
+import { faComments, faDeaf, faMobileAlt, faPhone, faTimesCircle } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import classNames from 'classnames';
+import vclLogo from '../assets/vcl-logo.png';
```
kalilsn: Same as above, should be changed to an SVG

tslint.json
```diff
@@ -18,7 +18,7 @@
       "severity": "error"
     },
     "no-console": {
-      "severity": "error"
+      "severity": "warning"
```
kalilsn: This shouldn't be committed

src/components/VeteransCrisisLine.tsx
mjlumetta: fixed the other one but don't have an svg for this one, unless @juliaelman knows where to find one

juliaelman: @mjlumetta you should be able to use this one: https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/assets/img/VCL-icon-white.svg

mjlumetta: i've replaced another icon with this one, but the VCL logo there is the one with colors.

PR #66, opened 2019-02-19
Bootstrap prettier and tslint rules for indentation.

dvogel:
As we [discussed on slack](https://lighthouseva.slack.com/archives/CAZF880HW/p1550506908019700) there's a strong desire for consistency of indentation, regardless of specific width. This PR adds a prettier config and tslint rule to enforce 2 space indentation. We don't need to continue to use prettier if folks dislike it but it was the quickest way I knew how to fix the existing indentation problems. Along with those indentation fixes comes a few extra "fixes" that I'd rather not enforce but prettier does by default. First among those is breaking template strings: https://github.com/prettier/prettier/issues/3368


src/actions/submitFormStub.ts
```diff
@@ -0,0 +1,18 @@
+// export const submitFormStub : ActionCreator<SubmitFormThunk> = () => {
```
mjlumetta: what is this for?

dvogel: Good catch. That was never supposed to be committed! I was using that as a local testing device a while back.

PR #67, opened 2019-02-20
Add sitemap to developer portal

nfasulo:
Work for department-of-veterans-affairs/vets-contrib#1123. Not complete, but I thought it would be a good time to get feedback. 

I've added the `sitemap.xml` and `robots.xml` to `public`. I am a little concerned about churn with `sitemap.xml`: every time `build` runs it will be regenerated, so when I add the `<lastmod>` tag if someone runs `build` locally, it will register has a change in git (once a day). Maybe this isn't a big deal, because people don't run `build` locally, but I wanted to point it out.

There is still more work to be done building out the `paramsConfig` in `sitemap-builder.js` to add more of the dynamic routes nested under `/explore`.

I have forked https://github.com/kuflash/react-router-sitemap to add the ability to create `<lastmod>` tags. Do we have a standard way of including such forks until changes are accepted upstream?


public/sitemap.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:news="http://www.google.com/schemas/sitemap-news/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml" xmlns:mobile="http://www.google.com/schemas/sitemap-mobile/1.0" xmlns:image="http://www.google.com/schemas/sitemap-image/1.1" xmlns:video="http://www.google.com/schemas/sitemap-video/1.1">
+<url> <loc>https://developer.va.gov/</loc> </url>
+<url> <loc>https://developer.va.gov/explore/terms-of-service</loc> </url>
```
kalilsn: I think we should make sure old URLs that we're redirecting don't show up in the sitemap. Is it possible to add that as a filter?

public/sitemap.xml
```diff
@@ -0,0 +1,13 @@
...
+<url> <loc>https://developer.va.gov/go-live</loc> </url>
+<url> <loc>https://developer.va.gov/terms-of-service</loc> </url>
+<url> <loc>https://developer.va.gov/apply</loc> </url>
+<url> <loc>https://developer.va.gov/applied</loc> </url>
+<url> <loc>https://developer.va.gov/beta</loc> </url>
+<url> <loc>https://developer.va.gov/beta-success</loc> </url>
+<url> <loc>https://developer.va.gov/explore/:apiCategoryKey?</loc> </url>
```
kalilsn: Looks like the dynamic routes didn't get substituted in here

src/sitemap-routes.jsx
```diff
@@ -0,0 +1,23 @@
...
+        <Route path="/go-live" />
+        <Route path="/terms-of-service" />
+        <Route path="/apply" />
+        <Route path="/applied" />
+        <Route path="/beta" />
+        <Route path="/beta-success" />
+        <Route path="/explore/:apiCategoryKey?" />
+        <Route path="/oauth" />
+    </Route>
+);
```
kalilsn: Can we move this to a general `Routes.tsx` and then use that as the single source of truth for the routes (ie. import and use it in `App.tsx`)? I worry that we won't be able to keep the two in sync if we duplicate it.

sitemap-builder.js
```diff
@@ -0,0 +1,30 @@
...
+    presets: [ '@babel/env', '@babel/react' ],
+});
+
+const path = require('path');
+const paths = require('./config/paths');
+const router = require('./src/sitemap-routes.jsx').default;
+const Sitemap = require('react-router-sitemap').default;
+
+// building paths from dynamic routes
+// will need to import apiDefs.ts to further build out dynamic paths
```
kalilsn: started writing a comment to this effect and then saw your comment 🥇

src/sitemap-routes.jsx
nfasulo: Yep, that is on my roadmap. I was running into an error when routes had a component attribute, so I will need strip that out programmatically.

public/sitemap.xml
nfasulo: Hmm, this had been working . . .

public/sitemap.xml
nfasulo: Sure! Besides `/explore/terms-of-service` are there any others I should filter? Specifically I wonder if I should exclude `/applied` and `/beta-success` which are submission redirects.

kalilsn: I don't think there's any other redirects yet, but I believe we'll be adding more in the future. Looks like react-router-sitemap only allows you to pass a regex as a filter, but if you're already forking it, maybe it could take a function too? Or just explicitly exclude it for now and flag that as some future work?

@dvogel do you have any thoughts about those successful submission routes? You can actually directly visit them, so I think we do want to make sure they don't show up in search results. However, I'm not sure if that's really a problem that should be solved at the level of the sitemap – maybe we should refactor so those aren't their own routes, or so they redirect you if you haven't actually submitted the form?

public/sitemap.xml
nfasulo: Fixed.

nfasulo:
@kalilsn 
> What's the concern with churn? If we're regenerating the same sitemap, is there something bad that will happen?

Nothing bad will happen. I just didn't want folks getting annoyed with git  saying they changed a random file if they `npm run build` locally.

> Finally, I restarted the build on jenkins to see if I could get it to deploy to a review bucket, but it failed at the build step:
> 
> ```
> /application/node_modules/sitemap/lib/sitemap.js:40
> 
>       throw new err.NoURLProtocolError();
> 
>       ^
> 
> NoURLProtocolError: Protocol is required
> ```

The problem was that I was using `paths.publicUrl` to get the protocol and hostname. The build job sets the env var `PUBLIC_URL`, which overrides the value I was expecting from `paths.publicUrl`. I will get it directly from `package.json` instead.

nfasulo:
I created department-of-veterans-affairs/vets-contrib/issues/1386 for building the sitemap using the existing code to determine the routes.

public/sitemap.xml
```diff
@@ -10,4 +10,20 @@
...
+<url> <loc>https://developer.va.gov/explore/facilities/docs/claims</loc> </url>
+<url> <loc>https://developer.va.gov/explore/facilities/docs/loan_guaranty</loc> </url>
+<url> <loc>https://developer.va.gov/explore/health/docs/benefits</loc> </url>
+<url> <loc>https://developer.va.gov/explore/health/docs/appeals</loc> </url>
+<url> <loc>https://developer.va.gov/explore/health/docs/claims</loc> </url>
+<url> <loc>https://developer.va.gov/explore/health/docs/loan_guaranty</loc> </url>
+<url> <loc>https://developer.va.gov/explore/verification/docs/benefits</loc> </url>
+<url> <loc>https://developer.va.gov/explore/verification/docs/appeals</loc> </url>
+<url> <loc>https://developer.va.gov/explore/verification/docs/claims</loc> </url>
+<url> <loc>https://developer.va.gov/explore/verification/docs/loan_guaranty</loc> </url>
```
kalilsn: While these URLs will all get you a page, a bunch of those permutations don't make sense (`/explore/facilities/docs/appeals`, for example).

nfasulo: Oops. I had it only pointing to benefits from an early test. Fixed.

nfasulo:
@kalilsn Ok, I'll take building the sitemap out of the build script for now. Would it be acceptable to dump the sitemap directly into the `build` directory or do I need to give it another home?

kalilsn:
I realize now that my last message was contradictory. If it's static, put it in `public`! Once it's being generated by a script, then it shouldn't be checked in at all.

PR #70, opened 2019-02-26
Changing Claims Title and Description

bastosmichael:
Quick nomenclature and description change for Claims


src/apiDefs.ts
```diff
@@ -29,9 +29,9 @@ export const apiDefs : IApiCategories = {
     apiKey: true,
     apis: [
       {
-        name: 'Benefits Intake',
-        shortDescription: 'Submit claims',
-        urlFragment: 'benefits',
+        name: 'Claims',
```
juliaelman: s Claims/Benefits Claims

src/apiDefs.ts
```diff
@@ -29,9 +29,9 @@ export const apiDefs : IApiCategories = {
...
+        shortDescription: 'Submit and Track claims',
```
juliaelman: s Track/track

src/apiDefs.ts
bastosmichael: Done

src/apiDefs.ts
bastosmichael: Done

PR #72, opened 2019-02-28
Style search for mobile

nfasulo:
Closes department-of-veterans-affairs/vets-contrib/issues/1392.

@mjlumetta was nice enough to style the full-size search box, this pull request is for the mobile version. I also plan to remove the feature flag for search so that once it is merged into master search will be shipped to prod.


src/components/Search.scss
```diff
@@ -1,9 +1,17 @@
+@import "../base";
+
 .vadp-search-form {
   display: flex;
   flex-flow: row nowrap;
   justify-content: center;
   align-items: center;
 
+  @media screen and (max-width: $large-screen - 1) {
```
mjlumetta: in many media query situations i've seen, we seem to use the pattern of making mobile the default style and using `min-width` queries with the variable (unmodified) to override styles that need to be different at larger screen sizes. this also allows us to use the `@include media($large-screen)` mixin for most of these scenarios. should we stick to that pattern here?

mjlumetta:
i think that maybe the search box should take up more like 100% of the width, or certainly a wider width in some scenarios. @juliaelman may want to chime in here and can answer questions about it. but see the mobile design here: https://www.figma.com/file/SGQZyWGE44uYM4lI1twVsw/IA-Updates-(Jan-2019)?node-id=21%3A19 

the other thing about that is that it looks very weird on the medium to medium-large type of screens:

mobilesearchwiderscreen

lastly, not sure whether the background should be white when it appears between the header and the hero - as in, when there is a hero? i'm removing that on most pages (ie, documentation pages) but not the home page. clarification on that seems like a good idea?

looks good overall though! should be good to go once these questions get cleared up

src/components/NavBar.tsx
```diff
@@ -94,38 +87,41 @@ export class NavBar extends React.Component<INavBarProps, INavBarState> {
...
+                  onMouseLeave={this.toggleDefaultNavLink.bind(this, true)}>
                   Support
                 </a>
               </li>
             </ul>
           </div>
         </nav>
+        <MediaQuery query={UNDER_LARGE_SCREEN_QUERY}>
+          <Search />
+        </MediaQuery>
```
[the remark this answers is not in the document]
nfasulo: I was thinking it would be simplest to keep  both uses of the `Search` in the same file, but if it makes more sense to take the mobile search out of `NavBar` that's fine by me. Are you considering moving it `App`?

src/components/Search.scss
nfasulo: This makes sense. I'm certainly not a styling guru. :slightly_smiling_face:

nfasulo:
@mjlumetta @juliaelman I agree that the getting the mobile search box to take up more width makes sense. In terms of what to do with medium to medium-large screens: I do think it would be best the search style to switch at the same width that the rest of the NavBar changes happen, but maybe that could happen at a narrower width.

src/components/NavBar.tsx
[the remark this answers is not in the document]
mjlumetta: yeah, i think either would be fine, just curious if you'd considered it? i think that makes more sense visually, but i also see the point of keeping them together, and it's kind of a push between those two for me. Julia's answer about the layout of the bar next to a page hero may affect this choice, but other than that, either are good by me, whichever you feel is better

juliaelman:
Great job @nfasulo! A few notes:

1. Add more space between the VCL banner and the search box, see Screenshot 1 below.
1. Height for the search box should change from 44px to 42px.
1. 🤔 can we add placeholder text in the search box that says "Search developer.va.gov"?
1. I am wondering if we can re-think the 950px to mobile layout later on, rather than address now. It's one of those items were I feel that perfect is the enemy of done, as it does not look terrible. I'd like to be able to come up with an optimal solution.

## Screenshot 1
screen shot 2019-03-04 at 2 54 42 pm

nfasulo:
@juliaelman I shifted both the search box and Get Started button slightly. I think they are better in like with the logo now. I can go further if you wish.

I wonder if the search box should be slightly smaller than 42px. Maybe I'm too wild, but I kinda like 38px.

I'm definitely fine with leaving the mobile break as is for now.

src/components/NavBar.scss
```diff
@@ -52,6 +52,7 @@
       display: flex;
       flex-flow: row nowrap;
       align-items: center;
+      margin-top: 0.3rem;
```
mjlumetta: i'd suggest that this might be more appropriate on the `.usa-nav-secondary` selector one level above (in the markup)?

src/components/Search.scss
```diff
@@ -1,8 +1,19 @@
...
   justify-content: center;
   align-items: center;
+  background-color: #FFFFFF;
+  padding-top: 3.0rem;
+  padding-bottom: 3.0rem;
+
+  @include media($large-screen) {
+    background-color: inherit;
+    padding-top: inherit;
+    padding-bottom: inherit;
```
mjlumetta: i don't think padding should be inherited from the parent; it's preferable to simply set it to 0, imo. background-color is fine to inherit on principle, but it doesn't make sense to me to specify that in an exported component like this, especially since the background-color isn't set on the parent element of the desktop search form. think it's fine to omit that one.

src/components/Search.scss
```diff
@@ -19,14 +30,20 @@
 
   input.usagov-search-autocomplete {
     font-size: 16px;
-    width: 240px;
+    height: 42px;
+    max-width: 80%;
```
mjlumetta: do we ever expect it to be under 80% for some reason? if we just want it to be 80% all the time, can we make this a `width` instead of a `max-width`?

juliaelman:
@nfasulo pulled down the changes. Looks like the spacing between the VCL banner and the search box hasn't been implemented. Please let me know if you want to pair on that!

src/components/NavBar.scss
nfasulo: I moved it into the `.usa-nav-secondary` and then realized I was actually fighting with the `top` value in there so I adjusted that and removed `margin-top`.

src/components/Search.scss
nfasulo: I have changed these to explicit values. The background-color is needed so that you can't see white around the rounded edges of the search button.

src/components/Search.scss
nfasulo: If I don't set the `max-width` it is defaulting to 460px, which looks awkwardly narrow on medium screens.

mjlumetta: i still prefer to setting `width: 80%;` and `max-width: none` to setting the width with `max-width` as it's more expressive of the actual intent, but that's small enough that it's up to you

nfasulo: Yep, that is definitely more clear!

PR #75, opened 2019-03-04
Separate API category pages and remove hero on documentation pages

mjlumetta:
## Overview

PR for ticket 1157: https://github.com/department-of-veterans-affairs/vets-contrib/issues/1157

This branch contains changes to separate API category pages. Previously, the pages for Benefits, Facilities, Health, and Veteran Verification were all rendered on the same page and scrolled to match the correct route. Now, each API category gets its own page. There is a new documentation overview page at `/explore` used by the Overview link at the top of side nav. This page and each API category page now have their links styled in cards and laid out horizontally across the top of the page. The hero on the documentation pages has also been removed.

## Testing

* Verify that the API category pages are rendered alone on their respective routes and that the documentation page exists. That means you should check the following routes:
    * `/explore`
    * `/explore/benefits`
    * `/explore/health`
    * `/explore/facilities`
    * `/explore/verification`
* Content should be left-aligned and appropriately spaced on all pages.
* Verify the layout on both mobile and desktop, as well as responsive sizing at intermediate sizes.
* Headings on documentation pages should be in [Bitter font](https://fonts.google.com/specimen/Bitter).
* The documentation for specific API pages (ie Benefits Intake, Appeals Status, Claims Status for Benefits) should remain the same as before, with the exception of left-alignment and header font.
* The cards on the documentation overview page and API category pages should wrap neatly as the screen shrinks.

### Desktop

* The documentation overview should appear like this: 

documentationoverviewdesktop

* For each API category page, the page should have a header, subheader, and then the cards for the specific APIs. Below that is the content previously included in the API category page. 

benefitsapipagedesktop

* When the screen width is wider than the page content `max-width` of 1200px, there should be 3 tiles in one row.

### Mobile

* The documentation overview should appear like this: 

documentationoverviewmobile

* For each API category page, the page should have a header, subheader, and then the cards for the specific APIs. Because it is mobile, the cards will be stacked vertically and take up the whole width. Below that is the content previously included in the API category page. 

benefitsapipagemobile

* At 767px and below, the cards on any given page should take the full width minus padding. That is, any time the cards wrap so that there is only one card per row, they should take up the whole space.

## Notes

* Any suggestions for unit or e2e testing welcome.
* The saving of the old version of `ApiPage` as `ApiPageWithScroll` is intentional; see the AC on the ticket.


mjlumetta:
@juliaelman are these changes only supposed to happen on the Explore pages (for now, anyway)? if not, should there be a new ticket for some of that?

juliaelman:
@mjlumetta good question! It should be across the board, all pages. Not sure we need to have a new ticket, as it should be pretty straight forward to incorporate into this PR.

PR #76, opened 2019-03-06
Add Updates page

nfasulo:
For department-of-veterans-affairs/vets-contrib/issues/1456.

I added a super simple Updates page that just displays the contents of the placeholder file `content/release-notes/03-12-2019-release-notes.mdx` that Jeff created. Obviously we still have a way to go until we have an actual MVP, but I think this satisfies department-of-veterans-affairs/vets-contrib/issues/1456 while not having to convert the release notes from `mdx` to `md` so that we can to the file in GitHub.


jeff2d2:
I made one small change to remove the date from the file name and associated code.

nfasulo:
Need to get Updates to stay underlined after routing to it.

jeff2d2:
LGTM. Confirmed that the Updates tab now shows as active after it's been clicked.
Agree that we should not deploy this until the first set of release notes are written and merged into place.

jeff2d2:
Updates the `release-notes.mdx` file, although it is still not yet finalized.

PR #77, opened 2019-03-06
More lenient linting during development.

dvogel:
I've split the default tslint config and the config we run during CI tests in order to avoid developers having to change rules while developing. This opens up the possibility of pushing a PR for a set of changes that will fail the linter during CI but not locally. To me that seems annoying but less annoying than the status quo.  The `no-unused-variable` and `prefer-const` rules are very noisy in development if you comment out something temporarily. Both of these issues usually crop up during debugging when you comment out the only call of a function or the only time a variable is reassigned. The `no-console` rule should be obvious :)

### DIFF
--- a/tslint.ci.json
+++ b/tslint.ci.json
@@ -1,0 +1,15 @@
+{
+  "extends": ["./tslint.json"],
+  "rules": {
+    "no-unused-variable": {
+      "severity": "error"
+    },
+    "no-console": {
+      "severity": "error"
+    },
+    "prefer-const": {
+      "severity": "error"
+    }
+  }
+}
+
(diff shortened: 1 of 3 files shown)


tslint.ci.json
```diff
@@ -0,0 +1,15 @@
...
+  "extends": ["./tslint.json"],
+  "rules": {
+    "no-unused-variable": {
+      "severity": "error"
+    },
+    "no-console": {
+      "severity": "error"
+    },
+    "prefer-const": {
+      "severity": "true"
```
kalilsn:
```suggestion
      "severity": "error"
```

kalilsn: Looks like you fixed this before I finished the review 🙃

PR #78, opened 2019-03-06
Move OAuth documentation to side nav.

dvogel:
These changes complete this ticket: https://github.com/department-of-veterans-affairs/vets-contrib/issues/1163

This incorporates the OAuth page into the side-nav as the `Authorization` entry under both Health and Veteran Verification. That side-nav sub-menu is created based on the TOC of the OAuth page.


src/apiDefs.ts
```diff
@@ -100,24 +107,28 @@ export const apiDefs : IApiCategories = {
...
       {
         name: 'Veteran Confirmation',
+        openApiDocUrl: `${process.env.REACT_APP_VETSGOV_SWAGGER_API}/services/veteran_verification/docs/v0/status`,
         shortDescription: "Get confirmation of a Veteran's status",
         urlFragment: 'veteran_confirmation',
         vaInternalOnly: false,
       },
       {
         name: 'Address Validation',
+        openApiDocUrl: `${process.env.REACT_APP_VETSGOV_SWAGGER_API}/services/address_validation/docs/v1/api`,
```
dvogel: The `openApiDocUrl` changes are technically unrelated to the ticket. I made these changes to so I could see the structure of the code for the doc pages better while determining what changes I needed to make.

src/components/ApiPage.tsx
```diff
@@ -6,11 +6,12 @@ import { RouteComponentProps } from 'react-router';
 import { apiDefs, IApiDescription } from '../apiDefs';
 import { IApiNameParam } from '../types';
 import ApiCard from './ApiCard';
+import AuthorizationCard from './AuthorizationCard';
 
 export class ApiPage extends React.Component<RouteComponentProps<IApiNameParam>, {}> {
   public render() {
     const { apiCategoryKey } = this.props.match.params;
-    const { apis, name: categoryName, overview, longDescription: introText } = apiDefs[apiCategoryKey];
+    const { apiKey, apis, name: categoryName, overview, longDescription: introText } = apiDefs[apiCategoryKey];
```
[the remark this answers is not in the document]
dvogel: I had the same thought you did but it seems easy enough to fix when it's time to cross that bridge.

dvogel:
> Do you understand why clicking the sidebar links on the oauth page take you to the top before scrolling to the correct element?

I don't know, but I suspect it has something to do with the `NavHashLink` component we're using because it happens for me on the OAuth page in production.

src/containers/ExploreDocs.tsx
```diff
@@ -91,7 +133,10 @@ export class ExploreDocs extends React.Component<RouteComponentProps<IApiNamePar
           <div className="usa-width-two-thirds">
             <Route exact={true} path="/explore/" component={DocumentationOverview} />
             <Route exact={true} path="/explore/:apiCategoryKey" component={ApiPage} />
-            <Route exact={true} path="/explore/:apiCategoryKey/docs/:apiName" component={Explore} />
+            <Switch>
+              <Route exact={true} path="/explore/:apiCategoryKey/docs/authorization" component={OAuth} />
```
mjlumetta: i don't know that this really works with the dynamic routing because the `apiCategoryKey` prop is optional in the rendered component. even so, some API categories shouldn't show Authorization, correct?

anyway, see these URLs i visited:

bananasauthorization

benefitsauthorization

facilitiesauthorization

src/containers/ExploreDocs.tsx
```diff
@@ -40,17 +42,57 @@ function SideNavApiEntry(apiCategoryKey: string, api: IApiDescription) {
...
+      </li>
+    );
+}
+
 function SideNavCategoryEntry(currentUrl: string, apiCategoryKey: string, apiCategory: IApiCategory) {
   const subNavLinks = apiCategory.apis.map(api => {
     return SideNavApiEntry(apiCategoryKey, api);
   });
 
+  const authorizationEntry = apiCategory.apiKey ? null : OAuthSideNavEntry(apiCategoryKey, apiCategory);
```
mjlumetta: i don't understand why this boolean in the `IApiCategory` interface is named `apiKey`. can you clarify? i see what's supposed to happen here, but i don't know the logic behind it.

src/containers/ExploreDocs.tsx
```diff
@@ -40,17 +42,57 @@ function SideNavApiEntry(apiCategoryKey: string, api: IApiDescription) {
...
+    return props.to === location.hash;
+  };
+  const toWithoutHash = props.to.replace(/^#/, '');
+  let id = `hash-link`;
+  if (props.idSlug != null) {
+    id = `${id}-${props.idSlug}`;
+  }
+  id = `${id}-${toWithoutHash}`;
+  return (
+      <NavHashLink activeClassName="usa-current" id={id} isActive={activeCheck} to={props.to}>
```
mjlumetta: shift lines 56-59 a tab left

src/containers/Explore.tsx
```diff
@@ -21,68 +22,26 @@ const mapStateToProps = ({ routing }: IRootState) => {
...
-                docs = this.renderIndex();
+      let docsDom: JSX.Element | null = null;
+      if (this.props.match.params.apiName != null) {
+        const api = lookupApi(this.props.match.params.apiName);
+        if (api != null) {
+          docsDom = (
+                <Flag name={`hosted_apis.${api.urlFragment}`}>
+                  <SwaggerDocs url={api.openApiDocUrl} />
+                </Flag>
+              );
```
mjlumetta: this is a nice refactor

src/apiDefs.ts
```diff
@@ -138,3 +149,29 @@ export const apiCategoryOrder: string[] = [
...
+export function withApiDescription(urlFragment: string, fn: (apiDesc: IApiDescription) => any): any {
+  const api = lookupApi(urlFragment);
+  if (api == null) {
+    return null;
+  }
+
+  return fn(api);
+}
+
+export function lookupApi(urlFragment: string): IApiDescription | null {
```
mjlumetta: since the category key is in scope when you call this, can you include the category key as a parameter and eliminate the first loop?

src/containers/ExploreDocs.tsx
dvogel: Good question. I don't much like the name. In the future this should evolve into an enum instead of a boolean. That would make each value describe it's own purpose for existing. The mapping will be like so:

| `apiKey`        | `authModel` |
|------------------|-----------------|
| true               | `apikey`        |
| false              | `oauth  `       |

src/apiDefs.ts
dvogel: Technically, now, yes. My main goal in introducing this function though is to decouple the code in various components from the shape of the `apiDefs` structure. As the site evolves, that will become more difficult though.

For example, in the near term, we'll probably move the `address_validation` API from the `veteran_verification` category into the `facilities` category (or maybe a new category). Once we do that we'll want to use the categorization in apiDefs to structure the rendered navigation but, in order to preserve old URLs, we'll probably want to be able to render a given API regardless of the category in the URL (which is more aesthetic than structural AFAICT).

Hopefully further in the future this data will be in a proper database or pulled from the kong API itself. At that point we'll definitely want a clearly defined module to fetch data and return it with a stable(-ish) set of types exported. This is a baby step in that direction.

src/containers/ExploreDocs.tsx
dvogel: I'll get in touch on slack to discuss this one. Probably easier that way.

mjlumetta:
@dvogel you can make the target branch `redesign-integration-target` now, as the 1157 branch has been merged

src/containers/ExploreDocs.tsx
dvogel: We discussed this on slack. We settled on the new `ApiKeyAuth` component. It currently serves as sort of dead-end to avoid rendering `OAuth` for API categories that don't use OAuth. In the future it will be a good home to add content related to API key auth.

dvogel:
> @dvogel you can make the target branch `redesign-integration-target` now, as the 1157 branch has been merged

Done!

dvogel:
@juliaelman These changes are ready for you to review on Monday.

dvogel:
@juliaelman I fixed the side nav and the internal-only tag. Mind reviewing again?

PR #80, opened 2019-03-13
Add sitemap to webpack build

nfasulo:
Closes department-of-veterans-affairs/vets-contrib/issues/1386.

I have created a webpack plugin that uses the routes and api definitions from developer-portal to generate a sitemap.

Work still to do:
- Tap into feature flags to exclude APIs at that aren't in production


nfasulo:
Hmm, pulling `focusedRoutes` out of `App.tsx` appears to have broken the OAuth e2e test.

SitemapBuilderWebpackPlugin.js
```diff
@@ -0,0 +1,146 @@
...
+    const paramsConfig = {
+      '/explore/:apiCategoryKey/docs/:apiName': sitemapData.apiCategoryOrder.map(apiCategory => {
+        return {
+          apiCategoryKey: apiCategory,
+          apiName: sitemapData.apiDefs[apiCategory].apis.map((api) => api.urlFragment),
+        };
+      }),
+      '/explore/:apiCategoryKey?': [''].concat(sitemapData.apiCategoryOrder).map(apiCategory => {
+        return { 'apiCategoryKey?': apiCategory };
+      }),
```
dvogel: What problem is this `[''].concat` bit solving?

src/Routes.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+
+import { RouteComponentProps } from 'react-router';
+import { Redirect, Route } from 'react-router-dom';
+
+import { apiCategoryOrder, apiDefs } from './apiDefs';
+import { PageContent } from './components/PageContent';
+import { ApplyForm, ApplySuccess, BetaPage, BetaSuccess, ExploreDocs, Home, OAuth, RoutedContent } from './containers';
+
+export function focusedRoutes(props: RouteComponentProps<void>) {
```
dvogel: I've never understood why this is called "focused" routes. Does anyone know? If not, maybe we could change it to something that describes it's role in the site? `topLevelRoutes` seems to make sense to me.

SitemapBuilderWebpackPlugin.js
```diff
@@ -0,0 +1,146 @@
...
+      rules: [/index.html|\/explore\/terms-of-service|\/applied|\/beta-success/],
+    };
+
+    const paramsConfig = {
+      '/explore/:apiCategoryKey/docs/:apiName': sitemapData.apiCategoryOrder.map(apiCategory => {
+        return {
+          apiCategoryKey: apiCategory,
+          apiName: sitemapData.apiDefs[apiCategory].apis.map((api) => api.urlFragment),
+        };
+      }),
```
dvogel: Eventually this data will probably be large enough that we'll want to avoid embedding it in the source of the developer portal. At that time, it will be difficult to reach into `apiDefs` directly like this. It's probably worth considering how we could export a function from `apiDefs.ts` that provided the API required to access this data without understanding the internal data format. A super naive approach would be to use `lookupCategory(apiCategory)` here. That would be slower but probably not too bad since the data is still small. Eventually we'll probably want a `allApiCategories` function or something like that.

SitemapBuilderWebpackPlugin.js
nfasulo: Because react-router-sitemap doesn't handle `?` as an indication of an optional path, when it creates paths based on this route it doesn't include a default one for `/explore`. Adding an empty string to the beginning of the parameter list fixed this.

It's a hack, but I wanted to get the test passing. I plan on adding `?` handling to react-router-sitemap when I work on department-of-veterans-affairs/vets-contrib/issues/1553.

SitemapBuilderWebpackPlugin.js
nfasulo: Is this anywhere near what you have in mind?
```
export function lookupCategory(apiCategory: string, properties?: string | string[]) {
  return apiDefs[apiCategory].apis.map((api) => {
    let mappedApi = {};

    if (properties) {
      if (typeof properties === 'string') {
        if (api.hasOwnProperty(properties)) {
          mappedApi = api[properties];
        }
      }

      for (const property of properties) {
        if (api.hasOwnProperty(property)) {
          mappedApi[property] = api[property];
        }
      }
    } else {
      mappedApi = api;
    }

    return mappedApi;
  });
}

nfasulo: Never mind, I see the function `lookupApiCategory` in `redesign-integration-target` branch is the naive implementation you are referencing.

dvogel: > Never mind, I see the function `lookupApiCategory` in `redesign-integration-target` branch is the naive implementation you are referencing.

Yeah, we should be merging that branch into master today.

SitemapBuilderWebpackPlugin.js
nfasulo: Because react-router-sitemap doesn't handle `?` as an indication of an optional path, when it creates paths based on this route it doesn't include a default one for `/explore`. Adding an empty string to the beginning of the parameter list fixed this.

It's a hack, but I wanted to get the test passing. I plan on adding `?` handling to react-router-sitemap when I work on department-of-veterans-affairs/vets-contrib/issues/1553.

dvogel: That makes sense now that you explain it but I'm not sure I would have ever understood without an explanation. Could it work with an entry above `'/explore/': {}`?

dvogel:
It occurs to me now that I understand the direction of this PR because we discussed it prior to me seeing the code. I think it would be super helpful to add a comment at the top of `SitemapBuilderWebpackPlugin.js` explaining the factors that pushed us toward using a webpack plugin approach. We don't want anyone to needlessly rediscover that :smile:

SitemapBuilderWebpackPlugin.js
nfasulo: I tried adding an extra line like `'/explore/:apiCategoryKey?': { 'apiCategoryKey?': '' },` but it doesn't add a path. I think the way to clean up both `[''].concat` and the regex on the sitemap itself is to fork react-router-sitemap and add the ability to handle the `?` syntax for optional routes.

dvogel: I wasn't sure it would work. Thanks for trying it!

nfasulo: No problem. I just pushed up using my fork of react-router-sitemap and it definitely makes for easier to read code.

PR #81, opened 2019-03-13
Merge redesign-integration-target into master

mjlumetta:
## Overview

PR for [vets-contrib#1526](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1526)

Merges changes from the following tickets:
* [1392 - search functionality](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1392)
* [1157 - reorganize and restyle documentation pages](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1157)
* [1163 - move OAuth nav](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1163)

## Testing

Regression testing on the above tickets


src/containers/Explore.scss
```diff
@@ -1,262 +1,308 @@
...
 .Explore {
-    #api-documentation {
-        margin-top: 0;
+  padding-top: 5rem;
+  padding-bottom: 3rem;
+
+  h1,
+  h2,
+  h3 {
+    font-family: $font-bitter;
```
mjlumetta: @dvogel did you intentionally add this back or is it a merge error or sometime? also tagging @juliaelman

dvogel: After you merged your branch and I rebased onto the new tip of master, this section had a pretty significant conflict. I definitely intended to use this styling because it matched the mocks. Did I miss a discussion about the fate of this font?

mjlumetta: yeah, unfortunately - it is in the mocks but Julia got back to me saying that it should be Source Sans Pro. i think if you remove this block you should get the expected behavior.

and yeah, as a note - i realize there are probably things shared between components here, but in general, it seems like this file should be split up into a more component-by-component basis, if possible. i thought about doing it earlier on in the redesign changes but thought better of it because i thought i might be missing context or that it might be too big a change. in any case, think it would be good to look for an opportunity to do that when it comes up again.

as a result of removing the thing, you can remove the `$font-bitter` definition in `src/base.scss` and the stylesheet link in `public/index.html`, also

dvogel:
> Excited for this to go out! Before that happens, I'm wondering if we can fix two interactions that feel like bugs to me.
> 
>     * Can we add a height to the nav so that the content doesn't get pushed down when you hover over one of the nav items?
> 
>     * Now that the various documentation sections have their own pages, on small screens  if you scroll down on one, then use the menu to navigate to another item, you'll end up in the middle of that page. Can we scroll to the top whenever you navigate to a new page?

I also see the nav height change. I hadn't noticed it before because it works as expected on the documentation page. The home page and the apply page both suffer though.

I don't see the scroll issue. What specific pages does this happen on?

mjlumetta:
fyi @dvogel @kalilsn the height change issue is fixed for me locally, but like Drew i feel i could use a bit more insight into the scroll issue. i'll try to reproduce

dvogel:
> fyi @dvogel @kalilsn the height change issue is fixed for me locally, but like Drew i feel i could use a bit more insight into the scroll issue. i'll try to reproduce

I just figured it out. I first read "on small screens" as "on mobile" but that seems fine. If you stay in desktop mode but reduce the window height down to about 760px you can see it when transitioning from Benefits overview to Facilities overview.

kalilsn:
I did mean on "mobile" but I only tested on my laptop at about 400px. Sounds like you were able to reproduce it but let me know if you need me to demonstrate

dvogel:
LGTM

mjlumetta:
@hubgitian verbally approved this PR from the UX side during our standup, so merging now

mjlumetta:
Closing and merging with clean git history [on this PR](https://github.com/department-of-veterans-affairs/developer-portal/pull/85)

PR #83, opened 2019-03-14
Fix accessibility errors and re-enable a11y linter with level ERROR 

mjlumetta:
## Overview

PR for [vets-contrib#1044](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1044)

Fixes all outstanding accessibility errors in the developer portal and updates the `react-a11y-*` rules in `tslint.json` to have a severity of "error".

## Testing

* Verify that every `react-a11y` rule in `tslint.json` has a severity of "error"
    * Optionally, you can break some [accessibility rules](https://www.npmjs.com/package/tslint-microsoft-contrib) in 1+ components and verify that you get errors when you run `npm run lint`. Obviously be sure not to commit your changes if you do (though if it works, you won't be able to!!).
* Run `npm run lint` and be sure that you don't see any errors

## Acceptance Criteria
- [ ] `npm run lint` doesn't report any accessibility errors
- [ ] accessibility linting issues are _errors_ (not warnings)


src/containers/ApplyForm.tsx
```diff
@@ -135,20 +141,26 @@ class ApplyForm extends React.Component<IApplyProps> {
...
                   <input
                     type="checkbox"
+                    role="checkbox"
                     id="health"
                     name="health"
                     checked={apis.health}
-                    onChange={props.toggleHealth} />
+                    aria-checked={apis.health ? 'true' : 'false'}
+                    onChange={props.toggleHealth}
+                    placeholder="Checkbox for VA Health API" />
```
kalilsn: The placeholder attribute doesn't really make sense for a checkbox. Screenreaders ignore it entirely, and it's not displayed visually.

tslint.json
```diff
@@ -27,55 +27,55 @@
       "severity": "off"
     },
     "react-a11y-anchors": {
-      "severity": "warning"
+      "severity": "error"
```
kalilsn: Is it useful to you to have these explicitly defined in this file? If you just delete all of the `react-a11y-` rules from here they'll inherit their severity (which is already error).

src/containers/Beta.tsx
```diff
@@ -126,11 +126,11 @@ export class BetaPage extends React.Component<{}, IBetaPageState> {
...
-            <input id="company" maxLength={40} name="company" hidden={true} defaultValue="Intake Form" />
-            <input id="lead_source" maxLength={40} name="lead_source" hidden={true} defaultValue="Intake Form" />
-            <input hidden={true} name="recordType" id="recordType" defaultValue="0123500000012eiAAA" />
+            <input hidden={true} name="oid" defaultValue={idMappings[env].oid} /> {/* tslint:disable-line:react-a11y-input-elements */}
+            <input hidden={true} name="retURL" defaultValue={process.env.REACT_APP_SALESFORCE_RETURN_URL} /> {/* tslint:disable-line:react-a11y-input-elements */}
+            <input id="company" maxLength={40} name="company" hidden={true} defaultValue="Intake Form" /> {/* tslint:disable-line:react-a11y-input-elements */}
+            <input id="lead_source" maxLength={40} name="lead_source" hidden={true} defaultValue="Intake Form" /> {/* tslint:disable-line:react-a11y-input-elements */}
+            <input hidden={true} name="recordType" id="recordType" defaultValue="0123500000012eiAAA" /> {/* tslint:disable-line:react-a11y-input-elements */}
```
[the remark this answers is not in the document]
mjlumetta: yes, correct

src/containers/ApplyForm.tsx
```diff
@@ -135,20 +141,26 @@ class ApplyForm extends React.Component<IApplyProps> {
                 <div className="form-checkbox">
                   <input
                     type="checkbox"
+                    role="checkbox"
                     id="health"
                     name="health"
                     checked={apis.health}
-                    onChange={props.toggleHealth} />
+                    aria-checked={apis.health ? 'true' : 'false'}
```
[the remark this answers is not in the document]
mjlumetta: they say they need `aria-checked`... it said `role="checkbox"` was implied and stopped complaining about that when i removed them, so idk. do you think we should drop that rule?

src/containers/ApplyForm.tsx
mjlumetta: hmm okay - must be some weird thing with rule. i can investigate to see if it's configurable and otherwise turn it off/to a warning?

tslint.json
mjlumetta: ah got it - can you or @dvogel point me to where that's defined? i'll look, but if you happen to know

mjlumetta: update: so you can delete the **severity prop** but not the **rule itself**. if you do the former, it reports errors as expected. if you do the latter, it reports no errors

dvogel: I'm fine with these rules being here. I like the fact that we can omit the severity to represent the fact that we want the rule evaluated and we've made no explicit decision re: the severity. The reason I like that is because rules with an explicit severity then, by contrast, also tell me that we've considered the severity and made an explicit decision.

mjlumetta: cool, sounds good. that seems like what we have to do, so that's perfect

src/containers/ApplyForm.tsx
mjlumetta: okay - it looks like this error was found and it should be fixed in version 6.1.0. so it seems like an upgrade to the `tslint-microsoft-contrib` package should fix it. will upgrade and remove these changes

src/containers/ApplyForm.tsx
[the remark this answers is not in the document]
mjlumetta: looks like this issue was [reported here](https://github.com/Microsoft/tslint-microsoft-contrib/issues/409)... trying to find an associated PR

mjlumetta:
okay, so here are the outstanding issues on this PR:

* `react-a11y-input-elements` requires a placeholder or value on inputs, including of type `checkbox`. that error was [identified](https://github.com/Microsoft/tslint-microsoft-contrib/issues/749) and [fixed in a PR](https://github.com/Microsoft/tslint-microsoft-contrib/pull/754), but the fix still _requires a value_ for types where the placeholder is irrelevant. 
* `react-a11y-role-has-required-aria-props` requires `aria-checked` for elements of type `checkbox` that it says have an implicit `role="checkbox"`. because it has the native HTML `type="checkbox"`, `aria-checked` should be redundant. [this error](https://github.com/Microsoft/tslint-microsoft-contrib/issues/409) has been reported but never fixed, to my knowledge.

so that leaves us with the question of what we'd like to do in the short-term around these particular rules. here are some possible options:

1. set each of these rules to have a severity of "warning" again
2. disable the rules line by line or file by file
3. eliminate the rules from `tslint.json` permanently or temporarily (for the latter, comment out and leave a note)

in addition, if it seems like something that could be done relatively quickly and it's okay from a program standpoint, i'd be interested in trying a PR to `tslint-microsoft-contrib` to fix the latter problem for the long term, if it has not been fixed. 

anyway curious for your thoughts, @dvogel @kalilsn (when Kalil's back lol). in the meantime i'll continue learning more so i can offer a bit more informed opinions.

src/containers/Beta.tsx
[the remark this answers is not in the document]
mjlumetta: these look like they could be fixed by changing `defaultValue` to `value`. since the input is hidden, seems like it shouldn't change anything in terms of behavior?

dvogel:
I think we should disable both rules by setting the severity to `off`. I believe `tslint.json` allows extra keys, so you should be able to add a comment in an object key explaining why it's off.

```
    "react-a11y-role-has-required-aria-props": {
      "severity": "off",
      "REASON": "THIS IS BROKEN AS OF 6.1.0. IT REQUIRES ARIA ATTRS ON ELEMENTS WITH IMPLICIT ROLES."
    }
```

dvogel:
I also think we should consider fixing tslint, if it's obvious how to fix it.

src/containers/Beta.tsx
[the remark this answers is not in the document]
dvogel: I think that's the best path forward. AFAIK the `defaultValue` attribute isn't meant to be set via markup. That's generally an attribute exposed via javascript in order to determine whether `value` has been changed since the page was loaded.

mjlumetta:
@dvogel sounds good. i'll set both to off and push those changes. 

i do think that i have a pretty good idea of how to go about fixing both issues in `tslint-microsoft-contrib` - or at least starting, anyway, as there might be other issues i'm not anticipating. if we're willing to allocate a ticket to it, i'd be happy and interested to work on that.

one question that is still bugging me is about the errors on the `react-a11y-input-elements` rule because that one _did_ get a PR to fix the issue as originally identified. the thing is that the fix still requires that the `value` attribute be present for checkboxes, which doesn't seem to make sense given that we're interested in the boolean `checked` attribute only (seems like it would have to be present for radio buttons, unsure about file inputs). technically this may be a different issue than the one reported, but the example in the OP in the issue report would fail the rule like ours does. so i think i'd like to ask about that on that thread, but i'm also not 100% sure that my understanding is correct. i haven't been able to find anything that explicitly validates that from an accessibility perspective. Drew, what do you think?

mjlumetta:
@1Copenut would you pls be able to weigh in on the question from my above comment?

> one question that is still bugging me is about the errors on the react-a11y-input-elements rule because that one did get a PR to fix the issue as originally identified. the thing is that the fix still requires that the value attribute be present for checkboxes, which doesn't seem to make sense given that we're interested in the boolean checked attribute only (seems like it would have to be present for radio buttons, unsure about file inputs). technically this may be a different issue than the one reported, but the example in the OP in the issue report would fail the rule like ours does. so i think i'd like to ask about that on that thread, but i'm also not 100% sure that my understanding is correct. i haven't been able to find anything that explicitly validates that from an accessibility perspective. Drew, what do you think?

1Copenut:
@mjlumetta Sure thing, happy to look into these.

Tracing the errors back to open Typescript tickets, it appears there are errors being flagged where there should not be. First and foremost, any input that gets rendered as a native HTML checkbox or radio buttons should not need ARA attributes. Those are facades for custom elements, so I'll use that as my base for advisements. I'll need a bit to get up to speed on what each one's asking, and will add comments where I can provide value.

I appreciate the heads up.

src/containers/ApplyForm.tsx
1Copenut: I concur. Radio buttons, checkboxes, and file inputs should definitely not require a `placeholder`.

1Copenut:
@mjlumetta So I did some digging on checkboxes, and their `value` attribute. MDN as per usual, was a wealth of information: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value

> ...only checkboxes which are currently checked are submitted to the server, and the reported value is the value of the `value` attribute. If the value is not otherwise specified, it is the string `on` by default

This by itself does not suggest the `value` attribute is required. I also ran a snippet through the W3C HTML validator, with multiple checkboxes sans value attribute. This snippet below passed validation with no errors or warnings.

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <title>Test</title>
</head>
<body>
  <fieldset>
    <legend>Choose your interests</legend>
    <div>
      <input id="coding" name="interest[]" type="checkbox"> <label for="coding">Coding</label>
    </div>
    <div>
      <input id="music" name="interest[]" type="checkbox"> <label for="music">Music</label>
    </div>
  </fieldset>
</body>
</html>
```

To my way of thinking, the `value` attribute should not be required. I ran that snippet through an anonymous Code Pen and evaluated it with aXe. No errors were thrown. I would need to do more testing to be a useful "expert" voice on the subject, but as a daily practitioner, I think this should be an optional attribute in the React a11y rulesets.

mjlumetta:
@1Copenut thanks so much for the input! based on your answer, i feel confident enough at least to ask about in on the package's github page. if it's wrong for some reason, i'm sure they'll point that out there. thanks so much

src/containers/ApplyForm.tsx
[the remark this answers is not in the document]
1Copenut: These really should not need the `role="checkbox"` or `aria-checked` attributes as hard requirements. The rendered HTML gives an implicit role of checkbox in the Accessibility API, and checkboxes have the Boolean checked attribute. I think these are great checks for teams building custom components on non-semantic or repurposed HTML, but they're overly prescriptive here.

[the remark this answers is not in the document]
mjlumetta: this is an outdated version, sorry. Kalil caught that as well, and we determined that since the issue was in the `tslint-microsoft-contrib` package, we should do it the correct way like you said. so if you check the current version, these changes should be undone. thanks for looking out, Trevor

PR #84, opened 2019-03-15
Updated Path to Production link and /go-live contents

jeff2d2:
This PR is based on [this thread in Slack](https://lighthouseva.slack.com/archives/CE79R5SR1/p1552604538142600).

It updates the link that exists at the bottom of the developer portal homepage `3. Request production access` to point to https://developer.va.gov/go-live instead of https://github.com/department-of-veterans-affairs/vets-api-clients/blob/master/Path-to-Production.md#production-api-access.

It also updates the contents of `/go-live` to match what is located on the above Github link. A misspelling was fixed, and a couple backticks were added around URIs to make them `stand out`, but otherwise, it's an exact match.

- [x] Tested locally


jeff2d2:
Need to fix the URLs and email addresses not showing as links.

jeff2d2:
> Need to fix the URLs and email addresses not showing as links.

Fixed.

src/content/goLive.mdx
```diff
@@ -1,26 +1,17 @@
...
-The VA API uses token-based authentication. Clients must provide a token with each HTTP request. This token is issued by the VA. To receive a developer token to access this API in a test environment, please email Alex.Yale-Loehr@va.gov.
-
-## Sample Applications 
-
-Sample code and documentation is available at [https://github.com/department-of-veterans-affairs/vets-api-clients](https://github.com/department-of-veterans-affairs/vets-api-clients). 
+The VA API uses token-based authentication. Clients must provide a token with each HTTPS request. This token is issued by the VA. 
 
 ## Development API Access
-During the alpha phase of this project, if you want access to a development API token, email Alex.Yale-Loehr@va.gov.  We will be creating a self service capability for developer tokens in the near future. 
+To receive a developer token to access an API in a test environment, go to <https://developer.va.gov/apply>.  If you are having trouble using <https://developer.va.gov/apply> or have questions, email [api@va.gov](mailto:api@va.gov) and we will get back to you quickly!
```
kalilsn:
```suggestion
To receive a developer token to access an API in a test environment, [submit an application](/apply).  If you are having trouble with the form or have questions, email [api@va.gov](mailto:api@va.gov) and we will get back to you quickly!
```

src/content/goLive.mdx
```diff
@@ -33,19 +24,21 @@ First, you must provide some basic information including:
 Next, the following criteria needs to be met and verified by the API team. 
 
 - You are a US based company
-- You agree to the Terms of Service
+- You agree to the [Terms of Service](https://developer.va.gov/explore/terms-of-service)
```
kalilsn:
```suggestion
- You agree to the [Terms of Service](/explore/terms-of-service)
```

src/containers/Home.tsx
```diff
@@ -87,7 +87,7 @@ class Home extends React.Component {
                   <ol>
                     {request}
                     <li><a href="https://github.com/department-of-veterans-affairs/vets-api-clients/">Build your app</a></li>
```
kalilsn: Do we still want this link if we're moving to the developer portal being the sole source of this information? If we do, I think we should alter the old path to production doc there to link to https://developer.va.gov/go-live so they can't get out of sync

kalilsn:
I just remembered, the old path to production doc is also in the overview for the benefits and health sections, and should probably be removed:

https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/content/apiDocs/benefitsOverview.mdx#overview

https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/content/apiDocs/healthOverview.mdx#path-to-production-for-health-api-clients

src/content/goLive.mdx
jeff2d2: @kalilsn got with Alex about this. His suggestion: "To receive a developer token to access an API in a test environment, [submit a request](/apply).  If you are having trouble or have questions, email [api@va.gov](mailto:api@va.gov) and we will get back to you quickly!"

Would you be OK with that instead?

src/containers/Home.tsx
jeff2d2: From [the talks here](https://lighthouseva.slack.com/archives/CAZF880HW/p1552939423010000?thread_ts=1552927119.006700&cid=CAZF880HW), this repo isn't going away. We should move instructional type content like this off of it, and only use it for app samples and other code reference material.

jeff2d2:
@juliaelman I'd like to hear your input/thoughts/concerns around this. One concern might be that the `/go-live` page that was modified looks a little different from the rest of the portal, although it does match some of the other text-only pages like `/terms-of-service` (i.e. there is no left nav).

PR #86, opened 2019-03-18
Enable axe-core e2e accessibility tests on the developer portal

mjlumetta:
## Overview

PR for [this ticket](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1488)

Turns on the tests in `src/base.accessibility.ts` so that we have automated accessibility testing and fixes errors discovered by those tests, with the exception of one on the `/apply` page which is located in `formation-react` (TODO: make PR to that repository and make follow-up ticket for it - unless that work should go in the same ticket).

## Testing

- [ ] Run the tests by running `npm run test:accessibility`
- [ ] Verify the code and visual presentation changes make sense


src/containers/ExploreDocs.tsx
```diff
@@ -104,7 +104,7 @@ export function SideNav({ match: { url } } : RouteComponentProps<IApiNameParam>)
   const navLinks = apiCategoryOrder.map((key: string) => SideNavCategoryEntry(url, key, apiDefs[key]));
 
   return (
-    <ul role="navigation" aria-label="API Docs Side Nav" className="usa-sidenav-list">
+    <ul className="usa-sidenav-list">
```
dvogel: Does the test output explain why it wants us to remove `role="navigation"`? I thought that was an important hint for screen readers. Is it implied by something else?

mjlumetta: it does not. if you look around the file (and specifically at `ExploreDocs.render()`) you'll see that it's been moved up a level, not removed. i _think_ - and this is kind of hard, because i wasn't able to explicitly validate this - that it's saying it's inappropriate to have that role on a `ul`. but can't quite tell for sure.

dvogel: I hadn't noticed it being added to `ExploreDocs.render`. Having it one level up the tree makes sense to me. Thanks for explaining that.

PR #88, opened 2019-03-19
Remove legacy Path to Production instructions from the Benefits APIs page

jeff2d2:
https://github.com/department-of-veterans-affairs/developer-portal/pull/84 sets up the /go-live page as the SSOT for the path to production. On the Benefits APIs page remains the legacy instructions leftover from the Alpha phase of the program.

This PR removes those old instructions. We can always add a link to /go-live back into to this page if desired, once that PR is merged and /go-live is fixed.


dhall:
Big fan of removing the duplication.. I think the `Overview` section for each section needs some work (long term) for consistency and to help the user signup or get help. tagging @juliaelman and @jwoodman5 for visibility.

PR #92, opened 2019-03-21
fixes all terms-of-service links that were still pointing through /explore

jeff2d2:
This PR fixes all legacy `/explore/terms-of-service` links and changes them to `/terms-of-service`.

There is still code that was supposed to auto-redirect, but it wasn't working in `Routes.tsx` (`<Route path="/explore/terms-of-service" render={props1 => <Redirect to="/terms-of-service" />} />`) but I left that alone for now in case it should be fixed instead of deleted.

### DIFF
--- a/SitemapBuilderWebpackPlugin.js
+++ b/SitemapBuilderWebpackPlugin.js
@@ -114,7 +114,7 @@ class SitemapBuilderPlugin {
 
     const pathFilter = {
       isValid: false,
-      rules: [/index.html|\/terms-of-service|\/applied|\/beta-success/],
+      rules: [/index.html|\/explore\/terms-of-service|\/applied|\/beta-success/],
     };
 
     const paramsConfig = {
--- a/SitemapBuilderWebpackPlugin.js
+++ b/SitemapBuilderWebpackPlugin.js
@@ -114,7 +114,7 @@ class SitemapBuilderPlugin {
 
     const pathFilter = {
       isValid: false,
-      rules: [/index.html|\/explore\/terms-of-service|\/applied|\/beta-success/],
+      rules: [/index.html|\/terms-of-service|\/applied|\/beta-success/],
     };
 
     const paramsConfig = {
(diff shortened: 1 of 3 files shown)


SitemapBuilderWebpackPlugin.js
```diff
@@ -114,7 +114,7 @@ class SitemapBuilderPlugin {
 
     const pathFilter = {
       isValid: false,
-      rules: [/index.html|\/explore\/terms-of-service|\/applied|\/beta-success/],
+      rules: [/index.html|\/terms-of-service|\/applied|\/beta-success/],
```
nfasulo: This rule removes matching paths from the sitemap, so I don't think we want this change to be made.

jeff2d2: Fixed. Put it back the way it was.

src/content/apiDocs/benefitsOverview.mdx
```diff
@@ -37,7 +37,7 @@ First, you must provide some basic information including:
 Next, the following criteria needs to be met and verified by the API team.
 
 * You are a US based company
-* You agree to the [Terms of Service](/#/explore/terms-of-service)
+* You agree to the [Terms of Service](/#/terms-of-service)
```
mjlumetta:
```suggestion
* You agree to the [Terms of Service](/terms-of-service)
```

remove the #

src/content/apiDocs/benefitsOverview.mdx
```diff
@@ -1,6 +1,6 @@
 ## Overview
 
-Our goal is to make easier and more secure for VSOs and other organizations to interact with the VA on behalf of Veterans. When you’re ready for production, review these [Terms of Service](/#/explore/terms-of-service) and then request it [here](https://github.com/department-of-veterans-affairs/vets-api-clients/issues/new). Individual API documentation is available above. 
+Our goal is to make easier and more secure for VSOs and other organizations to interact with the VA on behalf of Veterans. When you’re ready for production, review these [Terms of Service](/#/terms-of-service) and then request it [here](https://github.com/department-of-veterans-affairs/vets-api-clients/issues/new). Individual API documentation is available above.
```
mjlumetta:
```suggestion
Our goal is to make easier and more secure for VSOs and other organizations to interact with the VA on behalf of Veterans. When you’re ready for production, review these [Terms of Service](/terms-of-service) and then request it [here](https://github.com/department-of-veterans-affairs/vets-api-clients/issues/new). Individual API documentation is available above. 
```

remove the #

src/content/apiDocs/benefitsOverview.mdx
jeff2d2: fixed

src/content/apiDocs/benefitsOverview.mdx
jeff2d2: fixed

PR #93, opened 2019-03-21
Wrap routes in switch

kalilsn:
Turns out the router renders all the matching routes, not just the first or most specific one, if you don't have `<Switch>` around them.

### DIFF
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { RouteComponentProps } from 'react-router';
+import { RouteComponentProps, Switch } from 'react-router';
 import { Redirect, Route } from 'react-router-dom';
 
 import { apiCategoryOrder, apiDefs } from './apiDefs';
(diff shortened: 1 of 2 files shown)


src/Routes.tsx
```diff
@@ -14,7 +14,7 @@ export function topLevelRoutes(props: RouteComponentProps<void>) {
         <Route exact={true} path="/index.html" component={Home} />
 
         {/* Legacy routes that we want to maintain: */}
-        <Route path="/explore/terms-of-service" render={props1 => <Redirect to="/terms-of-service" />} />
+        <Route path="/explore/terms-of-service" render={() => <Redirect to="/terms-of-service" />} />
```
kalilsn: I didn't actually mean to commit this, but I do think it's slightly clearer!

dvogel: Does it work? I remember trying something similar and being annoyed at an error. I don't remember if it was from tsc, tslint, or something in the browser console though.

src/components/PageContent.tsx
```diff
@@ -23,7 +23,9 @@ export class PageContent extends React.Component<IPageContentProps, {}> {
         return (
             <div className="PageContent">
               <div className="content-main" ref={(loader) => (this.loader = loader)} tabIndex={-1}>
-                {this.props.children}
+                <Switch>
+                  {this.props.children}
+                </Switch>
```
dvogel: Should this `Switch` be here or wrapping the `Routes` themself? What if someone uses `PageContent` with children that don't include a `Route`. Does that work?

src/Routes.tsx
kalilsn: It's working fine for me locally! Maybe the error was from trying to use `props`, which throws this tslint error: `Shadowed name: 'props' (no-shadowed-variable)`?

src/components/PageContent.tsx
kalilsn: Good point. I think you're right that it should be in Routes.tsx

PR #91, opened 2019-03-21
Update disclaimer banner

juliaelman:
This PR updates the disclaimer banner to:

- Only be on the homepage
- Update design to be in sync with new UI
- Add feedback link to GitHub issue templates on [vets-api-clients](https://github.com/department-of-veterans-affairs/vets-api-clients)
- Small fix to header spacing

See screencast below to see updated design + link to feedback form.

Screen Shot 2019-03-21 at 9 44 57 AM

PR #94, opened 2019-03-22
Update homepage layout with new IA

juliaelman:
⚠️ **The PR is downstream from https://github.com/department-of-veterans-affairs/developer-portal/pull/91 and should NOT be merged in until that has been merged.**

This pull request is for small updates to the layout of the homepage of the developer portal. They are based on the IA changes that happened in https://github.com/department-of-veterans-affairs/developer-portal/pull/85.

See screencast below for reference:

bLpDzNN5xE


juliaelman:
@jeff2d2 updated here caebc31

Screenshot FYR

Screen Shot 2019-03-22 at 1 57 28 PM

src/containers/Home.tsx
```diff
@@ -1,82 +1,89 @@
...
-                <div className="usa-grid">
-                  <div className="usa-width-one-half">
-                    <div className="video-wrapper">
-                      <iframe src="https://www.youtube.com/embed/at69_Uz05vA" frameBorder="0" allowFullScreen={true} />
-                    </div>
                   </div>
                 </div>
               </section>
+              <div className="usa-grid">
+                <div className="ruled-stars">{null}</div>
```
kalilsn: what's the `{null}`? was one of our linters complaining about the div not having content?

src/containers/Home.scss
```diff
@@ -63,11 +51,48 @@
...
-        top: 20px;
-        left: 0;
-        width: 100%;
-        height: 100%;
+          display: block;
+          position: absolute;
+          top: 20px;
+          left: 0;
+          width: 100%;
+          height: 100%;
```
kalilsn: looks like this block got overindented a bit

src/containers/Home.tsx
```diff
@@ -1,82 +1,89 @@
...
-              <section role="region" aria-label="API Top-Level List" className="usa-section api-section">
+
+              <section role="region" aria-label="Page Hero" className="usa-hero">
                 <div className="usa-grid">
-                  <h2>Explore VA APIs</h2>
-                  <p className="usa-font-lead">
-                    We want to empower our partners to build innovative solutions for Veterans. So we designed a framework to deliver a modern API development experience and rolled out our first set of standards-based APIs.
-                  </p>
+                  <div className="usa-hero-callout">
+                    <h1 className="">Put VA Data to Work</h1>
```
kalilsn: I'm guessing this `className` can just be deleted!

src/base.scss
```diff
@@ -15,6 +15,11 @@ $medium-large-screen: 768px;
 $medium-screen: 640px;
 $large-screen: 951px;
 
+.usa-grid {
+  max-width: 1200px !important;
+  margin-left: 0 !important;
```
kalilsn: can you add a comment about why these are `!important`?

src/containers/Home.tsx
juliaelman: Yep.

src/containers/Home.scss
juliaelman: Fixed here 900320d

src/containers/Home.tsx
juliaelman: Fixed here ae58a03

src/base.scss
juliaelman: Because it is an important override of the USWDS.

PR #95, opened 2019-03-22
Add Release Notes

nfasulo:
This pull request is for creating an MVP for department-of-veterans-affairs/vets-contrib/issues/1521.

At this point we are focusing on the Facilities API to figure out what we what Release Notes to look like. All other APIs are hidden and will be revealed in on the Release Notes page once they have content to be displayed.


nfasulo:
I'm trying to think through a strategy for getting the Release Notes merged into master. Should I remove the ability to navigate to Release Notes for APIs other than Facilities until they have a bit more in the way of Release Notes than the initial release? I'd rather not, but if that's what it will take to get to production . . .

src/apiDefs.ts
```diff
@@ -20,6 +26,7 @@ export interface IApiCategory {
   readonly overview: React.StatelessComponent;
   readonly shortDescription: string;
   readonly longDescription: string;
+  readonly releaseNotes?: React.StatelessComponent;
```
edmkitty: Not familiar with React-y things: what does the `?` do?

src/components/ApiPageReleaseNotes.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const { apis, releaseNotes } = apiDefs[apiCategoryKey];
+
+    const headerProps = {
+      halo: 'Release Notes',
+      header: apiDefs[apiCategoryKey].properName, 
+    };
+
+    let cardSection;
+
+    if (apis.length > 1) {
```
edmkitty: What exactly is apis here? Is it ok that an `apis` with a length of 1 is excluded?

src/components/ApiPageReleaseNotes.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import { apiDefs, IApiDescription } from '../apiDefs';
+import PageHeader from '../components/PageHeader';
+import { IApiNameParam } from '../types';
+import ApiCard from './ApiCard';
+
+export class ApiPageReleaseNotes extends React.Component<RouteComponentProps<IApiNameParam>, {}> {
+
+  public render() {
+    const { apiCategoryKey } = this.props.match.params;
```
edmkitty: Maybe add some comments to explain what these consts should look like?

src/apiDefs.ts
nfasulo: `?` makes property optional in TypeScript. Once we have release notes for all of the APIs we can make `releaseNotes` required, but right now I am using it as a feature flag so that we only display APIs on the Release Notes page that have release notes.

src/components/ApiPageReleaseNotes.tsx
nfasulo: This comes from `apiDefs.tsx` which describes the structure of our API offerings. `apiCategory` objects are used to define our top-level APIs (Benefits, Facilities, Health, Verification). Each of these then has an `api` array that lists the individual APIs included in that category.

The point of the code above is to only show cards on Release Notes pages for API categories that include more than one API. It's hard to see one the site right now, because only Facilities has a Release Notes page, but this allows Benefits and Verification to have sub-navigation like on the Documentation pages.

src/components/ApiPageReleaseNotes.tsx
nfasulo: `apiCategoryKey` is a param for this route that is passed in using React Router. This happens in a couple of other places in the app and I just copied it. :smile:

Eallan919:
Quick note- Under Facilities- at the bottom should there be a bullet point starting at... `VA_Facilities API 

August 13, 2018 | v0.0.0.188
Launch of the VA Facilities API
VA_Facilities API endpoint for external API consumers. Supports json/geojson results and a bulk "get all" endpoint in geojson and csv format.

src/containers/ReleaseNotes.scss
```diff
@@ -0,0 +1,313 @@
...
+  h2 {
+    font-size: 2rem;
+    font-weight: 700;
+    margin-bottom: 2rem !important;
+  }
+
+  h3 {
+    font-size: 1.6rem;
+    font-weight: 700;
+    margin-bottom: 2rem !important;
```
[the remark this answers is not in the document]
kalilsn: Also I think we should avoid using `!important` whenever possible! Using a more specific selector than `h3` will hopefully eliminate the need for `!important`, but if not, I would appreciate a comment in the code explaining why it was necessary.

src/containers/ReleaseNotesOverview.tsx
```diff
@@ -0,0 +1,30 @@
...
+            const cardUrl = this.props.parent + '/' + apiCategoryKey;
+            let apiCard;
+            if (apiDefs[apiCategoryKey].releaseNotes) {
+              apiCard = (
+                <ApiCard name={name} description={shortDescription} key={apiCategoryKey} vaInternalOnly={false}
+                  url={cardUrl} />
+              );
+            }
+            return apiCard;
+          })}
```
kalilsn: Could we move this logic up a little? Either into its own function or before the `return` in the `render` function? Github (and my editor ☹️) does a bad job with syntax highlighting with this much JS

src/components/NavBar.tsx
```diff
@@ -95,6 +95,14 @@ export class NavBar extends React.Component<INavBarProps, INavBarState> {
                 </MediaQuery>
                 {this.state.visibleSubNavs.documentation && this.renderDocumentationSubNav()}
               </li>
+              <li className="main-nav-item">
+                <NavLink to="/release-notes" className="usa-nav-link" activeClassName="default-nav-link"
+                  isActive={this.checkActiveNavLink}
+                  onMouseEnter={this.toggleDefaultNavLink.bind(this, false)}
+                  onMouseLeave={this.toggleDefaultNavLink.bind(this, true)}>
```
kalilsn: Do you mind explaining what this does? It looks like the `Documentation` link doesn't have these `onMouse` event handlers and I wonder if that's a bug? If you navigate to the release notes page and then mouse over `Support` and `Documentation`, two different things happen. Maybe caused by this?

src/content/releaseNotes/facilities.mdx
```diff
@@ -0,0 +1,46 @@
+<div id="facilities" />
```
kalilsn: I'm guessing these `div`s are for navigating between different APIs within the same category? If that's the case then I would like to move these outside of the markdown files, so we can still keep the same split between presentation and content.

Also, we have a plugin that creates an ID for markdown headers automatically, so you probably don't need these empty divs if you add a `## Facilities` line

src/content/releaseNotes/facilities.mdx
```diff
@@ -0,0 +1,46 @@
...
+**Standardize closed days to "Closed" in Facilities API [#2802](https://github.com/department-of-veterans-affairs/vets-api/pull/2802)**
+- Changes any casing of "closed" and "-" to "Closed" for a consistent convention in the Facilities API
+
+---
+
+## **January 16, 2019** | v0.0.0.287
+
+**Fix facilities api bugs [#2735](https://github.com/department-of-veterans-affairs/vets-api/pull/2735)**
+- Allows users to pass both a list of IDs and a lat/long in the same request.
+- Fixes typo in the existing rake task, [view specific change](https://github.com/department-of-veterans-affairs/vets-api/pull/2735/files#diff-a0253f0a6c07463fb11a513df82f21d2).
```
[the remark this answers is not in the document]
nfasulo: Finding and massaging content for the release notes has certainly been trying. I get your point about describing the how fixing the typo affects consumers instead of the fixing of the typo itself.

I definitely like the idea of including samples as structure of requests/responses change.

docs/release-notes.md
```diff
@@ -0,0 +1,38 @@
...
+
+This interactive page let's you see how the markdown syntax in mdx files are converted into html: 
+[Markdown Component Loader - Online Editor](https://ticky.github.io/markdown-component-loader/repl.html)
+
+## Format of a Release Note
+
+```
+## **Month DD, YYYY** | v<api version>.<vets-api release number>
+
+**PR Name or other high-level description [#<PR number>](link to PR in GitHub)**
```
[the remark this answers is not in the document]
nfasulo: Hey, you read the documentation! :smile:

But seriously, I'm sure folks other than me have a better idea of how to pull these release notes together than I do. We discussed monthly roll-ups of changes, but it seemed like because deployments happen just about everyday it would make sense to try to follow that. Definitely open to other methods that make more sense.

src/containers/ReleaseNotes.scss
```diff
@@ -0,0 +1,313 @@
...
+    top: 20px;
+    display: none;
+
+    @include media($medium-screen) {
+      display: block;
+      position: sticky;
+    }
+  }
+
+  .swagger-ui {
```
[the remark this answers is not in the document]
nfasulo: You're not missing anything, I totally copied `Explore.scss`. Drew suggested extracting out a shared set of styles for Documentation, Release Notes, and What's New, which I think makes sense.

src/containers/ReleaseNotes.scss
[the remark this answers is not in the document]
nfasulo: This override was my solution to get each release note's header to be large and partially bold/partially regular weight without using HTML in the markdown.

e.g. make 
## **November 20, 2018** | v0.0.0.255
like below but with larger text
**November 20, 2018** | v0.0.0.255

Luckily, if we're not going to include versions in the release notes header this hack it is unnecessary.

src/components/index.ts
```diff
@@ -5,6 +5,8 @@ export * from './Footer';
 export * from './PageContent';
 export * from './PageHero';
 export * from './ApiPage';
+export * from './ApiPageReleaseNotes';
+export * from './PageHeader';
```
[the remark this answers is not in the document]
nfasulo: Yep, I haven't changed the Documentation pages yet, but I made sure that this PageHeader will handle not having halo text gracefully.

src/components/PageHeader.scss
```diff
@@ -0,0 +1,19 @@
...
+.header {
+  // padding-bottom: 40px;
+
+  // h1 {
+  //   margin-top: 0;
+  // }
+  
+  // h2 {
+  //   font-weight: bold;
+  // }
```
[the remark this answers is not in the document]
nfasulo: Just some stolen styling from `OAuth.scss` I need to remove.

src/containers/ReleaseNotesOverview.tsx
nfasulo: Yep, that would help with readability.

src/components/NavBar.tsx
nfasulo: The `onMouseEnter` and `onMouseLeave` allow the active class to be removed from/added to the actual active link while the mouse is hovering over the a different link in the NavBar, giving us a sense of action when the NavBar exploring with the mouse. I think we do want them added to `Documentation` as well, assuming that is how we want to show users that clicking on a different link will redirect them to that page. Good catch (of course!).

src/content/releaseNotes/facilities.mdx
nfasulo: Yep, I can get rid of those tags, that plugin gives me exactly what I need for navigating between different APIs on the same page. (Of course it's a moot point if we break these on to separate pages.)

nfasulo:
Thanks for the code review, @kalilsn! I appreciate the time you have taken in not only looking directly at the code, but in thinking through our larger goals.

I don't have a problem with hiding this behind a feature flag while discussion continues. We knew that this could prove to be a feature that aroused strong opinions which is one reason why we are trying to start with a small piece and get it out for people to see as quickly as possible.

I share your concern about multiple APIs being on the same page becoming problematic as the amount of release notes increases and this is something the has team discussed. We came to the conclusion that using a flatter navigation made sense until release notes matures and we have more content to share. But that was a week ago, so maybe it's just growing up quickly. :slightly_smiling_face: 

And of course a meeting of folks who are interested in the direction of release notes sounds fine to me.

src/containers/ReleaseNotes.scss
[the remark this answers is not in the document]
nfasulo: Sorry, my comment above is referring specifically to the styling at the bottom of the file: 
```
.api-release-notes {
  h2 {
    font-weight: normal !important;
  }
}
```
All of the `h#` styling above was copied from `Explore.scss`, so it won't go away if drop the version from each release note's header.

dhall:
@kalilsn @nfasulo I have a strong desire to ship things to production as they have value for our users. I think that this definitely has value for users in it's initial (mvp) form. I'd rather not do a feature flag, but 'make it work, then make it work better'. I'm up for us chatting about the next steps, of course, but lets deliver this. 👍 👍

dvogel:
I think all of this feedback is a little more compatible than it first seems. If I can paraphrase and summarize a few thoughts I see here:

We know we want release notes. We're not entirely sure that the final form will be. We should ship an MVP to start learning which aspects work for consumers and which ones don't. We also need to ensure that the MVP is indeed viable. It doesn't help to receive feedback that users like an aspect of the MVP if that aspect isn't sustainable. The suggestion for removing the PR link and the vets-api release numbers are focusing on making this sustainable. As we make this work better we'll expand it to other APIs. In short, less is more. Making it more minimal makes it more viable and more valuable by boosting the signal to noise ratio and ensuring consistency as we improve it.

We seem to all agree that we want to move forward with this PR in some way. I think the question of how best to do that comes down to when we plan to make improvements. Feature flagging this allows us to merge this PR sooner, then figuring out a different way to communicate versioning, etc and following up with a smaller PR in that direction. The alternative seems to be updating the release notes in this PR to remove the release numbers and the PR links in order to make it truly minimal, in which case going all the way to production won't carry with it any reputational risk of removing info from release notes or changing the versioning scheme, which could look bad to anyone wondering whether they should trust us.

dvogel:
> @kalilsn @nfasulo I have a strong desire to ship things to production as they have value for our users. I think that this definitely has value for users in it's initial (mvp) form. I'd rather not do a feature flag, but 'make it work, then make it work better'. I'm up for us chatting about the next steps, of course, but lets deliver this. +1 +1

@dhall to catch you up when you come back from retreat, this conversation [migrated over to slack](https://lighthouseva.slack.com/archives/CDNA4T352/p1553802524004700?thread_ts=1553695399.251500&cid=CDNA4T352). We came up with a plan to ship most of this. Quoting that slack thread:

> @kalilsn: After discussing with Julia we've decided to remove the version numbers for now, but otherwise leave the contents of the release notes for the facilities API as they are. We may add in version numbers in the future, pending research findings.
> @nfasulo: This is a good summary of the call. Thanks, Kalil. I'll get a push a commit with the versions stripped out shortly.

src/containers/ReleaseNotes.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { RouteComponentProps } from 'react-router';
+import { NavLink, Route } from 'react-router-dom';
+
+import { ApiPageReleaseNotes } from '../components';
+import { IApiNameParam } from '../types';
+import ReleaseNotesOverview from './ReleaseNotesOverview';
+
+import { apiCategoryOrder, apiDefs, IApiCategory, IApiDescription } from '../apiDefs';
+
+import './Explore.scss';
```
kalilsn: I think this is fine for now, but could you throw in a todo comment here indicating that the explore styles that are going to be applied across a bunch of pages should be extracted to a more generic file?

kalilsn: Actually scratch that, I'm realizing this is almost entirely for the side nav, so we should just create a ticket for extracting a full side nav component

PR #96, opened 2019-03-22
Report JS errors to sentry.vetsgov-internal

kalilsn:
This is the developer portal side of https://github.com/department-of-veterans-affairs/vets-contrib/issues/880

See https://github.com/department-of-veterans-affairs/devops/pull/4382 for adding the route to the rev proxy. The request looks like a POST to `https://dev-developer.va.gov/sentry/api/15/store/?sentry_key=7d7f3fde3b744e7099dc5e8cc65b01f7&sentry_version=7`

It's likely that there will be more customization to do to make these reports as useful as possible, but it's hard to do that before we deploy this. Open to any suggestions now if you have some sentry experience, otherwise we can iterate on this later.


src/index.tsx
```diff
@@ -9,10 +10,17 @@ import store from './store';
...
+
 ReactDOM.render(
-    <Provider store={store}>
-        <App />
-    </Provider>,
-    document.getElementById('root') as HTMLElement,
+  <Provider store={store}>
+    <App />
+  </Provider>,
+  document.getElementById('root') as HTMLElement,
```
dvogel: It seems like you're still iterating on this, so feel free to defer this, but should we wrap this `ReactDOM.render` in a try-catch block (instead of or in addition to the `componentDidCatch` approach) in order to capture exceptions coming out of react itself, `Provider`, etc?

kalilsn:
Switched to try/catch per @dvogel's suggestion. That will catch a slightly larger set of errors, and it appears that sentry gets a little more information that way:

try/catch: http://sentry.vetsgov-internal/vets-gov/developer-portal-dev/issues/59334/ (gets "breadcrumbs")
error boundary: http://sentry.vetsgov-internal/vets-gov/developer-portal-dev/issues/59333/

PR #98, opened 2019-03-22
Correct OAuth Docs

dvogel:
While supporting an API consumer I discovered the OAuth docs omitted [the `offline_access` scope](https://developer.okta.com/docs/api/resources/oidc/#response-properties-2). This is required to get the refresh token, which is included in some example responses.

While working through the documented steps, I also realized that some of the examples were inaccurate, so I did my best to correct those as well. The hybrid flow examples were incorrect but they felt duplicative of the other examples, so rather than correct them I removed them.


src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -190,6 +168,7 @@ Scopes define the API endpoint your application is allowed to access. The VA API
 Existing Scopes are:
 
 * `profile` - granted by default allows access to a Veteran's first and last name and email.
+* `offline_access` - this scope causes the authorization server to provide an refresh token when the access token is requests.
```
patrickvinograd:
```suggestion
* `offline_access` - this scope causes the authorization server to provide a refresh token when the access token is requested.
```

PR #99, opened 2019-03-25
Update OAuth docs

jeff2d2:
This PR addresses [this issue](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1475).

Old copy: "This anonymous function asynchronously retrieves the client information then starts the express server. Now you should be able to run node index.js and navigate to http://localhost:8080/ in your browser. You should see 'Hello World!' on your screen."

New copy: "This anonymous function asynchronously retrieves the client information then starts the express server. Now you should be able to run node index.js and navigate to \<yourRedirectURL\> in your browser (⚠️Important: Be sure to use the redirect URL that you registered with). You should see 'Hello World!' on your screen."


jeff2d2:
@dvogel I'm realizing now that this isn't correct. `http://localhost:8080` is located 6 more times in the OAuth documentation. Should I change all of those to \<yourRedirectURL\> with the same warning as I already did for this one instance?

dvogel:
> @dvogel I'm realizing now that this isn't correct. `http://localhost:8080` is located 6 more times in the OAuth documentation. Should I change all of those to <yourRedirectURL> with the same warning as I already did for this one instance?

I think that would be best, yeah. There's even more URLs to fix in the API example requests and responses. For example:

```
&redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb
```

dvogel:
In the examples, the path is already part of the redirect URL they register with us. For example the default redirect URL is `http://localhost:8080/auth/cb` so `<yourRedirectURL>/auth/cb` would end up including the path twice. 

I can't wait until we can have people login to the portal because then the portal will be able to lookup the redirect URL the user registered with and insert it into this documentation. That would clear a lot of this up ;) Thanks for cleaning this up in the meantime!

src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -31,8 +31,9 @@ Initiate the OpenID Connect authorization by directing Veterans using your appli
 
 Example Authorization URL:
 ```plaintext
-https://dev-api.va.gov/oauth2/authorization?client_id=0oa1c01m77heEXUZt2p7&redirect_uri=http://localhost:8080/implicit/callback&response_type=id_token token&response_mode=fragment&state=1AOQK33KIfH2g0ADHvU1oWAb7xQY7p6qWnUFiG1ffcUdrbCY1DBAZ3NffrjaoBGQ&nonce=o5jYpLSe29RBHBsn5iAnMKYpYw2Iw9XRBweacc001hRo5xxJEbHuniEbhuxHfVZy&scope=openid profile email veteran_status.read
+https://dev-api.va.gov/oauth2/authorization?client_id=0oa1c01m77heEXUZt2p7&redirect_uri=<yourRedirectURL>/implicit/callback&response_type=id_token token&response_mode=fragment&state=1AOQK33KIfH2g0ADHvU1oWAb7xQY7p6qWnUFiG1ffcUdrbCY1DBAZ3NffrjaoBGQ&nonce=o5jYpLSe29RBHBsn5iAnMKYpYw2Iw9XRBweacc001hRo5xxJEbHuniEbhuxHfVZy&scope=openid profile email veteran_status.read
 ```
+(⚠️Be sure to use the redirect URL that you registered with)
```
nfasulo: I didn't realize you could use that warning symbol in markdown. Nice.
```suggestion
(⚠️Be sure to replace `<yourRedirectURL>` with the redirect URL that you registered with)
```
A minor quibble, but I prefer "Be sure to replace `<yourRedirectURL>` with the redirect URL that you registered with". I think it just makes it a little easier to find what needs to be replaced in more verbose situations like this one.

jeff2d2: Went with "Be sure to replace <yourRedirectURL> with the redirect URL that you provided during registration" to avoid the double "with".

kalilsn:
I agree that changing these makes sense for most of the instances of localhost:8080, but towards the end, we're explicitly going through a tutorial which starts by directing you to sign up with that specific redirect URL. Shouldn't we leave those as they are for clarity?

dvogel:
> I agree that changing these makes sense for most of the instances of localhost:8080, but towards the end, we're explicitly going through a tutorial which starts by directing you to sign up with that specific redirect URL. Shouldn't we leave those as they are for clarity?

Seems like a horse apiece to me with the current arrangement. I think we should split that entire section out into the readme for the sample app and then link from the oauth directions to the sample app. That would make the split much clearer and it would help us make the oauth docs more concise.

PR #100, opened 2019-03-25
Add What's New page

nfasulo:
This pull request is for sharing work in progress for department-of-veterans-affairs/vets-contrib/issues/1520.


Eallan919:
Looks fantastic! I checked that all the links worked to the media, etc.  Great job, Nick!

src/components/NavBar.tsx
```diff
@@ -95,6 +95,14 @@ export class NavBar extends React.Component<INavBarProps, INavBarState> {
...
                 {this.state.visibleSubNavs.documentation && this.renderDocumentationSubNav()}
               </li>
+              <li className="main-nav-item">
+                <NavLink to="/whats-new" className="usa-nav-link" activeClassName="default-nav-link"
+                  isActive={this.checkActiveNavLink}
+                  onMouseEnter={this.toggleDefaultNavLink.bind(this, false)}
+                  onMouseLeave={this.toggleDefaultNavLink.bind(this, true)}>
+                  What's New
+                </NavLink>
+              </li>
```
dvogel: Should this be a fancy quote? Do we have a policy on that for the dev portal? /cc @juliaelman

src/containers/WhatsNew.scss
```diff
@@ -0,0 +1,307 @@
+@import '../base';
+
+.WhatsNew {
```
dvogel: A lot of the styling here looks like it was copied from the Explore page. Is there an opportunity to extract a shared set of styles? E.g. the side nav that's based on the USWDS side nav styling.

nfasulo: I stole all of the styling for this and the New Release page from `Explore.scss` (New Release has one additional piece of styling). I would be happy to use one combined stylesheet, I only made separate pages because I didn't want to rely on `Explore.scss` and then have a change only meant for the Documentation page have a negative impact on one of the other pages.

src/components/NavBar.tsx
nfasulo: Oh man, it took me way too long to realize that you meant `What&rsquo;s New`!

src/containers/WhatsNew.scss
nfasulo: I got rid of the separate stylesheet and am just importing at `Explore.scss` for now. @kalilsn pointed out that there it is really only the SideNav that needs the styling and that it would make sense to turn that into it's own component. There are actually empty files for `SideNav.scss` and `SideNav.tsx` in `component` already.

src/components/NavBar.tsx
nfasulo: I went ahead and used the fancy quote. @juliaelman @dvogel

PR #101, opened 2019-03-26
Fix slow jenkins builds

kalilsn:
We've been having some jenkins builds take up to an hour to complete, or just fail altogether, because a mystery firewall is blocking the port used by the `git` protocol. This PR ensures that we install our forked `react-router-sitemap` over `https`. It also removes an unused dependency from the dockerfile.

See https://lighthouseva.slack.com/archives/CDNA4T352/p1553542908233700 and https://lighthouseva.slack.com/archives/CAZF880HW/p1553628651025600 for context on this instance of the issue.

See https://dsva.slack.com/archives/C37M86Y8G/p1550258183033500 for a previous instance, with a similarly unsatisfying conclusion.


Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN groupadd --gid 504 jenkins \
 
 USER jenkins
 
-RUN npm install -g npm s3-cli
+RUN npm install -g npm
```
dvogel: Why remove s3-cli?

kalilsn: We're not using it for anything! When Ed first wrote the Jenkinsfile we were uploading the builds from inside the container, which doesn't have the aws cli. I changed that a while back but forgot to remove this node package from the Dockerfile.

PR #105, opened 2019-04-01
Add Verification release notes 

rtravitz:
This pull request addresses: https://github.com/department-of-veterans-affairs/vets-contrib/issues/1634

- Adds current release note language for the verification apis
- Enables Address Verification API Flag
- Fixes an issue where the sidebar and api cards were not scrolling to the anchor links on click.


src/App.tsx
```diff
@@ -28,7 +28,7 @@ function isHostedApiEnabled(shortName: string, defaultValue: boolean): boolean {
 
 export const flags = {
   hosted_apis: {
-    address_validation: isHostedApiEnabled('address_validation', false),
+    address_validation: isHostedApiEnabled('address_validation', true),
```
dhall: was this intentional? seems unrelated to the release notes

rtravitz: It was intentional. The verification section includes release notes for the address validation API, however the sidebar nav and the card to scroll to those notes will not be displayed while this flag is disabled. I can turn it back off, but then should we consider removing release notes for the address validation API?

nfasulo: This is my fault, I wasn't clear when describing how to show this API. But this does raise the question: is there a reason we need `address_validation` to default to `false` anymore, since it is set to `true` in the `.env` file for all three environments?
@rtravitz I'm guessing the reason you weren't seeing this API card when running locally is because `REACT_APP_ADDRESS_VALIDATION_API_ENABLED=true` is missing from your `.env.local`.

rtravitz: Ah, yep @nfasulo, it is missing. With that in `.env.local` it displays the API card and sidebar entry even with the flag in App.tsx set to false. I can change it back permanently or not based on what you and @dhall prefer.

src/content/releaseNotes/verification.mdx
```diff
@@ -1,34 +1,29 @@
-<div id="disability_rating" />
-
 ## Disability Rating
```
nfasulo: I think these API names should be marked with `#` instead of `##`.

src/content/releaseNotes/verification.mdx
```diff
@@ -1,34 +1,29 @@
...
 
-**November 2018 | v0.0.1**
-Launched v0.0.1 of the Disability Rating API.
-
-<br />
+### November 8, 2018
+### Launched the Disability Rating API
```
nfasulo: To match the Facilities API I would suggest:
```
## November 8, 2018

**Launched the Disability Rating API**
- Disability Rating API endpoint for external API consumers. Retrieves disability rating of an authorized Veteran.
```

src/content/releaseNotes/verification.mdx
```diff
@@ -1,34 +1,29 @@
...
 
-<div id="service_history" />
+---
 
 ## Service History
+ 
+### **June 11, 2018**
```
nfasulo: Now that we aren't showing the version we don't need to force the date to be bold.
```suggestion
### June 11, 2018
```

src/App.tsx
nfasulo: @rtravitz I say leave the flag as is for a minute. Only because I really am curious as to whether there is there a reason to default `address_validation` to `false` at this point, since it is set to `true` in the `.env` file for all three environments? @dhall @kalilsn @dvogel

Eallan919:
LGTM! Tagging @nfasulo and @hubgitian for review as well!

src/App.tsx
nfasulo: Drew gave a :+1:  in slack for changing the default value to `true`, so we should be good on this. I did realize there is one more piece of cleanup we should do before merging this though: `REACT_APP_ADDRESS_VALIDATION_API_ENABLED=true` can be removed from the `.env` files for `dev`, `staging`, and `production`. Will you make that change? Thanks.

rtravitz: Perfect, thanks for asking! Just pushed that change.

PR #109, opened 2019-04-11
Tell IE11 to disable compat mode and bring in polyfills.

dvogel:
This fixes two issues with the developer portal when run in IE11. On VA machines IE11 is configured to run *.va.gov in compatibility mode. We don't want to use compatibility mode because it effectively limits us to ES3 like it's 2008. The manifestation of that errors was a `SCRIPT1010` error in the IE console.

Once running in IE11 mode (which provides ES5) we ran into an issue where IE11 doesn't ship a native `Object.values` method. That's slated for the next ES version, so we should feel safe using it. Rather than importing just the specific methods we'll use I've brought in `@babel/polyfill`. I chose that route because it allows us to specify `es2017` in our `tsconfig.json` file without worrying that the code has passed the type checker due to type definitions for a method that we failed to import a specific polyfill for. This increases the compressed bundle size about 5% (28k over 481k).

### DIFF
--- a/config/polyfills.js
+++ b/config/polyfills.js
@@ -1,6 +1,8 @@
 'use strict';
 
-// We rely on this for a few functions like Object.values.
+// We rely on this for a few functions like Object.values right now but in
+// general it is what allows us to confidently add `es2017` to
+// `compilerOptions.lib` in our typescript config.
 import '@babel/polyfill';
 
 if (typeof Promise === 'undefined') {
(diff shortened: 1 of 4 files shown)


dvogel:
Just for funsies, here's a map of what our production bundles are made of:

Screenshot from 2019-04-11 15-32-51

dvogel:
> looks good to me. were you able to verify this on the VA laptop?

Yeah. I couldn't figure out how to take a screenshot on the laptop though.

PR #113, opened 2019-04-22
applied page support fixes

jeff2d2:
This PR addresses https://github.com/department-of-veterans-affairs/vets-contrib/issues/1839. It removes both references to api@va.gov from `/applied`.


src/containers/ApplySuccess.tsx
```diff
@@ -16,14 +15,9 @@ const mapStateToProps = (state : IRootState) => {
...
 
-function AssistanceEmail() {
-  return <a href={`mailto:${ASSISTANCE_EMAIL_ADDRESS}`}>{`${ASSISTANCE_EMAIL_ADDRESS}`}</a>;
-}
-
 function AssistanceTrailer() {
   return (
-    <p>If you need any assistance, please email us at <AssistanceEmail /></p>
-  );
+    <p>If you would like to report a bug or make a feature request, please open an issue on our <a href="https://github.com/department-of-veterans-affairs/vets-api-clients/issues/new/choose">GitHub page</a>.</p>  );
```
erluti: Should the right parens be on a newline?

jeff2d2: Whoops! Fixed. Thank you!

rtravitz:
Could you still do something similar to the AssistanceEmail component from before instead make it a link to Github Issues?

Also, what kind of new issue would I create if I wanted to start developing for an API that requires OAuth?

jeff2d2:
@rtravitz 
1. I definitely could, if you think that would be better? This is a stop-gap until the Support page is created. If you think it will be useful once we move to that as well, then I'll go back and add the constant back in.
2. *I think* the text `If you decide to develop an application with the Health API or the Verification API, please open an issue on our GitHub page to setup OAuth credentials.` in theory will only be shown if something happened during the key creation process. @kalilsn can you confirm? Maybe it would be helpful to tell them to open a `Support Request` issue and link directly to that template?

rtravitz:
> 1. I definitely could, if you think that would be better? This is a stop-gap until the Support page is created. If you think it will be useful once we move to that as well, then I'll go back and add the constant back in.

No sense if it's just going away!

> 2. _I think_ the text `If you decide to develop an application with the Health API or the Verification API, please open an issue on our GitHub page to setup OAuth credentials.` in theory will only be shown if something happened during the key creation process. @kalilsn can you confirm? Maybe it would be helpful to tell them to open a `Support Request` issue and link directly to that template?

That would make more sense to me if I was directly in the Support Request template.

kalilsn:
I think that text is shown when you don't sign up for one of the OAuth APIs. I'm not actually sure I agree with the text though. Do we want people to email us or open an issue if they want to get OAuth credentials? I'm guessing we would just tell them to go back to the developer portal and sign up for access to an OAuth API.

rtravitz:
> I'm guessing we would just tell them to go back to the developer portal and sign up for access to an OAuth API.

For my own knowledge, would the additional signup add permissions to their current account? Or would the user then be managing two sets of credentials?

kalilsn:
We don't actually have any notion of accounts on the developer portal. If you don't sign up for any OAuth APIs, you'll have a consumer added to Kong and receive the associated API key but no Okta app will be created. If you then go back and sign up for an OAuth API and no other APIs, an Okta app will be created, and you'll receive the client id and secret by email.

LMK if that's confusing and I can try to explain better. In short, I don't think there's any issue with people signing up multiple times (aside from skewing our signup count).

At some point in the future, we will definitely want to make this easier, but we can't really do that until we have some authentication on the portal.

jeff2d2:
@kalilsn @ryanlink Circling back on this, how about just nixing that line completely? Do we need to tell them to come back and sign up for more APIs if they want to access more APIs?

ryanlink:
@jeff2d2 Only reason I can think of to keep that line is so they don't think they can use the same API key for multiple APIs (they shouldn't be able to, anyway).  We're putting a bit about that in the FAQ.

kalilsn:
I vote for getting rid of it. @dvogel, since you added this notice in, is there a reason we should keep it?

dvogel:
> I vote for getting rid of it. @dvogel, since you added this notice in, is there a reason we should keep it?

I believe we added it just so users had _some_ support path. I don't see any reason we can't remove it now. When we added it @jwoodman5 wanted to make sure this was the same as the email the user receives after signing up. Probably worth getting his opinion on whether it should also be changed in the email.

jeff2d2:
For context the current email looks like this:
>>>Welcome Jeff,

Thank you for your interest in our Benefits Intake API and VA Facilities API. We are excited to partner with you to improve the lives of veterans.

Here's your key for accessing the development environment:
0FMnpS6TtcJYtyyOHOzfGBSItSLFq214

You can use your key by including it as an HTTP request header
apiKey: 0FMnpS6TtcJYtyyOHOzfGBSItSLFq214
in your requests to the API. You can find additional documentation at developer.va.gov

If you find a bug or would like to make a feature request, please open an issue in our GitHub. We are continually working to improve our process and welcome feedback along your journey.

When you're ready to move to a production environment, please follow the steps outlined on our Path to Production page.

Thank you again,
VA API Platform Team

Read VA API Docs at: developer.va.gov
Get support: Create Github Issue
Email us at: api@va.gov

jwoodman5:
> I vote for getting rid of it. @dvogel, since you added this notice in, is there a reason we should keep it?

I agree with @kalil on this. Looks like it's already taken care of, but since I was mentioned just wanted to validate it is not necessary any longer.

ryanlink:
Worth mentioning that this email also references the api@va.gov which we've
been on/off about deprecating. I'm for keeping it (and will note that in
other related tickets) but know that we were trying to steer folks towards
Github in the interim while we figure out ownership and workflow for that
email address.

PR #114, opened 2019-04-23
Fix visual regression test timeouts

kalilsn:
On a dev's local machine, puppeteer will sometimes timeout on the first page load during the visual regression tests. This is because jest starts the tests and the webpack dev server at the same time, and therefore it starts compiling after the first request is sent. When these timeouts happen, it's impossible to update the snapshots!

My fix here is just making sure the first request doesn't timeout.


src/visualRegressionTest.ts
```diff
@@ -24,6 +24,15 @@ const checkScreenshots = async (page: Page) => {
 const paths = testPaths.filter(path => path !== '/');
 
 describe('Visual regression test', async () => {
+  it('renders the homepage properly', async() => {
+    // Set unlimited timeout on first request, since it may timeout while webpack is compiling.
+    await page.goto(`${puppeteerHost}`, { waitUntil: 'networkidle0', timeout: 0 });
```
kalilsn: `timeout: 0` is the bulk of this change

jest-puppeteer.config.js
```diff
@@ -6,5 +6,8 @@ module.exports = {
   },
   launch: {
     args: ['--no-sandbox'],
+    // Uncomment the following line to open a browser window with devtools open
+    // when you run puppeteer tests, which can be useful for debugging.
+    // devtools: true,
```
kalilsn: This is just for convenience. When I'm debugging issues with puppeteer tests, I've often found it useful to have an actual browser I can interact with.

PR #115, opened 2019-04-24
Add facilities csv fix release note

rtravitz:
Finishes [#1885](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1885)

### DIFF
--- a/src/content/releaseNotes/facilities.mdx
+++ b/src/content/releaseNotes/facilities.mdx
@@ -1,3 +1,9 @@
+## April 18, 2019
+
+Fix facilities error when returning CSV [#2974](https://github.com/department-of-veterans-affairs/vets-api/pull/2974)
+
+---
+
 ## February 13, 2019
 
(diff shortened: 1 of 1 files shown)


jeff2d2:
@rtravitz Could you be a bit more specific? I think it would be helpful to add in that it was a `500` error, and that it was affecting the `/facilities/all` bulk download endpoint.

rtravitz:
@jeff2d2 Definitely, I just took the current language from [the top of the issue](https://github.com/department-of-veterans-affairs/vets-contrib/issues/1885).

How about: Fix an error where the `/facilities/all` endpoint returned a 500 status code when a user requested `text/csv`. [#2974](https://github.com/department-of-veterans-affairs/vets-api/pull/2974)

jeff2d2:
Perfection!

rtravitz:
Ended up tweaking the format slightly to match up with other release notes in that section. It looked a bit odd to have that one title go on for so long, so I moved the bulk of the updated text to a bullet point.

PR #120, opened 2019-05-13
Host is parameterized for browser tests

jasalisbury:
department-of-veterans-affairs/vets-contrib#1596

e2e, accessibility and visual regression tests have a parameterized host. This will allow us to run these tests against production to inform us when production has UI errors. Url that tests use is determined by the environment variable `TEST_HOST`. If `TEST_HOST` is not set a server is spun up at localhost:4444, which matches how e2e tests function today.

I am not sure what the requirements for running tests on the script exporter are, other than it needs to be under 60 seconds. Not having to spin up a server to test against knocked off around 20 seconds from the tests on my local machine. We may be able to use the e2e test script as is? Thoughts @kalilsn and @wunderhund?


docs/testing.md
```diff
@@ -66,6 +67,7 @@ docker run --rm -itv "$PWD:/application" -v "/application/node_modules" develope
 
 It's important to run these tests inside the docker container, or else the tests will report false negatives due to font differences.
 
+`TEST_HOST` can be set for visual regression testing in the same way as the end to end tests.
```
kalilsn: This could be confusing for people new to this project, since those tests mostly won't pass against the live site as is. Maybe add a note saying this isn't useful yet?

jasalisbury:
> LGTM. It would be great to see a comment in `App.e2e.ts` letting people know that it's mean to be run in CI and against a running site.

@dvogel App.e2e.ts isn't a test I wrote to run against CI, it is a test that was already written for e2e tests.

jasalisbury:
>  thanks for taking care of this! Did you run into any timeouts when running against the live site? @wunderhund was getting some weird timeout errors a while ago, so I'm curious if you had to do anything to fix that.

@kalilsn I was seeing timeouts running tests against a deployed site as well. I had to update the version of `puppeteer` and `puppeteer-jest`. I believe updating fixed an issue with puppeteer not counting active connections correctly. The tests determines if a page has loaded by waiting for 0 active connections, so not correctly counting active connections caused the tests to timeout waiting for pages to load.

dvogel:
> > LGTM. It would be great to see a comment in `App.e2e.ts` letting people know that it's mean to be run in CI and against a running site.
> 
> @dvogel App.e2e.ts isn't a test I wrote to run against CI, it is a test that was already written for e2e tests.

Before this change we ran the e2e tests as part of our CI builds. With this change, it has assumed a secondary role as part of our monitoring. Right?

jasalisbury:
> > > LGTM. It would be great to see a comment in `App.e2e.ts` letting people know that it's mean to be run in CI and against a running site.
> > 
> > 
> > @dvogel App.e2e.ts isn't a test I wrote to run against CI, it is a test that was already written for e2e tests.
> 
> Before this change we ran the e2e tests as part of our CI builds. With this change, it has assumed a secondary role as part of our monitoring. Right?

Oh I see what you are saying now, misread what you wrote. I'll add a comment pointing out that `App.e2e.ts` is used for CI and for monitoring.

jasalisbury:
@dvogel @kalilsn @mjlumetta what would you suggest I do about the failing visual regression tests? None of the differences are big (most look like a few pixel differences here and there), and can all probably be attributed to updating puppeteer. Should I just update the test images?

kalilsn:
I think it's fine to update them. The latest jenkins failure doesn't appear to be about visual regression tests:
```
+ npm audit

npm ERR! code ENOAUDIT

npm ERR! audit Your configured registry (https://registry.npmjs.org/) does not support audit requests, or the audit endpoint is temporarily unavailable.
```

I got that on another branch this morning... hopefully it's just a temporary npm outage.

dvogel:
> @dvogel @kalilsn @mjlumetta what would you suggest I do about the failing visual regression tests? None of the differences are big (most look like a few pixel differences here and there), and can all probably be attributed to updating puppeteer. Should I just update the test images?

Committing this screenshots seems right to me. There's also a tuning knob somewhere in the test that lets you adjust the sensitivity of the image diff. You could try lowering the sensitivity slightly.

config/jest/imageSnapshot.ts
```diff
@@ -1,5 +1,8 @@
 import { configureToMatchImageSnapshot } from 'jest-image-snapshot';
 
-const toMatchImageSnapshot = configureToMatchImageSnapshot({});
+const toMatchImageSnapshot = configureToMatchImageSnapshot({
+    failureThreshold: 0.01,
```
jasalisbury: @dvogel @kalilsn @mjlumetta This configures visual regression tests to fail if the difference between the generated and source page is greater than 0.01%. There was a test that occasionally failed with a difference of ~0.008%. I'm not sure why the image the test produced is sometimes different. Does this change seem appropriate?

PR #122, opened 2019-05-23
Add community care API to signup form in dev only

kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2134

Adds new Community Care Eligibility API to signup form (and documentation page, although there's nothing there). Feature flagged so it's only present in dev.

https://github.com/department-of-veterans-affairs/devops/pull/4854 takes care of the backend portion.


kalilsn:
>Do we not need to set the feature flag in .env.dev?

We do! That was a mistake on my part.

> Also unrelated to this PR... in src/apiDefs.ts the health api openApiDocUrl is set to staging-api.

I remember asking that a while ago, and I believe the answer at the time was that the environment that our staging env proxies to was the only one that had the OpenAPI spec available. Quite a bit has changed about how we proxy to their API though, so that's probably worth revisiting!

src/containers/ApplyForm.tsx
```diff
@@ -141,6 +145,18 @@ class ApplyForm extends React.Component<IApplyProps> {
                 <label htmlFor="health">VA Health API</label>
               </div>
 
+              <Flag key='community_care' name='hosted_apis.community_care'>
+                <div className="form-checkbox">
+                  <input
+                    type="checkbox"
+                    id="community_care"
+                    name="community_care"
```
kalilsn: Turns out this `name` attribute is not actually what determines the key in the payload. As it stands, this is sending `"communityCare":` rather than `"community_care":`

kalilsn: Since they're not used in the payload, is there any reason to keep these `name` attributes?

jasalisbury: Does `name` do anything for the UI in terms of accessibility?

kalilsn: Not as far as I can tell.

kalilsn: It's hard to prove that though!

dvogel: There was a point at which `name` was a requirement attribute for `input` elements. I believe the HTML5 spec changed that but a lot of linting tools still enforce it. That creates a strong incentive to add them because if you ever have other unrelated markup concerns the linter you turn to will be quite noisy if you don't include `input.name` attributes.

kalilsn:
This is actually not ready to merge, since checking the community care api box doesn't show + require the redirect URI field.

kalilsn:
This PR grew quite a bit because I wanted to refactor the apply form so that we weren't hardcoding which APIs used OAuth there. There's clearly a lot of boilerplate code required for adding a new API here that could be improved, but I don't want this to get any bigger.

src/apiDefs.ts
```diff
@@ -16,9 +16,13 @@ export interface IApiDescription {
   readonly vaInternalOnly: boolean;
 }
 
+export interface IApiDescriptions {
+  [key: string]: IApiDescription;
```
kalilsn: Is there a better way to do this?

src/apiDefs.ts
```diff
@@ -16,9 +16,13 @@ export interface IApiDescription {
...
+}
+
 export interface IApiCategory {
   readonly apiKey: boolean;
-  readonly apis: IApiDescription[];
+  readonly apis: IApiDescriptions;
```
kalilsn: Looking back at this, I wonder if I should have just left this as an array and added a `name:` property to each of them. Doing it this way feels like it makes it clearer what the primary identifier is, but I'm not wild about how much we have to use `Object.keys`/`Object.values` because of it.

kalilsn: To be honest, I didn't even remember at first that you can't call `map` or `length` etc. on an object in JS. If I had, I probably would have made a different choice.

src/apiDefs.ts
```diff
@@ -159,6 +165,7 @@ export const apiDefs : IApiCategories = {
     shortDescription: 'Empowering Veterans to take control of their data and put it to work.',
   },
 };
+// tslint:enable:object-literal-sort-keys
```
dvogel: I would like to stay away from using an object to store this data. It will incentivize us to rely on that key in places we shouldn't, simply because it will be so easy. That's actually how we ended up with the `urlFragment` field being spread throughout the code without having a formal name.

We can model the data and lookups much like a database table in order to avoid that trap. If getting O(1) lookups is the primary concern motivating, we can achieve that without adding these explicit keys. This would require building a second object that serves as an index into the array of records. Both approaches leave us with an object of the same shape, with the same values as keys in the object. The important difference IMO is that building that index (either at build time or runtime) requires us to name the field and use that name in a function name like so:

```
# exported method
lookupApiByNewNameField(pkey: string)

# internal to the module
function indexApisByNewNameField() {
  ...
}
const apisByNewNameField = indexApisByNewNameField();
```

src/apiDefs.ts
```diff
@@ -196,3 +203,13 @@ export function lookupApiCategory(categoryKey: string): IApiCategory | null {
   return apiDefs[categoryKey];
 }
 
+export function isOauthApi(apiName: string): boolean {
```
dvogel: Can we find a more precise name for this `apiName` parameter? Each API record already has a field called `name` but it's `Veteran Confirmation` whereas this looks like it would be `veteranConfirmation`.

src/containers/ApplyForm.tsx
```diff
@@ -207,7 +208,7 @@ class ApplyForm extends React.Component<IApplyProps> {
   }
 
   private renderOAuthFields() {
-    if (this.props.inputs.apis.health || this.props.inputs.apis.verification) {
+    if (this.anyOAuthApisSelected()) {
```
dvogel: :+1:

src/apiDefs.ts
kalilsn: That makes sense to me. My primary concern was just making sure future devs use consistent keys for looking up apis, which I think is solved by exporting this new function.

What do you think about still using `name` and changing the existing name field to `displayName`? Or maybe `machineName`?

dvogel: I like the existing values for `name`. If there's going to be a field called `name` it should come with all the ambiguities of human interpretation :smile: I'm not sure what a better name would be right now. Can you help me understand where this camelcase name is coming from? Is this naming used in the backend lambda API request? If so then `signupApiName` might be good. If it's coming from our form then it might make more sense to translate from those form field names to the existing `urlFragment` field in whatever module interprets that form data.

jasalisbury: If we are using `name` as an id why not just rename it to `id`? or maybe `key` (would rather `id` though)? Do we really need to worry about O(1) lookups? What is the max number of APIs we would ever need to loop through, 100s?

kalilsn: I agree that performance is not really the concern here! Ended up deciding that making changes to this data structure isn't really the right path forward, so instead there's now a map from the names on the form to url fragments in ApplyForm.tsx.

src/apiDefs.ts
```diff
@@ -189,3 +196,18 @@ export function lookupApiCategory(categoryKey: string): IApiCategory | null {
...
+    for (const api of cat.apis) {
+      if (apiList.includes(api.urlFragment)) {
+        categories.push(cat);
+      }
+    }
+  }
+  return Array.from(new Set(categories));
+}
+
+export function includesOauthAPI(apiList: string[]): boolean {
```
kalilsn: I moved this logic to `apiDefs.ts` and didn't export `categoriesFor` so we can move away from other modules needing the `IApiCategory` type. There are still a bunch of other cases of that, so I didn't stop exporting it yet, but I didn't want to add another.

src/apiDefs.ts
```diff
@@ -189,3 +196,18 @@ export function lookupApiCategory(categoryKey: string): IApiCategory | null {
   return apiDefs[categoryKey];
 }
 
+function categoriesFor(apiList: string[]): IApiCategory[] {
+  const categories = [];
```
kalilsn: tslint told me to change this from `let` to `const`. I kind of hate it because I didn't think you should change the value of a `const` after it was declared, but apparently that's not how it works!

dvogel: Yeah, I think that linting rule is silly. Typescript has constant references, in that they can only ever reference a single object. That object is not immutable though, which is what a lot of people want intuit "const" to mean. I'd be 100% OK disabling that linting rule. Does anyone really like it?

src/apiDefs.ts
```diff
@@ -189,3 +196,18 @@ export function lookupApiCategory(categoryKey: string): IApiCategory | null {
...
+  for (const cat of Object.values(apiDefs)) {
+    for (const api of cat.apis) {
+      if (apiList.includes(api.urlFragment)) {
+        categories.push(cat);
+      }
+    }
+  }
+  return Array.from(new Set(categories));
+}
```
dvogel: Could `categories` be initialized as a `Set`?  That would save memory by doing the uniqueness check while accumulating the set. It would also avoid an extra conversion from Array to Set.

src/apiDefs.ts
kalilsn: I don't feel very strongly either way. The linter did force me to learn something new, so it's not all bad.

PR #123, opened 2019-05-28
Add warning banner for outage

nfasulo:
Adding warning banner for outage caused by network outage in Austin.


src/containers/Home.tsx
```diff
@@ -16,6 +16,18 @@ class Home extends React.Component {
                 </a>
               </div>
 
+              <div className="columns small-12">
+                <div className="form-warning-banner">
+                  <div className="usa-alert usa-alert-warning">
+                    <div className="usa-alert-body">
+                      <h3 className="usa-alert-heading">You may have trouble accessing some of our APIs</h3>
+                      <div className="usa-alert-text">We’re sorry. We’re working to fix a problem that affects some parts of our site. If you have trouble signing in or using any tools or services, please check back soon.</div>
```
kalilsn: I think we want slightly different messaging, right? We don't have signin functionality and this is not really an issue for the site as much as it is the APIs. Maybe a list of the APIs + environments that are affected would be more helpful?

nfasulo: Yep, I just wanted to get something up as a starting point. I think @ryanlink is going to put together something better shortly.

kalilsn: Sorry, just saw all the slack messages about this 🙃

ryanlink:
Agree with @kskeene and Andrew F - let's use:

>You may have trouble accessing some of our APIs.

with a subtext of 

> Due to a datacenter outage as of Tuesday evening, May 28, the **Address Validation** and **Appeals Status** APIs may not be responding properly. We apologize for the inconvenience and are working to resolve the issue as quickly as possible.  Please check back here for updates.

PR #124, opened 2019-05-29
Fix styling for alert-warning

juliaelman:
This pull request compliments the work started on #123. Fixes the position of the warning icon + white space between the alert warning + beta disclaimer.


src/base.scss
```diff
@@ -123,3 +123,19 @@ p {
 span.mono {
   font-family: monospace;
 }
+
+.usa-alert {
```
kalilsn: I'm guessing in the future we'll want this styling to be specific to top of the page alert banners, just in case we use `usa-alert` for anything else?

kalilsn: Not saying we should change anything here though!

juliaelman: @kalilsn yep! Overall, I think we need to address alerts and their life cycle overall. I'd like to figure out a way to streamline adding alert and how they are used on the developer portal.

//cc @kskeene

PR #125, opened 2019-05-29
Remove warning banner

nfasulo:
This PR ~disables~ removes the outage warning banner that was added (department-of-veterans-affairs/developer-portal/pull/123) and styled (department-of-veterans-affairs/developer-portal/pull/124) last night to warn consumers about the API outage.

~I am using a simple `if` statement to determine whether to render the warning flag, which will allow us to update a single file to update and display the warning message across all environments.~

This file was also split between 2 and 4 spaces for indentation. Sorry for the noise in the diff that normalizing the indentation caused. ~The only real change is adding `warning banner`.~


src/containers/Home.tsx
```diff
@@ -4,110 +4,120 @@ import { Link } from 'react-router-dom';
...
+    function warningBanner() {
+      let banner;
+      if (false) {
+        banner = (
+          <section aria-live="polite" role="alert" className="usa-alert usa-alert-warning">
+            <div className="usa-alert-body">
+              <h3 className="usa-alert-heading">You may have trouble accessing some of our APIs</h3>
+              <div className="usa-alert-text">
+                As of Tuesday evening, May 28, the <strong>Address Validation</strong>, <strong>Appeals Status</strong>, and <strong>Health</strong> APIs may not be responding properly. We apologize for the inconvenience and are working to resolve the issue as quickly as possible. Please check back here for updates.
+                        </div>
```
dvogel: this closing tag is misaligned now

kalilsn: this indentation looks a little funky!

src/containers/Home.tsx
```diff
@@ -4,110 +4,120 @@ import { Link } from 'react-router-dom';
 import './Home.scss';
 
 class Home extends React.Component {
-    public render() {
-        return (
-            <div className="Home">
+  public render() {
+    function warningBanner() {
+      let banner;
+      if (false) {
```
kalilsn: If you want to leave the code in here, I think just commenting it will be less confusing in the future than `if (false)`. I don't necessarily think it's even worth leaving it in though. This felt like a quick solution because we were in a hurry, but I think now we'll want to build something more robust for the future.

kalilsn: By "robust", I mean things like:
- Shows on all pages
- Isolated into its own component
- Potentially relies on some outside data source so we don't have to do a deploy to put up a banner?

nfasulo:
In the interest of having a better implementation for future warn messages, I've removed the banner and its styling completely.

PR #127, opened 2019-06-06
Remove service worker

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2228

This will remove the service worker, and any caching the developer portal has. Keeping as a draft until we decide we actually want to remove the service worker.


src/index.tsx
```diff
@@ -20,7 +20,7 @@ try {
     </Provider>,
     document.getElementById('root') as HTMLElement,
   );
-  registerServiceWorker();
+  unregister();
```
dvogel: Do we know if this is takes effect immediately or is it subject to the same browser tab closing issue as the updates?

kalilsn: I was under the impression that we're no longer subject to multiple tab issues ever since we started using [skipWaiting](https://github.com/department-of-veterans-affairs/developer-portal/blob/master/config/webpack.config.prod.js#L403).

dvogel: :winning:

jasalisbury: Unfortunately I don't think `skipWaiting` helps us here. `skipWaiting` allows a service worker in the `waiting` state to automatically take control over the old service worker. Doesn't seem to apply as unregistering never puts a service worker in the `waiting` state. From my understanding the unregistered service worker will continue to be active until all tabs are closed, so yes Drew I think it behaves the same as updates. Checkout the [spec](https://w3c.github.io/ServiceWorker/#navigator-service-worker-unregister).

kalilsn: I read through the spec a bit, but I'm still confused. Does this mean that even if we deploy this change, the service worker won't unregister until users refresh?

jasalisbury: The unregister happens immediately but all active service workers stay active. Say the service worker is unregistered, if the tab is closed or the user navigates away from the developer portal the next time they browse to the developer portal the service worker will not start up. If they refresh their page the service worker will stay active, even though it has been unregistered.

kalilsn:
Could you add a few comments and/or a note in the readme about what happened here? Maybe a short explanation that we had a service worker and now we need this code to unregister it but aren't using it for any functionality?

PR #129, opened 2019-06-10
Add ability to temporarily disable signups

kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2331

Adds a message that is displayed in place of the signup form when `REACT_APP_SIGNUPS_ENABLED` is set to `false`. I also added a .editorconfig file and renamed the prettier config file since it wasn't taking effect.

Here's a screenshot of the disabled /apply page:

Signups Disabled


juliaelman:
@kalilsn can you tell me a little bit more about this work? when would there be an instance when sign ups would be disabled?

kalilsn:
@juliaelman There are two times I foresee: in a few days we're going to need to migrate the API gateway to the new DVP environments, which will be pretty hard to do without preventing signups for a little while (probably an hour or less). This is probably overly cautious for this migration, but later on when we want to use `sandbox` for external developers instead of `dev`, we'll need a similar function.

juliaelman:
@kalilsn thanks for clarifying! A few notes:

1. The width of the content should not span past 2/3 of the page. Anything greater makes it harder to read.
1. I wonder if we need to disable the `Get Started` along with disabling the form. Why provide users a link that goes to a page that is disabled? This could possibly erode user trust by providing a `Get Started` link that doesn't work.

Open to ideas and suggestion!

src/Routes.tsx
```diff
@@ -3,25 +3,50 @@ import * as React from 'react';
...
+  BetaPage,
+  BetaSuccess,
+  ExploreDocs,
+  Home,
+  OAuth,
+  ReleaseNotes,
+  RoutedContent,
+  WhatsNew,
+} from './containers';
+import { DisabledApplyForm } from './containers/DisabledApplyForm';
```
mjlumetta: shouldn't this be added to the previous import statement?

kalilsn:
Thanks for reviewing @juliaelman!

>The width of the content should not span past 2/3 of the page. Anything greater makes it harder to read.

Is this something that's generally true for our whole site? And I'm guessing this only applies above a certain window width?

Right now I'm just using the styling from our `.usa-grid` class, which is set to `max-width: 1200px`. That choice was just based on matching what we did on https://developer.va.gov/oauth.

> I wonder if we need to disable the Get Started along with disabling the form. Why provide users a link that goes to a page that is disabled? This could possibly erode user trust by providing a Get Started link that doesn't work.

What do you envision disabling that button would look like? Removing it with no explanation seems like it might be frustrating to users. I'm also not sure how we would we deal with this section of the homepage: 

Screen Shot 2019-06-11 at 9 23 37 AM

juliaelman:
@kalilsn great questions!

1. We use the class name `usa-width-two-thirds` in several areas on the site (e.g. homepage, docs). It's not documented at the moment, but creating those guidelines is something that is something we are currently working on 😄 
1. Disabling = removing the button. That section will be removed with the new homepage launch. 

I think we can we move forward with this once the two thirds layout is implemented. Don't want to hold up the work, as I know you are needing this in a few days, but I think that we need to think more about what the process looks like when we disable this form (in the future).

kalilsn:
@juliaelman I made the 2/3 change but didn't remove the button, since it sounds like you think that makes sense in tandem with future work. Let me know if I misunderstood and I can also hide the button here!

juliaelman:
Yep 🚀

PR #130, opened 2019-06-11
Curlify swagger plugin

ATeal:
Pull over curlify code from the swagger-ui project
include usage / example in the comments
swagger client type definition

for https://github.com/department-of-veterans-affairs/vets-contrib/issues/2176


src/components/swaggerPlugins/curlify.js
```diff
@@ -0,0 +1,69 @@
...
+          curlified.push(`"${k}=@${v.name}${v.type ? `;type=${v.type}` : ""}"`);
+        } else {
+          curlified.push(`"${k}=${v}"`);
+        }
+      }
+    } else {
+      curlified.push("-d");
+      curlified.push(JSON.stringify(request.get("body")).replace(/\\n/g, ""));
+    }
+  }
```
dvogel: Is `requestOptions` the raw sub-document of the OpenAPI definition? If it is, this looks like it's only compatible with OAS v2. The OAS v3 spec [did away with `body` and uses `requestBody` instead](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#operation-object). The structure of the new request body object is more varied because it has a `content` key that maps to an object with keys that represent the content-type of the request body.

ATeal: I'll have to look into that. My gut tells me that the swagger client build request handles both OAS v2 and OAS v3, but I will do some testing

ATeal: Sorry for the delay got busy working on some other stuff. 

As it stands now, Appeals is on OAS v3 spec. I tested generating a curl with these values:
```
{
  operationId: "getAppealStatus",
  securities: {
    authorized: {
      apikey: 123
    }
  }, 
  parameters: {
    "X-VA-SSN": 123,
    "X-VA-User": "test user"
  },
  spec: json_spec
}
```
and it generated: `curl -X GET "https://dev-api.va.gov/services/appeals/v0/appeals" -H  "X-VA-SSN: 123" -H  "X-VA-User: test user"`

PR #132, opened 2019-06-13
Art.version selector

ATeal:
Implement a version selector, loads the metadata endpoint, refreshes the swagger docs based on selected version.

resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/2348


src/components/swaggerPlugins/ExtendedLayout.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from "react";
+import { VersionSelectBox } from "./VersionSelectBox";
+
+export interface IExtendedLayoutProps {
+  getSystem: any;
+  getComponent: any;
+}
```
dvogel: I'm unclear on what these props are for. It seems like they come from swagger-ui. Is that right? Could you add a command here explaining what each is, with maybe a link to where they are documented by swagger-ui? Even just knowing for sure that they come from swagger-ui would let me put them in the "don't worry about these right now" bucket.

src/components/swaggerPlugins/VersionSelectBox.tsx
```diff
@@ -0,0 +1,56 @@
...
+      .getSystem()
+      .versionActions.updateVersion(
+        this.buirlUrlFromMeta(versionMetadata),
+        value,
+      );
+  }
+
+  public buirlUrlFromMeta(metaObject: any) {
+    return `${process.env.REACT_APP_VETSGOV_SWAGGER_API}${metaObject.path}`;
+  }
```
dvogel: Was this supposed to be `buildUrlFromMeta`?

src/components/swaggerPlugins/ExtendedLayout.tsx
ATeal: Yes they do come from swagger. I'll add some comments

src/components/swaggerPlugins/VersionSelectBox.tsx
ATeal: It sure was. Haha auto complete hiding spelling mistakes from me.

ATeal:
@dvogel this is post prettier, if you see something that stands out as against the rules, maybe nvim isn't picking up the project config, can you confirm that so I have a line to test against?

src/containers/Explore.tsx
```diff
@@ -1,55 +1,56 @@
-import * as React from 'react';
+import * as React from "react";
```
nfasulo: An simple example of the prettier config not being followed is the change from single quotes to double quotes at the top of `src/containers/Explore.tsx`.

dvogel:
> @dvogel this is post prettier, if you see something that stands out as against the rules, maybe nvim isn't picking up the project config, can you confirm that so I have a line to test against?

That's probably what's happening. If I checkout your branch and run:

```
prettier --write $(git diff --name-only HEAD master src/containers src/components)
src/components/SwaggerDocs.tsx 236ms
src/components/swaggerPlugins/ExtendedLayout.tsx 20ms
src/components/swaggerPlugins/VersionActions.ts 7ms
src/components/swaggerPlugins/VersionReducers.ts 5ms
src/components/swaggerPlugins/VersionSelectBox.tsx 24ms
src/components/swaggerPlugins/VersionSelector.ts 5ms
src/components/swaggerPlugins/index.ts 9ms
src/containers/Explore.tsx 14ms
```

There's some quote mark changes, as well as some wrapping changes. Here's a sample of that output.

```
git diff | tail -n 30
-import { Flag } from "flag";
-import { connect } from "react-redux";
-import { RouteComponentProps } from "react-router";
+import { Flag } from 'flag';
+import { connect } from 'react-redux';
 
-import { lookupApiByFragment } from "../apiDefs";
-import { SwaggerDocs } from "../components";
-import ExplorePage from "../content/explorePage.mdx";
-import { IApiNameParam, IExternalSwagger, IRootState } from "../types";
+import { lookupApiByFragment } from '../apiDefs';
+import { SwaggerDocs } from '../components';
+import ExplorePage from '../content/explorePage.mdx';
+import { IApiNameParam, IExternalSwagger, IRootState } from '../types';
 
 export interface IExploreProps extends RouteComponentProps<IApiNameParam> {
   argonaut: IExternalSwagger;
@@ -28,10 +28,7 @@ class Explore extends React.Component<IExploreProps, {}> {
       if (api != null) {
         docsDom = (
           <Flag name={`hosted_apis.${api.urlFragment}`}>
-            <SwaggerDocs
-              url={api.openApiDocUrl}
-              apiName={this.props.match.params.apiName}
-            />
+            <SwaggerDocs url={api.openApiDocUrl} apiName={this.props.match.params.apiName} />
           </Flag>
         );
       }
```

ATeal:
Sorry for that formatting stuff y'all. My nvim config wasn't soaking up the contextual config

tslint.json
```diff
@@ -47,6 +47,10 @@
       "REASON": "RULE REQUIRES aria-* ATTRIBUTES ON ELEMENTS WITH IMPLICIT ROLES. BROKEN AS OF 6.1.0."
     },
     "react-a11y-role-supports-aria-props": {},
+    "react-a11y-no-onchange": {
```
kalilsn: Instead of disabling this globally, can you just use a pragma in the place where it's a problem? Even though it's buggy, I think this rule is useful in other places

src/components/swaggerPlugins/VersionSelect.tsx
```diff
@@ -0,0 +1,55 @@
...
+import './VersionSelect.scss';
+
+export interface IVersionSelectBoxProps {
+  getSystem: any;
+  apiMetadata: any;
+}
+
+export class VersionSelect extends React.Component<IVersionSelectBoxProps, {}> {
+  public handleSelectChange(value: any) {
+    this.props.getSystem().versionActions.updateVersion(value);
```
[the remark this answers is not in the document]
ATeal: True, this is a holdover from when selecting changed the page so it was important to keep version there. I'll update this.

src/components/SwaggerDocs.tsx
```diff
@@ -5,41 +5,109 @@ import { SwaggerPlugins } from './swaggerPlugins';
...
+      const plugins = SwaggerPlugins(this.handleUrlChange.bind(this));
+      const ui = SwaggerUI({
+        dom_id: '#swagger-ui',
+        layout: 'ExtendedLayout',
+        plugins: [plugins],
+        url: this.state.url,
+      });
+      ui.versionActions.setApiMetadata(metadata);
+      ui.versionActions.setApiVersion(this.state.version);
+      return ui;
```
mjlumetta: should either return the UI in both cases or neither. it seems unnecessary?

src/components/swaggerPlugins/ExtendedLayout.tsx
```diff
@@ -0,0 +1,35 @@
...
+// These two props are handed in via swagger-ui
+// getSystem allows access to the swagger-ui state
+// getComponent just allows a helper function to get the base layout
+// see: https://github.com/swagger-api/swagger-ui/blob/master/docs/customization/custom-layout.md
+export interface IExtendedLayoutProps {
+  getSystem: any;
+  getComponent: any;
+}
+
+export class ExtendedLayout extends React.Component<IExtendedLayoutProps, { version: string }> {
```
mjlumetta: could this be renamed to `VersionedLayout`, since that's the specific extension? also where is `this.state.version` used?

src/components/swaggerPlugins/ExtendedLayout.tsx
```diff
@@ -0,0 +1,35 @@
...
+  public render() {
+    const { getComponent, getSystem } = this.props;
+
+    const apiMetadata = getSystem().versionSelectors.apiMetadata();
+
+    const BaseLayout = getComponent('BaseLayout', true)!;
+    return (
+      <div>
+        {apiMetadata && Object.keys(apiMetadata).length !== 0 && (
+          <VersionSelect getSystem={getSystem} apiMetadata={apiMetadata} />
```
mjlumetta: if `apiMetadata` can be computed from `getSystem()`, while pass it as a prop? is it expensive to get?

src/components/swaggerPlugins/VersionActions.ts
```diff
@@ -0,0 +1,30 @@
+export const VersionActions = (handler: any) => {
```
mjlumetta: can `handler` be renamed to `updateUrlHandler` or similar? it is currently pretty vague, given the size of this actions object

src/components/swaggerPlugins/VersionSelect.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <div id="version-select">
+        <select
+          aria-label="Version Selection"
+          value={this.props.getSystem().versionSelectors.apiVersion()}
+          onChange={e => this.handleSelectChange(e.target.value)}
+          onBlur={e => this.handleSelectChange(e.target.value)}
+        >
+          {this.props.apiMetadata.meta.versions.map((metaObject: any, index: number) => {
+            return (
+              <option value={metaObject.version} key={index}>
```
mjlumetta: don't use indices as keys - use the version string instead. especially since i believe the metadata ultimately is set from a state update higher up the chain? this can lead to React rendering issues unless you are sure that the iterated objects won't ever change

src/e2eHelpers.ts
```diff
@@ -1,56 +1,63 @@
...
+  "/",
+  "/apply",
+  "/terms-of-service",
+  "/go-live",
+  "/oauth",
+  "/explore",
+  "/explore/benefits/docs/benefits", // Only include a few swagger pages since they're all pretty similar
+  "/explore/benefits/docs/appeals",
+  "/release-notes",
+  "/whats-new",
```
mjlumetta: incorrectly using double quotes in this file, looks like it slipped through

src/e2eHelpers.ts
```diff
@@ -1,56 +1,63 @@
...
+      body: JSON.stringify(mocks[req.url()]),
+      contentType: "application/json",
+      headers: { "Access-Control-Allow-Origin": "*" },
+    });
+  } else if(req.url() in metadataMocks) {
+    req.respond({
+      body: JSON.stringify(metadataMocks[req.url()]),
+      contentType: "application/json",
+      headers: { "Access-Control-Allow-Origin": "*" },
+    });
```
mjlumetta: since the blocks are the same except for the body object, can you make the if-else if-else block set the body and then do `req.respond` outside of the conditional?

src/components/SwaggerDocs.tsx
```diff
@@ -5,41 +5,109 @@ import { SwaggerPlugins } from './swaggerPlugins';
...
+      json: props.json!,
+      metadata: {},
+      url: props.url!,
+      version: '',
+    };
+  }
+
+  public handleUrlChange(url: string, version: string) {
+    this.setState({
+      ...this.state,
```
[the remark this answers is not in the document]
ATeal: I didn't realize that was the case, probably some historical thing that I'm remembering from a past life updating state on either an older version or some other framework, regardless will fix.

src/components/SwaggerDocs.tsx
```diff
@@ -5,41 +5,109 @@ import { SwaggerPlugins } from './swaggerPlugins';
...
-        );
+  public componentDidUpdate(prevProps: ISwaggerDocsProps, prevState: object) {
+    if (prevProps.apiName !== this.props.apiName && this.props.url) {
+      this.setState({
+        ...this.state,
+        metadata: null,
+        url: this.props.url,
+      });
+      this.loadMetaDataAndRender();
+    } else {
```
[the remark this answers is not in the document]
ATeal: It would fall back to just rendering the swagger ui

Calling `this.renderSwaggerUI` will then check if url is present (in this case no) and assume that json is present as a fall back. If neither are present well then it just renders a blank page, but that's existing behavior:

image

src/components/SwaggerDocs.tsx
```diff
@@ -5,41 +5,109 @@ import { SwaggerPlugins } from './swaggerPlugins';
...
+      this.renderSwaggerUI();
     }
+  }
 
-    public render() {
-        return (
-            <div id="swagger-ui" />
-        );
+  public componentDidUpdate(prevProps: ISwaggerDocsProps, prevState: object) {
```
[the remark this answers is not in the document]
ATeal: This gets called in several ways which is probably why it's confusing.

1) Page change to different api. This would be the first case in the conditional. apiName has changed, so basically reset the state and re-render swagger-ui
2) Version is switched. url is updated and swagger-ui has to re-render again


`this.state.json` would be handled in the else, although it isn't exactly clear that that's the case. Even though it would do the same thing it might be more clear if I add:
```
else if(this.state.json){
  this.renderSwaggerUI();
}
```

Honestly though I _think_ the only scenario where the json prop is used is in testing. I could be wrong about that.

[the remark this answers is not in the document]
ATeal: Curious to hear your thoughts on that

src/components/SwaggerDocs.tsx
ATeal: good call, will remove. 

I blame ruby and thinking about implicit returns :P

src/components/swaggerPlugins/ExtendedLayout.tsx
ATeal: 1) My thought on calling it ExtendedLayout was that in the future we'll likely be changing/tweaking the swaggerui layout further, so I didn't wanna tie it to a specific feature
2) unused, removing

ATeal: Happy to have more discussion on 1) leaving it unresolved for now

src/components/swaggerPlugins/ExtendedLayout.tsx
ATeal: Good call removing.

src/components/swaggerPlugins/VersionActions.ts
ATeal: yup will do

src/components/swaggerPlugins/VersionSelect.tsx
ATeal: good call, updating.

src/e2eHelpers.ts
ATeal: Yup was a result of my local config not picking up the prettier config. Should be fixed when I refactor the other comment!

src/e2eHelpers.ts
ATeal: Yup will refactor

src/components/swaggerPlugins/VersionSelect.tsx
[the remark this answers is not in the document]
ATeal: got this changed

src/components/swaggerPlugins/VersionActions.ts
```diff
@@ -0,0 +1,30 @@
...
+          type: 'API_VERSION_SET',
+        };
+      },
+      updateUrl: (url: string, version: string) => {
+        updateUrlHandler(url, version);
+        return {
+          type: 'API_URL_UPDATED',
+        };
+      },
+      updateVersion: (version: string) => {
```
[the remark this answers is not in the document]
ATeal: Will remove

src/components/swaggerPlugins/VersionReducers.ts
```diff
@@ -0,0 +1,13 @@
+export const VersionReducers = {
+  reducers: {
+    API_METADATA_SET: (state: any, action: any) => {
+      return state.set('apiMetadata', action.payload);
+    },
+    API_VERSION_SET: (state: any, action: any) => {
+      return state.set('apiVersion', action.payload);
+    },
+    API_VERSION_UPDATED: (state: any, action: any) => {
```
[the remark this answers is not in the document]
ATeal: Yup I'll get rid of some of the unused stuff here

src/components/swaggerPlugins/VersionSelect.tsx
```diff
@@ -0,0 +1,76 @@
...
+      .meta.versions.find((metaObject: any) => {
+        return metaObject.version === currentVersion;
+      });
+    this.props
+      .getSystem()
+      .versionActions.updateUrl(this.buildUrlFromMeta(versionMetadata), currentVersion);
+  }
+
+  public buildUrlFromMeta(metaObject: any) {
+    return `${process.env.REACT_APP_VETSGOV_SWAGGER_API}${metaObject.path}`;
```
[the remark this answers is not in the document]
ATeal: hrm not a bad idea, I'll shift some things around.

src/components/swaggerPlugins/VersionSelect.tsx
```diff
@@ -0,0 +1,76 @@
...
+            .versionSelectors.apiMetadata()
+            .meta.versions.map((metaObject: any) => {
+              return (
+                <option value={metaObject.version} key={metaObject.version}>
+                  {this.buildDisplay(metaObject)}
+                </option>
+              );
+            })}
+        </select>
+        <button onClick={e => this.handleButtonClick()}>Select</button>
```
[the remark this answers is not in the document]
ATeal: I'm totally fine with that, but that is what we landed on in discussion with @juliaelman  / @bastosmichael . If they're cool with it I am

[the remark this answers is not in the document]
bastosmichael: @juliaelman picked "Select" when we discussed the design above many other options so I'm going to believe that this is the best compromise we could make with the uxteam on this issue given that "Update Version" or even "Update" could give a myriad of assumptions and impressions about what that button would actually do as well as "Select" covers the space concerns considering that there are flags that can appear next to it such as Internal or Deprecated etc. Unless the UX team tells us that they believe "Update Version" or something else better is better in this case then I'm going to say we can move forward to with "Select" for the time being.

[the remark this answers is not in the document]
kalilsn: I know this was a choice @juliaelman made, but it's an accessibility issue that I noticed when testing this out with a screen reader. Even if we can't fix this before merging this PR, please leave this conversation unresolved because I intend to go back through this PR to put together an issue for fixing accessibility issues around version selection.

PR #133, opened 2019-06-14
Run everything through prettier

kalilsn:
Now that our prettier config is named correctly, I don't see any reason to hold off on doing this. It should reduce the noise in other pull requests.


dvogel:
> LGTM. I'm not a fan of a few of the decisions `prettier` made to expand a single in line into multiple lines, but it will be nice to have everything standardized.

I originally setup the prettier config to standardize on 2-space indentation. I had to choose _some_ tool and prettier was the first one I grabbed. We don't have to stick with it. I'm perfectly fine coming up with an alternative. I just got sick of seeing PRs ping-pong between 2-spaces and 4-spaces.

kalilsn:
I'm also not a huge fan of how aggressively prettier breaks up lines. I would be fine with choosing a new tool, but I'd still like to go ahead and format all our files with it when we do. Unrelated formatting noise is making it pretty difficult to review moderately sized PRs.

nfasulo:
I'm not trying to derail this PR and make us look for a new tool! I just thought if I didn't voice my complaint now, it would always grate on me. :grin:

kalilsn:
>I'm not trying to derail this PR and make us look for a new tool

I opened this PR because it was bugging me and I wanted to see what people thought. Choosing a new tool feels like a perfectly valid outcome of that, not a derailment!

mjlumetta:
> LGTM. I'm not a fan of a few of the decisions prettier made to expand a single in line into multiple lines, but it will be nice to have everything standardized.

i'm not sure this happens that much? i'd like to respectfully disagree with @nfasulo here - it seems like it only happens once or twice (`App.e2e.ts` and maybe `index.ts`?), and i feel that it generally adds readability in those cases. but ofc it is a matter of preference, this is just mine.

PR #135, opened 2019-06-17
Initial framework for multiple FHIR docs - #2124

mjlumetta:
PR for the multiple FHIR docs. See [vets-contrib #2124](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2124).

Refactors `apiDefs` to allow for multiple doc sources for one API. If that is the case, then the documentation for that API renders a tab layout with one tab panel per doc source.


src/apiDefs.ts
```diff
@@ -97,8 +98,17 @@ export const apiDefs : IApiCategories = {
     apis: [
       {
         name: 'Veterans Health API',
-        openApiDocUrl: "https://staging-api.va.gov/services/argonaut/v0/openapi.json",
+        openApiDocUrl: [
```
jasalisbury: Seems easy for this array and the array below (`tabLabels`) to get out of sync. Any way we can store the `tabLabel` and corresponding `openApiDocUrl` in a way where we don't have to worry about the order of two arrays?

mjlumetta: yeah i was thinking about this because you're right about that. i think the thing is, because we're using typescript, that it makes sense to make the array version of `openApiDocUrl` to be an array of objects of form `{ url: string, label: string}`. that would simplify a lot

dvogel:
It seems like these changes could mix strangely with the changes @ATeal is making for the metadata URLs [the PR #132](https://github.com/department-of-veterans-affairs/developer-portal/pull/132). Could you try applying these changes to another branch that uses the `art.version-selector` branch as a base and then setup a time to review the behavior from the user's POV?

src/apiDefs.ts
dvogel: In addition to that change, could you make the `openApiDocUrl` field consistently one shape? The "either string or array" type can cause a lot of downstream incidental complexity. I'd like to start thinking about the data in `apiDefs.ts` as a database. That's in part because it's likely to move to a database in the future, but it's also because that will help us make sure we're writing components that will deal with the proper types. As it's written, if you're testing a view component that assumes one of the types (string vs array) it will be incorrect globally but it will seem correct or incorrect depending on which API you test it with. A singular type for that field can help avoid the need to test comprehensively.

mjlumetta:
> It seems like these changes could mix strangely with the changes @ATeal is making for the metadata URLs the PR #132. Could you try applying these changes to another branch that uses the art.version-selector branch as a base and then setup a time to review the behavior from the user's POV?

@dvogel Alex has merged to master now, so had to do anyway, but without the problem of making concurrent changes. working on this now

src/apiDefs.ts
mjlumetta: done in latest push

mjlumetta:
@juliaelman and i will pair to do a visual polish, also adding her as a reviewer

jasalisbury:
@mjlumetta think we can close this one.

mjlumetta:
@jasalisbury yup definitely

mjlumetta:
@jasalisbury yes definitely will do

PR #139, opened 2019-06-19
Enable community care in prod

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2139

Let me know if I missed any other feature flag I need to set to enable signup for community care.


dvogel:
This change is technically correct. Let's get Jason's opinion on whether this should go live before merging though.

mjlumetta:
couple of questions: 

* if this is to enable Community Care for prod, should it also be enabled in staging?
* regardless of the answer to the previous question, should this change be made by changing the default value in `App.tsx`? the setting `REACT_APP_COMMUNITY_CARE_API_ENABLED` is explicitly set to false in `.env.staging`

jasalisbury:
> if this is to enable Community Care for prod, should it also be enabled in staging?
Yeah I think so

> regardless of the answer to the previous question, should this change be made by changing the default value in App.tsx? the setting REACT_APP_COMMUNITY_CARE_API_ENABLED is explicitly set to false in .env.staging
Yes to this as well

dvogel:
> > if this is to enable Community Care for prod, should it also be enabled in staging?
> 
> Yeah I think so

Ideally, yes. Unfortunately we don't have a backend environment to proxy to right now. I'm tracking that in [the new status document](https://github.com/department-of-veterans-affairs/vets-contrib/pull/2472/files?short_path=83d2ddd#diff-83d2dddfe81bdbaac92d6bdbcb156c3c). For now I think this can proceed before staging.

jasalisbury:
Community care was adding in https://github.com/department-of-veterans-affairs/developer-portal/pull/160

PR #144, opened 2019-07-03
Adding Claims V1 to submission Oauth request process Second Attempt

bastosmichael:
Reverts department-of-veterans-affairs/developer-portal#143

Also resolves ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/2524 once and for all.


src/containers/ApplySuccess.tsx
```diff
@@ -93,7 +93,7 @@ function ApplySuccess(props: IApplication) {
                     ? null
                     : <ApiKeyNotice email={email} token={token} selectedApis={selectedApiNames(apis)} />;
 
-  const oAuthNotice = ((apis.health || apis.verification) && clientID && clientSecret)
+  const oAuthNotice = ((apis.health || apis.verification || apis.claims) && clientID && clientSecret)
```
bastosmichael: @kalilsn from my understanding of the code, this is all I needed to add in order to get an application out to the user upon registration, what am I missing here?

src/actions/index.ts
```diff
@@ -40,6 +40,10 @@ export interface IToggleBenefitsApi extends Action {
   type: constants.TOGGLE_BENEFITS_CHECKED;
 }
 
+export interface IToggleClaimsApi extends Action {
```
mjlumetta: this question applies throughout the PR, but i'll leave it here. is the Claims API now being considered separate from Benefits, that is, not categorized under it? because that implies changes elsewhere. the biggest concern is that this form has to follow the logic of `apiDefs.ts` or errors will occur. for example, if you select only Claims, you won't see the field for the OAuth redirect url that conditionally renders when an OAuth API is selected, because `apiDefs` doesn't think `claims` refers to/is part of an OAuth API

src/containers/ApplyForm.tsx
```diff
@@ -62,7 +64,8 @@ const mapStateToProps = (state : IRootState) => {
 // Mapping from the options on the form to url fragments for APIs
 const formFieldsToFragments = {
   appeals: 'appeals',
-  benefits: ['benefits', 'claims'],
+  benefits: 'benefits',
+  claims: 'claims',
```
mjlumetta: similar comment here - you can separate them into separate fields, but that doesn't mean the developer portal considers `claims` an OAuth API

src/containers/ApplySuccess.tsx
mjlumetta: related to the other comments, seems like this should really reference `apiDefs` as the source of truth for determining which APIs are OAuth APIs and which are not. not your fault, something we've been moving toward more recently as we've grown, but you should be able to check out `includesOauthAPI` or something similar. in either case, this should reference that data

src/actions/index.ts
bastosmichael: Claims is now seperate from Benefits because Claims V1 is now Oauth.

src/containers/ApplyForm.tsx
bastosmichael: Claims is now seperate from benefits and is Oauth Based in the V1 version.

src/containers/ApplySuccess.tsx
bastosmichael: again claims and benefits are separate in terms of auth.

src/actions/index.ts
bastosmichael: So I updated apiDefs to also check api's for whether they are oAuth, is there anything else I'm missing?

src/containers/ApplyForm.tsx
bastosmichael: So I updated apiDefs to also check api's for whether they are oAuth, is there anything else I'm missing?

src/containers/ApplySuccess.tsx
bastosmichael: So I updated apiDefs to also check api's for whether they are oAuth, is there anything else I'm missing?

src/apiDefs.ts
```diff
@@ -60,11 +62,12 @@ export const apiDefs: IApiCategories = {
...
       {
         metadataUrl: `${process.env.REACT_APP_VETSGOV_SWAGGER_API}/services/claims/metadata`,
-        name: 'Benefits Claims',
+        name: 'Claims',
+        oAuth: true,
         openApiDocUrl: `${process.env.REACT_APP_VETSGOV_SWAGGER_API}/services/claims/docs/v0/api`,
         shortDescription: 'Submit and track claims',
         urlFragment: 'claims',
-        vaInternalOnly: true,
+        vaInternalOnly: false,
```
ATeal: Maybe not a problem for this PR, but given that internal / external (and maybe auth type should?) exist on the metadata endpoint shouldn't we actually call it to determine some of this stuff? 

In general we have the question of where some of this data should live. Don't want to have a situation where by if we add another version we have to duplicate some of this info across multiple places and projects. Something to consider.

src/apiDefs.ts
```diff
@@ -233,6 +236,18 @@ function categoriesFor(apiList: string[]): IApiCategory[] {
...
+        apis.add(api);
+      }
+    }
+  }
+  return Array.from(apis);
+}
+
 export function includesOauthAPI(apiList: string[]): boolean {
-  return categoriesFor(apiList).some(category => !category.apiKey);
+  return categoriesFor(apiList).some(category => !category.apiKey) || apisFor(apiList).some(api => api.oAuth || false);
```
ATeal: undefined should be false-y so I don't think you have to do ` || false`

src/apiDefs.ts
```diff
@@ -233,6 +236,18 @@ function categoriesFor(apiList: string[]): IApiCategory[] {
   return Array.from(categories);
 }
 
+function apisFor(apiList: string[]): IApiCategory[] {
```
ATeal: how is this any different from `categoriesFor`?

src/apiDefs.ts
dvogel: IMO all of this data (both the apiDefs.ts and the data served via the metadata endpoints) should live in a dedicated backend that acts as an orchestration point between the developer portal and the various components that will need to coordinate for the various self-service features. That's one of the reasons we've extracted all of the API metadata into `apiDefs.ts`. It's a precursor to moving it into a database later.

src/apiDefs.ts
```diff
@@ -31,6 +32,7 @@ export interface IApiCategory {
   readonly shortDescription: string;
   readonly longDescription: string;
   readonly releaseNotes?: React.StatelessComponent;
+  readonly oAuth?: boolean;
```
dvogel: Making this flag available on both IApiDescription and IApiCategory opens the door to a logical inconsistency wherein the API's `oAuth?` flag disagrees with the category's flag. In line with the effort to push this data toward a database, this should be denormalized with the flag only existing on each IApiDescription.

There's a second inconsistency here, which is that the `apiKey == false` currently means that the category requires OAuth. Since you haven't changed the function that submits the API request to sign the developer up, the `oAuth?` flag you've added won't affect whether an OAuth client application is created within Okta.

src/apiDefs.ts
ATeal: Hrm, that's kind of the idea behind the metadata endpoints, but given that on the developer portal we're responsible for serving more than just vets-api api's documentation, I could see it being lifted out to it's own backend.

src/apiDefs.ts
bastosmichael: I specifically tested it and it returns undefined, not false-y and thus fails if missing from the json.

src/apiDefs.ts
bastosmichael: categoriesFor returns and compares the name of the category against the name in the form, apisFor does the same thing but for the apis listed within the categories allowing us to separate one from the other.

src/apiDefs.ts
bastosmichael: @dvogel so Mike L and I talked about this last week but he wanted to have sparkle motion work on a conversion from it being in the category to it being in the description so I told him I'd do something that worked with both for now until @mjlumetta and his team is ready for a refactor unless you'd like me to do the refactor myself in this PR which I can. 

Also on the point of the second inconsistency, I wanted to know what other parts of this am I missing since the Lambda portion has already been pushed up and it sounds like you said I'm still missing the request, I was under the impression that this is where that happened but it sounds like it's somewhere else. 

https://github.com/department-of-veterans-affairs/developer-portal/pull/144/files#diff-0e55b3d90680817f9a70fe2e40d7b7cbR96

If you can point me to where I'm missing this request I'll patch it up and finalize this PR right away. Thanks

src/apiDefs.ts
bastosmichael: @dvogel We had an RFC from Alex (https://github.com/department-of-veterans-affairs/vets-contrib/pull/1895) where everyone agreed on a while back that stated that all metadata would eventually make its way into a metadata end point per API that is currently statically committed on a per controller basis to git and not in a DB that would list things currently also listed in the apiDefs.ts and that we would work towards migrating that over once every API had a metadata end point etc. And it sounds like you also have a vision of what sounds like a different implementation where we potentially load the data already in apiDefs.ts into a postgres DB at some point somewhere and store this data dynamically somewhere thus removing the need for every API to have a metadata end point in the first place? @dhall is this a possible change of direction or do we need to bring back that RFC and discuss this more?

dvogel: @bastosmichael Yeah, this is an extension of the conversation on that RFC. At that time I wasn't able to respond [to @ATeal's invitation to comment](https://github.com/department-of-veterans-affairs/vets-contrib/pull/1895#discussion_r281168393) (sorry!). However, I wouldn't have been arguing for an alternative. I don't see my comment above as a change of direction as much as an extension of the same idea. At the time of that RFC we needed a way for the developer portal to get this information without having to update both the API and the developer portal when a new version was published. We didn't have a dedicated backend for the developer portal or any other appropriate component, so there wasn't an alternative path at that time.

The basic problem of having to coordinate development and deploys of multiple components that each have their own understanding of APIs is a problem broader than what the metadata endpoints were seeking to address. For example, having to coordinate changes to the developer portal and the self-service lambda function is also problematic. As we develop more components, we'll need to solve this coordination problem generally. Having a centralized component that defines parameters of APIs is the simplest approach to that problem. The presence of such a component doesn't necessarily negate the metadata endpoints themselves or the idea behind them. It might make more sense for some of that data to live in the new component. It's likely that one of the parameters of each API would be it's metadata endpoint.

bastosmichael: Understood, so we would still extrapolate data from the apiDefs.ts to the metadata end points but then this component would be where we store the metadata end point links themselves and what associations they have as well as groupings and titles etc. I think that idea deserves another RFC for clarity but I like it as well.

src/apiDefs.ts
```diff
@@ -19,6 +19,7 @@ export interface IApiDescription {
   readonly urlFragment: string;
   readonly shortDescription: string;
   readonly vaInternalOnly: boolean;
+  readonly oAuth?: boolean;
```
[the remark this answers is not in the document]
bastosmichael: Per our discussion keeping this for now until refactor.

src/apiDefs.ts
```diff
@@ -233,6 +235,18 @@ function categoriesFor(apiList: string[]): IApiCategory[] {
   return Array.from(categories);
 }
 
+function apisFor(apiList: string[]): IApiDescription[] {
```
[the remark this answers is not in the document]
bastosmichael: Per our discussion keeping this for now until refactor.

src/apiDefs.ts
```diff
@@ -233,6 +235,18 @@ function categoriesFor(apiList: string[]): IApiCategory[] {
...
+        apis.add(api);
+      }
+    }
+  }
+  return Array.from(apis);
+}
+
 export function includesOauthAPI(apiList: string[]): boolean {
-  return categoriesFor(apiList).some(category => !category.apiKey);
+  return categoriesFor(apiList).some(category => !category.apiKey) || apisFor(apiList).some(api => api.oAuth || false);
```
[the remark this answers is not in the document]
bastosmichael: Per our discussion keeping this for now until refactor.

src/containers/ApplySuccess.tsx
```diff
@@ -36,12 +36,12 @@ interface IOAuthCredentialsNoticeProps {
...
 const apisToEnglishList = (apis: string[]): string => {
   return sentenceJoin(apis.map((k) => {
-    return apiDefs[k].properName;
+    return (apiDefs[k] ? apiDefs[k].properName : k.charAt(0).toUpperCase() + k.slice(1));
   }));
 }
 
 function OAuthCredentialsNotice({ clientID, clientSecret, email, selectedApis } : IOAuthCredentialsNoticeProps) {
-  const apiListSnippet = apisToEnglishList(selectedApis.filter((k) => !apiDefs[k].apiKey));
+  const apiListSnippet = apisToEnglishList(selectedApis.filter((k) => !(apiDefs[k] ? apiDefs[k] : false)));
```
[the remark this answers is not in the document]
bastosmichael: Removed

src/containers/ApplySuccess.tsx
```diff
@@ -36,12 +36,12 @@ interface IOAuthCredentialsNoticeProps {
 
 const apisToEnglishList = (apis: string[]): string => {
   return sentenceJoin(apis.map((k) => {
-    return apiDefs[k].properName;
+    return (apiDefs[k] ? apiDefs[k].properName : k.charAt(0).toUpperCase() + k.slice(1));
```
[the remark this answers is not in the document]
bastosmichael: Changed and removed.

src/apiDefs.ts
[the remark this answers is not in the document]
bastosmichael: Per our discussion keeping this for now until refactor.

PR #145, opened 2019-07-08
Add support for view-only API data for navigation purposes

mjlumetta:
Completes [vets-contrib #2607](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2607). 

Adds support for the concept of a "view only" API within `apiDefs` that can be highlighted in the navigation but will not be counted as a separate API in its own right. Also refactors `apiDefs` to be a directory with multiple files rather than a single large file. 

Some UI questions remain on this ticket. For example:

* What should the URLs for each link be?
* What should be highlighted in the navigation when multiple nav items reference the current page (i.e. FHIR API and Urgent Care API)?


jasalisbury:
@mjlumetta can probably close out this PR

PR #147, opened 2019-07-15
Long description replaced with intro, a mdx document

jasalisbury:

Different formatting requirements for the intro requires use of mdx files. As long description was only used for the intro seems alright to remove it. Removed some dead code I found and fixed an unrelated typo I forgot to commit before now.


src/apiDefs/index.ts
```diff
@@ -41,7 +45,7 @@ export interface IApiCategory {
   readonly name: string;
   readonly overview: React.StatelessComponent;
   readonly shortDescription: string;
-  readonly longDescription: string;
+  readonly intro: React.StatelessComponent;
```
mjlumetta: if there are multiple mdx components being exported from the same source - i.e., `src/content/apiDocs/index.ts` - then i would suggest that we actually define them as their own interface and import them as such. for example: 

```
interface IApiCategoryContent {
  readonly intro:  React.StatelessComponent;
}
```

thoughts? could define the interface either here or in the `src/content/apiDocs/index.ts` file - but either way, export them in that format from there and then we can just be like,, "ah, one line for all the mdx files required per category" in here

src/content/apiDocs/index.ts
```diff
@@ -1,25 +1,33 @@
...
+import HealthIntro from './healthIntro.mdx';
 import HealthOverview from './healthOverview.mdx';
 import HealthStarted from './healthStarted.mdx'; 
 import UrgentCareApiIntro from './urgentCareApiIntro.mdx';
+import VerificationIntro from './verificationIntro.mdx';
 import VerificationOverview from './verificationOverview.mdx';
 import VerificationStarted from './verificationStarted.mdx';
 
-export { BenefitsOverview,
+export { BenefitsIntro,
```
mjlumetta: other end of my `apiDefs/index.ts` comment in here

src/apiDefs/health.ts
```diff
@@ -40,27 +43,28 @@ const healthApis : IApiDescription[] = [
...
         label: 'R4',
-        openApiUrl: 'https://dev-api.va.gov/services/fhir/v0/r4/openapi.json',
+        openApiUrl: 'https://api.va.gov/services/fhir/v0/r4/openapi.json',
       },
       {
         apiIntro: FhirDSTU2ApiIntro,
         key: 'dstu2',
         label: 'DSTU2',
-        openApiUrl: 'https://dev-api.va.gov/services/fhir/v0/dstu2/openapi.json',
+        openApiUrl: 'https://api.va.gov/services/fhir/v0/dstu2/openapi.json',
```
dvogel: It seems to me like these should be using the environment variable that provides the hostname for each environment:

```
REACT_APP_VETSGOV_SWAGGER_API=https://dev-api.va.gov
```

Am I missing something?

jasalisbury: We can't use the environment's swagger api because of staging. Since there are no health APIs in staging the developer portal will break in staging. If you remember to a few weeks ago (maybe a month) we switched using `dev-api.va.gov` and started using `api.va.gov` because you can see the swagger url on the developer portal. We decided it was confusing seeing `dev-api.va.gov` on the production developer portal. Now that I'm thinking about it a bit more we should definitely add a comment explaining why we use `api.va.gov`, because I've asked the same question you are asking at least 2 other times.

mjlumetta: what Josh said is true and the primary reason. i would also add that this is frankly not that big of a deal, imo; even if ppl are visiting the staging and dev env portal for some reason, having the prod URL is still okay because, presumably, they are trying to get to prod eventually. also they can't use staging anyway. so while i agree that consistency is preferable, i don't think this _hurts_.

dvogel: Can we split the different by continuing the pattern (referring to an environment variable) but introduce a new environment variable to reflect the fact that not all APIs can rely on the same host for docs in all environments? E.g. `REACT_APP_VETSGOV_SECONDARY_SWAGGER_API`. My hope is that these concerns will move into the docs service once we start building it. My concern here is making that as easy to move toward in the future.

src/apiDefs/health.ts
```diff
@@ -1,20 +1,30 @@
+/*
+  Note the use of `api.va.gov` instead of `${swaggerHost}` for openApiUrl. Health APIs do not have a staging
+  environment setup, as such we can only use `dev-api.va.gov` or `api.va.gov` for the openApiUrl host. Using
+  `${swaggerHost} would break staging. The swagger url is shown in the UI. In order to avoid the confusion of
+  having a `dev-api.va.gov` url shown in production `dev.va.gov` is used in all developer portal environments.
```
mjlumetta:
```suggestion
  having a `dev-api.va.gov` url shown in production `api.va.gov` is used in all developer portal environments.
```

src/apiDefs/health.ts
```diff
@@ -1,20 +1,33 @@
+/*
+  Note the use of the secondary swagger api instead of the primary for openApiUrl. Health APIs do not have a
+  staging environment setup, as such we can only use `dev-api.va.gov` or `api.va.gov` for the openApiUrl host.
+  The primary swagger api is tied to the environment. The secondary swagger api always points to production.
+  Using the primary swagger api would break staging. The swagger url is shown in the UI. In order to avoid the
+  confusion of having a `dev-api.va.gov` url shown in production `api.va.gov` (the secondary swagger api) is
+  used in all developer portal environments for health documentation.
+*/
+
```
dvogel: :heart: :heart: :heart: :heart: :heart:

PR #148, opened 2019-07-15
Agronaut API is deprecated

jasalisbury:

Merges into the [Long description replaced by mdx file branch](https://github.com/department-of-veterans-affairs/developer-portal/pull/147#pullrequestreview-262069986). Adds the ability to mark an API as deprecated and does so for the Agronaut API if the new FHIR API is enabled.


mjlumetta:
this deprecation flash message is kind of wonky-looking:

WonkyDeprecationMessage

specifically:

* the icon is big and i'm not sure that it's placed well? see `background-image` on `.usa-alert-info`, as well as `background-position-[x|y]` and `background-size`. you can style this however you want, but i'd suggest tweaking those. you'll also probably want to remove the `margin-top` on the `h3`
* the URLs are plain text, not the `code` style. do we want this to be a markdown component so we can take advantage of that here?
* need either a `margin-bottom` on `.usa-alert` or a `margin-top` on `.swagger-ui`

src/containers/Explore.tsx
```diff
@@ -46,25 +47,50 @@ class Explore extends React.Component<IExploreProps, IExploreState> {
...
   public render() {
     let docsDom: JSX.Element | null = null;
+    let depreacted: JSX.Element | null = null;
+    let header: JSX.Element | null = null;
+
     const api = this.getApi();
     const category = this.getCategory();
     if (api != null) {
       docsDom = this.renderApiDocs(api);
+      if (api.depreacted) {
```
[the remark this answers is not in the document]
jasalisbury: Yeah a lot of other conditional content is shown based on if there is content, not off an explicit boolean. Makes sense to me to remove the explicit boolean.

jasalisbury:
> * the URLs are plain text, not the `code` style. do we want this to be a markdown component so we can take advantage of that here?

I like the idea of making it a markdown component, I'll play around with that idea and see what it looks like.

We can see what styling issues there are if we switch to a markdown component, but will probably need you @mjlumetta or @juliaelman to take care of them.

src/content/apiDocs/health/argonautDeprecationNotice.mdx
```diff
@@ -0,0 +1,4 @@
+## Deprecation Notice
+Consumers currently using the <code>https://api.va.gov/services/argonaut/v0</code> (and <code>https://dev-api.va.gov/services/argonaut/v0</code>) endpoint, please migrate your applications to use the new Veteran Health API (FHIR) endpoints: <code>https://api.va.gov/services/fhir/v0/argonaut/data-query/</code> and <code>https://dev-api.va.gov/services/fhir/v0/argonaut/data-query/</code>
```
mjlumetta: so - this does work, but it's not _quite_ what i had in mind. what this does is defers the styling to the _HTML_ level, which does create the UI we want! but i was thinking, since we are in a markdown file, that we would use the markdown `` style ([reference](https://www.markdownguide.org/basic-syntax/#code)). i feel like that would be more semantically correct, imo. but that's kinda on me, i didn't want to say "the backtick thingy"

src/apiDefs/health.ts
```diff
@@ -4,10 +4,13 @@ import {
   FhirArgonautApiIntro,
   FhirDSTU2ApiIntro,
   FhirR4ApiIntro,
+  HealthArgonautDeprecation,
   UrgentCareApiIntro,
 } from "../content/apiDocs";
 
 const isNewFhirApiEnabled =  process.env.REACT_APP_FHIR_API_ENABLED === 'true';
+const argonautDeprecatedDesc = 'Both the legacy API endpoints and this legacy documentation will no longer be accessible beginning Oct 1, 2019.';
```
dvogel: Is this date legit? Are we committed to this? This makes me wonder if the `IApiDescription` record should have a date field indicated the end of life for an API. The portal could then automatically hide all EOL APIs on the correct day without future code changes.

jasalisbury: We will probably want to double check the date. It is the date from the mockups. I like the idea of having APIs removed without code changes. Now that I think about it there are other changes we'd need to do once the API is deprecated (remove routes from kong, monitoring, statuspage updates, sure there are more) so we will definitely want to confirm that date.

src/content/apiDocs/health/argonautDeprecationNotice.mdx
jasalisbury: oh yeah didn't even know you could do that. Just looked at how the authorization page shows links... I'll update to use backticks.

jasalisbury:
Styling of the deprecation notice info box was done in the [quickstart PR](https://github.com/department-of-veterans-affairs/developer-portal/pull/149)

src/content/apiDocs/health/argonautDeprecationNotice.mdx
```diff
@@ -0,0 +1,4 @@
+## Deprecation Notice
```
mjlumetta: minor but i _think_ you usually want a blank line after headers

PR #149, opened 2019-07-15
Add quickstart for health APIs

jasalisbury:

Merges into the branch that [deprecates the legacy Argonaut API](https://github.com/department-of-veterans-affairs/developer-portal/pull/148). Adds a quickstart for health APIs, navigable via the left nav. If the FHIR API isn't enabled for the environment quickstart will not be shown. This should be the last health doc update PR.


src/components/Quickstart.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export class Quickstart extends React.Component<RouteComponentProps<IApiNameParam>, {}> {
+  public render() {
+    const { apiCategoryKey } = this.props.match.params;
+    const { quickstart, properName } = apiDefs[apiCategoryKey];
+
+    return (
+      <div role="region" aria-labelledby="api-documentation">
+        <PageHeader halo={properName} header="Quickstart" />
+        {quickstart && quickstart({})}
```
jasalisbury: What should we do if quickstart is null here? Currently will display the quickstart header with no content.

src/containers/ExploreDocs.tsx
```diff
@@ -80,19 +80,32 @@ function OAuthSideNavEntry(apiCategoryKey: string, apiCategory: IApiCategory) {
...
+  );
+}
+
 function SideNavCategoryEntry(currentUrl: string, apiCategoryKey: string, apiCategory: IApiCategory) {
   const subNavLinks = apiCategory.apis.map(api => {
     return SideNavApiEntry(apiCategoryKey, api);
   });
 
   const authorizationEntry = apiCategory.apiKey ? null : OAuthSideNavEntry(apiCategoryKey, apiCategory);
+  const quickstart = apiCategory.quickstart ? QuickstartNavEntry(apiCategoryKey) : null;
```
dvogel: A small change to help this read easier would adding a suffix to this local `quickstart` variable (e.g. `quickstartEntry`) to avoid confusion with the `IApiCategory` field of the same name. Typescript makes it so easy to alias interface fields locally that it's usually best to avoid overloading them IMO.

src/content/apiDocs/healthQuickstart.mdx
```diff
@@ -0,0 +1,38 @@
+VA’s Health API allow a user/application to make queries that will return Veteran health records stored in a synthetic Veteran health database without interacting with or understanding the inner workings of VA systems.
```
dvogel: Seems like either "Health API" or "allow" should be plural. @juliaelman can you clarify whether [this text in the mock up](https://app.moqups.com/F4fi0Zw2mT/view/page/a77cc1ef0) has typo or not?

src/content/apiDocs/healthQuickstart.mdx
```diff
@@ -0,0 +1,38 @@
...
+The following basic information should be provided:
+
+ * Your Name
+ * Email address
+ * Organization name
+ * OAuth Redirect URI
+
+### What happens after I am approved?
+
+You will receive an email from the VA API team notifying you of approval. You will then receive a new Client ID and Secret for your application. The base URI for the Health API endpoint is: https://dev-api.va.gov/services/argonaut/v0/. Accordingly, the FHIR conformance statement can be retrieved from https://dev-api.va.gov/services/argonaut/v0/metadata.
```
dvogel: Since this is tied to the new FHIR APIs, I'm guessing this API base URL is supposed to be `/services/fhir`. Let's leave it for now but be sure to follow up with the Health team before this gets merged to master (I assume we'll have them look at a review instance?) or before the feature flag is enabled in production.

src/content/apiDocs/healthQuickstart.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+You will also be provided with a set of test accounts to use that will allow you to access specific synthetic data patient records.
+
+## Developer Guidelines
+
+Below are guidelines you should follow to be successful in your VA API integration.
+
+### Authorization
+
+The Health API require you to leverage our authentication protocol. We have a specific guide for authorization, [go to our Authorization documentation](/explore/health/docs/authorization).
```
dvogel: Another singular vs plural issue here with "require".

src/components/Quickstart.tsx
dvogel: To answer this narrowly, I think a redirect to the category overview would be most appropriate. Hopefully the quickstart link for a category wouldn't be generated in the nav unless the `quickstart` field was non-null and thus any URLs to it would likely be the result of a bookmark or external link to a now-deprecated page.

Looking at this more broadly, I think there's two items solving three problems here and we could introduce a new boundary to clear things up a bit. The `Quickstart` class here is meant to handle building the content for a category of pages. [My reading of the original components vs containers](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) article suggests there might be two classes here (despite the admittedly small size of the class). The redirect suggestion I made above feels a little weird because in general you wouldn't expect components (as defined in that article) to redirect to another page. At the same time, most of the render method here is concerned with the appearance of the quickstart page. It ensures that the content for each quickstart page has a consistent structure around it. Put a pin in that for a moment.

Since this class is called `Quickstart` and the `IApiCategory` field is named `quickstart` and the value of that field (for this particular category) is an instance of `HealthQuickstart`, it's tempting to think that `HealthQuickstart` is a subclass or specialization of `Quickstart` and wonder why `quickstart` can't always be rendered. None of those inferences align with the role those pieces are playing. So that begs the question, is there a different arrangement of this code that addresses all of these things?

I'm curious whether anyone else thinks this would be clearer:

* `IApiCategory.quickstartContent` -- a reference to the same markdown component as today, just a new name.
* `containers/QuickstartPage` -- handles lookup of content and either redirects or renders `QuickstartWrapper` with an attribute `content=quickstartContent`.
* `components/QuickstartWrapper` -- same `return` value as the existing class but since it's no longer doing the lookup `content` cannot be `null`.

*Side note: Please don't take that containers vs components article as gospel. I never agreed with it 100% and the author doesn't even agree with it entirely anymore but to the extent we've followed it so far it has been helpful, so we should continue to follow it as much as it makes sense until we decide to abandon the advice altogether.*

mjlumetta: okay - i think the structure that @dvogel suggested for Quickstart is basically what we do with `ApiKeyAuth`. i think we can do that, but i could also see it going the other way if you wanted (ie just adding a `Redirect` here if `!quickstart`)

src/content/apiDocs/index.ts
```diff
@@ -9,7 +9,8 @@ import FhirDSTU2ApiIntro from './fhirDSTU2ApiIntro.mdx'
 import FhirR4ApiIntro from './fhirR4ApiIntro.mdx';
 import HealthIntro from './healthIntro.mdx';
 import HealthOverview from './healthOverview.mdx';
-import HealthStarted from './healthStarted.mdx'; 
+import HealthQuickstart from './healthQuickstart.mdx';
+import HealthStarted from './healthStarted.mdx';
```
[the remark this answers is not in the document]
jasalisbury: I could see having each API category having it's own directory.

src/components/Quickstart.tsx
jasalisbury: I think splitting the redirect/lookup of content from the display of the quickstart page is the way to go. 

@mjlumetta I like the idea of having a content interface to hold all the mdx components. I'll think about how to add it to all these PRs but we may need to wait until this batch is merged in before adding it.

src/content/apiDocs/healthQuickstart.mdx
jasalisbury: @dvogel @mjlumetta @kalilsn I'm going to gather up any copy comments we have (that don't have obvious fixes) and move them over to the mockup for answers.

src/components/QuickstartWrapper.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import PageHeader from './PageHeader';
+
+interface IQuickstartWrapperProps {
+  halo: string,
+  quickstartContent: React.StatelessComponent
```
mjlumetta: semicolons after these properties, not commas/whitespace

src/components/QuickstartWrapper.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export default class Quickstart extends React.Component<IQuickstartWrapperProps> {
```
mjlumetta: if you're going to use the class, this might be a good time to try out using `React.PureComponent`. or this would be a good candidate for a functional component (it is a pure component, regardless)

src/containers/QuickstartPage.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { apiDefs } from '../apiDefs';
+import QuickstartWrapper from '../components/QuickstartWrapper'
+import { IApiNameParam } from '../types';
+
+export default class QuickstartPage extends React.Component<RouteComponentProps<IApiNameParam>, {}> {
+  public render() {
+    const { apiCategoryKey } = this.props.match.params;
+    const { content, properName } = apiDefs[apiCategoryKey];
+    const { quickstart: quickstartContent } = content;
```
mjlumetta: can deconstruct inline if you're comfortable with that

src/content/apiDocs/health/healthQuickstart.mdx
```diff
@@ -0,0 +1,38 @@
...
+The following basic information should be provided:
+
+ * Your Name
+ * Email address
+ * Organization name
+ * OAuth Redirect URI
+
+### What happens after I am approved?
+
+You will receive an email from the VA API team notifying you of approval. You will then receive a new Client ID and Secret for your application. The base URI for the Health API endpoint is: https://dev-api.va.gov/services/argonaut/v0/. Accordingly, the FHIR conformance statement can be retrieved from https://dev-api.va.gov/services/argonaut/v0/metadata.
```
mjlumetta: should these links get any special treatment? i understand that the developer would not want to _navigate_ to these, but should they be in the `code` style?

DeadLinksInMarkdown

i guess maybe this is a @juliaelman question. iirc, the mocks don't have anything distinguished in that markdown format, and i'm curious what, if anything, should be in that style.

mjlumetta:
the Health API intro is not feature-flagged in prod:

HealthApiIntroProd

tbh this is probably on me tho!! so, irony

jasalisbury:
> the Health API intro is not feature-flagged in prod

Thanks for pointing it out!

src/components/QuickstartWrapper.tsx
jasalisbury: I'll take a look at that tomorrow!

src/containers/QuickstartPage.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Redirect, RouteComponentProps } from 'react-router';
+
+import { apiDefs } from '../apiDefs';
+import QuickstartWrapper from '../components/QuickstartWrapper'
+import { IApiNameParam } from '../types';
+
+export default class QuickstartPage extends React.Component<RouteComponentProps<IApiNameParam>, {}> {
+  public render() {
+    const { apiCategoryKey } = this.props.match.params;
+    const { content: { quickstart: quickstartContent }, properName } = apiDefs[apiCategoryKey];
```
dvogel: `content.quickstart` being unwrapped into `quickstartContent` makes me happier than it really should. You've really brought balance to the force :)

PR #151, opened 2019-07-17
Add Community Care Eligibility OpenAPI spec - #2057

mjlumetta:
Adds the OpenAPI JSON for Community Care to complete [vets-contrib #2057](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2057).

### DIFF
--- a/src/apiDefs/health.ts
+++ b/src/apiDefs/health.ts
@@ -17,7 +17,7 @@ const healthApis : IApiDescription[] = [
     docSources: [
       {
         apiIntro: CommunityCareApiIntro,
-        openApiUrl: '',
+        openApiUrl: 'https://dev-api.va.gov/services/community-care/v0/eligibility/openapi.json',
       },
     ],
     name: 'Community Care Eligibility API',


src/apiDefs/health.ts
```diff
@@ -17,7 +17,7 @@ const healthApis : IApiDescription[] = [
     docSources: [
       {
         apiIntro: CommunityCareApiIntro,
-        openApiUrl: '',
+        openApiUrl: 'https://dev-api.va.gov/services/community-care/v0/eligibility/openapi.json',
```
jasalisbury: One thing I just noticed is we hardcode the host for health API docs. We probably want to use `swagger_host` for all of these APIs.

jasalisbury: Granted we can't do that right now for community care, I don't think it has routes in production.

mjlumetta: yeah, but i think the bigger problem is that none of them exist in staging? we could fix that but it's the issue atm

NoStagingHealthApiDocs

jasalisbury: gah you are right. You'd think I'd remember this is at least the 2nd time I've brought this up. :confused:

mjlumetta: hmm maybe can make a ticket to add them or something, idk

jasalisbury: I do recall we changed the current health api from using dev to using prod because the url for the swagger doc shows up in the UI. The thinking being seeing a `dev-api` url in prod is confusing.

PR #152, opened 2019-07-19
Create support pages

whuang85:
This resolves tickets:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2456
https://github.com/department-of-veterans-affairs/vets-contrib/issues/1903


nfasulo:
Is this really still WIP or is code review the only thing standing between this PR and a merge? Asking for a lazy friend. :smile:

whuang85:
> Is this really still WIP or is code review the only thing standing between this PR and a merge? Asking for a lazy friend. 😄

Lol, forgot to remove

whuang85:
> > Is this really still WIP or is code review the only thing standing between this PR and a merge? Asking for a lazy friend. 😄
> 
> Lol, forgot to remove

Well on second thought, I think it should be as this won't be released without a contact us page

src/containers/SupportOverview.tsx
```diff
@@ -0,0 +1,35 @@
...
+const headerProps = {
+  description: "Welcome to support for the VA API program. You can visit our FAQ page for answers to common questions. If you need to submit a support request or report a bug, you can contact us via Github. For general feedback use our 'Contact Us' form. Our customer support team is happy to help and will respond within one business day.",
+  header: "Support",
+};
+
+export default class SupportOverview extends React.Component<any, any, any> {
+
+  private cardsSections = this.props.sections.map((section: any) => {
+    return (
+      <NavLink key={section.id} to={`/support/${section.id}`} className="va-api-card">
```
mjlumetta: hmm, interesting choice to use the `NavLink` here. any particular reason that you chose this over the existing `ApiCard` component (which, to be fair, might need to be renamed to `Card`)?

src/components/GroupedAccordions.tsx
```diff
@@ -0,0 +1,77 @@
...
+    "width": "100%",
+}
+
+const headerStyle = {
+    "alignItems": "center",
+    "display": "flex",
+    "justifyContent": "space-between",
+}
+
+export interface IQuestion {
```
mjlumetta: this interface, and passing an array of questions to `GroupedAccordions` as props, doesn't make sense here. the questions are business logic that belongs in `src/containers` somewhere, whereas this is a display component that could be reused for other purposes. instead, you could rename this interface `IPanelContent` or something and then name the properties `title` and `body` or something like that.

meanwhile, if you want to wrap the accordions as questions, you can define a `SupportQuestions` component in `src/containers` that takes a string title and an `IQuestion[]` as props and renders them as `GroupedAccordions`. or whatever you like

src/components/GroupedAccordions.tsx
```diff
@@ -0,0 +1,77 @@
...
+    answer: any,
```
mjlumetta:
```suggestion
    answer: string | JSX.Element,
```

src/components/GroupedAccordions.tsx
```diff
@@ -0,0 +1,77 @@
+import CollapsiblePanel from '@department-of-veterans-affairs/formation/CollapsiblePanel';
+import * as React from 'react'
+
+declare let window: any;
```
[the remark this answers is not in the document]
whuang85: I need to access the window object to fix this:
https://github.com/department-of-veterans-affairs/va.gov-team/issues/640

I could declare it like this:
`let window: {VetsGov: object}`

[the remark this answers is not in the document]
mjlumetta: gotttcha. i think any of those are fine, including `any` - since it's a hack to fix somebody else's bug, it's lower stakes and imo an appropriate use for `any`. welp, hope they get it fixed soon, and thanks for filling me in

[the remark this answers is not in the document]
whuang85: I think I still need to keep the declare keyword since window is a global variable:
https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html

src/containers/SupportOverview.tsx
whuang85: I was looking at the What's New container to see how they did it. I did not know about the ApiCard component, will switch over.

src/components/GroupedAccordions.tsx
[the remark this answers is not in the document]
mjlumetta: ahhh got it, i didn't know that. thanks for sharing that!

src/components/GroupedAccordions.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    expect(wrapper.find('.form-review-panel').length).toEqual(1);
+    expect(wrapper.find('.usa-accordion-content').length).toEqual(0);
+  });
+
+  it('should toggle panels when  expand all / collapse all clicked', () => {
+    const wrapper = mount(<GroupedAccordions title="title" panelContents={contents}/>);
+    const toggeLink = wrapper.find('.toggle-panels');
+
+    toggeLink.simulate('click', event);
+    expect(wrapper.find('.usa-accordion-content').length).toEqual(1);
```
mjlumetta: this test (suite) is really nice and more enough as is - but maybe, in order to make the distinction between expand all and expand one, add a second panel object to `contents` and test for 2 here?

src/components/GroupedAccordions.tsx
```diff
@@ -0,0 +1,77 @@
...
+          {c.body}
+        </CollapsiblePanel>
+      )
+    });
+  }
+
+  private handleExpandCollapse(event: React.MouseEvent<HTMLElement>) {
+    event.preventDefault();
+    this.setState((prevState: IGroupedAccordionsStates) => ({ allCollapsed: !prevState.allCollapsed }), () => {
+      this.panelRefs.filter(r => r && (r.state.open === this.state.allCollapsed)).map(r => {
```
mjlumetta: why do we need to do this to the `panelRefs` and have the `panelRefs` in the first place? this seems like a bad pattern to engage in; if `CollapsiblePanel` only has an uncontrolled mode and not a controlled mode, maybe it isn't the right component for this job? if it has a controlled mode, we should definitely use that.

src/containers/Support.tsx
```diff
@@ -0,0 +1,78 @@
+import * as classNames from 'classnames';
+import * as React from 'react'
+import { Route } from "react-router";
+import { NavLink } from 'react-router-dom';
+import SupportFAQ from './SupportFAQ';
+import SupportOverview from './SupportOverview';
+
+export interface ISection {
+  component: React.ComponentType
```
mjlumetta: semi-colon, but perhaps more importantly - will this be a component other than `SupportFAQ` ever? if not, then we should probably remove this and replace it with `SupportFAQ` where it's consumed.

src/containers/Support.tsx
```diff
@@ -0,0 +1,78 @@
...
+  description: string;
+  id: string;
+  name: string;
```
mjlumetta: interface is looking good - one small touchup that we use elsewhere is modifying its members with `readonly` when we expect data to be static. i think that could apply here; see `src/apiDefs/index.ts` for examples.

src/containers/Support.tsx
```diff
@@ -0,0 +1,78 @@
...
+  );
+}
+
+export default class Support extends React.Component {
+
+  public render() {
+    return (
+      <div className="Explore">
+        <section className="usa-section">
+          <div className="Explore-main usa-grid">
```
mjlumetta: `classNames` would be useful here as well

src/containers/Support.tsx
```diff
@@ -0,0 +1,78 @@
...
+          </div>
+        </section>
+      </div>
+    )
+  }
+
+  private createSubRoutes() {
+    return sections.map((section, i) => {
+      return (
+        <Route key={i} exact={true} path={`/support/${section.id}`} component={section.component} />
```
mjlumetta:
```suggestion
        <Route key={section.id} exact={true} path={`/support/${section.id}`} component={section.component} />
```

src/containers/SupportFAQ.tsx
```diff
@@ -0,0 +1,136 @@
...
+  },
+]
+
+const headerProps = {
+  description: "We've compiled a list of FAQs for the VA API program; our goal is to get your question answered as soon as possible. If you need further help\u2014or have comments or feedback\u2014please contact us via Github or use our short 'Contact Us' form. Our customer support team will be happy to help.",
+  halo: "Support",
+  header: "FAQ",
+};
+
+interface ISupportQuestions {
```
mjlumetta:
```suggestion
interface ISupportQuestionsProps {
```

src/containers/SupportOverview.tsx
```diff
@@ -0,0 +1,33 @@
...
+  header: "Support",
+};
+
+interface ISupportOverviewProps {
+  sections: ISection[];
+}
+
+export default class SupportOverview extends React.Component<ISupportOverviewProps> {
+  
+  private cardsSections = this.props.sections.map((section: ISection) => {
```
mjlumetta: move into `render` method

src/components/GroupedAccordions.tsx
whuang85: Currently it is not controllable, please correct me if I'm wrong.
https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation-react/src/components/CollapsiblePanel/CollapsiblePanel.jsx

I assumed we are supposed to be using the formation components as much as possible so wasn't sure what the policy is for creating my own version which will be the same except controllable through props.

src/containers/Support.tsx
whuang85: In the near future we will be creating a contact us container component

src/components/GroupedAccordions.tsx
mjlumetta: we actually rely on formation very little for React components. we do rely on `formation` itself (as opposed to `formation-react`) for styles more frequently, as we are trying to keep the UI in line with VA and USWDS practices. *but* would generally say that for React components, you can feel free to create your own in any given case. 

i would say that, since you've already designed it this way, it's fine to move ahead with this solution. but if you want to rewrite it with a controlled component of our own, that's cool too. mostly was just alarmed to see a heavy reliance on refs coupled with using those refs to set state - in general that seems bad, but if there's an explanation, i'm ok with it. there are plenty of things in the dev portal that are less-than-ideal in the short term.

willhuang85: How do you feel about me creating a pull request to add the open/close state to be controllable in formation-react? Maybe in the future when we update our formation library to the latest then we can refactor this.

whuang85: How do you feel about me creating a pull request to add the open/close state to be controllable in formation-react? Maybe in the future when we update our formation library to the latest then we can refactor this.

src/components/GroupedAccordions.tsx
```diff
@@ -0,0 +1,77 @@
...
+  public componentWillUnmount() {
+    this.panelRefs = [];
+  }
+
+  public render() {
+    return (
+      <section className="va-grouped-accordion">
+        <div className="va-grouped-accordion-header">
+          <h2>{this.props.title}</h2>
+          <a className="toggle-panels" href='javascript:void(0)' onClick={(event) => this.handleExpandCollapse(event)}>{this.checkState()}</a>
```
mjlumetta: a `<button/>` or another element would be preferable to an anchor tag, since this isn't a link

src/components/GroupedAccordions.tsx
```diff
@@ -0,0 +1,77 @@
...
+        </div>
+        {this.createPanels()}
+      </section>
+    );
+  }
+
+  private checkState() {
```
mjlumetta: would prefer a more descriptive method name for this, something relating it to the button/toggle label

src/components/GroupedAccordions.tsx
mjlumetta: i think that would be great if you have the bandwidth! definitely an improvement. my guess would be that `vets-website` only needs the uncontrolled version, so that would be a big step up to, you know, make this more useful as a foundational component.

src/containers/Support.tsx
```diff
@@ -0,0 +1,78 @@
+import * as classNames from 'classnames';
+import * as React from 'react'
+import { Route } from "react-router";
+import { NavLink } from 'react-router-dom';
+import SupportFAQ from './SupportFAQ';
+import SupportOverview from './SupportOverview';
+
+export interface ISection {
+  readonly component: React.ComponentType
```
mjlumetta:
```suggestion
  readonly component: React.ComponentType;
```

src/components/GroupedAccordions.tsx
whuang85: I would need to style the button as a link then, is that preferable over just using an anchor tag?

mjlumetta: yup, that's good and expected

src/containers/Support.tsx
```diff
@@ -0,0 +1,78 @@
...
+          Overview
+        </NavLink>
+      </li>
+      {navSections}
+    </ul>
+  );
+}
+
+export default class Support extends React.Component {
+
```
mjlumetta: extra line

src/containers/SupportFAQ.tsx
```diff
@@ -0,0 +1,136 @@
...
+    }
+  });
+
+  return (
+    <GroupedAccordions panelContents={content} title={props.title} />
+  )
+}
+
+export default class SupportFAQ extends React.Component {
+
```
mjlumetta: extra line

src/containers/SupportOverview.tsx
```diff
@@ -0,0 +1,34 @@
...
+  header: "Support",
+};
+
+interface ISupportOverviewProps {
+  readonly sections: ISection[];
+}
+
+export default class SupportOverview extends React.Component<ISupportOverviewProps> {
+  
+  public render() {
```
mjlumetta: extra lines above and below

PR #146, opened 2019-07-12
Add urgent care documentation

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2607

This PR does a few things:

 * Splits out API definitions into separate files
 * Adds an API intro section to be shown above the swagger doc but below the title
 * Hides the title coming from swagger
 * Adds tabs to APIs that have multiple open API specs
 * Adds Urgent Care to the health API section

Some explanation on why we hide swagger's title... We need to add an intro section for health APIs. This intro is outside of swagger. The title for the API is currently placed on the page from the swagger doc, which means the title will show up under the intro section. `API Documentation` used to be at the top of all documentation page. Now it is the title of the API the user is viewing. This update means we no longer need to see the title from swagger, it is already shown at the top of the page.

I may be able to answer some questions on the PR but we may need to wait for @mjlumetta for most. This PR should have pages match the mockup for [Urgent Care](https://app.moqups.com/F4fi0Zw2mT/view/page/a8e73dd77) and the [Overview Page](https://app.moqups.com/F4fi0Zw2mT/view/page/a9f3639d2). The new FHIR pages should also match [the mockups](https://app.moqups.com/F4fi0Zw2mT/view/page/a875dcd18). The quickstart page and deprecation warning for the legacy health API, along with Community Care, will come in additional PRs.

PR #153, opened 2019-07-22
Selected tab shown in fragment

jasalisbury:

This PR makes the selected tab show up in the fragment of the URL. When visiting a tabbed API the first tab in the list is used as the fragment. Selecting tabs no longer changes the component state but will update the fragment, causing a rerender. The tab to show is selected based on the url fragment. The basic flow is:

 1) User selects a new tab (or loads the API documentation page)
 2) The URL fragment is updated, causing a re-render
 3) `getDerivedStateFromProps` sees the new URL fragment, setting the tab state accordingly
 4) The component is rendered again with the new tab state

Let me know if there is a better way to accomplish having the selected tab show up in the fragment of the url. I believe I am using the react life-cycle methods correctly. Rendering twice for the tab change may be a big deal or not. I don't have enough React experience to know if there is a better way.

### DIFF
--- a/src/containers/Explore.tsx
+++ b/src/containers/Explore.tsx
@@ -29,20 +29,41 @@ const mapStateToProps = ({ routing }: IRootState) => {
 };
 
 class Explore extends React.Component<IExploreProps, IExploreState> {
+  public static getDerivedStateFromProps(props: any, state: any) {
+    const api = lookupApiByFragment(props.match.params.apiName)
+
+    if (api) {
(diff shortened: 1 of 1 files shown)


mjlumetta:
@jasalisbury how do you want to go about this? i'll plan to review this in a sec unless you say otherwise

jasalisbury:
I think we can just use the changes you made in https://github.com/department-of-veterans-affairs/developer-portal/pull/146. I'll close out this PR.

PR #156, opened 2019-07-25
Benefits Minor release note changes

bastosmichael:
### Description

Minor last minute changes to Benefits Intake

### DIFF
--- a/src/content/releaseNotes/benefits.mdx
+++ b/src/content/releaseNotes/benefits.mdx
@@ -80,9 +80,6 @@
 ---
 
 ## July 23, 2018
-(v1) Announced production support for external use of v1, allowing:
-* Users with existing Power of Attorney to establish 526 claims on a Veteran's behalf
-* Authenticated Veterans to establish 526 claims on their own behalf
 (v0) Limit the size of individual attachments in uploads [#2094](https://github.com/department-of-veterans-affairs/vets-api/pull/2094)
 
 ---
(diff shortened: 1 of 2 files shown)


src/content/releaseNotes/benefits.mdx
```diff
@@ -80,6 +80,9 @@
 ---
 
 ## July 23, 2018
+(v1) Announced production support for external use of v1, allowing:
```
AubreySuter: Looks like these got added to the July 23 *2018* list - they should be at the top, for July 23, 2019

AubreySuter: Also, looks like it's under Intake, but should be Claims

bastosmichael: Oops sorry about that, should be fixed and pushed up.

src/content/releaseNotes/benefits.mdx
```diff
@@ -148,29 +148,36 @@ Add appeals service API [#1961](https://github.com/department-of-veterans-affair
 <br />
 <div id="claims" />
 
-# Benefits Claims
-### June 25, 2019
+# Claims
```
AubreySuter: I think we should leave it as "Benefits Claims" but if you think we should switch to just "Claims" I'd be interested to hear your thoughts.

bastosmichael: Fixed as well

PR #160, opened 2019-08-02
Enable FHIR, Urgent Care, and Community Care APIs in prod

mjlumetta:
Completes:
* [2126: Add new health documentation to developer portal in prod](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2126)
* [2129: Add community-care API documentation to the developer portal in prod](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2129)
* [2139: Add community-care API to sign-up in prod](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2139)j

Basically finishes enabling our Health API docs (and signup, in the case of Community Care) in production.


kalilsn:
Visual regression tests failed, and then jenkins failed to comment here indicating that ☹️.
Looks like it's expected changes to the apply page for the new API that's enabled:

mjlumetta:
@kalilsn will update (once i get my local docker container built)

kalilsn:
Are we not including the deprecation notice and the `(Legacy)` on the old argonaut API? It's present in dev right now (https://dev-developer.va.gov/explore/health/docs/argonaut) but not on the prod review instance.

mjlumetta:
> Are we not including the deprecation notice and the (Legacy) on the old argonaut API? It's present in dev right now (https://dev-developer.va.gov/explore/health/docs/argonaut) but not on the prod review instance.

@kalilsn i think it might be affirmatively based on the setting being included, which i did not do. let me fix that

dvogel:
@mjlumetta why did you revert the commit that relied on `isHostedApiEnabled`? Was it not working? Keeping those checks centralized seems wise to me. We recently tracked down a bug in kongfig to dispersed comparisons leading to an inconsistency `=== 'remove'` vs `=== `removed'`, which in turn led to a long and frustrating debugging process. I'd love to avoid that here, if possible.

mjlumetta:
> @mjlumetta why did you revert the commit that relied on isHostedApiEnabled? Was it not working? Keeping those checks centralized seems wise to me. We recently tracked down a bug in kongfig to dispersed comparisons leading to an inconsistency === 'remove' vs === removed'`, which in turn led to a long and frustrating debugging process. I'd love to avoid that here, if possible.

@dvogel i believe the reason i went with this is that the location of that function in `App.tsx` created some kind of circular dependency. i can play around with it and trying moving that to `apiDefs` somewhere. bc it was on friday i don't actually remember exactly what the error was, but i decided this was easier at that point.

dvogel:
> > @mjlumetta why did you revert the commit that relied on isHostedApiEnabled? Was it not working? Keeping those checks centralized seems wise to me. We recently tracked down a bug in kongfig to dispersed comparisons leading to an inconsistency === 'remove' vs === removed'`, which in turn led to a long and frustrating debugging process. I'd love to avoid that here, if possible.
> 
> @dvogel i believe the reason i went with this is that the location of that function in `App.tsx` created some kind of circular dependency. i can play around with it and trying moving that to `apiDefs` somewhere. bc it was on friday i don't actually remember exactly what the error was, but i decided this was easier at that point.

Moving that function to `apiDefs.ts` or moving it to it's own module seems like a good approach to me.

mjlumetta:
@dvogel i tried moving it to `apiDefs/index.ts`, but the same problem keeps happening, which i now remember. it's that `npm run test:unit` keeps crashing in a couple of test suites when the function is imported at one of the consumption locations, so... gonna try and figure that out.

i guess part of my thought here, though, is - i definitely want to time-cap this, because i don't want to get into a situation where i'm working with my head down like, "wow, hmm, let me figure out this technical problem with import statements in our Node project" when a simpler boolean still works and will let us get this deployed. especially bc it's not that bad in the less clean version and the developer portal is not that clean overall... i would be in favor of merging this PR as-is so we can get our expedited work out in the world and make our product owners happy and adding a follow-up commit to clean up the logic. what do you think?

dvogel:
I agree. Let's not let technical perfection get in the way here. I still
think we should clean this up. Seems like maybe a typescript/jest thing we
could all stand to learn more about so I'd be happy working through the
details with you in another PR. Consider this comment an approval for this
PR.

Side note: let's not let the current quality and character of the developer
portal anchor us to sub-par practices. Let's aim high ;)

mjlumetta:
thanks @dvogel! merging now.

> Side note: let's not let the current quality and character of the developer
portal anchor us to sub-par practices. Let's aim high ;)

i agree!! my position would be that it's okay to commit sub-optimal things _temporarily_ and _if they are small enough_. i definitely agree that should be our aim but think some recent needs have led to situations where it was expedient for product purposes (thinking of this and Claims v1 on the Apply page). perhaps a good retro topic

PR #161, opened 2019-08-08
Remove circular dependencies and abstract API environment check

mjlumetta:
We had a series of circular dependencies that were exposed by trying to pull this function out into its own module. This PR fixes them and adds a `src/apiDefs/env` module for environment-related checks for APIs.


src/apiDefs/index.ts
```diff
@@ -164,3 +165,19 @@ function apisFor(apiList: string[]): IApiDescription[] {
...
+  claims: isHostedApiEnabled('claims', true),
+  community_care: isHostedApiEnabled('community_care', true),
+  disability_rating: isHostedApiEnabled('disability_rating', true),
+  facilities: isHostedApiEnabled('facilities', true),
+  fhir: isHostedApiEnabled('fhir', true),
+  loan_guaranty: isHostedApiEnabled('loan_guaranty', false),
+  service_history: isHostedApiEnabled('service_history', true),
+  urgent_care: isHostedApiEnabled('fhir', true),
+  veteran_confirmation: isHostedApiEnabled('veteran_confirmation', true),
+};
```
dvogel: It makes a little more sense to me for `apiEnvFlags` to be defined in `src/apiDefs/env.ts` but I'm really fine either way.

mjlumetta: oh yeah, you're absolutely right. sorry, i didn't look too carefully at the other functions after i solved the problem! my bad

dvogel:
Thanks for digging in and sorting these out! Compilers are horrible at sorting out circular issues because it can't infer what you really meant so these are always painful. I'm positive this will help us avoid a lot of pain in the future :smile:

PR #163, opened 2019-08-14
add contact us

whuang85:
closes https://github.com/department-of-veterans-affairs/vets-contrib/issues/1836

This PR is adding the contact us form and the logic to call the lambda function, I thought it was overkill to use redux on this for state management but correct me if I'm wrong.


src/containers/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+  error: boolean;
+}
+
+interface ISubmissionErrorProp {
+  error: boolean;
+}
+
+const SubmissionError = (props: ISubmissionErrorProp) => {
+  const assistanceTrailer = (
+    <span>Need assistance? Create an issue on our <a href="https://github.com/department-of-veterans-affairs/vets-api-clients/issues/new/choose">Github page</a></span>
```
rtravitz: Should there be a period here at the end of `Github page`?

whuang85: This was copied over from the apply page.. I could add a period but then it would not be consistent with that or I can add a period to both places.

src/containers/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,258 @@
...
+        value: '',
+      },
+      organization: {
+        dirty: false,
+        value: '',
+      },
+    }
+  }
+
+  public render() {
```
charleystran: Can we make this a single component? Or a couple more components instead of one long function?

whuang85: @charleystran sure, any suggestions on the break down?  I could make the checkboxes one component and the 4 input texts above it one?

charleystran: I think the form should be its own component, which then could use input components. What do you think?

whuang85: currently the form is a component just under the container folder and is being used by the `SupportContactUs.tsx` container.

src/containers/SupportContactUs.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+interface ISupportContactUsState {
+  error: boolean;
+  sending: boolean;
+}
+
+const GithubSnippet = () => {
+  return (
+    <div className="va-github-snippet">
+      <h3>Submit an Issue via Github</h3>
```
nfasulo: Hmm, both this h3 and the text above it which is an h2 are both 20px. That doesn't seem right to me from a styling perspective and I don't like how it looks on the page.

whuang85: I'll double check the usages of h# in my components

src/containers/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,258 @@
...
+    return (
+      <div>
+        <form className="va-contact-us-form">
+          <fieldset>
+            <legend>Contact Us</legend>
+            <p>
+              Have a question? Use the form below to send us an email and we'll do the best to answer your question and get you headed in the right direction
```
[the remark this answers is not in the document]
whuang85: > I just have a couple of small comments below.
> 
> I'm definitely not for throwing things in Redux just because we can, but I would say there is value in keeping parity between this form and the Apply form.

I did notice a bug in the apply page that if you leave the page with an error (specifically the email field) and navigate away from the page and navigate back the error is still there.. I believe this is because redux is for global state of the application.

src/containers/SupportContactUs.tsx
whuang85: at some point the h2 in github snippet and the h2 in the page header were the same, haven't pinpointed the exact change and I noticed the change today when running visual regression that the font got bigger than before so I changed all the h2s to h3s.  I can change it back to h2s and adjust the font size via css but that seems weird.  Also the font size more or less matches the current figmas (besides the page header description text, it should not be bold)

src/containers/SupportContactUs.scss
```diff
@@ -0,0 +1,10 @@
+.va-github-snippet {
```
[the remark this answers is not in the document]
whuang85: I thought VA formation uses `vads-`?  Anyway yes I agree, I'll be more consistent with what is existing, I'm leaning towards `va-api-`

src/containers/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,258 @@
...
+            </p>
+
+            <div className="usa-grid">
+              <div className="usa-width-one-half">
+                <ErrorableTextInput
+                  errorMessage={null}
+                  label="First name"
+                  field={this.state.firstName}
+                  onValueChange={(field: IErrorableInput) => this.setState({ firstName: field })}
```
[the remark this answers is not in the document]
whuang85: I was mostly following the apply form here, but yes I can turn on presence validation.

src/containers/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,258 @@
...
+      </div>
+    )
+  }
+
+  private toggleApis(event: React.ChangeEvent<HTMLInputElement>) {
+    const name = event.target.name
+    const value = event.target.checked
+    const apis = this.state.apis;
+    apis[name] = value
+    this.setState({ apis })
```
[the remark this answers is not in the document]
whuang85: I thought setState only merges the top level and not nested objects?

src/containers/SupportContactUsForm.tsx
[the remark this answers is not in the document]
nfasulo: Oh, that's a good catch. I bet you are right about the cause.

src/components/GroupedAccordions.tsx
```diff
@@ -49,7 +49,7 @@ export default class GroupedAccordions extends React.Component<IGroupedAccordion
     return (
       <section className="va-grouped-accordion">
         <div className="va-grouped-accordion-header">
-          <h2>{this.props.title}</h2>
+          <h3>{this.props.title}</h3>
```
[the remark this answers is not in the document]
whuang85: for some reason, after bringing in master the font sizes changed.. not sure if it was because of the formation upgrade

src/containers/SupportContactUs.tsx
kalilsn: I can explain! There were some overrides to our h1-h6 font sizes in the `Explore.scss` which were being mistakenly applied throughout the site. Those styles were only ever intended for the documentation pages and shouldn't have been applied elsewhere. In general, the guidance I've received from @juliaelman is that we should stick to our base font sizes for headings, so I made sure those overrides were only being applied in that section.

In all cases, we should choose between header elements for their semantic meaning, not their styles. If the page has an h1 and the item you're adding would logically be the next level down in the outline, then use an h2. If it's important to match the figma mockup exactly (and that's sometimes worth checking in about!), then change that particular instance with CSS.	This is in accordance with the [design system's guidance](https://design.va.gov/design/typography#headings-and-accessibility), and is important for non-graphical clients like screen readers. The design system suggests using their classes for changing the font size, which will be the recommendation soon. Until then, adding some custom CSS is fine.

src/components/ApiSelection.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  constructor(props: IApiSelectionProps) {
+    super(props);
+    this.state = {
+      benefits: false,
+      claims: false,
+      communityCare: false,
+      facilities: false,
+      health: false,
+      verification: false,
```
kalilsn: The options on this form match the ones on the signup form. However, those options don't really match up with any other parts of our systems. For example, `communityCare` is actually an API in the Health category, and `claims` is just a different version of an API in the Benefits category. The mismatch between those options and how the rest of our systems treat the different APIs have caused a great deal of pain. I think we should treat the options on that form as tech debt and quarantine them from the rest of our code as best as possible, to avoid compounding our reliance on them.

Since this is for indicating which API you're having trouble with, could we instead render these options from the APIs in `apiDefs`?

src/containers/SupportContactUs.tsx
whuang85: I believe these are legitimate headers, the problem is I'm using the PageHeader component which I assume was designed for the documentation pages.  So when looking at the HTML, the page header has an h2 which is a different size than an h2 outside of the header but all on the same page

src/components/ApiSelection.tsx
whuang85: So what you are saying is the apis really are just these 4?

- 'benefits'
- 'facilities'
- 'health'
- 'verification'

In that case, @jeff2d2 the figma design has what's in the apply page, but should I just change the checkboxes to match just the above 4?

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,168 @@
...
+                name="organization"
+                field={this.state.organization}
+                onValueChange={(field: IErrorableInput) => this.setState({ organization: field })}
+                required={false} />
+            </div>
+          </div>
+        </fieldset>
+
+        <p>If applicable, please select any of the APIs pertaining to your issue.</p>
+        <ApiSelection onSelection={(apis) => this.setState({apis})}/>
```
mjlumetta: related to the above about generic form models: this is one of the things that would be great to avoid. in this case, data is flowing back up the tree via callback. i think that that should be an exceptional case; in general, data is meant to flow down in React. while the `ErrorableTextInput` does a similar thing, that makes more sense because each is a single input and therefore it follows the pattern of a controlled input. but for `ApiSelection`, where the data is a sub-model of the overall form, it becomes more complicated. 

again, don't think it's necessary to fix in this PR, but we should get as close to it as possible.

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,168 @@
+import ErrorableTextArea from '@department-of-veterans-affairs/formation-react/ErrorableTextArea';
+import ErrorableTextInput from '@department-of-veterans-affairs/formation-react/ErrorableTextInput';
+import * as React from "react";
+import { validateByPattern, validateEmail } from '../../actions';
```
mjlumetta: actually, just noticed this - you probably shouldn't use these here. if they're in `src/actions`, they're meant to be related to Redux actions, which these are not. _however_, i don't think the `validate*` functions are themselves actions anywhere, so maybe they could be abstracted out to a `src/utils/validators.ts` file or something like that. i would also suggest that `validatePresence` would go in that location as well, because that definitely will be reused.

src/components/ApiSelection.tsx
jeff2d2: The proposed layout of showing only 4 categories is how it was originally captured in the discovery ticket for "what should be on this form" (https://github.com/department-of-veterans-affairs/vets-contrib/issues/1800). That was changed to the newest design of 6 (mirroring the apply page) that was coded here through https://github.com/department-of-veterans-affairs/vets-contrib/issues/2919. Because those changes were suggested by @juliaelman, I'm going to defer to her for the reasoning/explanation.

This does bring up a few good points. I agree with @kalilsn in that our API categories, naming conventions, apply page checkboxes, and documentation IA is a bit all over the place and out of sync. Even trimming this back to what's in `apiDefs` doesn't totally "fix" it (Address Validation will probably be moved out from under Verification and into a Utilities category soon, and so will the Confirmation endpoint but into a new API).

What I'm getting at is that it will all be moved around and synced up (hopefully soon) – even the categories themselves are likely to change/expand in the near future. I think we're stacking tech debt no matter which way we turn here. What should go on this form should be whatever is most helpful in allowing @ryanlink do what he intends to do with the information collected.

If we ask them `select any of the APIs pertaining to your issue`: I count 9 different APIs if they're going off of the actual base URLs, 4 if the homepage or docs top-level nav, and 6 if the apply page. So while I agree with being consistent, we don't have anything right now in place to be consistent with.

My apologies for the diatribe...

jeff2d2: @whuang85 Hold off on this for now, please. I'd like for everyone tagged to have a chance to chime in. We'll get you a solid answer soon.

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,168 @@
...
+interface ISupportContactUsFormProps {
+  onSuccess: () => void;
+}
+
+export default class SupportContactUsForm extends React.Component<ISupportContactUsFormProps, ISupportContactUsFormState> {
+
+  constructor(props: ISupportContactUsFormProps) {
+    super(props);
+    this.state = {
+      apis: {},
```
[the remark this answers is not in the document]
nfasulo: Appreciate you thinking about the larger picture, @mjlumetta. I like the idea of reusable for form state management in Redux as the portal continues to grow in complexity. I'm going to request the @whuang85 not dive into a larger form refactor right now, since we have other priorities awaiting him once this PR is merged.

src/components/ApiSelection.tsx
ryanlink: For reference, when I imported the /apply form contacts into Salesforce, I included a custom field with a picklist of the 4 options we identified in https://github.com/department-of-veterans-affairs/vets-contrib/issues/1800 - these match up with current top-level nav on the left side of the dev portal.

Subsequently I expanded the picklist on the Salesforce Case object to the 9 API choices Jeff mentioned, plus one for general OAuth flow (perhaps redundant), for more specific metrics tracking.  

We do need to make some choices soon on reorganizing this info that we've known is inconsistent for months.

kalilsn: I don't have a strong opinion about exactly which options make sense to show here – I just don't want us to take the ones on the apply page and replicate them. Both a full list of every API, or the four categories can be taken from `apiDefs`, which reduces the risk of this page ending up drifting away from the rest of the site and implementing its own unique categorization.

> I think we're stacking tech+product debt no matter which way we turn here.
> So while I agree with being consistent, we don't have anything right now in place to be consistent with.

I think these are both true, unfortunately! But I do think using apiDefs as the source of truth at least avoids making the problem worse, even if it's not a perfect solution.

src/containers/support/SupportContactUs.scss
```diff
@@ -0,0 +1,10 @@
+.va-github-snippet {
```
kalilsn: I don't think this file is being imported anywhere

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,168 @@
...
+          accept: 'application/json',
+          'content-type': 'application/json',
+        },
+        method: 'POST',
+      },
+    );
+
+    const response = await fetch(request);
+    if (!response.ok) {
+      throw Error(response.statusText);
```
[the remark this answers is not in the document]
kalilsn: This `formSubmission` function is passed as `onSubmit` to the `Form` component, which wraps the call to `onSubmit` in a try/catch block, so it won't make it to the user. A failure message appears beneath the form when that happens, so I think throwing is fine here!

src/components/ApiSelection.tsx
jeff2d2: We're going with the 4 categories, and I love the idea of pulling them from `apiDefs`. Couple quick questions for @kalilsn before we do that though: If we populate the checkboxes directly from there, do you suggest using `name` or `properName` (2 are the same and 2 are different). What else already populates a "list of APIs" from this (if anything) and what do they use?

kalilsn: | Location | React Component | Property used |
| -------- | --------------- | ------------- |
| Documentation category page header | `ApiPage` | `name` |
| Release Notes category page header | `ApiPageReleaseNotes` | `properName` |
| Mobile menu documentation subnav | `NavBar` | `name` |
| Signup success message | `ApplySuccess` | `properName` |
| Documentation overview card nav | `DocumentationOverview` | `name` |
| Documentation section sidenav | `ExploreDocs` | `name` |
| Authorization docs header | `OAuth` | `name` |
| Documentation section quickstart header | `QuickstartPage` | `properName` |
| Release notes sidenav | `ReleaseNotes` | `name` |
| Release notes overview card nav | `ReleaseNotesOverview` | `name` |

Here are the places where the category names from `apiDefs` are used. It seems like `name` is used more often, but I couldn't really say why, or what the difference between them is. I'm not sure it matters all that much in this case. `name` makes slightly more sense to me, if only because `Benefits API` is slightly more accurate than `Benefits Intake API` as a category name (although both are a little confusing since there are multiple APIs within that category).

jeff2d2: Thank you!
@whuang85 Green light to go with the 4 categories as our checkboxes, pulling from the `name` field in `apiDefs`. Because of this change, there's also no reason for the `Standard APIs:` or `OAuth APIs:` headers currently present in the mockup. Let me know if you have any questions!

src/containers/support/SupportContactUsForm.tsx
whuang85: So would the alternative would be to pass in the state as props and have the `ApiSelection` component manipulate the `SupportContactUsForm` state?

src/containers/support/SupportContactUsForm.tsx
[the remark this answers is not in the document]
mjlumetta: @nfasulo i think that makes perfect sense. don't really think that should be on Will at all but should probably be a conversation for a non-PR forum

src/containers/support/SupportContactUsForm.tsx
mjlumetta: ummm no i think you're good here. i think this is one of the patterns we have to adopt until we're able to build a form model - your implementation is better for using `setState`

src/containers/support/SupportContactUs.tsx
```diff
@@ -0,0 +1,56 @@
...
+interface ISupportContactUsState {
+  sent: boolean;
+}
+
+export default class SupportContactUs extends React.Component<{}, ISupportContactUsState> {
+
+  constructor(props: {}) {
+    super(props);
+    this.state = {
+      sent: false,
```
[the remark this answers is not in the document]
whuang85: The problem here is that I need the confirmation page to replace both the form and the pageheader and snippet.  If I gave the `Form` component the ability to render the success, it would only be able to replace itself and not the other components that I need hidden away.  I'm open to suggestions on how to solve this problem because initially I would just going to give the user the ability to pass down a component so it could render the success.

src/containers/support/SupportContactUsForm.tsx
whuang85: I know formation-react has a ErrorableCheckboxGroup, I would have used it earlier but it had ally issues.  I think it might have been fixed since we now bring in the latest formation-react . I could switch to that instead of using the ApiSelection component I created.

mjlumetta: that's true! if that makes sense to you. i was assuming that you hadn't used that because these checkboxes aren't errorable

src/containers/support/SupportContactUs.tsx
```diff
@@ -0,0 +1,56 @@
...
+import SupportContactUsForm from './SupportContactUsForm';
+
+const GithubSnippet = () => {
+  return (
+    <div className="va-api-github-snippet">
+      <h3>Submit an Issue via Github</h3>
+      <a className="usa-button" href="https://github.com/department-of-veterans-affairs/vets-api-clients/issues/new/choose"><FontAwesomeIcon icon={faGithub} /> Submit an Issue</a>
+    </div>
+  );
+}
```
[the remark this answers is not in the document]
whuang85: i turned it on to fix all the Support related pages and components, but I don't think this PR should go through and fix all the other unrelated components. What are your thoughts?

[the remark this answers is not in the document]
mjlumetta: given how far you are on this, i wouldn't worry about it if there are a lot of errors. if it's only a few files, i would say let's fix it now and get it done. but if it'll be like 10-20 files that are gonna pollute this (almost done) PR, probably better to skip. ultimately i'd say up to you and your comfort level

src/containers/support/SupportContactUs.tsx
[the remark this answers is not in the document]
mjlumetta: yeah, so, i think this is the sort of problem that Redux is meant to solve. there is a boolean state value that needs to be shared across the page and isn't naturally able to be passed around as props or maintained as local state. currently `SupportContactUsForm` does the actual submitting, so that component would be responsible for dispatching an event for contact us submit success (whatever the best way to name that would be). the appropriate action and reducer to support this event would be defined in `src/actions/index.ts` and `src/reducers/index.ts`. then `SupportContactUs` would receive this boolean value as a prop via `connect()` and could make that decision without local state.

i would say that whether or not to make this change depends on your comfort level and the effort you feel it would take to complete it. unfortunately the state model `IApplication`, which is defined in `src/types/index.ts`, is overfitted to the Apply page, so that would likely require some small refactoring. if you do feel able, there is a substantial benefit imo to building on the Redux pattern and spreading the knowledge of how to do that across the program, as well as avoiding a later rewrite. but it's also fine as is if it would be difficult.

src/containers/support/SupportContactUs.tsx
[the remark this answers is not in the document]
whuang85: there were about 17 different files that had various semi colon violations so I left it out for another time as I didn't want it to pollute this PR

src/containers/support/SupportContactUs.tsx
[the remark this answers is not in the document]
whuang85: Good point, I think for now I would like to introduce redux into the reusable form component in another ticket:

https://github.com/department-of-veterans-affairs/vets-contrib/issues/3093

src/components/Form.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+    this.submittingForm = this.submittingForm.bind(this);
+  }
+
+  public render() {
+    const assistanceTrailer = (
+      <span>Need assistance? Create an issue on our <a href="https://github.com/department-of-veterans-affairs/vets-api-clients/issues/new/choose">Github page</a></span>
+    );
+
+    const alert = this.state.error ? <AlertBox status="error" headline={"We encountered a server error while saving your form. Please try again later."} content={ assistanceTrailer } /> : null;
```
mjlumetta: i think that it might be easier to read this rendered inline? like, where the `{alert}` is now, replace with

```
{this.state.error &&
  <AlertBox
    status="error"
    headline="headline goes here"
    content={assistanceTrailer} />
}
```

src/components/GroupedAccordions.tsx
```diff
@@ -78,7 +78,7 @@ export default class GroupedAccordions extends React.Component<IGroupedAccordion
     this.setState((prevState: IGroupedAccordionsStates) => ({ allCollapsed: !prevState.allCollapsed }), () => {
       this.panelRefs.filter(r => r.current && (r.current.state.open === this.state.allCollapsed)).forEach(r => {
         if (r.current) {
-          r.current.setState({ open: !this.state.allCollapsed })
+          r.current.setState({ open: !this.state.allCollapsed });
```
mjlumetta: thanks for getting all of these!!

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  private get initialApiState() {
+    return Object.keys(apiDefs).reduce((accumulator, currentValue) => {
+      accumulator[currentValue] = false;
+      return accumulator;
+    }, {});
+  }
+
+  private get defaultErrorableField(): IErrorableInput {
```
mjlumetta:
```suggestion
  private static get defaultErrorableField(): IErrorableInput {
```

can you do this? seems small put worthwhile

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -0,0 +1,188 @@
...
+    return Object.keys(apiDefs).map(api => {
+      return {
+        label: apiDefs[api].name,
+        value: api,
+      };
+    });
+  }
+
+  private get initialApiState() {
+    return Object.keys(apiDefs).reduce((accumulator, currentValue) => {
```
mjlumetta:
```suggestion
    return Object.keys(apiDefs).reduce((accumulator, apiId) => {
```

something a little more specific if possible?

PR #167, opened 2019-08-20
Update formation and formation react to latest versions

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/1669

All tests passed locally and everything seems to look the same just browsing around. The ticket mentions enabling the `axe` test for the `/apply` page but I can't remember how enabling that test was related to upgrading formation, anyone remember?


dvogel:
I believe department-of-veterans-affairs/vets-contrib/issues/1669 came out of department-of-veterans-affairs/vets-contrib/issues/1488. That ticket mentions a few issues with the `/apply` page. Any of them still relevant?

kalilsn:
>I believe department-of-veterans-affairs/vets-contrib/issues/1669 came out of department-of-veterans-affairs/vets-contrib/issues/1488. That ticket mentions a few issues with the /apply page. Any of them still relevant?

The issue Mike addressed in formation was invalid `autocomplete` attributes being set: https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/pull/44. I believe that was the only thing that we couldn't fix ourselves, so I think it's worth trying those accessibility tests on the apply page again!

jasalisbury:
The only accessibility error I saw for the `/apply` page was the header content didn't have an id. The `aria-labelled-by` requires an element to have a matching `id` I couldn't see a way to add the required `id` via `markdown-component-loader`, which is why I wrapped the markdown component in a div.

PR #168, opened 2019-08-20
News section rework

kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2615

This takes care of the content updates described in the issue, and makes some changes that will make embedding media possible (moving from mdx -> yml for storing the content). There were a few cleanup changes that I made throughout this work that are bundled in here:
	- Started towards extracting the sidenav as a component
	- Stopped using the explore styles on the What's New section
	- Increased the specificity of some explore styles that were causing issues throughout the site
	- Updated accessibility/VR tests to include some health pages and exclude the /oauth page (which was causing issues for me)

It's a big PR, let me know if walking through it together would be helpful.


kalilsn:
The tests aren't passing right now, but I'll fix that. I forced jenkins to build/deploy so we would get a review instance.

dvogel:
I'll dig deeper into this tomorrow morning but at a high level, I think the data file approach has some key advantages to the `.ts` approach. The main advantage is that we eventually want this to be data sourced from somewhere else. To reach that goals, it's important to make the data defined in the code have a regular, predictable shape. That creates an appropriate amount of friction to adding one-off exceptions for a given API. The `apiDefs` module started out with a bunch of those. I see it as an in an intermediate step on this path:

0. The jungle that it was.
0. Organized code.
0. Pure data in code repository.
0. Pure data source from a database/API.

src/containers/News.tsx
```diff
@@ -4,28 +4,12 @@ import { NavHashLink } from 'react-router-hash-link';
...
-    id: 'news',
-    name: 'News',
-  },
-  {
-    description: 'Media related to developers, APIs, Veterans, and the Department of Veterans Affairs',
-    id: 'media',
-    name: 'Media',
-  },
-];
+const sections = NewsData.sections.map((section: any) => ({ ...section, id: section.title.replace(/\s+/g, '-').toLowerCase()}));
```
dvogel: It would be good to pull this string manipulation into a generic function. In addition to the readability improvement on this line, there's additional characters that likely need to be accounted for (e.g. apostrophes) in order to truly consider this code detached from the specific values currently used in the `title` field of the data. A function like `toHtmlId(instr: string): string` would give us a good place to deal with other special characters. In fact, I bet there's a reliable method like this available on npm.

kalilsn: Agreed, I'll pull it out into a generic function.

The HTML 5 spec actually says the only restriction on an `id` is it can't have spaces or be the empty string. Do you know a reason that we need to support the more restrictive HTML 4 id spec? I'm having a hard time finding any actual info about actual browser compatibility for this particular change.

dvogel: I think we'd like to avoid characters that are hard to understand in the context or require special considerations. I look at it a lot like URLs. They can contain most characters (at least in the IRI form) but there's some characters that require percent encoding, so they are generally removed for URL slugs that are likely to be seen by humans. E.g. if you run this javascript:

```
var newPara = document.createElement("p");
newPara.setAttribute('id', '"'); // a double quote wrapped in single quotes
document.body.insertBefore(newPara, null);
```

The HTML, as serialized by the browser:

```
<p id="&quot;"></p>
```

That's legit HTML but it's unlikely to be pretty in the URL fragment that's added when clicking on the card.

That ID is also not a legitimate CSS id selector since the two standards disagree on the characters that can be used. We wouldn't ever want someone to directly style a particular ID like this, but that's where pretending like we don't know what data will be fed into the function comes back into play. Once we make a generic, reusable function it's likely to see other uses (and that's a good thing!)

src/toHtmlId.ts
```diff
@@ -0,0 +1,9 @@
+// Converts a string to an html id suitable for use in an anchor link
+function toHtmlId(phrase: string) {
+  return phrase
+    .replace(/^[^a-zA-Z]+/, '') // Ensure id starts with a letter
+    .replace(/\s+/g, '-')
+    .replace(/[^\w-_'?!()*]+/g, '')
```
kalilsn: This allows a subset of the characters that don't need % encoding in a URL. That subset is purely based on my aesthetic judgment about what characters are acceptable, so very open to change 😄

dvogel: I would go "full CSS" here. Docs for the id selector: https://www.w3.org/TR/CSS21/syndata.html#value-def-identifier

> In CSS, identifiers (including element names, classes, and IDs in selectors) can contain only the characters [a-zA-Z0-9] and ISO 10646 characters U+00A0 and higher, plus the hyphen (-) and the underscore (_); they cannot start with a digit, two hyphens, or a hyphen followed by a digit. Identifiers can also contain escaped characters and any ISO 10646 character as a numeric code (see next item). For instance, the identifier "B&W?" may be written as "B\&W\?" or "B\26 W\3F".
>
> Note that Unicode is code-by-code equivalent to ISO 10646 (see [UNICODE] and [ISO10646]).

dvogel: caveat: "full CSS" means be as restrictive or more than CSS. E.g. I don't think there's any reason we really need to support escaped characters the way CSS does.

src/containers/News.tsx
```diff
@@ -7,21 +7,21 @@ import toHtmlId from '../toHtmlId';
 
 import PageHeader from '../components/PageHeader';
 import { LocalNavHashLink } from '../components/SideNav';
-import * as NewsData from '../content/news.yml';
+import NewsData from '../content/news.yml';
```
kalilsn: Using the `import * as Alias` syntax to import all the exports of the module doesn't seem to work when my module has actual types (as opposed to simply `declare module '*.yml';`). That's fine, except that this syntax doesn't seem to work with jest.

After making this change, NewsData is undefined during the jest tests. Instead the imported object is showing up in a variable called `news_yml_1`. Anyone have any ideas what's going on there?

PR #169, opened 2019-08-21
Add statuspage link

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2929

 * Adds a link `API Status` above the search box.
 * Updates formation
 * Enables css source maps
 * Adds visual regression test for the header


kalilsn:
It looks like the loan_guaranty API is being enabled in the updated screenshots here. Do you have `REACT_APP_LOAN_GUARANTY_API_ENABLED=true` in your `.env.test`?

jasalisbury:
> It looks like the loan_guaranty API is being enabled in the updated screenshots here. Do you have `REACT_APP_LOAN_GUARANTY_API_ENABLED=true` in your `.env.test`?

I didn't have it enabled in `.env.test` but had it enabled in `.env.local`. Updated the screenshots with it disabled.

jasalisbury:
> (I'm kind of confused why `.env.local` would matter for the test, I'm double checking to make sure that the local env file is used for visual regression tests)

The test server is started with `"command": "PORT=4444 BROWSER=false npm run-script start",`. The start script forces the `NODE_ENV` to be development, which causes `.env.local` to be used for config.

src/components/NavBar.scss
```diff
@@ -1,11 +1,47 @@
...
+  margin-left: auto;
+}
+
+.header-right-content {
+  display: flex;
+}
+
+#get-started-button {
+  margin-right: 2rem;
+}
```
dvogel: A comment above this would be useful:
> a temporary patch until we adopt USWDS more fully

src/components/NavBar.tsx
```diff
@@ -53,33 +53,39 @@ export class NavBar extends React.Component<INavBarProps, INavBarState> {
...
           </div>
+          <div className="header-right-container">
+          <MediaQuery query={OVER_LARGE_SCREEN_QUERY}>
+            <a className="api-status-link" href="https://valighthouse.statuspage.io">API Status</a>
+            <div className="header-right-content">
+              <Link id="get-started-button" to="/apply" className="usa-button">Get Started</Link>
+              <Search />
+            </div>
+          </MediaQuery>
+          </div>
```
dvogel: This `MediaQuery` element is not indented. The one below is indent. Can you make them consistent?

src/visualRegressionTest.ts
```diff
@@ -33,6 +33,21 @@ describe('Visual regression test', async () => {
...
+      await page.setViewport(viewport);
+      await new Promise((resolve, reject) => setTimeout(resolve, 500));
+      const header = await page.$('header');
+      if(header) {
+        const screenshot = await header.screenshot();
+        expect(screenshot).toMatchImageSnapshot();
+      }
+    }
+  });
+
```
dvogel: One of the ways these visual regression tests weigh us down is by capturing the entire page. We've talked in the past about making these snapshot tests each use a more constrained viewport. Since updating the header is changing every one of these, maybe this ticket is the tipping point and we should make that change as part of this ticket?

jasalisbury: Splitting visual tests even further would be great. We had to adjust the failure threshold to be more lenient because some of the pages we compare are huge. If we had smaller snapshots we could revert that change. I could see us updating the current visual regression tests to exclude the header for this ticket. I feel splitting the snapshots more than just "header" vs "non-header" should be a part of a different ticket.

src/components/NavBar.scss
kalilsn: It should probably be `VA design system` instead of `USWDS`, right?

kalilsn:
>The test server is started with "command": "PORT=4444 BROWSER=false npm run-script start",. The start script forces the NODE_ENV to be development, which causes .env.local to be used for config.

Well that sounds like a bug. I _think_ that if we have `NODE_ENV=test`, then `.env.local` won't get loaded. But I'm even more confused – if NODE_ENV is development, is `.env.test` even being loaded?

juliaelman:
Took a look at this and have some comments about the layout:

1. Space should be added above and below the `API Status` link, see screenshot:

Screen Shot 2019-08-22 at 1 47 07 PM

You can reference Figma for exact measurements: https://www.figma.com/proto/SGQZyWGE44uYM4lI1twVsw/IA-Updates-(Jan-2019)?node-id=1%3A2&viewport=194%2C222%2C0.205882&scaling=min-zoom&redirected=1

2. The `API Status` copy should compute at 16px. It is currently 17px.

Please let me know if you have any questions or there is anything that I can do to help!

src/visualRegressionTest.ts
```diff
@@ -31,8 +34,19 @@ describe('Visual regression test', async () => {
     // Hide problematic video on homepage
     await page.evaluate('document.querySelector("iframe").style="visibility: hidden;"');
 
-    await checkScreenshots(page);
+    await checkScreenshots(page, '.main');
+  });
+
+  it('renders the header properly', async() => {
+    await page.goto(`${puppeteerHost}`, { waitUntil: 'networkidle0', timeout: 0 });
```
kalilsn: I don't think you need `timeout: 0` here. I think that's only necessary for the first request, which usually starts while webpack is still compiling.

src/visualRegressionTest.ts
```diff
@@ -31,8 +34,19 @@ describe('Visual regression test', async () => {
...
+    await checkScreenshots(page, 'header');
   });
+
+  it('renders the footer properly', async() => {
+    await page.goto(`${puppeteerHost}`, { waitUntil: 'networkidle0', timeout: 0 });
```
kalilsn: same as above re: `timeout: 0`

jasalisbury:
> > The test server is started with "command": "PORT=4444 BROWSER=false npm run-script start",. The start script forces the NODE_ENV to be development, which causes .env.local to be used for config.
> 
> Well that sounds like a bug. I _think_ that if we have `NODE_ENV=test`, then `.env.local` won't get loaded. But I'm even more confused – if NODE_ENV is development, is `.env.test` even being loaded?

So I took some time to track down what is really going on here yesterday. The test processes fires up with `NODE_ENV=test`. The env loader is required causing environment variables to be set from `.env.test`. After environment variables are loaded the server process is started with an entry point of `scripts/start.js`. This new process inherits environment variables from the parent process. In `script/start.js` `NODE_ENV` is set explicitly to `development`. The env loader is required for the server process. This causes environment variables to be loaded from `.env.local`. However the env loader library we use will not overwrite environment variables. `SWAGGER_HOST` is already set when `env.local` is loaded from the parent process, which loaded it from `.env.test`. The test server uses `.env.local` but since `SWAGGER_HOST` has already been set swagger calls are still mocked.

dvogel:
> > > The test server is started with "command": "PORT=4444 BROWSER=false npm run-script start",. The start script forces the NODE_ENV to be development, which causes .env.local to be used for config.
> > 
> > 
> > Well that sounds like a bug. I _think_ that if we have `NODE_ENV=test`, then `.env.local` won't get loaded. But I'm even more confused – if NODE_ENV is development, is `.env.test` even being loaded?
> 
> So I took some time to track down what is really going on here yesterday. The test processes fires up with `NODE_ENV=test`. The env loader is required causing environment variables to be set from `.env.test`. After environment variables are loaded the server process is started with an entry point of `scripts/start.js`. This new process inherits environment variables from the parent process. In `script/start.js` `NODE_ENV` is set explicitly to `development`. The env loader is required for the server process. This causes environment variables to be loaded from `.env.local`. However the env loader library we use will not overwrite environment variables. `SWAGGER_HOST` is already set when `env.local` is loaded from the parent process, which loaded it from `.env.test`. The test server uses `.env.local` but since `SWAGGER_HOST` has already been set swagger calls are still mocked.

Nice debugging! It sounds like the path forward depends on whether we want the server process to run with `NODE_ENV=test` or `NODE_ENV=development`. Is that right?

jasalisbury:
> > > > The test server is started with "command": "PORT=4444 BROWSER=false npm run-script start",. The start script forces the NODE_ENV to be development, which causes .env.local to be used for config.
> > > 
> > > 
> > > Well that sounds like a bug. I _think_ that if we have `NODE_ENV=test`, then `.env.local` won't get loaded. But I'm even more confused – if NODE_ENV is development, is `.env.test` even being loaded?
> > 
> > 
> > So I took some time to track down what is really going on here yesterday. The test processes fires up with `NODE_ENV=test`. The env loader is required causing environment variables to be set from `.env.test`. After environment variables are loaded the server process is started with an entry point of `scripts/start.js`. This new process inherits environment variables from the parent process. In `script/start.js` `NODE_ENV` is set explicitly to `development`. The env loader is required for the server process. This causes environment variables to be loaded from `.env.local`. However the env loader library we use will not overwrite environment variables. `SWAGGER_HOST` is already set when `env.local` is loaded from the parent process, which loaded it from `.env.test`. The test server uses `.env.local` but since `SWAGGER_HOST` has already been set swagger calls are still mocked.
> 
> Nice debugging! It sounds like the path forward depends on whether we want the server process to run with `NODE_ENV=test` or `NODE_ENV=development`. Is that right?

Yea. One option I haven't thought all the way through yet is to make `scripts/start.js` only set `NODE_ENV` if it hasn't already been set.

PR #170, opened 2019-08-27
Focus halo update

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2242

Almost all focus-able elements have the same yellow halo. The exceptions:

 * The search text box and button have no border when focused. The border is currently explicitly removed. I feel we should add the halo to the search button and _maybe_ to the search in put box as well.
 * Check boxes on the `/apply` page have the default focus styling. Apparently it is difficult to apply styling to check boxes. Any styles I try to apply are ignored.
 * The embedded video from youtube has it's own styling for when elements are focused. I imagine there is a way to overwrite those styles if desired.
 * The link in the **This is a beta site** notice uses the default focus styles because that element has a yellow background making the yellow halo invisible.

The border is shrunk in for some elements to make the halo "hug" the element a little tighter. We don't want to shrink in the border for all elements, mostly <a> tags, because the halo sometimes makes text hard to read.

Seems natural to define halo styles in `base.scss` because they apply to the whole site. If there is a better place (or way) to define those styles let me know!


1Copenut:
@jwoodman5 I pulled the code to my local machine and checked this branch for the yellow halo. LGTM and I appreciate the calls to modify to a dark halo on yellow background and avoid the embedded YouTube focus halo.

PR #171, opened 2019-08-29
Generalize ApiCard component

kalilsn:
During my work for vets-contrib#2615 I noticed that the ApiCard component's markup was being recreated in the News section. The ApiCard component was also being used in the support section in the same way as in the News section. Given that, I decided to make it slightly more general since it's a pattern we're using throughout the site.


jasalisbury:
> Visual regression testing failed. Review these diffs and then [update the snapshots](https://github.com/department-of-veterans-affairs/developer-portal/blob/master/docs/testing.md#visual-regression-testing). 
> 
> [renders-explore-benefits-properly-1-diff.png](https://s3-us-gov-west-1.amazonaws.com/developer-portal-screenshots/feb58f5ec84907c64f578af88ca4373005132670/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-properly-1-diff.png)

At least you get a screenshot this time!

kalilsn:
>At least you get a screenshot this time!

Turns out it actually caught an issue: I lost a commit when bringing this to a new branch!

src/components/CardLink.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import './CardLink.scss';
+
+export interface IApiCardProps {
+  name: string;
+  description: string;
+  url: string;
+}
+
+export class CardLink extends React.Component<IApiCardProps, {}> {
```
dvogel: Breaking this out into a more generic component makes a lot of sense to me. Since it is becoming more generic, I think we'd benefit from a block comment above the class declaration to guide it's future evolution. E.g. it might not be clear to everyone why children are rendered between the name and the description. Without any explanation the reader's best option is to look at how it's used. Right now they might see only the internal-only use and make an unhelpful inferrence. Is there a rationale for the visual structure that can be documented without mentioning the internal-only use?

kalilsn: I actually can't think of a good rationale other than "this is how it's being used now" 😄. Perhaps instead of using the children prop, I could use a `subhead: JSX.Element` prop? I think that would make it unsurprising that it's rendered after the name but before the description, at the cost of making it slightly less pleasant to use (IMO JSX props are kind of a pain to use – not sure if that opinion is shared tho!)

dvogel: I could get behind the `subhead` prop approach. Another approach I could see working well is moving the description into the children. The name of the component `CardLink` evokes the `<a>` tag, which expects it's description to be provided as a child node.

mjlumetta: yeah, like this decision too 😄 if you're interested in alternatives to passing a `JSX.Element` prop, an alternative might be passing a functional component that takes no props (`React.StatelessComponent`) as `subhead`. right now we only use `VAInternalOnlyTag`, which fits. this would be similar with what we do with the `mdx` components in `apiDefs` (in terms of use, not component definition)

[this article](https://medium.com/@martin_hotell/react-children-composition-patterns-with-typescript-56dfc8923c64) also offers some interesting suggestions for how to use other patterns to create more complex types of children. they get much less appealing as you go on but i'm intrigued by (2) - passing an object as `props.children` rather than a `ReactNode` or `ReactChild` or `JSX.Element` - and think we could consider whether we want to start trying that out

src/containers/DocumentationOverview.tsx
```diff
@@ -2,24 +2,28 @@ import * as React from 'react';
...
 
 export default class DocumentationOverview extends React.Component<RouteComponentProps, {}> {
   public render() {
     return (
       <div className="doc-overview">
-        <PageHeader header="Documentation" description="Explore usage policies and technical details about VA's API offerings."/>
+        <PageHeader
+          header="Documentation"
+          description="Explore usage policies and technical details about VA's API offerings."
+        />
```
mjlumetta: passing comment but the tag closes (improvements! ty) are reminding me that we need to work on our linter. i'm gonna start keeping a list

src/containers/News.tsx
```diff
@@ -45,31 +55,35 @@ export function SideNav() {
...
     return (
-      <NavHashLink key={section.id} to={`#${section.id}`} className="va-api-card">
-        <h3 className="va-api-name">
-          {section.title}
-        </h3>
-        <div className="va-api-description">
-          {section.description}
-        </div>
-      </NavHashLink>
+      <CardLink key={section.id} url={`#${section.id}`} name={section.title}>
```
mjlumetta: can you please inline this in `render()` now that it's smaller? unsure that there's substantial benefit to having this separate as a private class member

mjlumetta:
ahaha well nvm you already merged, but none of that was important anyway

kalilsn:
🤦‍♂ sorry, I should have waited! I had kind of forgotten about this PR and figured my changes were pretty minor. I appreciate your suggestions about the element props – very helpful article too! I'll see about inlining that function soon too. I think there's probably a lot of that sort of cleanup to be done.

As for the linter thing, if I have a sec, I'm going to open a(nother) PR that runs everything through prettier AND make sure our jenkinsfile actually checks the files via `prettier -l` so that we can enforce it.

PR #173, opened 2019-09-03
Upload source maps to sentry

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/1762

This change uploads source maps to sentry in a release when building on master and sets the release when sending errors to sentry. There are still some work to do:

 * Need a Sentry service account with an auth token
 * Need to add the service account's auth token to credstash and make it available inside Jenkins (via the same mechanism that other credentials are available to Jenkins jobs)

Some more open questions/problems that I don't have a strong opinion about:

 * Determine if we want to keep separate Sentry projects for each environment or roll all developer-portal Sentry projects into one Sentry project. Sentry has support for having multiple environments in a single project, have to set the environment when initializing Sentry.
 * `envName` can be `dev`, `staging` or `production` but the Sentry projects are `dev`, `staging` or `prod`. Will either need to change the project name in Sentry (might not be able to), create a new project in Sentry or create a map `Jenkins Env` -> `Sentry Env`
 * `upload_source_map.sh` is not specific to this repo and can be moved out (into Jenkins? unsure where or how)
 * Right now the developer-portal does not initialize Sentry if there is no DSN set. We could also prevent Sentry from being used if the release isn't set, solving [this ticket](https://github.com/department-of-veterans-affairs/vets-contrib/issues/2418)

I also need to look at [testing Jenkins](https://github.com/department-of-veterans-affairs/devops/blob/master/docs/Testing%20Jenkins%20Changes.md) to make sure my changes actually work before pushing.


src/index.tsx
```diff
@@ -11,6 +11,7 @@ import store from './store';
 if (process.env.REACT_APP_SENTRY_DSN) {
```
jasalisbury: Could change this to
```
if (process.env.REACT_APP_SENTRY_DSN && process.env.REACT_APP_SENTRY_RELEASE) {
```
to only use Sentry if the DSN and release are set in the env file (preventing sentry from being used on review instances)

Jenkinsfile
```diff
@@ -226,11 +226,19 @@ node('vetsgov-general-purpose') {
...
-            sh "cd /application && NODE_ENV=production BUILD_ENV=${envName} npm run-script build ${envName}"
+            if(onDeployableBranch()) {
+              withCredentials([string(credentialsId: 'sentry_auth_token', variable: 'SENTRY_AUTH_TOKEN')]) {
+                def sentryRelease = "${envName} - ${env.BUILD_NUMBER}"
+                def sourceDir = "build/${envName}/static/js"
+                sh "cd /application && SENTRY_RELEASE=${sentryRelease} NODE_ENV=production BUILD_ENV=${envName} npm run-script build ${envName}"
+                sh "cd /application/scripts && SENTRY_RELEASE=${sentryRelease} SOURCE_DIR=${sourceDir} SENTRY_URL=http://sentry.vfs.va.gov/ SENTRY_AUTH_TOKEN=$TOKEN SENTRY_ORG=vets-gov SENTRY_PROJECT=developer-portal-backend-${envName} ./upload_source_map.sh"
+              }
+            } else {
+              sh "cd /application && NODE_ENV=production BUILD_ENV=${envName} npm run-script build ${envName}"
```
kalilsn: you can factor out the `cd /application` from all of these scripts with 
```groovy
dir('/application') {
    sh '...'
}
```

Similarly, you can factor out the shared env variable stuff with `withEnv(["k1=v1","k2=v2"]) {}`

Jenkinsfile
```diff
@@ -226,11 +226,19 @@ node('vetsgov-general-purpose') {
...
       try {
       def builds = [:]
-
       envNames.each{ envName ->
         builds[envName] = {
           dockerImage.inside(args) {
-            sh "cd /application && NODE_ENV=production BUILD_ENV=${envName} npm run-script build ${envName}"
+            if(onDeployableBranch()) {
+              withCredentials([string(credentialsId: 'sentry_auth_token', variable: 'SENTRY_AUTH_TOKEN')]) {
+                def sentryRelease = "${envName} - ${env.BUILD_NUMBER}"
```
kalilsn: Is there a reason not to use the git hash for this? We name our build artifacts with that, so I think it would be nice to keep that identifier consistent.

kalilsn:
>envName can be dev, staging or production but the Sentry projects are dev, staging or prod. Will either need to change the project name in Sentry (might not be able to), create a new project in Sentry or create a map Jenkins Env -> Sentry Env

Not sure yet about the other questions (IDK how multi-environment stuff looks on sentry), but I should be able to update the sentry project name, and that seems like the easy road!

src/index.tsx
kalilsn: I don't have a strong opinion about how to solve this one. That sounds like it would work, but would be worth a comment explaining why. Otherwise, we could set the DSN env variable in the `Jenkinsfile` like you're doing for the release

Jenkinsfile
jasalisbury: Build number seems easier to reason about. If we use the build number it is easy to look up a running build of the developer portal in Jenkins. I feel It is easier to remember "I need to look at build number 100" instead of "I need to look at git hash `eaf015...`". If you see an error in Sentry it is still easy to checkout that build to reproduce (as the commit is tagged with the build number). However using the build number instead of git hash is purely convenience since we tag built commits with the build number, so I'd really be fine either way.

jasalisbury:
Looking at this after the weekend I am confused why I decided to upload source maps manually instead of using the webpack plugin. I'm going to start looking into just using the Sentry webpack plugin

jasalisbury:
> Looking at this after a night of sleep I am confused why I decided to upload source maps manually instead of using the webpack plugin. I'm going to start looking into just using the Sentry webpack plugin

I now remember why I didn't use the webpack plugin. I didn't see an obvious way to prevent source map upload for review instance builds when using the webpack plugin. We could pass in some variable and conditionally upload source maps but it feels like the better option is to keep upload of source maps outside of the actual build process to prevent additional complexity in the webpack config.

Jenkinsfile
kalilsn: >the commit is tagged with the build number

It is? I see that we're uploading it to S3 according to the hash, and our releases do have a number with them (e.g. https://github.com/department-of-veterans-affairs/developer-portal/releases/tag/developer-portal%2Fv0.0.198), but that's not the same as the Jenkins build number.

I think the `BUILD_NUMBER` will just be an integer count of the number of times that job has run. That means that if we ever were to lose our jenkins 
 build history (from changing the job, or losing the EBS volume, etc.), we'd wind up with a collision.

jasalisbury: Oh gotcha. Yeah in that case the commit hash is the way to go!

jasalisbury:
> Not sure yet about the other questions (IDK how multi-environment stuff looks on sentry), but I should be able to update the sentry project name, and that seems like the easy road!

Cool that seems to be the way to go if you can update the project name!

PR #178, opened 2019-09-16
Update News page with 2 news articles

stephaniemace:
2 news articles mentioning VA Lighthouse were published on September 12. Updating the news page with both articles.


nfasulo:
I talked to @kalilsn about paths forward. We agreed that the best solution available on the horizon would be to mock out the content of this page (and similar pages like Release Notes) for the purposes of visual regression testing. That way PRs like this will not fail until the images are updated, but we will still be protected against changes to components and styling inadvertently affecting the page and continue to run end to end and accessibility tests on it. This would also set us up nicely for a rosy future where the content of a page like this is maintained in a separate database.

In the meantime I have updated the visual regression tests for this PR.

PR #180, opened 2019-09-16
Auto-fix npm audit errors.

dvogel:
I ran `npm audit --fix` and the tests passed locally. None of the changes in the diff jumped out at me as surprising.


jasalisbury:
https://github.com/department-of-veterans-affairs/developer-portal/pull/179 (Someone needs to say jinx right?)

kalilsn:
I think @mjlumetta just merged these changes in with #3012 so this might not be necessary anymore?

PR #182, opened 2019-09-17
Set rewrite flag to true when uploading maps to sentry

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/1762

The default `--no-rewrite` is going to soon be deprecated by Sentry. To make sure our sourcemap file is formatted the way Sentry wants the upload has the `--rewrite` flag. This should add any missing source information into our source map file. It also validates the source map file stricter than Sentry's `--validate` option. 

https://docs.sentry.io/platforms/javascript/sourcemaps/#uploading-source-maps-to-sentry
https://docs.sentry.io/cli/releases/


jasalisbury:
> I don't completely understand why the `--no-rewrite` deprecation is relevant, and I take it fixing this in webpack is more difficult?
> 
> Those are both just curiousity questions though, this seems good to me!

Sorry I wasn't very clear... Right now if you don't provide a `--no-rewrite` or `--rewrite` flag the upload will default to `--no-rewrite`. This default is being deprecated; you will have to explicitly provide `--rewrite` or `--no-rewrite` when uploading source maps in the future. If this doesn't work looking at fixing it in webpack is probably the next step.

PR #185, opened 2019-09-18
Report signup errors

kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3144

To help with future investigations of issues like the Firefox/CORS one, this explicitly sends an event to sentry whenever a signup fails. I've set the level to warning (default is error), which should allow us to configure a separate notification option if they're not useful.

I've also just removed the `fetchWithRetry` function because it's not working. I didn't spend too much time debugging it because I'm having a hard time seeing the value. If the form submission fails, it's simple to hit the retry button since none of the form fields are cleared out.


dvogel:
> I've also just removed the `fetchWithRetry` function because it's not working. I didn't spend too much time debugging it because I'm having a hard time seeing the value. If the form submission fails, it's simple to hit the retry button since none of the form fields are cleared out.

I believe the original intent of the retry was to help users avoid errors due to the rate limiting performed by kong on behalf of the sign-up lambda.

kalilsn:
> I believe the original intent of the retry was to help users avoid errors due to the rate limiting performed by kong on behalf of the sign-up lambda.

We rate limit by IP address now (10/hr, 2/minute), so that seems like another good reason to remove this!

PR #186, opened 2019-09-18
Use optimize-css-assets-webpack-plugin to eliminate CSS duplication

mjlumetta:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3012

weirdly, `OptimizeCSSAssetsPlugin` was already in `webpack.config.prod.js`, but it wasn't being used. this cuts the CSS bundle size down to 172K from 355K.

we shouldn't need this in development because in development resource loading is managed by hot module replacement.


dvogel:
> If this is our solution to the duplicate CSS problem, I'd like to see the previous changes (`uswdsPatch.scss` etc.) reverted so we can prove that we no longer need to worry about duplicate imports.
> 
> I also would like it enabled in dev, since the actual browser parsing time for that huge bundle of CSS was a significant pain for me.

Enabling it in dev should speed up the e2e tests too, since the tests start up a dev server -- right?

kalilsn:
>Enabling it in dev should speed up the e2e tests too, since the tests start up a dev server -- right?

Assuming there's not a significant cost to running this plugin, I would hope so! In general I think keeping the dev/prod config the same should be our default.

dvogel:
> > Enabling it in dev should speed up the e2e tests too, since the tests start up a dev server -- right?
> 
> Assuming there's not a significant cost to running this plugin, I would hope so! In general I think keeping the dev/prod config the same should be our default.

That's a good point. There's certainly a cost to this plugin. Not sure how large that cost is, but I think it's worth making us reconsider removing the usdsPatch as part of this merge. Removing the usdsPatch now will force this plugin to do a lot more work. I'd still like to see a branch (off this branch) showing that it would work before we commit to this path. Removing the usdsPatch forever seems best left to https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2893 because that will result in a lower level of duplication in the CSS text that this plugin would have to remove.

mjlumetta:
>> Assuming there's not a significant cost to running this plugin, I would hope so! In general I think keeping the dev/prod config the same should be our default.
>
> Not sure how large that cost is, but I think it's worth making us reconsider removing the uswdsPatch as part of this merge. Removing the uswdsPatch now will force this plugin to do a lot more work.

yeah, so here's the thing: our dev and prod config in the SCSS/CSS pipeline are already different at the end. in prod, we use `MiniCssExtractPlugin.loader` as the final loader in the chain, after `style-loader`, and that loader is responsible for merging all the CSS chunks into `static/css/main.[contenthash].css`. in dev, the chain ends with the `style-loader`, and under `plugins` we have the `HotModuleReplacementPlugin`. i'm pretty sure that the HMR plugin does its own magic to serve and that therefore we can't go beyond that point in terms of optimizing size (nor should we have to). i'm not 100% sure of that, though, so i will do some more reading and try to come up with something more definitive. 

i agree with Drew, though, that this probably means we should reconsider our approach to the `uswdsPatch`. removing it now means that some of our SCSS files will need to import `base.scss` again, which will inflate the size of some chunks substantially, and consequentially dev performance will suffer. so i think removing it in 2893 is fine.

> I also would like it enabled in dev, since the actual browser parsing time for that huge bundle of CSS was a significant pain for me.

i'm not sure the CSS bundle is your problem. it might have been prior to the `uswdsPatch`, but now that it's down to 355K (in prod), i don't think that's the issue - especially since the Javascript bundle is 2.2M:

JavascriptBundleDev

so maybe that's something to look at. i'm also not sure if the CSS bundle is _ever_ compiled into one file/module with HMR and Webpack dev server, so i'm not sure if there was ever a problem there? if you know otherwise, would love to hear the explanation.

kalilsn:
>i'm not sure the CSS bundle is your problem. it might have been prior to the uswdsPatch, but now that it's down to 355K (in prod), i don't think that's the issue - especially since the Javascript bundle is 2.2M

Absolutely! It's fine now, with the patch solution. But I'm not comfortable with this as our long term solution to duplication if we can't run it in dev. Even if we don't get rid of that patch file now, we will eventually get back to a state closer to where we were before the patch, which makes the app extremely unwieldy in a browser. I'm convinced that it's fine not to remove that patch in this PR, but I would still like the plugin enabled in dev.

jasalisbury:
> yeah, so here's the thing: our dev and prod config in the SCSS/CSS pipeline are already different at the end

Do they need to be? (not suggesting we make them the same for this PR, just a question)

>  especially since the Javascript bundle is 2.2M

One technique I've seen is bundling dependencies separate from application code. Since dependencies don't change as often the browser can just use a cached version of the dependencies most of the time and only download the smaller application bundle.

mjlumetta:
@kalilsn will research HMR and whether this can be done in dev. i think this relates to what @dvogel was saying before demo, that compiler level solutions aren't a replacement for writing good modular code and won't necessarily solve problems; this might be one of those problems. i will look into it and get a more definite answer as to whether it is possible or not. in any case, if the chunks are never merged, i'm not sure what solution would be possible, and i'm not sure it would matter. if that's true, then only keeping chunks small will improve the performance at that point. which ought to get a long-term solution in #2893.

mjlumetta:
@kalilsn @dvogel @jasalisbury i manually checked the sizes of the CSS blob files in the network tab with and without `OptimizeCSSAssetsPlugin` and the file sizes were unchanged. hopefully that information is useful in terms of future improvements that we might make for local development. i might try to do a bit of reading this afternoon or next week about the `WebpackDevServer` API and whether there's anything there that might be useful - not for this PR per se, but to see if there's anything that might be useful in the future.

PR #188, opened 2019-09-24
Refactor some developer portal components in prep for deprecating Argonaut

mjlumetta:
this PR is a precursor to solving [3138](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3138). it does some pretty significant refactoring of where things live and what they're named on the dev portal without altering the logic. changes that i was trying to make here:

* extract schema (i.e., typescript interfaces) from `src/apiDefs/index` to move a little bit towards having a database for our API definitions
* rename components to better reflect their content (e.g., distinguishing between `Category*` and `Api*` components, replacing `Explore` with `Documentation` in component names)
* move some components from `src/components` to `src/containers` to better reflect that pattern 
  * i'm agnostic on this pattern, and if i had to choose, i'd lean toward picking a new pattern. but as long as we have it, i think we should try to follow it. use of `apiDefs` is a pretty good heuristic for whether a component is a container (yes) or a presentational component (no).
* create `documentation` and `releaseNotes` subdirectories within `src/containers`
* remove environment checks for FHIR API, as it should be enabled everywhere

i want to open and ideally merge this PR quickly if we can agree on the changes. if other people have open or merge PRs, this will create heinous merge conflicts for somebody. it would be great to avoid that if possible.


src/Routes.tsx
```diff
@@ -13,7 +13,7 @@ import {
   BetaPage,
   BetaSuccess,
   DisabledApplyForm,
-  ExploreDocs,
+  DocumentationPage,
```
dvogel: Maybe `DocumentationPage` should be more specific re: what it's documenting?

src/apiDefs/benefits.ts
```diff
@@ -1,7 +1,7 @@
-import { IApiDescription } from '.';
+import { IApiDefinition } from './schema';
```
dvogel: "description" vs "definition" feels like it's approaching bike shed territory. We probably won't ever have a perfect name and those two seem super close to me.

src/content/apiDocs/index.ts
```diff
@@ -30,9 +26,9 @@ const facilitiesContent: IApiCategoryContent = {
 };
 
 const healthContent: IApiCategoryContent = {
-  intro: isNewFhirApiEnabled ? HealthIntro : LegacyHealthIntro,
+  intro: HealthIntro,
   overview: HealthOverview,
-  quickstart: isNewFhirApiEnabled ? HealthQuickstart : undefined,
+  quickstart: HealthQuickstart,
```
dvogel: I'm in favor of refactoring the code as we need to in order to keep it well organized as we work on tickets. It would be nice to see these specific ticket AC-related items split out into a separate PR though. That makes it easier to discuss the merits of each separately.

src/apiDefs/schema.ts
```diff
@@ -0,0 +1,48 @@
...
+  likely move this data to a non-Typescript location, i.e. a database.
+*/
+
+import * as moment from 'moment';
+
+export interface IApiCategoryContent {
+  readonly intro: React.StatelessComponent;
+  readonly overview: React.StatelessComponent;
+  readonly quickstart?: React.StatelessComponent;
+}
```
dvogel: :heart: re: the block comment at the top. It's probably worth amending it to explain that `React` component use should be moved out of this module and new ones shouldn't be added because they will only add to the pile of work required to move this stuff to a database.

src/Routes.tsx
mjlumetta: how about `ProgramDocumentation`? `ApiPlatformDocumentation`? `ProgramApisDocumentation`?

dvogel: I just took a closer look at the `ExploreDocs` component and it doesn't really document anything else. It's a bit like a layout in that it provides the nav and then routes to specific pages for the content. In that way `DocumentationLayout` makes sense to me. That's not a particularly common term in React land though. We've been discussing getting rid of the `containers` vs `components` directory split but this might be a place to retain the container term with `DocumentationContainer`. I could also be convinced to use a generic bespoke term like `DocumentationRoot` since it handles routing the root of the doc routes.

PR #191, opened 2019-09-25
Routing readme update

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3133


config/webpack.config.prod.js
```diff
@@ -1,3 +1,6 @@
+// If you are looking to add an additional file hosted at the root of the project see the readme
+// section `Revproxy Routing`
```
mjlumetta: add link to the README in github?

README.md
```diff
@@ -38,3 +38,7 @@ updates to the site were made. It took multiple visits to the site for the updat
...
 to display an incident banner on the developer portal. Most users will never see the banner
 because of the service worker. To prevent this issue in the future the service worker was removed.
+
+## Revproxy Routing
+
+The developer portal sits behind an Nginx reverse proxy. Nginx is configured to route all requests to `/static` through to `/static`. All other paths are routed to `index.html`. This allows react to be in control of `404`ing any bad paths. However there are a group of paths at the root that need to be routed to (favicon.ico, google analytics, etc...). Ngnix is configured to explicitly route to these files. If you need to add an additional file hosted at the root of the developer portal do so [here](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/revproxy-vagov/vars/developer_portal_root_routes.yml). The vars in that file are used where the developer-portal [is configured](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/deployment/config/revproxy-vagov/templates/nginx_revproxy.conf.j2#L668) in the revproxy (search for `developer_portal_root_routes`)
```
mjlumetta: do we need to add a 404 to React along with this ticket? the behavior after the devops ticket is merged will be rendering the site with no content where the `Route`s are rendered, which is in my opinion a sideways move if not an improvement from the current behavior. but it would be good to get a real 404 page in there and relatively simple to add a basic one. the ticket is https://github.com/department-of-veterans-affairs/vets-contrib/issues/2342

dvogel: My memory of the conversations around this revproxy ticket and the 404 page ticket is that if we did the 404 page ticket first, we wouldn't be able to test it until this revproxy ticket was done.

jasalisbury: Adding a 404 page feels like a different PR to me.

mjlumetta: that's correct. here's my thinking: 

* while the `nginx` change provides value to us by addressing tech debt and unblocking that other ticket, it does not provide any value to the end user of our site
* we have a PR associated with this change open in this repo
* it would be trivial to add a placeholder that informs the end user that the route isn't valid without doing all of 2342 (i.e., our ideal 404 page)

it would be simple to add a [`Redirect`](https://reacttraining.com/react-router/web/api/Redirect) to a new `/notfound` route as the catch-all route within the developer portal - that's only one line of code and it does produce the desired value for the end user, albeit in an imperfect form. slightly more effort would be doing the above and adding the actual `Route` for `/notfound` that renders a placeholder `NotFound` component whose content is just an `h1` that says "404 Not Found". that's sufficiently small enough of a change that delivers great enough of a value that i think we should do that, and it can be tested locally even if the other change isn't deployed. i'm fine doing a quick commit that demonstrates implements this.

jasalisbury: Just added a 404 page, want to pull it down and take a look @mjlumetta? Does the addition match what you expected?

mjlumetta: yeah definitely. only thing i'd say is to add a wrapper on the `NotFound` component, because the `h1` is on the left side of the browser. if you don't want to do this though i don't think it's necessary

dvogel: :+1: for avoiding `Redirect`. I'm pretty sure UX best practices are for someone getting an error page to be able to copy/paste the mismatched URL out of their browser location bar.

PR #193, opened 2019-09-26
Additional changes to support API schema changes

mjlumetta:
this branch contains

* extraction of API lookup functions from `index` into `query` and deletion of `index`
* always using functions to access API definitions rather than accessing constants directly


src/App.tsx
```diff
@@ -20,10 +20,11 @@ history.listen(location => {
 import 'highlight.js/styles/github.css';
 
 const flags = {
-  hosted_apis: apiEnvFlags,
+  ... getApiFlags(),
```
dvogel: It feels strange to me that the `hosted_apis` key would be defined by `getApiFlags()`. That bit seems endemic to `App.tsx` and unrelated to other components.

mjlumetta: yeah, i think that's fair. i'll move this calculation into `App.tsx`.

src/Routes.tsx
```diff
@@ -74,8 +75,10 @@ export function topLevelRoutes(props: RouteComponentProps<void>) {
 
 export function sitemapConfig() {
   function getApiRouteParams(route: string, apiCategory: string): string[] {
-    const routeParams = apiDefs[apiCategory].apis.reduce((result: string[], api) => {
-      if (apiEnvFlags.hasOwnProperty(api.urlFragment) && apiEnvFlags[api.urlFragment]) {
+    const apiDefs = getApiDefinitions();
+    const apiFlags = getApiFlags();
+    const routeParams = apiDefs[apiCategory].apis.reduce((result: string[], api: IApiDescription) => {
```
jasalisbury: Instead of `apiDefs[apiCategory]` should we use `lookupApiCategory(categoryKey)` defined in `apiDefs/query`?

src/containers/documentation/CategoryPage.tsx
```diff
@@ -18,7 +19,7 @@ export default class CategoryPage extends React.Component<RouteComponentProps<IA
       apis,
       name: categoryName,
       content: { intro, overview },
-    } = apiDefs[apiCategoryKey];
+    } = getApiDefinitions()[apiCategoryKey];
```
jasalisbury: Seems like this can be replaced with `lookupApiCategory`

src/containers/documentation/DocumentationOverview.tsx
```diff
@@ -15,7 +18,7 @@ export default class DocumentationOverview extends React.Component<RouteComponen
         />
         <div className="va-api-container">
           {apiCategoryOrder.map((apiCategoryKey: string) => {
-            const { name, shortDescription } = apiDefs[apiCategoryKey];
+            const { name, shortDescription } = apiDefinitions[apiCategoryKey];
```
jasalisbury: Can this be changed to use `lookupApiCategory`?

src/containers/documentation/OAuth.tsx
```diff
@@ -24,7 +23,7 @@ export class OAuth extends React.Component<IOAuthProps, {}> {
 
   public render() {
     const { apiCategoryKey } = this.props;
-    const category = apiDefs[apiCategoryKey] || {};
+    const category = getApiDefinitions()[apiCategoryKey] || {};
```
jasalisbury: `lookupApiCategory`?

src/apiDefs/data/categories.ts
```diff
@@ -64,4 +64,5 @@ const apiDefinitions : IApiCategories = {
   },
 };
 
+export const apiCategoryOrder: string[] = ['benefits', 'facilities', 'health', 'verification'];
```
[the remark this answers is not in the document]
dvogel: This ordering is likely to be further detached from definition of the rest of the API data since it's really a parameter of the portal UI.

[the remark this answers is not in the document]
jasalisbury: I would think the API is in charge of ordering and the UI could instruct the API how it wants the data ordered. But my thinking comes from the perspective of doing as much work as possible in the API as opposed to the UI.

src/Routes.tsx
mjlumetta: i don't think so, necessarily? `lookupApiCategory` just evaluates the same expression on the object returned by `getApiDefinitions()`. so we're still referencing `apiDefs/query`... unless that doesn't address your concern

src/apiDefs/deprecated.test.ts
```diff
@@ -1,12 +1,16 @@
...
-// let apiDefinitions = jest.genMockFromModule('./data/categories');
-// jest.mock('./data/categories');
-// import * as categories from './data/categories';
+jest.mock('./query');
+const query = require('./query'); // tslint:disable-line:no-var-requires
+const {
+  getAllApis,
+  lookupApiByFragment,
+} = query;
+getAllApis.mockReturnValue([]);
```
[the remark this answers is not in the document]
mjlumetta: kind of. it is on this branch but i fixed it on #194. the thing that i was doing was computing the deprecated flags inline in the module so that it would be calculated on startup and then just referenced in memory. as a result, `getAllApis()` would be called when the module was loaded, so if it returns `undefined`, there'd be an error and the test suite won't run. i decided that wasn't a good idea though and put the calculation within a function, which fixed the problem.

src/containers/documentation/DocumentationOverview.tsx
mjlumetta: it could be! i think those are roughly equivalent (`lookupApiCategory` will do the same thing internally). the easier thing here, i believe, is that we don't need to worry about `null` error cases (would probably write `lookupApiCategory(apiCategoryKey) || {}`. lmk which you prefer, either is fine

jasalisbury: ahh gotcha, at this point we know `apiCategoryKey` is valid which is why we don't use `lookupApiCategory` (because it can return null). I'm fine either way really was just wondering why we had `lookupApiCategory` and weren't using it in some places.

jasalisbury: Also you can ignore my other `lookupApiCategory` comments now that I understand why it wasn't used in those places.

mjlumetta:
@jasalisbury made changes to feedback on #194. i'm going to merge that one into this one since you approved. re: `lookupApiCategory` vs `getApiDefinitions` the reason for the latter is basically type safety (insofar as that is a thing). the linter/compiler complain about the return value of `lookupApiCategory` because it could be null, and we can't safely `!` it in components that are passed to `Route`s. hope that explanation helps.

PR #198, opened 2019-10-01
Fix form fields in ApplyForm

mjlumetta:
in reference to https://lighthouseva.slack.com/archives/CEXC8GV2S/p1569948668238100


kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3277

PR #199, opened 2019-10-01
Replace reference to USWDS in base.scss with VADS

mjlumetta:
department-of-veterans-affairs/vets-contrib#2893

This is the starting point for our replacement of USWDS with Formation. There are still some direct references to USWDS classes that may be removed in the future, but this PR allows us to use the utilities from formation. There are some minor changes to the actual style of the site: slightly smaller font size, larger header logo, and some minor spacing adjustments. These were mostly made in the interest of consistency: removing overrides of the design system where possible. The larger header simply matches the mockups more closely.


src/base.scss
```diff
@@ -33,6 +33,19 @@ footer {
...
+// Override <a> styles from formation. The non-hover transition
+// is defined in formation/sass/base/_va.scss and the hover styles
+// can be found in formation/sass/base/_b-mixins.scss
+a {
+  transition: none;
+
+  &:hover {
+    transition: none;
+    background-color: rgba(0,0,0,0);
+    text-decoration: none;
```
kalilsn: @jasalisbury Did you add `text-decoration: none` to fix a particular issue you found? It's changed the interaction so that the underline disappears whenever you hover over a link.

Sorry, should have checked this before I approved your PR!

jasalisbury: Yeah the text in api card links was being underlined. It looks like links on prod aren't underlined when they are hovered over so it seems fine, unless I'm misunderstanding something

kalilsn: Most of them are underlined: look at the `Send us your feedback` link or the ones in `Ready to get started?` at the bottom of the homepage. I see why you made the change though. I think I'm going to remove this from here and instead apply `text-decoration: none` only to the `CardLink` components – how does that sound?

jasalisbury: Oh I see it now. Yeah go for it!

src/components/MarkdownPage.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 
 export default function MarkdownPage(Component: MarkdownComponent): JSX.Element {
   return (
-      <section className="usa-section">
+      <section className="vads-u-padding-y--6">
```
kalilsn: `usa-section` isn't part of formation so we lost padding on the markdown pages. This isn't totally consistent with the other kinds of pages, but I think that should be refactored further down the line.

kalilsn:
@mjlumetta @dvogel @jasalisbury I think this is ready to go: once y'all approve I'll update these snapshots and merge. I forced it to give us a review instance too, let me know if anything seems broken!

mjlumetta:
@kalilsn this looks good to me. can't approve because i opened it initially but yeah

jest-puppeteer.config.js
```diff
@@ -11,7 +11,7 @@ if(!process.env.TEST_HOST) {
   puppeteerConfig.server = {
     "command": "PORT=4444 BROWSER=false npm run-script start",
     "port": 4444,
-    "launchTimeout": 60000,
+    "launchTimeout": 120000,
```
dvogel: This is a getting a little concerning. Fine for the purposes of this PR but we should discuss whether there's a better alternative for the future. It seems like requiring a dev build before the test suite is run would be a reasonable change.

kalilsn: Yeah, I'm going to write a ticket specifically for addressing this because I don't want it to get forgotten.

PR #201, opened 2019-10-01
Fix footer

jasalisbury:

Formation doesn't seem to have styles for .usa-footer-logo defined. Created a `va-api-footer-logo` class and added padding.

### DIFF
--- a/src/components/Footer.scss
+++ b/src/components/Footer.scss
@@ -10,6 +10,15 @@
     width: 200px;
   }
 
+  // Copied from uswds uswds/src/stylesheets/components/_footer.scss
+  .usa-footer-logo {
+    padding: 1rem 0;
+
+    @include media($medium-screen) {
+      padding: 2rem 0;
+    }
+  }
+
   .usa-footer-links {
 
     float: left;
--- a/src/components/Footer.scss
+++ b/src/components/Footer.scss
@@ -10,15 +10,6 @@
     width: 200px;
   }
 
-  // Copied from uswds uswds/src/stylesheets/components/_footer.scss
-  .va-api-footer-logo {
-    padding: 1rem 0;
-
-    @include media($medium-screen) {
-      padding: 2rem 0;
-    }
-  }
-
   .usa-footer-links {
 
     float: left;
--- a/src/components/Footer.scss
+++ b/src/components/Footer.scss
@@ -11,7 +11,7 @@
   }
 
   // Copied from uswds uswds/src/stylesheets/components/_footer.scss
-  .usa-footer-logo {
+  .va-api-footer-logo {
     padding: 1rem 0;
 
     @include media($medium-screen) {
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -8,7 +8,7 @@ export default function Footer() {
     return (
         <footer className="usa-footer usa-footer-medium" role="contentinfo">
           <div className="usa-grid">
-            <div className="usa-footer-logo usa-width-one-third">
+            <div className="va-api-footer-logo usa-width-one-third">
               <a href="https://www.va.gov">
                 <img src={logo} alt="Department of Veterans Affairs" />
               </a>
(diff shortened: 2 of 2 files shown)


dvogel:
Copying styles from uswds feels like one point in favor of using uswds as a direct dependency alongside formation. Anyone disagree?

kalilsn:
>Copying styles from uswds feels like one point in favor of using uswds as a direct dependency alongside formation. Anyone disagree?

I think I disagree, but it might be because I don't completely understand how you see that being implemented. I'd be a little worried about future compatibility if we continue using a different version of uswds than formation (currently we're on 1.6.12, they're on 1.6.10). Also, in my mind, a big part of the value of this work is getting us to a place where we can write (or borrow!) clear guidelines about how we work with our design system. Including uswds as a direct dependency makes that harder to do because the choices formation makes aren't always the same as uswds. If formation isn't suiting our needs, I would rather just move to uswds 2.x instead of formation.

Regardless, I don't think that we actually need to copy styles from uswds in this case. Formation provides classes that can replace these styles: https://design.va.gov/utilities/padding#padding-top-and-bottom and https://design.va.gov/utilities/padding#responsive-prefixes. That's exactly the kind of thing I see happening in https://github.com/department-of-veterans-affairs/vets-contrib/issues/3011, but IMO we should start doing that now where possible!

src/components/Footer.scss
```diff
@@ -10,6 +10,15 @@
     width: 200px;
   }
 
+  // Copied from uswds uswds/src/stylesheets/components/_footer.scss
```
mjlumetta: i think we can remove this comment since we intend to decouple it from USWDS

src/components/Footer.scss
```diff
@@ -10,6 +10,15 @@
...
+  .va-api-footer-logo {
+    padding: 1rem 0;
+
+    @include media($medium-screen) {
+      padding: 2rem 0;
```
mjlumetta: do we have a feeling about `rem` vs `px` for sizing? right now there is no rhyme or reason to when each is used

kalilsn: The main reason I can think of to use one or the other is how browser zoom will work, which is something we're going to need to address soon anyways as part of our 508 work. If we were changing the base font size with different screen widths, then that would also be a consideration, but I don't think we do that? IMO we should probably default to `rem`, but I'm not 100% sure.

dvogel:
> > Copying styles from uswds feels like one point in favor of using uswds as a direct dependency alongside formation. Anyone disagree?
> 
> I think I disagree, but it might be because I don't completely understand how you see that being implemented. I'd be a little worried about future compatibility if we continue using a different version of uswds than formation (currently we're on 1.6.12, they're on 1.6.10).

I don't see us deviating much. I don't see the current mismatch as an issue because we are ahead of them and it's only a higher patch level. Presumably they are using semver, so that shouldn't break anything. If they are, chances are we can also make due with 1.6.10 versus 1.6.12 for a short period while we work to resolve the conflict.

My suggestion comes from a bit of induction. I see this [`.usa-footer-logo` class defined in 1.6.12] (and 1.6.10). If that's not coming through to our CSS bundle, I assume that's because formation is shielding us from the definition of uswds styles not used by formation itself. Am I misunderstanding something there?

kalilsn:
> > > Copying styles from uswds feels like one point in favor of using uswds as a direct dependency alongside formation. Anyone disagree?
> > 
> > I think I disagree, but it might be because I don't completely understand how you see that being implemented. I'd be a little worried about future compatibility if we continue using a different version of uswds than formation (currently we're on 1.6.12, they're on 1.6.10).
> 
> I don't see us deviating much. I don't see the current mismatch as an issue because we are ahead of them and it's only a higher patch level. Presumably they are using semver, so that shouldn't break anything. If they are, chances are we can also make due with 1.6.10 versus 1.6.12 for a short period while we work to resolve the conflict.
> 
> My suggestion comes from a bit of induction. I see this [`.usa-footer-logo` class defined in 1.6.12](https://github.com/uswds/uswds/blob/release-1.6.10/src/stylesheets/components/_footer.scss#L306) (and 1.6.10). If that's not coming through to our CSS bundle, I assume that's because formation is shielding us from the definition of uswds styles not used by formation itself. Am I misunderstanding something there?

Nope, I was misunderstanding. The version thing was a total red herring, 1.6.10 and 1.6.12 are actually identical releases. Formation only includes a subset of USWDS classes. However, I still stand by my other objections to including both!

PR #202, opened 2019-10-02
fix global a styles

jasalisbury:

### DIFF
--- a/src/base.scss
+++ b/src/base.scss
@@ -33,6 +33,16 @@ footer {
   flex-shrink: 0;
 }
 
+a {
+  transition: none;
+
+  &:hover {
+    transition: none;
+    background-color: rgba(0,0,0,0);
+    text-decoration: none;
+  }
+}
+
 .main {
   flex-grow: 1;
   background-color: #f9f9f9;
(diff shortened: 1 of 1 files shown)


src/base.scss
```diff
@@ -33,6 +33,16 @@ footer {
...
+a {
+  transition: none;
+
+  &:hover {
+    transition: none;
+    background-color: rgba(0,0,0,0);
+    text-decoration: none;
+  }
+}
+
```
dvogel: Can you add a comment here pointing to the global style that this is meant to override? I'd hate to get to a point where this breaks and someone comes up with a clever but ill-advised workaround.

dvogel:
I hit approve despite the test failures because it looks like those might have to be resolved by unrelated changes on `css-to-va-design-system`. Does anyone know for sure why those tests are failing? It certainly doesn't look like anything related to this PR.

jasalisbury:
> I hit approve despite the test failures because it looks like those might have to be resolved by unrelated changes on `css-to-va-design-system`. Does anyone know for sure why those tests are failing? It certainly doesn't look like anything related to this PR.

Looks to be accessibility failures, which make sense to me with the state of CSS on the developer portal makes the site look wonky.

PR #203, opened 2019-10-02
Add first draft of provider integration guide.

dvogel:
Addresses https://github.com/department-of-veterans-affairs/vets-contrib/issues/3146 by publishing a new, hidden markdown page. Since the page cannot be navigated to via a link, you cannot view it on a review instance and need to run the site locally.


src/containers/documentation/ProviderIntegrationGuide.jsx
```diff
@@ -0,0 +1,17 @@
+// This module cannot use TypeScript because the compiler does not have enough
+// type information about the component returned by the markdown component
+// loader to understand that it should accept the standard `className`
+// attribute.
```
kalilsn: I think the problem here is actually that the compiler is being supplied with incorrect information in the type definition: https://github.com/department-of-veterans-affairs/developer-portal/blob/dv-3146-provider-integration-guide/src/types/md.d.ts#L2

Changing that to `React.StatelessComponent<React.HTMLAttributes<HTMLDivElement>>` satisfies my compiler!

kalilsn: I guess that probably prevents us from passing any custom props to these components, so it could be changed to `React.FunctionComponent<any>`. But that restriction might actually be an advantage for making it easier to move this kind of content to a database eventually.

mjlumetta: yeah, agree with Kalil. i think defining types for props is a good move, and you can define specific types in that definition file Kalil linked. i don't like the ida of putting `React.[modifier]Component<any>` because at that point it's probably more honest to use Javascript since they're functionally equivalent.

ps anybody

dvogel: > Changing that to React.FunctionComponent<React.HTMLAttributes<HTMLDivElement>> satisfies my compiler!

Interesting. :thinking:

> so it could be changed to React.FunctionComponent<any>

I tried something similar that did not work. I'll investigate this more.

> i don't like the ida of putting React.[modifier]Component<any> because at that point it's probably more honest to use Javascript since they're functionally equivalent.

I think it's at least useful to be explicit about the `any` parameter, if nothing else than to have a better home for the comment I left in my component module. I think it's going to be difficult to be both explicit and correct about this since the props type is generated at runtime based on the config: https://github.com/adhocteam/markdown-component-loader/blob/develop/src/formatters/module.js#L12-L28

src/containers/documentation/ProviderIntegrationGuide.jsx
```diff
@@ -0,0 +1,17 @@
...
+import * as React from 'react';
+import IntegrationGuideContent from '../../content/providers/integrationGuide.mdx';
+
+export default function ProviderIntegrationGuide() {
+  return (
+      <section className="usa-section">
```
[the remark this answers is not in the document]
dvogel: I agree re: a `Page`-esque component. That would be useful for situations where we just need to wrap a component in some DOM elements to remain consistent with the rest of the layout elements. I'm not particularly keen to move the instantiation into the route itself though. The purpose of this `ProviderIntegrationGuide` component is to be a natural home for additional provider content. Similar to the catch-all `RoutedContent` but for provider content.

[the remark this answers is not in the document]
mjlumetta: i see - so you envision this having more child components in the future, then? that's totally fine with me then. i think that having a generic wrapper is still potentially useful but if you envision the component growing in the future then i agree it doesn't apply here.

src/containers/documentation/ProviderIntegrationGuide.jsx
kalilsn: Hm, looking at that it seems like we actually just want `React.FunctionComponent<{className?: string, style?: string}>` since those are the only two props that are always included. If someone uses a markdown component and defines additional custom props in the front matter, it seems reasonable that they should have to define a specific type for those props (which could inherit from the default one).

dvogel: I used the `React.HTMLAttributes<HTMLElement>` approach because that best approximates the set of attributes that we would be eager add to the list.

src/containers/documentation/ProviderIntegrationGuide.jsx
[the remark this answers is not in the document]
dvogel: I added a `MarkdownPage` component that ties markdown content to the base DOM structure and styling. That pushes us gently in the right direction without getting too far afield from the ticket.

dvogel:
@kalilsn @mjlumetta this is ready to review again. The two remaining visual regression failures are intended. If y'all approve of those I will update the snapshots to satisfy the tests.

src/containers/documentation/ProviderIntegrationGuide.jsx
kalilsn: This seems like it could bite someone in the future. Right now the compiler will happily let you define an `id` attribute on the generated component, but none will appear in the markup. I'd prefer using the more restrictive type and including a comment explaining how to use other standard attributes. If we don't do that, can we at least add a comment to the type definition explaining that only `style` and `className` will actually show up as attributes on the wrapper element?

package.json
```diff
@@ -50,7 +50,7 @@
     "jest-puppeteer": "^4.1.1",
     "js-yaml-loader": "^1.2.2",
     "lodash.debounce": "^4.0.8",
-    "markdown-component-loader": "^1.1.0",
+    "markdown-component-loader": "github:adhocteam/markdown-component-loader#dv-3146-from110",
```
jasalisbury: Will we want to update this to point to master once `dv-3146-from110` is merged?

dvogel: I'm not planning on changing that because I'm not sure about the future of that library. The pull request I submitted upstream hasn't received any feedback and has not been merged. I think there's a more drastic set of changes we'll need to make (like adopting the library, renaming it, etc) if the changes can't be pushed upstream.

PR #204, opened 2019-10-02
Fix header spacing

kalilsn:


src/components/Search.scss
```diff
@@ -10,7 +10,7 @@
   padding-bottom: 3.0rem;
 
   @include media($large-screen) {
-    background-color: $color-primary-darkest;
+    background: none;
```
kalilsn: This is a fix for an issue that wasn't introduced by formation. I just noticed it because I was looking closely at the search form.

PR #205, opened 2019-10-03
Override 16px base font size from formation

kalilsn:

### DIFF
--- a/src/base.scss
+++ b/src/base.scss
@@ -18,6 +18,7 @@ body {
   min-height: 100vh;
   flex-direction: column;
   color: #222;
+  font-size: 1.7rem;
 }
 
 .usa-grid {
--- a/src/base.scss
+++ b/src/base.scss
@@ -18,7 +18,6 @@ body {
   min-height: 100vh;
   flex-direction: column;
   color: #222;
-  font-size: 1.7rem;
 }
 
 .usa-grid {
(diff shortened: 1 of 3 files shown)


mjlumetta:
@kalilsn is there any particular rationale for doing this?

kalilsn:
>@kalilsn is there any particular rationale for making this change?

There's a weak one: the current site uses 17px because that's what's in USWDS. I don't feel very strongly about it either way, so I'm happy to drop this if others feel that's what's best.

mjlumetta:
@kalilsn i would prefer not to in general, because it's a 1 pixel difference and i would prefer to keep following the principle that we don't override the design system defaults unless we have a good UI/UX reason to do so. i can see one short-term benefit to this, though, which is that it would substantially reduce noise in the visual regression tests when we're trying to figure out what's okay and what we still need to fix.

kalilsn:
> @kalilsn i would prefer not to in general, because it's a 1 pixel difference and i would prefer to keep following the principle that we don't override the design system defaults unless we have a good UI/UX reason to do so. i can see one short-term benefit to this, though, which is that it would substantially reduce noise in the visual regression tests when we're trying to figure out what's okay and what we still need to fix.

Cool, I agree 100%. Given that, do you think I should make the api status link 16px as well?

mjlumetta:
@kalilsn yeah, definitely. i think that'd be an appropriate thing to do here

PR #206, opened 2019-10-03
Fix NavBar mess

mjlumetta:


config/webpack.config.prod.js
```diff
@@ -458,7 +458,8 @@ module.exports = (envName) => {
     },
     performance: {
       hints: 'error',
-      maxAssetSize: 300000,
```
jasalisbury: Why did these values have to change?

src/App.tsx
```diff
@@ -15,10 +15,10 @@ import { history } from './store';
...
 import './App.scss';
 
-let currentPath = history.location.pathname;
-history.listen(location => {
-  currentPath = location.pathname;
-});
+// let currentPath = history.location.pathname;
+// history.listen(location => {
+//   currentPath = location.pathname;
+// });
```
jasalisbury: any reason to comment this snippet instead of removing it?

src/components/NavBar.scss
```diff
@@ -1,217 +1,176 @@
 @import '../styles/variables';
 
-.header-content {
-  max-width: none;
+.va-api-nav {
+  background: $color-white;
+  color: #323A45;
+  border-top: 1px solid #DBDBDD;
```
jasalisbury: Should we be using a color variable here?

src/components/NavBar.test.tsx
```diff
@@ -6,17 +6,17 @@ import NavBar from './NavBar';
...
 describe('NavBar', () => {
   it('should render the navbar', () => {
-    const documentation = shallow(<NavBar hideLinks={false} />).contains('Documentation');
+    const documentation = shallow(<NavBar />).contains('Documentation');
     expect(documentation).toBeTruthy();
   });
 
   it('should use "is-hidden" when hidelinks is passed', () => {
-    const result = shallow(<NavBar hideLinks={true} />);
+    const result = shallow(<NavBar />);
```
kalilsn: I think this test should just be removed entirely since the `hidelinks` functionality was deleted.

src/components/Header.scss
```diff
@@ -0,0 +1,65 @@
+/*
+  This file is here to prep for separating NavBar into two components, NavBar and Header. This file is for 
+  Header, which would currently be the top-level component. It is actually called NavBar, but that naming is
+  confusing, given that the top levels tags and CSS classes use "header" naming terminology. Anyway, in the 
+  future, there will be a component Header to accompany these styles; currently they are used in NavBar.
```
kalilsn: 👏

src/containers/ApplyForm.tsx
```diff
@@ -279,8 +281,8 @@ class ApplyForm extends React.Component<IApplyProps> {
   }
 
   private anyOAuthApisSelected() {
-    const apiUrlFragments = this.selectedApis().flatMap((formField) => formFieldsToFragments[formField]);
-    return includesOauthAPI(apiUrlFragments);
+    const apiIdsByField = this.selectedApis().map((formField) => formFieldsToFragments[formField]);
+    return includesOauthAPI(flatten(apiIdsByField));
```
kalilsn: Is this related to the formation navbar stuff? If not can we put it in a different PR?

config/webpack.config.prod.js
kalilsn: 😢

kalilsn: ^ that's assuming this change is because the assets have increased in size

mjlumetta: yeah it isn't specific to this PR it's a result of the USWDS -> VADS change. it may be fixable later?

src/App.tsx
mjlumetta: that was a mistake haha, thanks for catching

src/components/NavBar.scss
mjlumetta: yes we should, but i don't think it belongs in this PR. i think we should catch all those in one PR

src/components/NavBar.test.tsx
mjlumetta: yeah, you're right - i hadn't really come back to these after making it compile

src/containers/ApplyForm.tsx
mjlumetta: yeah i'll open up a new PR

package.json
```diff
@@ -50,6 +50,7 @@
     "jest-puppeteer": "^4.1.1",
     "js-yaml-loader": "^1.2.2",
     "lodash.debounce": "^4.0.8",
+    "lodash.flatten": "^4.4.0",
```
kalilsn: I think this should to go too (ditto the accompanying change in package.lock)

src/App.tsx
```diff
@@ -9,18 +9,12 @@ import { getEnvFlags } from './apiDefs/env';
...
 
-import './App.scss';
-
-let currentPath = history.location.pathname;
-history.listen(location => {
-  currentPath = location.pathname;
-});
-
 import 'highlight.js/styles/github.css';
+import './App.scss';
```
kalilsn: Moving this after the highlight.js styles seems like a good choice in general, but was there a bug that this fixed?

package.json
kalilsn: Oops! Just realized the actual flatmap change is still in here too :)

src/App.tsx
mjlumetta: nope! this was more remove a blank line -> linter yells @ u to alphabetize things. but because, as you say, it's already a good idea to move it and there shouldn't be collisions, why not

package.json
mjlumetta: yup definitely will merge in `master` and resolve those conflicts

src/components/Header.scss
```diff
@@ -0,0 +1,75 @@
...
+*/
```
kalilsn: Looks like this comment can actually be removed now

PR #208, opened 2019-10-03
Use map + lodash.flatten instead of flatMap for compatibility reasons

mjlumetta:
`flatMap` was causing an error somewhere towards the end of the build script. this replaces it with a more broadly compatible version of the same operations

### DIFF
--- a/SitemapBuilderWebpackPlugin.js
+++ b/SitemapBuilderWebpackPlugin.js
@@ -28,12 +28,12 @@ const vm = require('vm');
 const fs = require('fs');
 
 class SitemapBuilderPlugin {
-
   constructor(options) {
     if (typeof options === 'undefined' || !options.routesFile) {
       console.error('To use SitemapBuilderPlugin you must provide an options object with routesFile as a property')
     }
     this.routesFile = options.routesFile;
+    this.polyfillsFile = options.polyfillsFile;
     this.verbose = options.verbose || false;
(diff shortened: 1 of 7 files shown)


dvogel:
These changes are ok but we could avoid the dependency by using this pattern: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#Alternative

Alternatively, do we just need a polyfill for this function? AFAIK it's a standardized function for esNext.

mjlumetta:
> These changes are ok but we could avoid the dependency by using this pattern: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#Alternative
>
> Alternatively, do we just need a polyfill for this function? AFAIK it's a standardized function for esNext.

yeah @dvogel any of these are fine. i just realized why this error is coming up in the build at the end of the build - it's coming from the `SitemapBuilderWebpackPlugin`, which doesn't have the polyfills file as an entry point. so the polyfill isn't entirely straightforward, but i can do it. and it might already be covered by the babel polyfill but polyfills aren't present in that context either way. so, can do the reduce/concat quickly if you'd prefer, and will take a look at the polyfills.

mjlumetta:
@dvogel never mind, the polyfill thing turned out to be pretty easy

PR #209, opened 2019-10-03
Media items update

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3075 

 * If a media source is youtube an embedded youtube player is displayed
 * If a media source is not youtube a 'play' image is displayed next to the description and link to the source
 * Date/Description moved below link for all news items

Youtube css was copied straight from `home.scss`. Styles defined in `News.scss` will be replaced with formation styles when that is merged into master. Also there may be better ways to accomplish what those styles are doing (specifically centering the description vertically with `margin-top/bottom: auto`). Also added typescript types for items in `content/news.yml`. Not sure if I like how `EmbeddedYoutubeVideo` handles a non-youtube URL. Trying to embed any URL seems wrong as it'll just show a blank youtube player. Also not sure if there is a better way to bind this to functions (in `HoverImage.tsx`.


src/containers/News.tsx
```diff
@@ -29,8 +49,64 @@ function NewsSideNav() {
...
+  return (
+    <div className="va-api-media-row va-api-media-item">
+      <div aria-hidden={true}>
+        <a href={item.url} tabIndex={-1}>
+          <HoverImage
+            imagePath={require('../assets/video-player.png')}
+            hoverImagePath={require('../assets/video-player-hover.png')}
+          />
+        </a>
+      </div>
```
jasalisbury: I was running into a lot of problems trying to get the accessibility tests to pass while hiding the link/image. AXE didn't like it when I tried to move the `aria-hidden` to the link. We will want to take a closer look at this section to make sure it is compliant.

kalilsn:
@jasalisbury I've updated the snapshots, but the tests will probably keep failing because of the youtube videos. To handle that on the homepage I just set the tests to hide that element before taking the screenshot. We'll probably want to generalize that so it hides these too.

src/components/HoverImage.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+
+interface IImageProps {
```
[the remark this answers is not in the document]
jasalisbury: No reason, I missed it. Thanks for catching!

src/containers/News.tsx
```diff
@@ -29,8 +49,64 @@ function NewsSideNav() {
   );
 }
 
+function renderNewsItem(item: INewsItem, media: boolean) {
```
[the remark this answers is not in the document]
jasalisbury: I like this! It also has the added benefit of moving some `key={...}` props out of a couple of awkward places which feels right to me.

src/components/EmbeddedYoutubeVideo.scss
```diff
@@ -0,0 +1,14 @@
...
+  padding-bottom: 56.25%; /* 16:9 */
+  width: 100%;
+  height: 0;
+
+  iframe {
+    display: block;
+    position: absolute;
+    left: 0;
+    width: 100%;
+    height: 100%;
```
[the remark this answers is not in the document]
jasalisbury: So I made this change but it feels a little goofy to me. I couldn't find VADS classes that matched up with all the homegrown styles (but I may have missed them). Since we can't replace all styles with VADS classes we still have to have the homegrown styles and the extra `classnames` sections in `EmbeddedYoutubeVideo.tsx`. I could be missing something though.

src/containers/News.tsx
```diff
@@ -29,8 +49,57 @@ function NewsSideNav() {
   );
 }
 
+function NewsItem({ item, media } : {item: INewsItem, media: boolean}) {
+  return (
+    <div>
+      {media ? mediaItem(item) : itemDescription(item)}
```
mjlumetta: oh yeah i guess i should've said this explicitly, but i think the same change you made to `NewsItem` re: making it a functional component could also be applied to `mediaItem` and `itemDescription` if you'd like

src/components/EmbeddedYoutubeVideo.scss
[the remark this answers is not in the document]
mjlumetta: i think that's def a fair point and probably something we should discuss together in terms of how strictly we should use VADS in situations like this. the main benefit that i can see of using the maximal possible use of the VADS classes is that it restricts the domain of possible styles to a set of standard named styles. so like, margin/padding styles in intervals of 4 or 8, colors with names, specific font sizes, etc. trying to stick to an API for styling enforces greater consistency across the site, but this is a case where that could be overkill. so i think we should talk about it.

src/containers/News.tsx
```diff
@@ -2,18 +2,88 @@ import classNames from 'classnames';
...
+const sections = NewsData.sections.map((section: ISection) => ({
   ...section,
   id: toHtmlId(section.title),
 }));
 
+function NewsItem({item, media} : {item: INewsItem, media: boolean}) {
+  return (
+    <div>
+      {media ? <MediaItem item={item} /> : <ItemDescription item={item} />}
+    </div>
```
kalilsn: Is this `div` doing anything? I'd like to drop it if not!

src/components/HoverImage.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+describe('HoverImage', () => {
+  it('should render base image', () => {
+    const wrapper = mount(<HoverImage imagePath={props.imagePath} hoverImagePath={props.hoverImagePath} />);
+    expect(wrapper.find('img').prop('src')).toEqual(props.imagePath);
+  });
+
+  it('should render hover image when hovered over', () => {
+    const wrapper = mount(<HoverImage imagePath={props.imagePath} hoverImagePath={props.hoverImagePath} />);
+    wrapper.simulate('mouseenter');
```
kalilsn: this is cool!

src/containers/News.tsx
kalilsn: This reads fine for me in voiceover! The extra div is annoying, but I think AXE is being overly cautious because some screen readers will ignore `aria-hidden` on any links. I think `tabindex="-1"` disables that behavior though, so I think this is good.

PR #210, opened 2019-10-04
Add Appeals API Documentation

erluti:
Adding the Appeals information to the developer portal.  Currently it has only the "Decision Reviews API" and it's set to disabled because Appeals is not ready for official release.

Ticket https://app.zenhub.com/workspaces/developervagov-5ac272fff41bfa50c0e46c48/issues/department-of-veterans-affairs/vets-contrib/2889

Feature Changes:
* `categories.ts` now exports `apiCategoryOrder` filtered so that only API Categories with enabled apis will show up in the list.
* `Home.tsx` now shows APIs based on the `apiCategoryOrder`.
  * `IApiCategories` now have a `placardText` property that is used to display the information on the Home page.


src/containers/Home.tsx
```diff
@@ -35,30 +39,15 @@ class Home extends React.Component {
...
-                </Link>
+                {apiCategoryOrder.map((apiCategoryKey: string) => {
+                  const { name, placardText } = apiDefinitions[apiCategoryKey];
+                  return (
+                    <Link className="va-api-card" to="/explore/appeals" key={apiCategoryKey}>
+                      <h3 className="va-api-name">VA {name}</h3>
+                      <p className="va-api-description">{placardText}</p>
+                    </Link>
+                  );
+                })}
```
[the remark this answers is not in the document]
mjlumetta: i think you can cut this down even further by replacing this with `ApiCard` from `src/components`? not sure if there are styling differences but hopefully it's something we can consolidate

src/content/releaseNotes/index.ts
```diff
@@ -1,7 +1,8 @@
+import AppealsReleaseNotes from './appeals.mdx';
 import BenefitsReleaseNotes from './benefits.mdx';
 import FacilitiesReleaseNotes from './facilities.mdx';
 import HealthReleaseNotes from './health.mdx';
 import VerificationReleaseNotes from './verification.mdx';
 
-export { BenefitsReleaseNotes, FacilitiesReleaseNotes, HealthReleaseNotes, VerificationReleaseNotes };
+export { AppealsReleaseNotes, BenefitsReleaseNotes, FacilitiesReleaseNotes, HealthReleaseNotes, VerificationReleaseNotes };
```
mjlumetta: small style thing but maybe expand this out to a multiline object since it's getting big?

src/containers/Home.tsx
```diff
@@ -35,30 +39,15 @@ class Home extends React.Component {
...
-                <Link className="va-api-card" to="/explore/verification">
-                  <h3 className="va-api-name">VA Veteran Verification API</h3>
-                  <p className="va-api-description">
-                    Build tools to help Veterans verify their Veteran status electronically on job sites, e-commerce sites, and third-party benefit sites.
-                        </p>
-                </Link>
+                {apiCategoryOrder.map((apiCategoryKey: string) => {
+                  const { name, placardText } = apiDefinitions[apiCategoryKey];
+                  return (
+                    <Link className="va-api-card" to="/explore/appeals" key={apiCategoryKey}>
```
mjlumetta: also need to get `to` from the category definition haha

src/apiDefs/data/appeals.ts
```diff
@@ -0,0 +1,20 @@
+import { IApiDescription } from '../schema';
+
+const swaggerHost : string = process.env.REACT_APP_VETSGOV_SWAGGER_API!;
+const appealsApis : IApiDescription[] = [
+  {
+    description: 'The Decision Reviews API allows you to interact with a Veteran’s Decision Review requests, also known as benefit appeals.',
+    docSources: [
+      {
+        // metadataUrl: ,// metadata endpoint is not yet exposed
```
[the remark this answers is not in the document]
erluti: yes, it's written but hasn't been exposed through kong yet

src/containers/Home.tsx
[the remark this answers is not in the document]
kalilsn: +1 to this! JSYK it's been renamed to `CardLink` since it's not really specific to APIs in any way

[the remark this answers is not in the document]
erluti: `ApiCard` uses the `description` text, but these Links were using some hardcoded text that I moved to `placardText` because they were phrased differently.

src/content/apiDocs/appeals/appealsQuickstart.mdx
```diff
@@ -0,0 +1,19 @@
+VA’s Appeals APIs allow a user/application to manage a Veteran's Appeal information.
+
+## Which API should I use?
+
+The **Appeals Status API** is for internal VA use only.  It allows for retrieving the status of all Appeals associated to a Veteran.  These statuses are read-only, and can include Appeals within the legacy process as well as the new Appeals Modernization Act (AMA) process.  You can learn more about the AMA [here](https://benefits.va.gov/benefits/appeals.asp).
+
+The **Decision Reviews API** is currently for internal VA use as well, though will include OAuth authentication in the future.  This API allows for submitting Decision Review requests in accordance with the AMA process.  It also allows for retrieving the details and status of Decision Review requests associated to a Veteran.
```
edmangimelli: Should we more explicitly say that the Decision Reviews API will be available externally in the future?

>The **Decision Reviews API** is currently only for internal VA use as well, but in the future will be accessible externally via both API key and OAuth authentication.

src/content/apiDocs/appeals/appealsQuickstart.mdx
```diff
@@ -0,0 +1,19 @@
...
+
+## Developement API Access
+
+To gain access to the Decision Reviews API you must request an API Key. This API Key is then provided as a header on all requests.
```
edmangimelli: tiny suggestion --just making it a little more clear that _they_ are the one's providing the key:

>Your API Key will need to be included as a header on all of your requests.

src/content/apiDocs/index.ts
```diff
@@ -12,32 +18,59 @@ import FhirR4ApiIntro from './health/fhirR4ApiIntro.mdx';
...
+
+const appealsContent: IApiCategoryContent = {
+  intro: AppealsIntro,
+  overview: AppealsOverview,
+  placardText: 'Build tools to help Veterans electronically manage, submit, and track appeals.',
+  quickstart: AppealsQuickstart,
+  releaseNotes: AppealsReleaseNotes,
+  shortDescription:
+    'Enables managing benefit decision appeals on behalf of a Veteran.',
+};
```
edmangimelli: does the term Decision Review need to be in here at all?

If it was up to me both terms would be used together in most places, with Appeals in parentheses 🤷‍♀ 

> Decision Review (Appeal)

src/content/apiDocs/appeals/appealsQuickstart.mdx
kellylein: Andrew and I talked about this when reviewing the original copy and decided that since the OAuth version will come soon after, we are ok not going into too much detail yet.

src/content/apiDocs/appeals/appealsQuickstart.mdx
kellylein: Works for me.

src/content/apiDocs/index.ts
kellylein: So far we have been going off the assumption that if someone is integrating with our API, they have the base knowledge that a Decision Review is the same thing as an Appeal.  Let's discuss this with Maria when she is back to see if/how we could test this assumption.

src/apiDefs/data/appeals.ts
```diff
@@ -0,0 +1,20 @@
...
+        openApiUrl: `${swaggerHost}/services/appeals/docs/v3/decision_reviews`,
```
dvogel: What went into the decision to use this path? It seems very strange to me that we would have a new category with a new API that is beginning life with `v3` in the path.

src/apiDefs/data/appeals.ts
```diff
@@ -0,0 +1,20 @@
...
+      },
+    ],
+    enabledByDefault: false,
+    name: 'Decision Reviews API',
+    urlFragment: 'decision_reviews',
+    vaInternalOnly: true,
```
dvogel: I was under the impression that this API was going to serve multiple audiences, both inside and outside the VA. Why is this set to `true`?

src/apiDefs/data/categories.ts
```diff
@@ -59,10 +54,8 @@ const apiDefinitions : IApiCategories = {
     content: verificationContent,
     name: 'Veteran Verification API',
     properName: 'Veteran Verification API',
-    releaseNotes: VerificationReleaseNotes,
-    shortDescription: 'Empowering Veterans to take control of their data and put it to work.',
```
dvogel: Whats up with these attributes being removed from unrelated APIs? Is this branch out of sync with it's target branch?

src/apiDefs/schema.ts
```diff
@@ -11,6 +11,9 @@ import * as moment from 'moment';
 export interface IApiCategoryContent {
   readonly intro: React.StatelessComponent;
   readonly overview: React.StatelessComponent;
+  readonly releaseNotes: React.StatelessComponent;
+  readonly shortDescription: string;
+  readonly placardText: string;
```
dvogel: Adding `releaseNotes` here is out of alignment with the direction set forth in this comment: https://github.com/department-of-veterans-affairs/developer-portal/pull/210/files#diff-5b0bab2a3a76b3eabb84b699458ce813L1-R7 Could you refactor how these release notes are associated with an API elsewhere in the code in order to avoid adding this new React component dependency here?

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -108,25 +109,30 @@ export default class SupportContactUsForm extends React.Component<ISupportContac
...
   }
-  
+
   private static get initialApiState() {
-    const apiDefs = getApiDefinitions();
-    return Object.keys(apiDefs).reduce((accumulator, api) => {
+    const categoryFlags = getCategoryFlags();
+    return Object.keys(categoryFlags).filter((category) => {
+        return categoryFlags[category];
+    }).reduce((accumulator, api) => {
```
dvogel: ISTM this type of filtering would be more at home in the `apiDefs/query` module. Doing it inside components like this is going to naturally lead us to duplicate the logic. If this is frequently used logic, we should centralize it with the rest of our data queries so as to keep these components focused on their presentation responsibilty.

src/apiDefs/data/categories.ts
erluti: @mjlumetta asked me to move these fields into content.

src/apiDefs/data/appeals.ts
erluti: Our API is being built in Caseflow, where it is version 3 of Caseflow's API.  We assumed the version number wouldn't impact the usability of the API for the consumer, so we decided to start at the same version number so it'd be easier to discuss internally as a team.  Maria is going to do some testing to be sure our assumption is correct that starting at v3 has no impact on the consumer.

edmangimelli: Interestingly, I'm willing to bet that our API consumers won't so much as blink at the fact that the API starts at v3 (for all they know they've simply come to the party late), but now I nearly want to do the version translation just lighthouse engineers stop asking the question. lol
I'm interested in what the testing shows. Even if the question _did_ pop into the head of an API consumer, it's still very debatable whether this has any effect on usability.

edmangimelli: Incidentally, a much bigger worry for me is the synonyms (or near synonyms) _decision review_ and _appeal_.

src/apiDefs/data/appeals.ts
erluti: The MVP is to support va.gov.  We planned to drop this tag once we've gotten past the MVP.  Does this flag have other implications?

src/containers/support/SupportContactUsForm.tsx
erluti: So the categoryFlags are generated with data from `apiDefs/env`, which is itself dependent on `apiDefs/query` so I was trying not to create a circular dependency.  The categoryFlags are used as Flags in most places, but I don't think flags work when trying to just make a list for the checkboxes.

If I were to move logic to query, what would it be?  Would it make sense to have helper function in SupportContactUsForm?

src/apiDefs/schema.ts
erluti: That link is working for me, I'm not sure to what you're referring.  I was just moving the value from IApiCategory to IApiCategoryContent.  These are based on .mdx files, so I'm not sure what removing the react component dependency would do.  This was part of the change @mjlumetta suggested to move things into content.

dvogel: I think that original link is for lines that default to a collapsed state in the PR view. Here's a link that goes to the block comment at the top of the file: https://github.com/department-of-veterans-affairs/developer-portal/blob/225d92b928cd65210f481f8b333f8fa36a6594b7/src/apiDefs/schema.ts#L4-L6. It's describing a line in the sand that we're using to avoid further-entangling this data with the code required to render it.

src/containers/support/SupportContactUsForm.tsx
dvogel: I see what you mean. It's definitely good to avoid a circular relationship. Personally I think that relationship might be inverted from what we want but it's pretty well established for now. I think a helper function in SupportContactUsForm is a good option to avoid the duplication for now.

src/apiDefs/data/appeals.ts
mjlumetta: it's for display purposes only, so sounds like you're good? and just change it later?

src/apiDefs/schema.ts
mjlumetta: @dvogel yeah, so, i did ask Philip to move it here. my reasoning behind that, since the release notes are already a function component defined in the schema, it would be beneficial to group it together with the other functional components. if you disagree, i don't think this is very important, but don't think it violates that stipulation in that it's already on the wrong side of that line.

src/apiDefs/env.ts
```diff
@@ -18,3 +18,12 @@ export const getEnvFlags = () => {
   }, {});
   return envFlags;
 };
+
+export const getCategoryFlags = () => {
+  const apiDefinitions = getApiDefinitions();
+  const categories = {};
+  getApiCategoryOrder().forEach((category) => {
+    categories[category] = apiDefinitions[category].apis.filter( api => isHostedApiEnabled(api.urlFragment, api.enabledByDefault) ).length > 0;
```
mjlumetta:
```suggestion
    categories[category] = apiDefinitions[category].apis.some( api => isHostedApiEnabled(api.urlFragment, api.enabledByDefault) );
```

slightly shorter syntax if you'd like to use it. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/containers/Home.tsx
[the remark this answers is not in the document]
mjlumetta: should still be able to use `CardLink` here, i think?

src/containers/documentation/ApiPage.tsx
```diff
@@ -32,14 +32,14 @@ export default class ApiPage extends React.Component<RouteComponentProps<IApiNam
     const isDeprecated = isApiDeprecated(api);
     const category = lookupApiCategory(params.apiCategoryKey)!;
     return (
-      <Flag name={`enabled.${api.urlFragment}`}>
+      <Flag name={`enabled.${api.urlFragment}`} fallbackComponent={ExplorePage}>
```
mjlumetta: 💯 🎉

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -24,6 +25,13 @@ interface ISupportContactUsFormProps {
 
 export default class SupportContactUsForm extends React.Component<ISupportContactUsFormProps, ISupportContactUsFormState> {
 
+  private static apiCategories() {
```
mjlumetta: could this method be abstracted into `env` as `getEnabledCategories` or something along those lines? might be something that other components would use as well

src/containers/Home.tsx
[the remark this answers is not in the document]
erluti: I can use it here, except the font is actually slightly different (like 1 point smaller).  Is this okay?  I think it will end up causing the test to fail.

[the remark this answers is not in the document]
mjlumetta: yeah definitely, you can feel free to do that and update the snapshot, i think. especially if the final font is 16px, i think that's what we're moving toward because it's what VA Design System uses. in general, you can feel free to update those snapshots for small changes that result from standardizing React components or CSS classes, no worries.

src/apiDefs/env.ts
```diff
@@ -18,3 +18,19 @@ export const getEnvFlags = () => {
...
+export const getCategoryFlags = () => {
+  const apiDefinitions = getApiDefinitions();
+  const categories = {};
+  getApiCategoryOrder().forEach((category) => {
+    categories[category] = apiDefinitions[category].apis.some( api => isHostedApiEnabled(api.urlFragment, api.enabledByDefault) );
+  });
+  return categories;
+};
+
+export const getApiCategories = () => {
```
mjlumetta: can you add "enabled" or "env" in some form to this name? i don't think it would be clear to me how this was different from the similar functions exported from `query` without going back to their definitions. but yeah, thanks for defining this in here, will be useful down the road

src/apiDefs/env.ts
```diff
@@ -18,3 +18,19 @@ export const getEnvFlags = () => {
...
+  const categoryFlags = getCategoryFlags();
+  return Object.keys(categoryFlags).filter((category) => {
```
mjlumetta: can inline this if you want:

```suggestion
  return Object.keys(categoryFlags).filter((category) => categoryFlags[category]);
```

another small thought that i have here is that it might be useful to use `getApiCategoryOrder()` rather than `Object.keys(categoryFlags)` here? that way this can operate as an analogous function that cuts down that array to the relevant categories, as you had originally written it.

src/containers/Documentation.scss
```diff
@@ -181,6 +181,14 @@
         border: none;
       }
     }
+
+    blockquote {
+      line-height: 1.5;
+      color: #323a45;
+      background: #e1f3f8;
+      padding: 1rem;
+      padding-right: 2rem;
```
mjlumetta: this is something that's written in the markdown, right? if so, this is good - if not want to suggest a small change based on our recent adoption of the VA design system

erluti: I added a "quote block" in the markdown, which renders as a `blockquote` html element.  So this is for that.  The colors were taken from here https://design.va.gov/components/featured-content

src/apiDefs/schema.ts
erluti: I agree with Matt on this, that it's not making things any worse.  I also am not sure how to extract it from a mdx file, so maybe I'm biased.

src/containers/Documentation.scss
erluti: It's the markdown that comes from the swagger doc description. https://dev-api.va.gov/services/appeals/docs/v3/decision_reviews
Searching that color guide `$color-gray-dark` and `$color-primary-alt-lightest` are those two colors, but those variables aren't available in this file.

mjlumetta: okay, sounds good 👌this block is the right place for these styles then. 

as for the colors, that is an excellent point. for now, could you please add them to the `src/styles/colors.scss` file? they'll then be available through the `styles/variables` import above. we're currently in the process of refactoring our CSS to rely on the VA design system, and `colors.scss` is sort of a polyfill for those variables in the meantime. does that sound okay?

src/containers/Home.tsx
```diff
@@ -1,15 +1,22 @@
 import classNames from 'classnames';
+import { Flag } from 'flag';
 import * as React from 'react';
 import { Link } from 'react-router-dom';
+import { getApiCategoryOrder, getApiDefinitions } from '../apiDefs/query';
+import CardLink from '../components/CardLink';
+
```
mjlumetta: extra line but don't bother pushing for it, only if there end up being other things to fix

src/content/apiDocs/index.ts
erluti: I'm closing this comment and we'll make a ticket in the future if this is no good.

PR #211, opened 2019-10-04
Fully adopt VADS and clean up old USWDS styles

mjlumetta:


dvogel:
It looks like the CI tests are failing due to an accessibility issue re: color contrast. The error doesn't make it very clear which colors. Have you identified what is causing those? Is this a new issue injected by VADS?

src/components/MarkdownPage.tsx
```diff
@@ -1,12 +1,15 @@
...
-        <div className="usa-grid">
-          <Component className="markdown-wrapper" />
-        </div>
-      </section>
-    );
+    <div className={classNames(
+      'vads-u-width--full',
+      'vads-u-margin--2',
+      'medium-screen:vads-u-margin--4',
+    )}>
```
dvogel: What advantages do we gain from using this `classNames` helper method over the JSX syntax? I thought this helper was for including nullable variable class names and handling dynamic class name lists but I don't see that happening here (ditto for some additional cases in this PR).

package.json
```diff
@@ -50,7 +50,7 @@
     "jest-puppeteer": "^4.1.1",
     "js-yaml-loader": "^1.2.2",
     "lodash.debounce": "^4.0.8",
-    "markdown-component-loader": "^1.1.0",
+    "markdown-component-loader": "github:adhocteam/markdown-component-loader#dv-3146-from110",
```
dvogel: I suspect this diff is due to branching off master (where this has been merged) but the PR targeting the css-to-va-design-system branch. Rebasing this branch off the css-to-va-design-system branch or merging master into css-to-va-design-system would likely make this diff go away (and the `enabledMarkdownItRules` diff) go away.

src/components/MarkdownPage.tsx
mjlumetta: well, it does do the dynamic inclusion and is useful for that, but i personally like to use it wherever multiple class names are required. imo, it improves readability because it allows you to read each class name as its own string; the function syntax is easier for me to read. also, since each string is on its own, it allows us to break up that declaration across multiple lines to avoid overly long strings, especially since the VADS BEM names tend to be on the long side. it's obviously a very small utility, and i'm not sure if others feel like this is an advantage or significant enough to bother with, but my preference is to take advantage of the fact that JSX attributes can be strings or Javascript to slightly improve readability. open to feedback on that.

package.json
mjlumetta: cool, yeah, that should be my next move

PR #215, opened 2019-10-08
Fix sentry initialization

kalilsn:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3277

We weren't alerted to any failures from 3277 because sentry wasn't initialized: `REACT_APP_SENTRY_RELEASE` is set to `""` in dev/staging/prod. This removes that requirement on startup and replaces it with a line in the jenkinsfile to remove those env vars from the review instance `.env` files.


Jenkinsfile
```diff
@@ -103,7 +103,9 @@ node('vetsgov-general-purpose') {
 
       if (prNum) {
         envNames.each{ envName ->
-          sh "echo PUBLIC_URL=/${reviewBucketPath()}/${envName} >> ./.env.${envName}"
+          def envFile = "./.env.${envName}"
+          sh "echo PUBLIC_URL=/${reviewBucketPath()}/${envName} >> ${envFile}"
+          sh "sed -i '/REACT_APP_SENTRY_/d' ${envFile}"
```
dvogel: It seems like sentry could be useful for review instances. That's where we would most easily be able to demonstrate bugs to each other. Any reason we shouldn't just add a sentry project for the review instances to share?

kalilsn: I can't see a reason not to, and that shouldn't be difficult. Would that project send notifications anywhere?

dvogel: > Would that project send notifications anywhere?

Probably not? Feels like a place we'd check if we encountered an error. In general though we don't want people distracted by notifications for errors in review instances that are deployed for a PR someone else opened.

kalilsn:
I'm going to merge this as is so we can get sentry working again now! I'll open up a new PR to use the new sentry project once https://github.com/department-of-veterans-affairs/devops/pull/5679 is merged.

PR #216, opened 2019-10-10
Make changes for new nav

mjlumetta:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2862

blue background on desktop nav, rearranged mobile nav. not too large.


jasalisbury:
Some discrepancies from the mockup: 
 * The bar between the nav and header doesn't go all the way to the edge of the page
 * The menu button on mobile isn't as wide
 * `Request an API Key` has white background with blue(ish) text
 * Order of menu items is different

I could see some of these changes being made in a different PR, specifically the updates to the `Request an API Key` button and maybe the order of menu items.

jasalisbury:
> > The bar between the nav and header doesn't go all the way to the edge of the page
> 
> ^ can you please clarify that one, though? not sure i understand

Sure! The horizontal 1px border between the header and nav does not extend the full length of the page in the mockup. In the mockup that line starts 34px from the left edge of the screen and ends 34px from the right edge of the screen.

mjlumetta:
> Sure! The horizontal 1px border between the header and nav does not extend the full length of the page in the mockup. In the mockup that line starts 34px from the left edge of the screen and ends 34px from the right edge of the screen.

ohhh i see, thank you!!! for some reason i read that as the opposite, like it should be all the way across in the mockup and wasn't in the code... whoops haha

PR #219, opened 2019-10-10
Misc header updates

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2858


src/components/Search.scss
```diff
@@ -15,6 +14,37 @@
...
+  *::-moz-placeholder {
+    /* FF 19+ */
+    color: $color-off-white;
+    opacity: 1;
+  }
+  *:-ms-input-placeholder {
+    /* IE 10+ */
+    color: $color-off-white;
+  }
+  *::-ms-input-placeholder {
```
mjlumetta: i don't think these should be * - since the `placeholder` pseudo element only applies to `input` and `textarea` elements (see https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder), we can be specific about that. VADS only uses these for `input`, so i think we can say that for now.

src/components/Search.scss
```diff
@@ -15,6 +14,37 @@
...
+    /* Microsoft Edge */
+    color: $color-off-white;
+  }
+  *::placeholder {
+    /* modern browser */
+    color: $color-off-white;
+  }
```
mjlumetta: yeah, i agree with you that this feels icky. i would suggest defining a mixin for this, either in `base.scss` or a new `mixins.scss` module if we want to make our own mixins module. something like `input-color` or `form-input-color`, depending on whether you want it to be applied at the input or form level, with the placeholder/input color as a parameter. does that sound good?

src/components/Search.scss
```diff
@@ -37,13 +67,19 @@
...
     @include media($large-screen) {
       width: 240px;
       max-width: inherit;
     }
   }
 
+  .va-api-search-icon {
+    position: relative;
+    left: 25px;
+    color: $color-white;
```
mjlumetta:
```suggestion
    color: $color-white;
```

in the design, this is some shade of gray: #C3C8CF. i see where you got it from design.va.gov, but what they're really doing is using white + opacity to get this color: `rgba(255, 255, 255, 0.75)`. we can do that or use the hex value, either way.

src/components/Header.scss
```diff
@@ -66,3 +61,8 @@
     }
   }
 }
+
+.button-inverse {
```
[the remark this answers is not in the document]
jasalisbury: Actually don't even need that style anymore

src/styles/colors.scss
```diff
@@ -7,5 +7,7 @@ $color-primary-darkest: #1b2b3e;
 $color-white: #ffffff;
 $color-base: #212121;
 $color-gray-lighter: #d6d7d9;
+$color-gray: #C3C8CF;
```
mjlumetta: this is going to conflict with VADS's variable of the same name: #5b616b. i'm not sure what to name this exactly, but probably shouldn't be this. maybe `$color-alt-gray`?

src/components/Search.scss
```diff
@@ -15,6 +17,11 @@
     padding-bottom: 0;
   }
 
+  input {
+    background-color: #12223A;
+    color: $color-off-white;
```
mjlumetta: i think there's an open question as to whether this line should be part of the mixin. it's conceivable to me that you'd want the input text color and placeholder color to be the same for the same input, in which case that consistency could be part of the value the mixin provides. up to you. but, if you do do that, would suggest removing "placeholder" from the mixin name.

src/styles/mixins.scss
```diff
@@ -0,0 +1,32 @@
+@mixin va-api-input-placeholder-color($color) {
```
mjlumetta: beautiful 💯 🎉 

only thing about this is the suggestion i made in `Search.scss`, which could go either way

kalilsn:
I hate to say this at such a late stage, but removing the submit button from the search form is an accessibility problem. https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-unpredictable-change.html makes it clear that we either need to provide a submit button, or provide clear directions about how the form works (which would mean visible text saying "Type in your search terms and press the enter key to navigate to the search results page"). I suspect that latter option isn't really appealing to anyone, so we should probably discuss how to handle this.

hubgitian:
I reached out to @1Copenut on this and he said that ideally we should have a button (but that the form should still listen for  `enter`  keypress.) I've developed a solution which I believe honors the spirit of the original design but gives us a solution that is in line with best practice of having a button and a rollover state that uses the existing highlight color. 


searchbox

1Copenut:
@hubgitian Looking at the comp above, I think the `button:focus` state will need to be adjusted slightly. The input won't retain the yellow halo, so maybe we could use the red hit space you identified for a yellow border "halo" or solid background.

jasalisbury:
@1Copenut @hubgitian checkout the review instance above to see the update. I misread your comment @1Copenut and the button has the yellow halo on focus and on hover. I just pushed a version that only has the yellow halo on focus, there should be a review instance below this comment soon.

hubgitian:
@jasalisbury This looks great on OSX/Safari! There's one thing I've noticed on OSX/Chrome is that searchbox doesn't extend all the way past the magnifying glass. (The top image is Chrome, the bottom is OSX/Safari)

Screen Shot 2019-10-15 at 9 13 04 AM

jasalisbury:
> These changes have been deployed to an S3 bucket. A build for each environment is available:
> 
> https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/36b627b/dev/index.html
> https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/36b627b/staging/index.html
> https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/36b627b/production/index.html
> 
> Due to S3 website hosting limitations in govcloud you need to first navigate to index.html explicitly.

@hubgitian Can you give one of the review instances above a try?

hubgitian:
@jasalisbury whatever you did, it worked! Looks great in OSX chrome now too! 🚀

src/components/Header.tsx
```diff
@@ -24,6 +24,7 @@ export default class Header extends React.Component<{}, INavBarState> {
 
   public render() {
     const navBarCloseHandler = this.toggleMenuVisible.bind(this);
+    const buttonClassnames = classNames('vads-u-background-color--white', 'vads-u-color--primary-darkest', 'usa-button', 'vads-u-margin-right--2');
```
mjlumetta: i would suggest putting `usa-button` first in order to let the other classes override it (they should anyway but just in case)

src/components/NavBar.tsx
```diff
@@ -95,7 +95,7 @@ export default class NavBar extends React.Component<INavBarProps, INavBarState>
           </ul>
           <MediaQuery query={UNDER_LARGE_SCREEN_QUERY}>
             <div className="va-api-nav-secondary">
-              <Link to="/apply" className="usa-button">Request an API Key</Link>
+              <Link to="/apply" className="vads-u-width--full usa-button">Request an API Key</Link>
```
mjlumetta: prefer to use `classNames` with multiple classes for readability

src/components/Search.scss
```diff
@@ -1,11 +1,46 @@
...
-.vadp-search-form {
+.va-api-search-form--inverse-color {
+  @include va-api-input-text-color($color-off-white);
+
+  input {
+    background-color: $color-dark-blue;
+  }
+}
+
+.va-api-search-form {
```
mjlumetta: use VADS for these styles?

src/components/Search.scss
```diff
@@ -1,11 +1,46 @@
...
+  display: flex;
+  flex-flow: row nowrap;
+}
+
+.va-api-search-form--transparent-submit {
+  button[type=submit] {
+    appearance: none;
+    background-color: Transparent;
```
mjlumetta:
```suggestion
    background-color: transparent;
```

src/components/Search.scss
```diff
@@ -1,11 +1,46 @@
...
+    background-repeat: no-repeat;
+    border: none;
+    color: $color-alt-gray;
+    cursor: pointer;
+    overflow: hidden;
+    margin: 0px;
```
mjlumetta:
```suggestion
    margin: 0;
```

src/components/Search.scss
```diff
@@ -1,11 +1,46 @@
...
+    padding: 0px;
+    position: relative;
+    &:hover {
+      background-color: none !important;
+      color: $color-gold;
+    }
+  }
+}
+
+.va-api-search-wrapper {
```
mjlumetta: can these be moved to VADS?

src/components/Search.tsx
```diff
@@ -6,28 +6,36 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import './Search.scss';
 
-export default class Search extends React.Component<{className?: string}> {
+export default class Search extends React.Component<{inMenu?: boolean, className?: string}> {
   public render() {
+    const { inMenu } = this.props;
+    const wrapperClassName = classNames("va-api-search-wrapper", this.props.className);
+    const formClassName = classNames("va-api-search-form", {"va-api-search-form--transparent-submit": !inMenu});
+    const wrapperClassNameModified = classNames(wrapperClassName, {"va-api-search-form--inverse-color": !inMenu});
```
mjlumetta: you can compute this in one call to `classNames` - it's pretty handy for that

src/components/Search.tsx
```diff
@@ -6,28 +6,36 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
...
+    const iconClassName = classNames({"va-api-search-icon": !inMenu});
+    const buttonClassName = classNames({"va-api-search-submit": inMenu});
+    const placeholderText = inMenu ? "" : "Search...";
```
mjlumetta: you could maybe inline some of these computed attributes if you feel like it. it's pretty arbitrary, but i usually prefer to inline when possible so that the JSX markup is more self-contained/don't have to reference variable assignment as much

src/components/Search.tsx
```diff
@@ -6,28 +6,36 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
...
+            acceptCharset="UTF-8"
+            method="get"
+            className={formClassName}>
+          <input name="utf8" type="hidden" value="&#x2713;" />
+          <input type="hidden" name="affiliate" id="affiliate" value="developer.va.gov" />
+          <input type="text"
+            name="query"
+            id="query"
+            autoComplete="off"
+            className='usagov-search-autocomplete'
```
mjlumetta:
```suggestion
            className="usagov-search-autocomplete"
```

does the linter not yell at you for this? i didn't know you could do that haha

src/styles/mixins.scss
```diff
@@ -0,0 +1,41 @@
+@mixin va-api-input-text-color($color) {
+  input {
+    color: $color;
+  }
+
+  ::-webkit-input-placeholder {
+    color: $color;
+  }
```
kalilsn: Do we actually need to manually include all of these different prefixed properties? Shouldn't our autoprefixer take care of this?

src/components/Search.scss
```diff
@@ -1,11 +1,46 @@
...
+.va-api-search-form {
+  display: flex;
+  flex-flow: row nowrap;
+}
+
+.va-api-search-form--transparent-submit {
+  button[type=submit] {
+    appearance: none;
+    background-color: Transparent;
+    background-repeat: no-repeat;
```
kalilsn: `background: none` can replace these two

src/components/Search.scss
```diff
@@ -1,11 +1,46 @@
...
+    cursor: pointer;
+    overflow: hidden;
+    margin: 0px;
+    outline: none;
+    margin-left: -44px;
+    width: 44px;
+    padding: 0px;
+    position: relative;
+    &:hover {
+      background-color: none !important;
```
kalilsn: `none` isn't a valid value for `background-color`, so this isn't doing anything

src/components/Header.tsx
kalilsn: I don't think the order of classes in the attribute has any effect on precedence.

mjlumetta: that's true unless `usa-button` happens to have `!important` on any of these properties. i sort of prefer it that way regardless because it reads more naturally like the utility classes are being applied on top of `usa-button`, but that's entirely trivial 🤷‍♂

src/styles/mixins.scss
jasalisbury: It does! I wasn't aware we used autoprefixer but the first example on the [npm page](https://www.npmjs.com/package/autoprefixer) is for placeholder color!

dvogel:
When I try to access the review instance I get a 403 (likely a 404 behind the scenes) for the JS bundle.

Screenshot from 2019-10-17 17-22-04

jasalisbury:
> When I try to access the review instance I get a 403 (likely a 404 behind the scenes) for the JS bundle.
> 
> Screenshot from 2019-10-17 17-22-04

Same for `dev` and `staging`. `prod` works for some reason. I'll start digging in.

src/components/Header.tsx
dvogel: Order within the attribute shouldn't matter. AFAIK it was never meant to matter but some browsers used to incorrectly do so. This is the reference I always use: https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_and_inheritance#Understanding_the_cascade

src/base.scss
```diff
@@ -3,6 +3,7 @@
 
 // Asset reference - must import before VADS
 @import './styles/variables';
+@import './styles/mixins';
```
kalilsn: Does this need to be before the vads import? If not, I think we should default to putting stuff after.

src/components/NavBar.tsx
```diff
@@ -95,7 +95,7 @@ export default class NavBar extends React.Component<INavBarProps, INavBarState>
           </ul>
           <MediaQuery query={UNDER_LARGE_SCREEN_QUERY}>
             <div className="va-api-nav-secondary">
-              <Link to="/apply" className="usa-button">Request an API Key</Link>
+              <Link to="/apply" className={classNames("vads-u-width--full usa-button")}>Request an API Key</Link>
```
kalilsn:
```suggestion
              <Link to="/apply" className={classNames('vads-u-width--full', 'usa-button')}>Request an API Key</Link>
```

kalilsn: or
```suggestion
              <Link to="/apply" className="vads-u-width--full usa-button">Request an API Key</Link>
```

kalilsn: I have no real preference between these options

PR #220, opened 2019-10-10
Update site header title

mjlumetta:
pretty small update, just changing the title and standardizing the font size. second AC of this ticket: https://github.com/department-of-veterans-affairs/vets-contrib/issues/2858


mjlumetta:
> Looks good but do you know why snapshots for the explorer and support pages needed to be updated?

@jasalisbury not entirely sure. i went ahead with it because they were extremely marginal. might have to do with changing the title text size? or version control, taking screenshots with slightly different sources and then comparing them. not sure.

jasalisbury:
> @jasalisbury not entirely sure. i went ahead with it because they were extremely marginal. might have to do with changing the title text size? or version control, taking screenshots with slightly different sources and then comparing them. not sure.

Oh yeah I could see the title being a different size causing the screenshots to be different.

PR #223, opened 2019-10-10
Rewrite site styles to rely primarily on VADS

mjlumetta:

@jasalisbury @kalilsn @dvogel this is the staging branch for 3011. this work can be parallelized like the previous ticket, so you can open PRs against this branch if you plan to work on the ticket. we should get a thread going on slack or comment in this PR to track who's working on what. my thinking is that we should split the work by component or directory rather than by VADS utility (like, say, changing all padding or font sizes across the site), but i'm open to suggestions.


src/components/CardLink.scss
```diff
@@ -1,36 +1,14 @@
...
 
   &:hover {
     border-top: 5px solid $color-gold;
-    text-decoration: none;
   }
 
-  &:visited {
-    color: #323A45;
+  @include media($medium-screen) {
+    width: 250px !important;
```
kalilsn: Why do we need this `!important`?

src/styles/variables.scss
```diff
@@ -6,7 +6,7 @@ $site-max-width: 1200px;
 
 // Responsive media
 $medium-large-screen: 768px;
-$medium-screen: 640px;
+$medium-screen: 600px;
```
kalilsn: What's the reason for this change?

src/components/CardLink.tsx
```diff
@@ -18,11 +19,29 @@ export interface ICardLinkProps {
...
-        <div className="va-api-description">{this.props.children}</div>
+        <div className={classNames('va-api-description', 'vads-u-color--base')}>
+          {this.props.children}
+        </div>
       </NavHashLink>
     );
   }
 }
+
+export const getCardContainerClasses = () => {
```
kalilsn: Can we add a comment explaining what this function is for? It seems to me it's for a wrapper around multiple `CardLink` components. If that's the case, is there a reason not to just make this a full component rather than just a function for classes?

kalilsn:
After reviewing this, I don't think there's a reason for us to merge other PRs into here before merging it. It looks pretty good! I only see a few things that would need changing before we could merge this in, and I don't think those changes are related to the work that's being done in https://github.com/department-of-veterans-affairs/developer-portal/pull/226 and https://github.com/department-of-veterans-affairs/developer-portal/pull/227

mjlumetta:
@kalilsn thanks for the review! i would actually say that it would be useful to merge #226 into this one, because it addresses 3/4 (?) of your comments. yesterday when i thought a staging branch would be useful i fixed a lot of those things in that branch. so if you want to take a look at that, we can merge it in here, or merge this and then that if they're both ready.

src/components/CardLink.tsx
mjlumetta: in #226 i rewrote this as a utility called `defaultFlexContainer` in a new TS module in `src/styles/`. so if we merge that one PR with this one, it'll be fixed. or can push that change to here if necessary

src/components/CardLink.scss
mjlumetta: `!important`: i wrote the mobile style with the `vads-u-width--full` (i think that one, but specific class not important), and all the VADS utilities use `!important`. it would be unnecessary if we added the mobile style back into this file. i don't particularly care which, with a slight preference for this way because i think it sets us up to make further changes down the road which could allow us to remove a lot of component stylesheets entirely.

similar to other things, the width is fixed in #226 as well. i copied this style from `Home.scss` - if you visit the site in production, you'll notice that the width in the Home page is 250px. i changed the mobile vs desktop styles because the design for mobile had width = 100%.

src/styles/variables.scss
mjlumetta: i was trying, unsuccessfully in this case, to align these variables with the VADS breakpoints. this is also fixed in #226. my goal here was to create temporary consistency even if tablet-ish/medium screens aren't perfect, with the goal of cleaning up responsive media in a later PR. how's that sound to you, any feedback?

kalilsn:
I see now that there are changes in 226 that are related to this. But there's also a bunch of other stuff in 226 that isn't. Can we realign these PRs so that they can be merged separately? I don't think it matters whether that is pulling out the breaking changes from here and moving them to 226 or moving the relevant fixes from 226 over to here. I'd just like to see this PR be the minimum amount of code required for us to merge all of the other parts of vets-contrib#3011 directly into master rather than into a temporary branch.

src/styles/variables.scss
kalilsn: I think that sounds like it belongs in its own PR, unless it's fixing something specific that's introduced here!

src/components/CardLink.scss
kalilsn: >all the VADS utilities use `!important`

That's really unfortunate!

src/styles/variables.scss
mjlumetta: it's not specific to anything in this PR but it is going to be necessary across all PRs for the VADS change so that we can have a consistent breakpoint in the middle. i don't think this change is harmful; i can do a check through that all components using this variable look okay. 

the other way we could do this would be to front-load the work of tracing and updating how we get variables from formation. my inclination is not to do that now and instead to do _that_ in its own PR later. reason to do that later, in my view, is that we don't want to do the variables resolution until we've ported over as many styles as we can so that we know what variables we want to use and need. and then we could do another follow-up PR once we have a consistent responsive system if we'd like to fine-tune some things.

TLDR our current responsive media isn't very good in the first place and this won't break anything. it will ensure that `medium-screen:` breakpoints and `@include media($medium-screen)` styles are consistent, so even though it seems counterintuitive, doing this hack in the short term will bring us closer to a more consistent design system. open to feedback on this.

src/components/CardLink.scss
mjlumetta: >>all the VADS utilities use `!important`
>
>That's really unfortunate!

i don't think so, necessarily. i'm guessing the reason for that there happen to be conflicting styles, it won't take a very strong selector to override the VADS class - for example, any nested classes will do it. another point is that, if the VA design system is supposed to be the source of truth, any utility classes you apply should be absolute. this is inconvenient where the VADS falls short of covering all use cases, as it does here. i would agree with you that use of `!important` is a bad practice in general, but i think it makes conceivable sense in this case for those reasons, essentially because it is an opinionated design system. i have some ideas to mitigate this in the long run but those should wait until later PRs.

src/types/constants.ts
```diff
@@ -51,8 +51,8 @@ export type TOGGLE_ACCEPT_TOS = typeof TOGGLE_ACCEPT_TOS;
 
 // Media query values - see src/base.scss and node_modules/uswds/src/stylesheets/core/_variables.scss
 export const MEDIUM_SCREEN = '640px';
-export const OVER_MEDIUM_SCREEN_QUERY = 'screen and (min-width: 640px)';
-export const UNDER_THAN_MEDIUM_SCREEN_QUERY = 'screen and (max-width: 639px)';
+export const OVER_MEDIUM_SCREEN_QUERY = 'screen and (min-width: 600px)';
+export const UNDER_THAN_MEDIUM_SCREEN_QUERY = 'screen and (max-width: 599px)';
```
dvogel: This is unrelated but I'm just noticing now that we probably meant to change `UNDER_THAN_` to `UNDER_`. Seems like a good opportunity to leave this a bit cleaner than we found it.

mjlumetta: yeah, that's probably true. i don't think i'm going to clean this up now just because i think we should get rid of these constants, period. i added `MediaQuery` from `react-responsive` a while ago because we didn't have a framework for responsive styles beyond rewriting media queries with `display: none` and `display: block` wherever necessary. but now we have the VADS `display` utilities and the corresponding responsive breakpoints for that instead, so we can eliminate `MediaQuery` as a dependency and cut these constants out entirely.

PR #225, opened 2019-10-11
Add Facilities API release note for mobile and active_status

jeff2d2:
This adds a new release note to the dev portal for a change to the Facilities API response to include flags for `mobile` and `active_status`.

Addresses https://github.com/department-of-veterans-affairs/vets-contrib/issues/3341


jeff2d2:
Preview: 
image

nfasulo:
Not a required addition, but I wonder if we should add something about how you shouldn't trust the address listed for a mobile facility? The documentation says: "To get the exact current location, end consumers should contact the mobile facility directly."

src/content/apiDocs/facilities/facilitiesReleaseNotes.mdx
```diff
@@ -1,3 +1,10 @@
+## September 30, 2019
+You can now see if Health facilities are mobile or temporarily closed (for power outages, weather, etc.)
+- Added `mobile` and `active_status` flags to the API response for Health facilities (other facility types will return `null` for these fields). Options for `mobile` are `true` or `false`; options for `active_status` are `A` for Active or `T` for Temporarily closed. [View code change(s)](https://github.com/department-of-veterans-affairs/vets-api/pull/3339/commits)
+  - If a facility is mobile, the address is subject to frequently change. To get the exact current location, consumers should contact the mobile facility directly.
```
nfasulo:
```suggestion
  - If a facility is mobile, the address is subject to frequent changes. To get the exact current location, consumers should contact the mobile facility directly.
```

davem-va:
Thanks @jeff2d2 

Temporarily closed may not be correct.  VHA refers to this as temporarily deactivated.  This may make more sense.  My understanding is that the facility is not accepting appointments while in this "T" state but there may be medical staff reporting for duty so it's not really "closed."

Also, the #'s listed for mobile facilities are sometimes those for the mobile facility itself and other times for the parent facility.  To avoid confusion, should we say to call the number listed rather than "call the mobile facility directly?" It may not be possible to speak with the mobile facility directly in all cases.

jeff2d2:
@davem-va Thank you. The language has been updated based on your feedback. I removed the part about *why* a facility might show as temporarily deactivated as it can be many reasons, and I didn't want to mention "weather or power outages" as that implies someone might not be there when that may not be the case.

PR #226, opened 2019-10-11
Convert Banner to VADS

mjlumetta:
part of 3011: https://github.com/department-of-veterans-affairs/vets-contrib/issues/3011

in the TSX file, there are still some classes related to VA crisis line that are defined in that component's stylesheet and also by VADS. i'm going to leave resolution of that for a later pull request, when the VCL component/stylesheet get converted.


src/styles/variables.scss
```diff
@@ -6,7 +6,7 @@ $site-max-width: 1200px;
 
 // Responsive media
 $medium-large-screen: 768px;
-$medium-screen: 600px;
+$medium-screen: 768px;
```
jasalisbury: Same question as I had on https://github.com/department-of-veterans-affairs/developer-portal/pull/227. This does change the point at which the layout changes from 'small' to 'medium'

PR #227, opened 2019-10-11
Convert Footer to VADS styles

mjlumetta:


src/components/Footer.tsx
```diff
@@ -5,9 +5,17 @@ import './Footer.scss';
...
 import logo from '../assets/lighthouseVaLogo.png';
 
+const footerLinkStyles = classNames('vads-u-font-size--sm', 'vads-u-color--white');
+const listItemStyles = classNames('medium-screen:vads-u-padding-left--4', 'medium-screen:vads-u-margin-bottom--0');
+
 export default function Footer() {
   return (
-    <footer className="va-api-footer" role="contentinfo">
+    <footer role="contentinfo" className={classNames(
+      'va-api-footer',
```
jasalisbury: Does this still need the `va-api-footer` class?

src/styles/variables.scss
```diff
@@ -6,7 +6,7 @@ $site-max-width: 1200px;
 
 // Responsive media
 $medium-large-screen: 768px;
-$medium-screen: 600px;
+$medium-screen: 768px;
```
jasalisbury: This was changed in a previous PR why does it need to be changed again? Also `768px` matches the value of `$medium-large-screen`, is that intentional?

src/components/Footer.tsx
mjlumetta: nope, removing

src/styles/variables.scss
mjlumetta: yes, so, the previous value was wrong for what i was trying to do. i wanted to get a consistent breakpoint between small and large screens that split in the middle, and this is what the `medium-screen:` prefix does. i want to do PRs later to (a) clean up variable resolution and (b) improve our mobile styles as necessary

PR #229, opened 2019-10-11
Use VADS styles in NavBar

mjlumetta:


kalilsn:
I like what you've done with the breakpoints here a lot – should be much simpler to work with in the future! There's one small bug I'm seeing:

Screen Shot 2019-10-18 at 3 17 26 PM

Both the logo and the request an api key button wrap to a new line between 768 and 786px width. Can we just stop those from ever wrapping?

jasalisbury:
> Both the logo and the request an api key button wrap to a new line between 768 and 786px width. Can we just stop those from ever wrapping?

768 vs 786 makes me suspicious that the last two digits got transposed somewhere

mjlumetta:
> Both the logo and the request an api key button wrap to a new line between 768 and 786px width. Can we just stop those from ever wrapping?

@kalilsn i think that it might be slightly better to wrap the right-hand content (wrap the search under the request button) and not wrap the title. regardless, after merging with master, i'm not seeing this error? but i can still apply that change in case it slips back in again, to say that the right div should `wrap` and the title should not wrap.

PR #232, opened 2019-10-14
Hero update

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2863

Still a few style issues to sort out:

 1. Mobile homepage hero should flip the VA logo and text (logo should be above the text). There are still some other little style issues in the hero on smaller screens that will get sorted out once the logo is above the text.
 2. Might need to add some margins to the homepage content between the images and text (on medium sized screens)

For reference the [mockup](https://www.figma.com/file/O0J3wAWWz8S5H06I2CbdYP/Developer-Portal-Homepage-Final?node-id=0%3A1)


src/containers/Home.tsx
```diff
@@ -1,63 +1,94 @@
...
+import { OVER_LARGE_SCREEN_QUERY } from '../types/constants';
 import './Home.scss';
 
 class Home extends React.Component {
   public render() {
-    const apiDefinitions = getApiDefinitions();
-    const apiCategoryOrder = getApiCategoryOrder();
+    const leftColumnClasses = classNames('medium-screen:vads-l-col--4', 'small-screen:vads-l-col--6', 'va-api-u-margin-y--auto');
+    const rightColumnClasses = classNames('medium-screen:vads-l-col--8', 'small-screen:vads-l-col--6', 'va-api-u-margin-y--auto', 'va-api-margin-y--auto');
+    const flexContainer = classNames('vads-l-grid-container', 'vads-u-margin-top--6', 'vads-u-margin-x--auto');
```
mjlumetta:
```suggestion
    const flexContainer = classNames('vads-l-grid-container', 'vads-u-margin-top--6');
```

i know that this is how we currently have our home page and the design shows it, but should we be overriding the default grid styles? is the home page special?

src/containers/Home.tsx
```diff
@@ -1,63 +1,94 @@
...
+import Hero from '../components/Hero';
+import { OVER_LARGE_SCREEN_QUERY } from '../types/constants';
 import './Home.scss';
 
 class Home extends React.Component {
   public render() {
-    const apiDefinitions = getApiDefinitions();
-    const apiCategoryOrder = getApiCategoryOrder();
+    const leftColumnClasses = classNames('medium-screen:vads-l-col--4', 'small-screen:vads-l-col--6', 'va-api-u-margin-y--auto');
+    const rightColumnClasses = classNames('medium-screen:vads-l-col--8', 'small-screen:vads-l-col--6', 'va-api-u-margin-y--auto', 'va-api-margin-y--auto');
```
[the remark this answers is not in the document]
mjlumetta: also re: `va-api-u-margin-y--auto`, i wonder if this might be accomplished with `vads-u-align-items--center` instead. the grid system is defined as a flexbox system, and this takes advantage of flex capabilities and slightly more explicitly conveys the meaning of what it's doing. pretty much a wash though, not v important

src/styles/colors.scss
```diff
@@ -10,4 +10,5 @@ $color-gray-lighter: #d6d7d9;
 $color-gold: #fdb81e;
 $color-red: #cd2026;
 $color-primary-alt-lightest: #e1f3f8;
-$color-gray-dark: #323a45;
\ No newline at end of file
+$color-gray-dark: #323a45;
+$color-blue-dark: #12223A;
```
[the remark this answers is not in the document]
jasalisbury: Yea `color-cool-blue` looks a little too far off. We probably need to look through the colors we use and see if VADS has any colors close that we can use instead. As for the namespace it makes sense to me to have `va-api` in the name the same way we prefix css classes.

src/containers/Home.tsx
jasalisbury: It makes sense to me to override the default grid for the homepage. The homepage in this PR doesn't look as good left aligned vs centered but it isn't _that_ bad left aligned.

mjlumetta: ok then, this seems like the right way to do it! no objections, thanks for clarifying

src/components/Hero.tsx
```diff
@@ -0,0 +1,37 @@
...
+    "small-desktop-screen:vads-u-flex-direction--row",
+    "vads-u-flex-direction--column-reverse",
+    "vads-l-grid-container",
+    "vads-u-margin-x--auto",
+  );
+
+  return (
+    <section role="region" aria-label="Page Hero" className={heroClasses}>
+      <div className={containerClasses}>
+        <div className={classNames("", "va-api-margin-y--auto")}>
```
kalilsn: Why the empty string?

src/containers/Home.tsx
```diff
@@ -1,130 +1,130 @@
...
-          'vads-u-padding-left--4',
-          'site-disclaimer',
-        )}>
-          <strong>This is a beta site.</strong>
-          &nbsp;We are always looking to make improvements.&nbsp;
-            <Link to="/support">
-            <strong>Send us your feedback</strong>
-          </Link>
-        </div>
+const leftColumnClasses = classNames('medium-screen:vads-l-col--4', 'small-screen:vads-l-col--6', 'va-api-margin-y--auto');
```
kalilsn: Is the margin-y--auto class is necessary here?

src/containers/Home.tsx
```diff
@@ -1,130 +1,130 @@
...
-            <div className="va-api-hero-callout">
-              <h1>Put VA Data to Work</h1>
-              <p>Empowering our partners to build innovative, Veteran-centered solutions.</p>
+function ApiPlatform() {
+  return (
+    <section role="region" aria-label="An API platform">
+      <div className={flexContainer}>
+        <div className="vads-l-row">
+          <div className={leftColumnClasses}>
+            <img src={padlockImg} alt="padlock-image" />
```
kalilsn: This image (and the others added here) should have `alt=""` since they're purely decorative. Also in general alt text should be human readable and doesn't need to say "image" or "graphic".  Just "padlock" would be fine too.

src/components/Hero.tsx
jasalisbury: missed it, thanks for catching!

src/components/Hero.tsx
```diff
@@ -0,0 +1,37 @@
...
+          <div className="vads-u-max-width--100">
+            <h1 className="vads-u-color--white">A Veteran-centered API platform for securely accessing VA data.</h1>
+            <Link id="hero-request-key" to="/apply" className="usa-button vads-u-width--full medium-screen:vads-u-width--auto">Request an API Key</Link>
+          </div>
+        </div>
+        <div className={classNames("", "vads-u-margin-x--auto", "va-api-margin-y--auto")}>
```
kalilsn: Empty string here too :)

src/containers/Home.tsx
jasalisbury: It keeps the two images in the main content centered in the row on small screens.

kalilsn: Ah I see now. I was only looking at it on the largest size and on the smallest size where the images are stacked vertically with the content

src/components/Hero.tsx
```diff
@@ -0,0 +1,37 @@
...
+import logo from '../assets/hero-logo.png';
+
+export default function Hero() {
+  const heroClasses = classNames(
+    'vads-u-background-color--primary-darkest',
+    'vads-u-padding-y--5',
+    'vads-u-padding-x--0',
+  );
+
+  const containerClasses = classNames(
```
mjlumetta: personally have a slight preference for inlining classes unless they're used multiple times. but this is trivial

src/containers/Home.tsx
```diff
@@ -1,130 +1,130 @@
...
+const leftColumnClasses = classNames('medium-screen:vads-l-col--4', 'small-screen:vads-l-col--6', 'va-api-margin-y--auto');
+const rightColumnClasses = classNames('medium-screen:vads-l-col--8', 'small-screen:vads-l-col--6', 'va-api-margin-y--auto');
+const flexContainer = classNames('vads-l-grid-container', 'vads-u-margin-top--6', 'vads-u-margin-x--auto');
 
-        <section role="region" aria-label="Page Hero" className="va-api-hero">
-          <div className="vads-l-grid-container">
-            <div className="va-api-hero-callout">
-              <h1>Put VA Data to Work</h1>
-              <p>Empowering our partners to build innovative, Veteran-centered solutions.</p>
+function ApiPlatform() {
```
mjlumetta: i think i was a little unclear about this earlier, my bad. i was thinking about extracting this and `ReliableApi` into one generic component that could abstract out the specific parts as props. so i was thinking the following props could cover it (in order): `ariaLabel`, `image`, `title`, `children` (the main text)

src/containers/Beta.scss
```diff
@@ -1,4 +1,3 @@
-@import '../styles/hero';
```
mjlumetta: this does break the hero on the Beta page. i'm not sure whether we're even supposed to still be maintaining this page? but barring definitive word that we shouldn't, i would default to maintaining those styles. not sure if you were planning to do that in VADS on this PR.

kalilsn: I've asked about that page a few times before and was unable to find out whether there's a reason for us to keep it. I hope that means we can just delete it?

mjlumetta: that would be amazing, i've never heard anybody refer to this page as a program need since i've been here. do we need to preserve the route in any way if we do, and who can we check with to confirm that it's okay?

jasalisbury: I've added the hero styles back in so the Beta page doesn't break. We can remove those styles and other Beta page related items in a different PR if we determine we don't need the page anymore. @mjlumetta I'm not sure who to check with either, any ideas @dvogel?

kalilsn: https://lighthouseva.slack.com/archives/CE79R5SR1/p1552606595143600?thread_ts=1552604538.142600&cid=CE79R5SR1

It was for the health team. I didn't get a response the last two times I asked about it.

PR #234, opened 2019-10-15
Use VADS in VA-related components

mjlumetta:


jasalisbury:
Looks good! Just a couple questions:

 1. The crisis modal is smaller on large screens which seems fine but want to make sure is intentional
 2. The links on the crisis modal are no longer underlined. Do they need to be underlined for accessibility reasons?

mjlumetta:
@jasalisbury yes to 1 - that was the result of using the `formation-react` `Modal` instead of a custom one. re: 2, i'm not sure - i will check that out and update as necessary.

dvogel:
These visual regression test failures look like the result of intentional changes. We should probably update them but I'd like a second set of eyes.

PR #242, opened 2019-10-17
App uses VADS

jasalisbury:


src/App.tsx
```diff
@@ -15,17 +15,19 @@ import { topLevelRoutes } from './Routes';
...
+    // the double flex container only exists and is flexed to
+    // address a bug in IE11 where min-height is only respected
+    // if the parent of a flex container is also a flex container.
     return (
       <FlagsProvider flags={appFlags}>
         <ConnectedRouter history={history}>
-          <div className="app-container">
-            <div className={classNames('app', 'vads-u-width--full')}>
+          <div className="vads-u-display--flex">
+            <div className={classNames("vads-u-display--flex", "vads-u-flex-direction--column", "vads-u-min-height--viewport", "vads-u-flex--1", "vads-u-width--full")}>
```
mjlumetta:
```suggestion
            <div className={classNames("vads-u-display--flex", "vads-u-flex-direction--column", "vads-u-min-height--viewport", "vads-u-width--full")}>
```

don't really need this. since it doesn't have any siblings, `flex-grow` is irrelevant. can leave this off unless it is also necessary for that IE11 bug.

PR #244, opened 2019-10-17
Use VADS in Apply

mjlumetta:


src/containers/ApplySuccess.tsx
```diff
@@ -36,17 +34,27 @@ interface IOAuthCredentialsNoticeProps {
...
 // Mapping from the options on the form to Proper Names for APIs
-const apisToEnglishOauthList: any = {
+const apisToEnglishOauthList = {
   claims: 'Benefits Claims API',
   community_care: 'Community Care API',
   health: 'VA Health API',
   verification: 'Veteran Verfication API',
 };
 
+// this isn't great but we need to do it in the short term since Apply isn't based on the API definitions
```
kalilsn: Can we just put this in its own module so we only define it once?

kalilsn: and include the caveat in there that absolutely no other components should import it

dvogel: Why would we define it more than once? I don't see another definition of it.

kalilsn: 🤦‍♂ I thought this was the same as `formFieldsToFragments`. Nevermind!

kalilsn: in my defense... the whole thing is confusing

src/containers/Apply.scss
```diff
@@ -1,12 +0,0 @@
-@import '../styles/variables';
-
-.form-required-span {
-  color: $color-red;
-}
-
```
dvogel: Was this required class unused? I don't see it removed from anywhere in this PR.

mjlumetta: i don't remember if it was removed in a previous PR, but i couldn't find it anywhere else when i grep'd for it. so yeah, i think not used.

kalilsn: It's a duplicate of a class that's included in formation. I'm assuming we didn't have it back when we were just on uswds?

src/containers/ApplySuccess.tsx
mjlumetta: haha yeah, that is a good defense. either way it's horrible... but yeah, not sure that this ever worked in the first place?

PR #247, opened 2019-10-17
env.js fix

jasalisbury:
env.js now uses BUILD_ENV to determine which `env.local` file to load. Before it used NODE_ENV, which is always set to production in Jenkins. Because NODE_ENV is always `production` when building in Jenkins (regardless of the environment the developer portal is being built for) and the local env file to use was determined by NODE_ENV all environments would use `env.production.local`. This recently caused an issue for review instances where PUBLIC_URL was set using `env.${environment}.local` and all review instance environments used `env.production.local`. This caused the javascript bundle in `dev` and `staging` to not be found. It helps to work through an example:

A recent batch of review instances for https://github.com/department-of-veterans-affairs/developer-portal/pull/219#issuecomment-543361242...
 * The javascript bundle found in `production`'s `index.html` is https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/ff236a4/production/static/js/main.f146abc6.js. This file exists in s3. 
 * The javascript bundle found in `dev`'s `index.html` is https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/ff236a4/production/static/js/main.124b4583.js (Note the `production` in the URL). This file does not exist in s3. However https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/ff236a4/dev/static/js/main.124b4583.js does.


kalilsn:
I don't think there's any good reason to select env files based on `NODE_ENV`, so perhaps we should remove https://github.com/department-of-veterans-affairs/developer-portal/blob/js-properly-load-local-envs/config/env.js#L34

jasalisbury:
> I don't think there's any good reason to select env files based on `NODE_ENV`, so perhaps we should remove https://github.com/department-of-veterans-affairs/developer-portal/blob/js-properly-load-local-envs/config/env.js#L34

Yeah good catch. With loading env files based on NODE_ENV `.env.production` is loaded for all environments, basically making the variables in `.env.production` defaults for any other variables that haven't been defined already.

config/env.js
```diff
@@ -22,16 +22,16 @@ if (!NODE_ENV) {
...
 var dotenvFiles = [
   // e.g. .env.test.local
-  `${paths.dotenv}.${NODE_ENV}.local`,
+  BUILD_ENV !== null && `${paths.dotenv}.${BUILD_ENV}.local`,
 
   // Don't include `.env.local` for `test` environment
   // since normally you expect tests to produce the same
   // results for everyone
   NODE_ENV !== 'test' && `${paths.dotenv}.local`,
+  NODE_ENV === 'test' && `${paths.dotenv}.test`,
```
kalilsn: good catch

mjlumetta: does this fix that visual regression issue??? regardless, good call

jasalisbury: > good catch

The tests actually caught it for me, so :+1: for tests!

> does this fix that visual regression issue??? regardless, good call

I'm pretty sure it won't fix the visual regression issue (having variables in `.env.dev` impact the visual regression tests). NODE_ENV and BUILD_ENV are both set to `test` when the test script fires up and then `dev` when the test server comes up. `.env.${BUILD_ENV}` will cause both `.env.test` and `.env.dev` to be loaded for visual regression tests.

kalilsn: 🤔 instead of using `NODE_ENV` here for the test env files, we could instead set `BUILD_ENV=test

jasalisbury: >  instead of using `NODE_ENV` here for the test env files, we could instead set `BUILD_ENV=test`

I like that! Gets rid of the explicit check for `NODE_ENV=test` (at least one of them)

mjlumetta: gotcha, that's helpful. i think i figured out the source of the bug i was thinking of, actually. i think that you're right that `.env.dev` variables affect the tests and are a bug, but my bigger problem has been `.env.local` (mine has some settings mismatches with the test env, buggy or not, that deterministically cause failures). regardless of whether we use `BUILD_ENV` or `NODE_ENV` to perform the check, neither will affect the test server because neither are set on startup: https://github.com/department-of-veterans-affairs/developer-portal/blob/master/jest-puppeteer.config.js#L12

i'm gonna try that out and see if it works. it does seem like `BUILD_ENV` is preferable, because `start.js` uses `NODE_ENV`.

PR #248, opened 2019-10-18
Use NODE_ENV=test for Puppeteer host in visual regression tests

mjlumetta:
fixes a bug i have in my dev environment where visual regression tests will fail if i run them from `master` with no changes due to use of `.env.local` in the dev environment. the process that runs `visualRegressionTest.ts` has `NODE_ENV=test`, but the process running the web server has `NODE_ENV=development`.


jest-puppeteer.config.js
```diff
@@ -9,7 +9,7 @@ const puppeteerConfig = {
 
 if(!process.env.TEST_HOST) {
   puppeteerConfig.server = {
-    "command": "PORT=4444 BROWSER=false npm run-script start",
+    "command": "USE_LOCAL=false PORT=4444 BROWSER=false npm run-script start",
```
jasalisbury: Do we want to set `USE_LOCAL` false in other `scripts/test.js` as well?

mjlumetta: no, i think that the default of false for scripts other than `start.js`. there's really no point to using `.env.local` in contexts other than the development server. and since the check in `env.js` requires an explicit `true` value, i think it's preferable to ignore `USE_LOCAL` in `test.js` and `build.js`, since it is (at least presently) irrelevant in those contexts.

jasalisbury: Yea you are right, I got mixed up and thought `start.js` was ran when running tests for some reason.

mjlumetta:
retroactively created https://github.com/department-of-veterans-affairs/vets-contrib/issues/3470 for this fix

PR #250, opened 2019-10-18
Use VADS in Search

mjlumetta:


src/components/Search.scss
```diff
@@ -1,58 +1,47 @@
...
+  }
+
+  input.va-api-search-autocomplete {
+    width: 80%;
+  
+    @include media($large-screen) {
+      width: 240px;
+    }
+  
+    // &:focus {
```
mjlumetta: @jasalisbury commented this out but haven't removed it - it wasn't taking effect in the current version. was it supposed to take effect, or is it left over? can do whatever is necessary with it

src/components/Search.tsx
```diff
@@ -20,24 +20,44 @@ export default class Search extends React.Component<{inMenu?: boolean, className
...
+                'vads-u-margin--0',
+                'vads-u-padding--0',
+              ],
+              !inMenu && [
+                'vads-u-margin-bottom--0',
+                'vads-u-margin-left--neg5',
+                'vads-u-margin-right--0',
+                'vads-u-margin-top--0',
+                'vads-u-padding--0',
+              ],
```
dvogel: Using the ternary syntax here would be clearer to me. At first I assumed you must have a typo or something because the two independent expressions are mutually exclusive.

```
              inMenu ? [
                'va-api-search-submit',
                'vads-u-margin--0',
                'vads-u-padding--0',
              ] : [
                'vads-u-margin-bottom--0',
                'vads-u-margin-left--neg5',
                'vads-u-margin-right--0',
                'vads-u-margin-top--0',
                'vads-u-padding--0',
              ],
```

src/components/Search.scss
jasalisbury: Can't remember why I didn't just remove it, I'd say it's okay to just take it out at this point.

PR #253, opened 2019-10-18
Fix documentation link on homepage

rtravitz:
This link is broken on the homepage of the new developer portal. Just a quick fix!


kalilsn:
JK, just realized what link this is!

PR #255, opened 2019-10-21
Update statuspage link location

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3149


src/components/Header.tsx
```diff
@@ -34,6 +34,7 @@ export default class Header extends React.Component<{}, INavBarState> {
           defaultFlexContainer(true),
           'vads-u-justify-content--space-between',
           'medium-screen:vads-u-padding-x--4',
+          'medium-screen:vads-u-margin-y--3',
```
jasalisbury: without the statuspage link inside this container the header was shorter, only as tall as the header title and `request a key` button/search box.

jasalisbury:
> @jasalisbury this looks great! I love the placement and mobile looks perfect. I just have one request...
> 
> I know that the link was bold in the mockup so this is not your fault but can we make the "API Status" `font-weight:400;` like the other main nav menu items for the sake of visual consistency?
> Screen Shot 2019-10-21 at 9 30 51 AM

@hubgitian Yep that should be easy! One potential issue though is I don't think there is a good way to change the font weight of the icon. Font awesome icons come in a single font weight so (unless someone else has a better idea) I'll start changing the font size color to see if I can get something that looks good.

hubgitian:
> I'll start changing the font size color to see if I can get something that looks good.

@jasalisbury I actually don't mind that the external link icon is a tad heavier. I think it calls attention without being too heavy handed. I think we're gtg as is.

PR #257, opened 2019-10-21
Add "Lighthouse" to Support Overview

ryanlink:
Quick PR to add the word "Lighthouse" to the overview paragraph on the Support portal.


src/containers/support/SupportOverview.tsx
```diff
@@ -6,7 +6,7 @@ import { ISection } from './Support';
 
 const headerProps = {
   description:
-    "Welcome to support for the VA API program. You can visit our FAQ page for answers to common questions. If you need to submit a support request or report a bug, you can contact us via GitHub. For general feedback use our 'Contact Us' form. Our customer support team is happy to help and will respond within one business day.",
+    "Welcome to support for the VA Lighthouse API program. You can visit our FAQ page for answers to common questions. If you need to submit a support request or report a bug, you can contact us via GitHub. For general feedback use our 'Contact Us' form. Our customer support team is happy to help and will respond within one business day.",
```
stephaniemace: VA Lighthouse program, instead of VA Lighthouse API program.

ryanlink: Can we have "API" elsewhere in the paragraph then?  If not, "API" doesn't appear anywhere on the page except in the small print on the FAQ card.  My understanding was that "VA Lighthouse" describes the API program.  We just don't want "Lighthouse" used without "VA" in front of it.  I could be wrong, though.

stephaniemace: Sure thing. I agree that API should be inserted in the intro paragraph. Dave and Andrew's guidance back on May 10th was that we should refer to the program as VA Lighthouse, not VA Lighthouse API to avoid confusion. 

One suggestion: we could add "Veteran-centered API platform" to mirror the language on the new homepage.  Or simply:
**You can visit our FAQ page for answers to common questions about VA's API platform.**

PR #258, opened 2019-10-21
Use VADS in VersionSelect

mjlumetta:

i forgot to open this PR earlier, my bad.


mjlumetta:
> Personally, I love big buttons (especially on my phone), so I'm not opposed to making it a bit larger, but could we keep this a little closer to the original size?

@kalilsn yeah definitely. i wasn't really sure what to do here but setting a max width seems appropriate.

PR #259, opened 2019-10-21
Remove unnecessary styles from base.scss

mjlumetta:

should be the last conversion PR other than variable resolution


src/containers/releaseNotes/ReleaseNotesOverview.tsx
```diff
@@ -7,34 +7,40 @@ import CardLink from '../../components/CardLink';
...
     return (
       <div>
-        <PageHeader
-          halo={this.props.halo}
-          header={this.props.header}
-          description={this.props.description}
-        />
-        <div className={defaultFlexContainer()}>{cards(this.props.parent)}</div>
+        <PageHeader halo="Release Notes" header="Overview" />
+        <h2 className="vads-u-font-weight--normal">
```
kalilsn: `<div>` and `<p>` aren't legal children of an `<h2>` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements). Even aside from the html spec issue, this isn't really a header, it's just introductory text. It doesn't seem like the h2 is really necessary either, since we're controlling the font size with a utility class anyways. Can we just drop it?

mjlumetta: that is what i wanted to do initially, but it created an accessibility issue with `h1` to `h3`. so you are one hundo right 💯, but need to think of a way around this, which is not immediately obvious to me

mjlumetta: tbh the answer is probably that we shouldn't use `h3`s in `CardLink` and should style those directly as `div`s or whatever

kalilsn: Yeah I think that's reasonable. `<p class="vads-u-font-size--h3">` perhaps?

kalilsn: reusable components + not skipping heading levels is a huge pain

mjlumetta: definitely, i think that's a good thing to avoid unless it truly needs to be a header element or if being a header is inherently part of the component (ie `PageHeader`, site-wide components, etc). if we _truly_ needed to for some component, we could allow the element type to be passed as a prop and instantiate it, but there's almost always gonna be a simpler solution. anyway, i'll push that commit

PR #261, opened 2019-10-22
Use variables from VADS instead of defining our own

mjlumetta:

we have a weird mismatch of variables that we defined ourselves and never really import from the places they're supposed to. this PR is for making VADS the source of truth on those variables.


src/styles/colors.scss
```diff
@@ -1,17 +1,17 @@
...
-$color-base: #212121;
-$color-dark-blue: #12223A;
-$color-gray-lighter: #d6d7d9;
-$color-alt-gray: #C3C8CF;
-$color-gold: #fdb81e;
-$color-red: #cd2026;
-$color-off-white: #F1F1F1;
-$color-primary-alt-lightest: #e1f3f8;
-$color-gray-dark: #323a45;
-$color-blue-dark: #12223A;
```
jasalisbury: :boom: was able to remove so many colors!

src/styles/uswdsPatch.scss
```diff
@@ -1,23 +1,11 @@
...
-
-// Responsive media
-$small-screen: 481px;
-
-// Fonts
-$helvetica: "Helvetica Neue", "Helvetica", "Roboto", "Arial", sans-serif;
-$font-sans: 'Source Sans Pro', $helvetica;
-
-// polyfill media and display-icon mixins
-// see https://github.com/uswds/uswds/blob/v1.6.10/src/stylesheets/core/_utilities.scss
```
jasalisbury: Do we want to keep this comment?

mjlumetta: the info is essentially covered by the top-level comment now, since this in-line content applies to everything that's left after removing the variables at the top. hoping that's sufficient, let me know if there's anything that could make it clearer!

src/styles/swagger.scss
```diff
@@ -112,12 +112,4 @@
...
     }
   }
-
-  blockquote {
-    line-height: 1.5;
-    color: $color-gray-dark;
-    background: $color-primary-alt-lightest;
-    padding: 1rem;
-    padding-right: 2rem;
-  }
```
kalilsn: This style is actually in use by the appeals team. Compare https://dev-developer.va.gov/explore/appeals/docs/decision_reviews to the same docs on the review instance

mjlumetta: got it, i will add that back. this raises some process questions for me, though... this file is extremely difficult to reason about

kalilsn: more so than other parts of the codebase? what prompted you to remove this rule?

mjlumetta: yeah, for 2 reasons: (1) harder to think about styles for 3rd party components in general, and (2) because we don't actually have control over this content. i guess the latter kinda invalidates what i did, but i stand by how i got here. i was basically auditing this file to see if there was anything that could be removed (bc of redundancy with `base`), and i couldn't find any `blockquote` elements in any of the swagger docs. BUT did not have Decision Reviews enabled locally, and that's how i got there

this file isn't going to be a big problem because we won't touch it a lot, but maybe it's something to keep in the back of our minds going forward

PR #263, opened 2019-10-23
Add release note for facilities sort order fix

nfasulo:
Release note for the fix in department-of-veterans-affairs/vets-api/pull/3461
Closes department-of-veterans-affairs/vets-contrib/issues/2501


src/content/apiDocs/facilities/facilitiesReleaseNotes.mdx
```diff
@@ -1,3 +1,9 @@
+## October 23, 2019
+When including latitude and longitude parameters in a request to  `/facilities` the results are now sorted in ascending order by distance.
+-  Previously the facilities in the response had be sorted by facility type first and then distance, now the closest facilities to the requested location will be returned first regardless of facility type [View code change(s)](https://github.com/department-of-veterans-affairs/vets-api/pull/3461)
```
whuang85: `Previously the facilities in the response had be sorted`, is that supposed to be `has been`?

rtravitz: A couple of thoughts with a suggested change underneath:
- "had be" seems like a typo but can be changed more clearly to "were"
- "facility type" can be replaced by "type." The "facility" part is implicit
- Remove the comma splice after distance and make a new sentence

```
Facilities in the response were previously sorted by type first and then distance. Now the closest facilities to the requested location will be returned first regardless of type.
```

nfasulo: Have I ever mentioned that I **love** having a real-life copy-editor on the team?

PR #264, opened 2019-10-23
Improve CSS documentation

mjlumetta:
documentation follow-up for https://github.com/department-of-veterans-affairs/vets-contrib/issues/3011

there are a couple blank sections but i don't want to spend any more time on this ticket.


docs/guides/css.md
```diff
@@ -0,0 +1,131 @@
...
+| `va-crisis-panel-body` | Styles the inside of the VCL modal | [`_m-crisis-line.scss`](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation/sass/site/) | `VeteransCrisisLinePanel` |
+| `vcl` | Styles the white VCL flag icon on the left of the VCL button | [`_m-crisis-line.scss`](https://github.com/department-of-veterans-affairs/veteran-facing-services-tools/blob/master/packages/formation/sass/site/) | `VeteransCrisisLine` |
+
+#### `uswds` in `formation`
+
+[`uswds`](https://github.com/uswds/uswds), the `npm` package for the [US Web Design System](https://designsystem.digital.gov/), is a direct dependency of `formation`, and `formation` exports some parts of `uswds` as part of its library. As a result, we use some USWDS styles on the developer portal. Most, if not all, of these styles are documented in the [VADS component docs](https://design.va.gov/components). We should only use USWDS styles that are exported and documented by USWDS.
```
jasalisbury: >  We should only use USWDS styles that are exported and documented by USWDS.

Should this be `We should only use USWDS styles that are exported and documented by VADS`?

docs/guides/css.md
```diff
@@ -0,0 +1,131 @@
...
+| `usa-button-primary` | Variation on `usa-button` used for the primary action within a given context | [`_buttons.scss`](https://github.com/uswds/uswds/blob/release-1.6.10/src/stylesheets/elements/_buttons.scss) | `Form`, `ApplyForm` | [https://design.va.gov/components/buttons](https://design.va.gov/components/buttons) |
+| `usa-font-lead` | Sets a few properties for leading or significant paragraphs: `font-size`, `font-weight`, `line-height`, `max-width` | [_typography.scss](https://github.com/uswds/uswds/blob/release-1.6.10/src/stylesheets/elements/_typography.scss) | `ApplyHeader`, `ApplySuccess` | NOT DOCUMENTED, probably don't use again |
+| `usa-form` | Sets some basic form properties |  | `ApplyForm`, `swaggerPlugins/Servers` | [https://design.va.gov/components/form-controls](https://design.va.gov/components/form-controls)
+| `usa-sidenav-list` | Provides some styles for lists of items within the side nav | [_sidenav.scss](https://github.com/uswds/uswds/blob/release-1.6.10/src/stylesheets/components/_sidenav.scss) | `SideNav` | [https://design.va.gov/components/sidenav](https://design.va.gov/components/sidenav) |
```
dvogel: Is the value in listing these out to provide examples of where it makes sense or is it to be an exhaustive list? It seems like it will be difficult to keep current. If it's the former, maybe it would make sense to pare it down to be more digestible?

mjlumetta: i was going for a few different things here:

- give examples of where to use it/existing patterns
- show what parts of USWDS we use
- show what parts of USWDS are included in VADS
- make it easier to change our usage of USWDS in the future, for example:
  - if we want to import specific modules from USWDS instead of importing through VADS
  - if VADS deprecates any of these `usa-*` classes and we need to update/remove them

i don't know that any of these is any sort of absolute need, and i see where all of this together would be confusing. my overarching goal would be to have some sort of reference on where USWDS is in VADS. i think this could be split out into 2 tables (or a table and a list):

1. USWDS in VADS - links to documentation and source files
1. where we use USWDS - the same list but just classes and components. i agree that this shouldn't be considered exhaustive but i don't think it hurts to include everything that's present now, if it will be digestible in a different form.

let me know if you have thoughts on this.

PR #271, opened 2019-10-30
Prettify authorization docs styles

mjlumetta:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2569

pretty up the authorization docs in accordance with this design: https://www.figma.com/file/eCLoBkuMRD4IR31BdttRtjJd/Documentation-Readability?node-id=0%3A1

@hubgitian i have a few follow-up questions i'll post here tomorrow.


mjlumetta:
@hubgitian starting my issue list. issue 1: code syntax highlighting. this is already inconsistent, but we have some new issues with the dark background...

**HTTP example**

HTTPDarkSyntaxHighlights

**Javascript example**

JavascriptDarkSyntaxHighlights

you'll notice that the syntax highlighting = bad on dark background. a little background (haha, sorry): the source of these syntax styles is from [`highlight.js`](https://highlightjs.org/static/demo/), which is included via the Markdown component loader. the markdown loader automatically styles the code elements with syntax highlighting. in the first example, we've specified that the snippet is HTTP, so it styles the content as such. in the second, hightlight.js has automatically inferred that the code is Javascript, so it applies those styles. in either case, it doesn't look good.

contrast that result with the following HTTP snippet:

HTTPNoSyntaxHighlights

here we've _also_ specified HTTP as the language, but we get no syntax highlights. i'm guessing that's because highlightjs couldn't parse/recognize this snippet as an HTTP response. unsure, but the takeaway from this example is that we would need to do some massaging to get the automatic syntax highlighting working for all snippets (not sure how important that is).

here are some possible options available to us:

1. override problematic highlights in `AuthorizationDocs.scss` (worst, going to have to fix later)
1. temporarily use `plaintext` as the language on either (a) all snippets or (b) every problematic snippet. this will remove all syntax highlighting, but it will keep things consistent for the time being. defers the work of syntax highlighting, however important that is to us.
1. see if `markdown-component-loader` allows you to configure the highlightjs style/theme - i don't think that this is an option, but verify that (best option if it's possible)
  a. if this is possible, or if we make it possible with the next point, i'd suggest that we update that acceptance criteria to use a particular highlightjs theme rather than background color and color
1. extend `markdown-component-loader` to allow specifying the highlightjs theme (best if previous option isn't possible). @dvogel has already forked that repo to provide other functionality that we needed, and i think it's quite conceivable that we could do it, it just requires the most work. which leads to the option of... 
1. defer this work to a later ticket

mjlumetta:
@hubgitian issue 2: inline code wrapping. i've tentatively chosen not to wrap across lines:

InlineCodeBlockNowrap

the alternative is, because this is an inline element, that the background doesn't have any padding at line end/start. so, before i did this, the spacing was like `  Authorization: Bearer` and `{token}  ` (whitespace shows where horizontal padding is. either way is fine, but want you to be aware of it in case you have preferences or thoughts on creative solutions.

hubgitian:
@mjlumetta this looks great so far! Already way better.
### issue one
On the issue of syntax highlighting, are we able to load "Atom One Dark Reasonable" or VS2015" instead of the "Default" version of hilight.js?

### issue two
While I'm usually not a fan of big chunked up paragraphs I think you're making the right call here clarity is way more important than a symmetrical looking paragraph.

### issue three
I'm glad we'll potentially be able to implement this using classes but I also am interested in you concerns with extensibility and want to make sure I don't paint us into a corner we'll have to back out of.

kalilsn:
This is so much easier to read already 🎉!

@mjlumetta I'm pretty sure that the markdown loader does tokenize and apply hljs classes to the code blocks, but I don't think it actually does the styling. We should be able to adjust the theme just by changing this line:
https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/App.tsx#L17

As for the tips question, I think that has its own ticket (https://github.com/department-of-veterans-affairs/vets-contrib/issues/2576), so leaving them as is should be ok for this PR.

mjlumetta:
> @mjlumetta I'm pretty sure that the markdown loader does tokenize and apply hljs classes to the code blocks, but I don't think it actually does the styling. We should be able to adjust the theme just by changing this line:
https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/App.tsx#L17

thanks for the link @kalilsn! that makes sense - i was previously under the impression (assumption, really) that the theme was configured into highlightjs rather than imported as a stylesheet. this makes it super easy. the one thing that i'm going to add to that is adding a direct dependency on highlightjs, because that was part of my confusion and seems like best practice.

> As for the tips question, I think that has its own ticket (department-of-veterans-affairs/vets-contrib#2576), so leaving them as is should be ok for this PR.

whoops! somehow i tied them together in my head because of the proximity to code blocks. i'm going to push this commit to a new branch and it can be up there as a starting point for that ticket.

mjlumetta:
@jasalisbury thanks for the feedback! i'll fix the first and third issues, will wait for Ian on the second.

@hubgitian wanted to raise another UX issue (that wasn't introduced in this ticket), and which may also be an accessibility issue (but is an issue for all users regardless). for code blocks with horizontal overflow, there's no visual indication that there is overflow; you have to infer it from the content. i don't think that necessarily needs to go in this PR - could potentially go in 2576 - because it's not a new issue, but at the same time, i think we should address it now. this means we'll have to put some work into getting acceptable scrollbar styling (which may require some effort), or we should figure out an alternative way to arrange this content so that there isn't horizontal overflow (conceivably doable using other patterns we have on the page, but will make copy to clipboard more difficult, most likely). anyway, if you'd like to sync up on this, i can show you what i mean. you can also see it in production if you look at the example authorization URL under Initiating Authorization.

jasalisbury:
> @hubgitian wanted to raise another UX issue (that wasn't introduced in this ticket), and which may also be an accessibility issue (but is an issue for all users regardless). for code blocks with horizontal overflow, there's no visual indication that there is overflow; you have to infer it from the content. i don't think that necessarily needs to go in this PR - could potentially go in 2576 - because it's not a new issue, but at the same time, i think we should address it now. this means we'll have to put some work into getting acceptable scrollbar styling (which may require some effort), or we should figure out an alternative way to arrange this content so that there isn't horizontal overflow (conceivably doable using other patterns we have on the page, but will make copy to clipboard more difficult, most likely). anyway, if you'd like to sync up on this, i can show you what i mean. you can also see it in production if you look at the example authorization URL under Initiating Authorization.

I see horizontal scroll bars on long code blocks. Is the issue that it doesn't show up in all browsers or is the issue that it needs to be styled? We may be able to reduce the horizontal scroll on some by just adding some newlines to the content (since we control these code blocks) but that won't help in the general case and some blocks have lines that don't have obvious breakpoints.

mjlumetta:
here's what i see in Chrome on Mac:

Screen Shot 2019-11-01 at 12 31 24 PM

Screen Shot 2019-11-01 at 12 31 54 PM

for me, it's also not possible to focus (?) in that element, so the only way i can scroll is by selecting/highlighting text.

jasalisbury:
Gotcha I see this on linux chrome:

Screenshot from 2019-11-01 11-34-53

mjlumetta:
@jasalisbury ahh good to know that it's OS-specific. i also see it in firefox, so it's probably all browsers on mac. sounds like could maybe make a ticket and not address it in any of the current tickets, then. thanks for helping sort this one out

kalilsn:
Yeah this is a mac "feature" (that I find very hostile!). In system preferences -> general there's a setting about when to show scroll bars.

Screen Shot 2019-11-01 at 11 35 43 AM

I think it defaults to automatic, which gives me the following UI:

Screen Shot 2019-11-01 at 11 36 43 AM

When it's "always", I see this:
Screen Shot 2019-11-01 at 11 37 22 AM

I think this is just some bad behavior from Apple, probably not something for us to worry about.

mjlumetta:
thanks @kalilsn - that's great, love a good no-op. uhh, screw apple then, i guess? ... you learn something new every day

jasalisbury:
It feels like it is worth investigating a little bit (like 30 minutes or so max) to see if we can force some horizontal scroll bars for mac users. This is a problem for a lot of people so there must be a technique or library we can use to force the horizontal scroll bar to show up.

hubgitian:
@mjlumetta yes please remove the `<code>` tag from the `<h2>` level instance of "id token"

jasalisbury:
oh another thing I forgot to point out, in the mockup `<h2>`s (or `<h3>`s, not sure which it should be @hubgitian) are underlined but they aren't on the review instance.

mjlumetta:
> oh another thing I forgot to point out, in the mockup <h2>s (or <h3>s, not sure which it should be @hubgitian) are underlined but they aren't on the review instance.

@jasalisbury we ended up deciding to remove that AC when we met the other day. good catch, though

PR #281, opened 2019-11-19
directly link to api version

whuang85:


src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -71,41 +71,58 @@ export default class SwaggerDocs extends React.Component<ISwaggerDocsProps, ISwa
...
-    return metadata.meta.versions.find(selectCurrentVersion);
+  private getVersionInfo(version: string, metadata: any) {
+    if (!metadata) {
+      return null;
+    }
+    const selectCurrentVersion = (versionInfo: IVersionInfo) => versionInfo.status === SwaggerDocs.currentVersionStatus;
+    const selectVersion = (versionInfo: IVersionInfo ) => versionInfo.version === version;
+    let metadataInfo = metadata.meta.versions.find(selectVersion);
+    if (!metadataInfo) {
+      metadataInfo = metadata.meta.versions.find(selectCurrentVersion);
```
mjlumetta: is there any case where the first selection shouldn't succeed but this one should? if there is, should it still try to find the metadata or should it fall back to the current version?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -26,39 +27,38 @@ export interface IVersionInfo {
 }
 
 export default class SwaggerDocs extends React.Component<ISwaggerDocsProps, ISwaggerDocsState> {
+
```
mjlumetta: kind of a general comment on the changes to this component, but the control flow of this component is getting pretty confusing, imo. not really your fault because it was kind of already that way and therefore difficult to build on top of. i would make myself available to pair sometime in the next day or so if you'd like a second set of eyes on it; i realize this suggestion isn't very concrete. i think the general goal would be to consolidate the state management so that it's clear where the source of truth for version and metadata is. again, happy to talk about this further. hope that some of the other comments in this file are a little more concrete.

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -26,39 +27,38 @@ export interface IVersionInfo {
...
-    const versionMetadata = this.state.metadata.meta.versions.find((versionInfo: IVersionInfo) => {
-      return versionInfo.version === version;
-    });
+    const versionInfo = this.getVersionInfo(version, this.state.metadata);
     this.setState({
-      docUrl: this.buildUrlFromVersionInfo(versionMetadata),
-      version,
-    });
+      docUrl: versionInfo ? this.buildUrlFromVersionInfo(versionInfo) : this.props.docSource.openApiUrl,
+    }, () => this.props.updateApiVersion(version));
```
mjlumetta: can this whole method be moved into the redux action so that we can skip the two-step process of updating the local state and then the redux state? if that can be done, it would also have the benefit of moving the doc URL to the store, which streamlines things a bit

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -26,39 +27,38 @@ export interface IVersionInfo {
...
   public componentDidMount() {
-    this.loadMetaDataAndRender();
+    this.setSwaggerDocState();
   }
 
   public componentDidUpdate(prevProps: ISwaggerDocsProps, prevState: ISwaggerDocsState) {
     if (prevProps.apiName !== this.props.apiName) {
+      this.setSwaggerDocState();
+    }
+    if (prevProps.apiVersion !== this.props.apiVersion) {
```
mjlumetta: it seems like this method will probably set the state and render twice any time the API changes. i'm guessing that the API version will usually also change when the API changes. if that's the case, then the second block will re-run the logic of the first block, minus setting the metadata. setting the metadata is actually the only difference between the two blocks, so i think it can be refactored so that's the only conditional part of this method?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -114,8 +131,16 @@ export default class SwaggerDocs extends React.Component<ISwaggerDocsProps, ISwa
         plugins: [plugins],
         url: this.state.docUrl,
       });
-      ui.versionActions.setApiMetadata(metadata);
-      ui.versionActions.setApiVersion(this.state.version);
+      ui.versionActions.setApiVersion(this.props.apiVersion);
+      ui.versionActions.setApiMetadata(this.state.metadata);
+      this.updateURLHash();
```
mjlumetta: i'm a little confused here... why does updating the hash happen as part of the render method?

src/reducers/index.ts
```diff
@@ -112,5 +112,15 @@ export function application(
     default:
       return { ...state, inputs: applicationInput(state.inputs, action) };
   }
-  return state;
 }
+
+export function apiVersion(
```
mjlumetta: is there a closing brace and parentheses for this function? seems off  but not sure

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -114,8 +131,16 @@ export default class SwaggerDocs extends React.Component<ISwaggerDocsProps, ISwa
...
     }
   }
-}
+
+  private updateURLHash() {
+    const versionInfo = this.getVersionInfo(this.props.apiVersion, this.state.metadata);
+    const version = versionInfo ? versionInfo.version : null;
```
mjlumetta: it's a little bit confusing that the logic for getting the version here is to use `props.apiVersion` to get the version info and then using the version from the version info. is that there for a particular reason, or can it be simplified a bit?

whuang85: yea I should rename the `getVersionInfo`, it is trying to get the versionInfo for the `props.apiVersion` and if it is null, then it would fall back to the versionInfo with `status == `Current Version`

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -1,40 +1,61 @@
...
+
+class ApiDocumentation extends React.Component<IApiDocumentationProps, IApiDocumentationState> {
   public constructor(props : IApiDocumentationProps) {
     super(props);
     this.state = { tabIndex: 0 };
   }
 
   public componentDidMount() {
     this.setTabIndexFromFragment();
+    this.setApiVersionFromHash(location.hash);
```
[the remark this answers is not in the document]
whuang85: I will need to redesign, I totally forgot the fhir hashes for the tabs.. I don't think it would be right to have two hashes, I will be taking a poll in ao-engineering to see if people would mind if I used a query parameter for the version so something like `fhir?version=1.0.0#dstu2` so the changes you have commented on might be moot.

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -51,7 +72,7 @@ export default class ApiDocumentation extends React.Component<IApiDocumentationP
     return (
       <Flag name={`hosted_apis.${apiDefinition.urlFragment}`}>
         {apiDefinition.docSources.length === 1 
-          ? <SwaggerDocs docSource={apiDefinition.docSources[0]} apiName={apiDefinition.urlFragment} />
+          ? <SwaggerDocs docSource={apiDefinition.docSources[0]} apiName={apiDefinition.urlFragment} apiVersion={this.props.apiVersion} updateApiVersion={this.props.updateApiVersion} />
```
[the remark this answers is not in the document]
whuang85: so originally I had these two props passed in via redux but it seemed weird to me that the container (ApiDocumentation) and the child (SwaggerDoc) both had to call `connect` in them.. it's probably just me as I don't quite know the do's and don't's of react-redux

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -80,6 +101,10 @@ export default class ApiDocumentation extends React.Component<IApiDocumentationP
     );
   }
 
+  private setApiVersionFromHash(hash: string = '') {
+    this.props.updateApiVersion(hash.replace(/[^\d.-]/g, ''));
```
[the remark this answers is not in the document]
whuang85: the safety check is to mostly make sure the passed in hash in not null, but thinking about it, I don't think that could ever be the case..  On the tabbed pages, it would set the `apiVersion` to a blank string which actually signifies to me to use the current version for the docs.

src/containers/documentation/SwaggerDocs.tsx
mjlumetta: okay! i don't think you have to change the name, but if you have a better one, go for it. i guess my question then is can we remove this null check and just say `const version = this.getVersionInfo(...).version`?

src/containers/documentation/DocumentationRoot.tsx
```diff
@@ -16,14 +16,15 @@ import DocumentationOverview from './DocumentationOverview';
...
+import SwaggerDocs from './SwaggerDocs';
 
 function SideNavApiEntry(apiCategoryKey: string, api: IApiDescription) {
   return (
     <Flag key={api.urlFragment} name={`hosted_apis.${api.urlFragment}`}>
       <SideNavEntry
         key={api.urlFragment}
         exact={true}
-        to={`/explore/${apiCategoryKey}/docs/${api.urlFragment}`}
+        to={`/explore/${apiCategoryKey}/docs/${api.urlFragment}#${SwaggerDocs.currentHash}`}
```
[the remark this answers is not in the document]
whuang85: One of the requirements was when you navigate to any of those api document links via the side bar, it would mark it as the `current` version (so no version number just the word `current`). I actually treat all non numbers as a blank string which when getting the `versionInfo` would fall back to getting the `versionInfo` where the status is `Current Version`. 

Your suggestions seem like it's keeping track of the users navigation of the the docs which is something I don't care about because if they navigate via the side bar, it will always present them the current version.  If they navigated to a specific API via manipulating the url, they could specify the version in the url i.e. `docs/claims#v0.0.1`.  For the version selector, if they change the version, to a previous version, the url will be appended with the hash `#v0.0.1` and if it's the current version, then `#current`, this is just so they can bookmark or share the link to someone for a specific version. I hope this explanation makes sense.

src/containers/documentation/ApiDocumentation.tsx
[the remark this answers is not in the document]
mjlumetta: i also haven't used react-redux enough to say this with confidence, but based on experience with similar javascript state management systems, i'm pretty sure it's the way you had it initially, and i hope i can explain why. if we pass props from Redux to some parent and then directly from the parent to child components, then there are two ways to get values from Redux: as a child directly from a parent, and as a parent for some subtree in the component tree that uses a particular set of values. always using `connect` gives us a single source of truth, which can simplify the logic a bit. it also means that we have fewer parent-child dependencies in the component tree. this isn't really a problem when you only do it once, but as application state grows and becomes more complex, using `connect` is somewhat more scalable. i think a good general guideline is to use values from the store if the component can deterministically compute its own props (ie, in `mapStateToProps`) from the application state in the store, and to pass props directly if the parent container can make decisions about what the values should be. hope that makes sense.

src/containers/documentation/DocumentationRoot.tsx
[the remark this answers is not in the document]
mjlumetta: okay, gotcha, i think i get it. i see that we don't want to do anything complicated here with user navigation. so this is always "current"? if so, can we not use a variable at all?

i guess my other question would be why this needs to be in the link at all. if a user navigates to the URL with no fragment included, you append `#current` to the URL, right? if so, what's the benefit of including it in the link? this question is more for my understanding, i'm sure there's a motivation behind it but i'm not quite familiar with all the UX goals.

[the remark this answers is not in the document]
jeff2d2: The thinking is that by appending #current (or similar), we're leading the user to knowing that they are indeed grabbing a link that will always point to the current version. Should a new version become the new current version, then they should expect the contents of that link to change and won't be surprised when it does.

src/reducers/versioning-request.ts
```diff
@@ -0,0 +1,67 @@
+import { createSelector } from 'reselect';
+import { ISetInitialVersioning, ISetRequestedApiVersion } from '../actions';
+import { IVersionInfo } from '../containers/documentation/SwaggerDocs';
+import { IApiVersioningRequest } from '../types';
+import * as constants from '../types/constants';
+
+export const currentVersion = 'current';
```
mjlumetta: can this be renamed to `currentVersionIdentifier`, `currentVersionSelector`, `currentVersionValue`? i'm not sure what the best name would be, but since this is exported, i feel like it would be good to make clear that this is a constant and not the current version of some API if possible. another solution would be moving it into `types/constants`.

src/types/index.ts
```diff
@@ -41,7 +41,14 @@ export interface IApplication {
   clientSecret: string;
 }
 
+export interface IApiVersioningRequest {
```
mjlumetta: since this is part of the state tree and not an action, could we remove the "request" from this name (and similar types/variables/the reducer module)? to me, "request" sounds like it's referring to dispatching an action, and avoiding that ambiguity might be worthwhile.

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -1,40 +1,61 @@
...
+import { IRootState } from '../../types';
+
 
 import '../../../node_modules/react-tabs/style/react-tabs.scss';
 
 interface IApiDocumentationProps {
   apiDefinition: IApiDescription;
   categoryKey: string;
   location: Location;
+  requestedApiVersion: string;
```
mjlumetta:
```suggestion
  apiVersion: string;
```

see comment in `src/types/index.ts`

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -80,19 +101,27 @@ export default class ApiDocumentation extends React.Component<IApiDocumentationP
...
+  private getTabIndexFromQueryParams() : number {
+    if (this.props.location.search) {
+      const params = new URLSearchParams(history.location.search);
+
       const hasKey = (source : IApiDocSource) => !!source.key;
       const tabKeys = this.props.apiDefinition.docSources.filter(hasKey)
         .map(source => source.key!.toLowerCase());
-      const fromFragment = this.props.location.hash.slice(1).toLowerCase();
+      const tabQuery = params.get('tab');
+      const fromFragment = tabQuery ? tabQuery.toLowerCase() : '';
```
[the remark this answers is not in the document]
whuang85: unfortunately, VS code keeps telling me `params.get('tab')` can be null even though we have the params.has('tab') check

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -101,8 +130,14 @@ export default class ApiDocumentation extends React.Component<IApiDocumentationP
   }
 
   private onTabSelect(tabIndex: number) {
-    const hash = this.props.apiDefinition.docSources[tabIndex].key;
-    history.push(`${history.location.pathname}#${hash}`);
+    const tab = this.props.apiDefinition.docSources[tabIndex].key;
+    const params = new URLSearchParams(history.location.search);
+    if (tab) {
```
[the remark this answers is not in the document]
whuang85: this is another case where typescript is telling me that `tab` could be undefined and the `params.set()` function does not take undefined

src/containers/documentation/ApiDocumentation.tsx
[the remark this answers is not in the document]
mjlumetta: for future reference, you can override that intellisense with an exclamation point, as in `params.get('tab')!` that tells the typescript compiler that you're sure that a value will be non-null when the compiler can't determine that. not something you want to overuse but useful in such a situation

PR #282, opened 2019-11-19
Add all the necessary things for vaForms

ATeal:
This adds the vaForms API to the developer portal and allows users to request sign ups for it.

Part of https://github.com/department-of-veterans-affairs/vets-contrib/issues/3523


src/apiDefs/data/categories.ts
```diff
@@ -45,7 +47,16 @@ const apiDefinitions : IApiCategories = {
...
-    tabBlurb: "The VA's FHIR Health APIs allow consumers to develop applications using Veteran data. Please see the tabs below for the specific FHIR implementations.",
+    tabBlurb:
+      "The VA's FHIR Health APIs allow consumers to develop applications using Veteran data. Please see the tabs below for the specific FHIR implementations.",
+  },
+  vaForms: {
+    apiKey: true,
+    apis: vaForms,
+    buttonText: 'Stay Informed',
+    content: vaFormsContent,
+    name: 'Form API',
```
charleystran: I think this should be 'Forms' over 'Form' to be consistent

ATeal: true will change

PR #284, opened 2019-11-22
Fix va-forms signup

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3717


jasalisbury:
I'll rerun in Jenkins so we get a review instance once the current build finishes

PR #287, opened 2019-11-26
Standardize API labeling

jeff2d2:
### The What
This PR looks to solve for a disconnect that exists on the developer portal in labeling some API categories as `an API` when in reality they are categories that make up a larger `set of APIs`. It also addresses how some API names on the apply page don't match the API name in the docs.

### The Why
This was first raised as a concern and recommendation in the [Developer Portal IA Foundational Research](https://github.com/department-of-veterans-affairs/vets-contrib/blob/master/products/developer-portal/research/_foundational/sprintreport-devportaluxfixes.md) (see # 6 in [Findings](https://github.com/department-of-veterans-affairs/vets-contrib/blob/master/products/developer-portal/research/_foundational/sprintreport-devportaluxfixes.md#prioritized-usability-findings) and # 6 in [Recommendations](https://github.com/department-of-veterans-affairs/vets-contrib/blob/master/products/developer-portal/research/_foundational/sprintreport-devportaluxfixes.md#tactical-recommendations)), however it has also been discussed in various conversations regarding how to best represent the APIs on the platform and their respective categories.

This PR does NOT look to solve for everything at once, but instead take a step in the right direction.


src/apiDefs/data/categories.ts
```diff
@@ -29,8 +29,8 @@ const apiDefinitions: IApiCategories = {
     apis: benefitsApis,
     buttonText: 'Get Your Key',
     content: benefitsContent,
-    name: 'Benefits API',
-    properName: 'Benefits Intake API',
+    name: 'Benefits APIs',
+    properName: 'Benefits APIs',
```
mjlumetta: i'm not 100% sure, but i feel like it has been plural before and we've changed it back (as a historical note - i don't think a whole lot of thought went into that, though). i think that, if you are trying to make a definitive choice about how to pluralize these, it would be good to add a comment at the top of this `apiDefinitions` object outlining that pattern so that future editors of this file know not to change that.

also, "Veteran Verification API" below seems like it should be plural as well? if the pattern is to pluralize categories with multiple APIs and leave singular categories with one API

mjlumetta: also, a small word of warning - as categories with 1 API get a second API and therefore need to be pluralized, i feel like it will be very easy for that change to slip through the cracks. i can think of a way to safeguard against that so that it doesn't need to be changed manually and instead this file automatically figures it out, but it would require a small bit of code. i can push that here or help you make that change, if you'd like.

src/containers/ApplyForm.tsx
```diff
@@ -177,7 +177,7 @@ class ApplyForm extends React.Component<IApplyProps> {
                   checked={apis.benefits}
                   onChange={props.toggleBenefits}
                 />
-                <label htmlFor="benefits">VA Benefits API</label>
+                <label htmlFor="benefits">VA Benefits Intake API</label>
```
mjlumetta: i believe that this is meant to refer to all of the Benefits APIs collectively rather than the Benefits Intake API

jeff2d2:
Related conversation: https://lighthouseva.slack.com/archives/CEXC8GV2S/p1574801732025900

src/apiDefs/data/categories.ts
jeff2d2: through some additional [group discussion here](https://lighthouseva.slack.com/archives/CEXC8GV2S/p1574801732025900), we're now exploring if having either API or APIs in the category is unnecessary and too repetitive. I've made some updates to get a look at this.

src/containers/ApplyForm.tsx
jeff2d2: This checkbox on the apply page is specifically for the Benefits Intake API. The other Benefits APIs (Benefits Claims and Appeals) have their own checkboxes.

mjlumetta: that's not quite true - the Appeals checkbox refers to the new Decision Reviews API listed under "Appeals" added by @erluti. the Claims API is the Benefits Claims API but is listed separately because it uses OAuth rather than an API key as of v1, whereas Benefits Intake and Appeals (under Benefits) use the key. in any case, these things may need to change based on current needs - which you would know far better than i would - but that's historically how it has evolved.

PR #292, opened 2019-12-04
Move address validation to new utility APIs category

dvogel:
This is a demonstration PR to address department-of-veterans-affairs/vets-contrib/issues/2190. I've moved the `Address Validation API` out of the `Veteran Verification API` category. Then I introduced the long-planned `Utility APIs` category to house it.

In addition to moving the `Address Validation API`, it collapses the various psuedo-APIs in the `Veteran Verification API` category into a properly-defined `Veteran Verification` API nav entry. This brings the presentation of the separate OpenAPI specification files for the constituent parts of the API into the tabbed interface our governance process settled on for the FHIR API.

This highlights the need to rename the `Veteran Verification API` category and the `Veteran Verification` API nav entry. This discrepancy could be sufficiently addressed by moving `API` from the category names and perhaps moving the term down into the second level side nav entries for semantic accuracy. For a demonstration of that idea, see this PR: https://github.com/department-of-veterans-affairs/developer-portal/pull/287


afichter-va:
Thanks for putting the thought into this and creating a visual representation that can be reviewed.

While I personally like the tabbed interface, I want to ensure we are using it for consistent purposes. The tabs for the Health API represent different versions of the FHIR spec; they don't represent different APIs entirely. We incorporated the tabs specifically for this unique situation.

Additionally, for other top-level categories that have multiple APIs, we have the card display (for example, under Benefits). Using tabs for Veteran Verification would not be consistent in this case. However, it may make sense to consider (eventually) using tabs to represent the OAuth and api-key endpoints for Veteran Confirmation.

As for renaming the Veteran Verification nav entry, that is indeed being renamed as part of the work. It will be re-categorized as Veteran Data. You can refer to the latest mockup [here](https://www.figma.com/file/rcxafNa0GJxc2nnCWZV1W0/Veteran-Verification?node-id=8%3A0), for reference.

mjlumetta:
@dvogel we should hold off on reviewing this, right?

dvogel:
> @dvogel we should hold off on reviewing this, right?

Yeah. I should have made it a draft PR. It's intentionally incomplete, to leave it as a demonstration.

rtravitz:
Closing, but will continue to reference this PR in future conversations about category reorganization.

PR #294, opened 2019-12-09
get the real version number in addition to the pseudo 'current' version

whuang85:
this PR addresses issue: https://github.com/department-of-veterans-affairs/vets-contrib/issues/3844

### DIFF
--- a/src/containers/documentation/SwaggerDocs.tsx
+++ b/src/containers/documentation/SwaggerDocs.tsx
@@ -6,7 +6,7 @@ import { connect } from 'react-redux';
 import SwaggerUI from 'swagger-ui';
 import * as actions from '../../actions';
 import { IApiDocSource } from '../../apiDefs/schema';
-import { getDocURL, getVersion } from '../../reducers/api-versioning';
+import { getDocURL, getVersion, getVersionNumber } from '../../reducers/api-versioning';
 import { history } from '../../store';
 import { IRootState } from '../../types';
 import { SwaggerPlugins } from './swaggerPlugins';
@@ -22,6 +22,7 @@ export interface ISwaggerDocsProps {
   setInitialVersioning: (url: string, metadata: any) => void;
   setRequestedApiVersion: (version: string) => void;
   version: string;
+  versionNumber: string;
 }
 
 export interface IVersionInfo {
@@ -38,6 +39,7 @@ const mapStateToProps = (state : IRootState) => {
     location: state.routing.location,
     metadata: state.apiVersioning.metadata,
     version: getVersion(state.apiVersioning),
+    versionNumber: getVersionNumber(state.apiVersioning),
   };
 };
 
@@ -121,7 +123,7 @@ class SwaggerDocs extends React.Component<ISwaggerDocsProps> {
         plugins: [plugins],
         url: this.props.docUrl,
       });
-      ui.versionActions.setApiVersion(this.props.version);
+      ui.versionActions.setApiVersion(this.props.versionNumber);
       ui.versionActions.setApiMetadata(this.props.metadata);
     }
   }
--- a/src/containers/documentation/swaggerPlugins/CurlForm.tsx
+++ b/src/containers/documentation/swaggerPlugins/CurlForm.tsx
@@ -96,13 +96,7 @@ export class CurlForm extends React.Component<ICurlFormProps, ICurlFormState> {
     if (this.isSwagger2()) {
       spec.host = this.state.env ? `${this.state.env}-${spec.host}` : spec.host;
     } else {
-      let version = this.props.system.versionSelectors.apiVersion();
-      if (!version) {
-        version = '0';
-      } else {
-        version = version.substring(0, 1);
-      }
-
+      let version = this.props.system.versionSelectors.majorVersion();
       options.server = this.state.env;
       options.serverVariables = {
         version: `v${version}`,
--- a/src/containers/documentation/swaggerPlugins/CurlForm.tsx
+++ b/src/containers/documentation/swaggerPlugins/CurlForm.tsx
@@ -96,7 +96,7 @@ export class CurlForm extends React.Component<ICurlFormProps, ICurlFormState> {
     if (this.isSwagger2()) {
       spec.host = this.state.env ? `${this.state.env}-${spec.host}` : spec.host;
     } else {
-      let version = this.props.system.versionSelectors.majorVersion();
+      const version = this.props.system.versionSelectors.majorVersion();
       options.server = this.state.env;
       options.serverVariables = {
         version: `v${version}`,
--- a/src/containers/documentation/swaggerPlugins/VersionSelector.ts
+++ b/src/containers/documentation/swaggerPlugins/VersionSelector.ts
@@ -1,7 +1,12 @@
+import { createSelector } from 'reselect';
+
 export const VersionSelector = {
   selectors: {
     apiMetadata: (state: any) => state.get('apiMetadata'),
     apiName: (state: any) => state.get('apiName'),
     apiVersion: (state: any) => state.get('apiVersion'),
+    majorVersion: createSelector((state: any) => state.get('apiVersion'), (apiVersion: string) => {
+      return apiVersion ? apiVersion.substring(0, 1) : '0';
+    }),
   },
 };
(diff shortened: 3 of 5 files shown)


src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -22,6 +22,7 @@ export interface ISwaggerDocsProps {
   setInitialVersioning: (url: string, metadata: any) => void;
   setRequestedApiVersion: (version: string) => void;
   version: string;
+  versionNumber: string;
```
bastosmichael: won't this require adding an additional string to all of the configs so that if versionNumber doesn't exist then it would be blank? Might it not be better to just round up the last version so that 0.0.1 rounds up to v0 and 1.0.0 rounds up to v1 instead? This way no one has to add a new config option to each new API etc.

whuang85: I'm not sure I understand your concern.. What configs are you talking about?  The `versionNumber` here is a derived property from the metadata.

whuang85:
> Makes sense to me in general. Is there a way to add a test case for the original bug so it doesn't happen again?

Hmm, I've been thinking about it and trying to figure out a way to test to make sure it doesn't happen again.. The only thing I can think of is ensuring I pass back an empty string or null/undefined if the `versionInfo` is not available because the `curlForm` checks if the version that gets passed in is `falsey` then it will be `0`.

whuang85:
> could you explain how this bug occurred in the first place? i see from running it that it fixes the problem, but i'm not sure how. `curlify` is kind of opaque to me, and i'm a little confused about how it handles versions and specifically why replacing current with the empty sting fixes it. i'm gonna keep looking at it, but if you have any background, that could be helpful. thanks Will



> could you explain how this bug occurred in the first place? i see from running it that it fixes the problem, but i'm not sure how. `curlify` is kind of opaque to me, and i'm a little confused about how it handles versions and specifically why replacing current with the empty sting fixes it. i'm gonna keep looking at it, but if you have any background, that could be helpful. thanks Will

Sure, so originally before my version stuff was implemented, we would pass in the version from the versionInfo on the metadata or a blank string to the `versionActions`, in the `CurlForm`, it uses the redux store that Swagger plugin created to get the version and if it's blank, it's set to `0` and if not, it does a `substring(0,1)` on the version.

My implementation, passed in `current` as the version into the `versionActions` and when `CurlForm` parsed it it would get `c` from the substring operation and the version would turn into `vc`

mjlumetta:
@whuang85 cool, thanks, i see how that happened. i think the first thing i'd say is that this won't work for APIs where the current version is higher than 0. so i think that that probably needs to be fixed as part of this PR; that was a fair default to get this feature up and running, but it's not going to work now that API versions are more complex. so i think the goal would be to construct this `v0` (or `v1`, `v2`, ...) in the derived variable from `getVersionNumber`. the derived values from `getVersion` and `getVersionNumber` are a little too similar, and the derived variable from `getVersionNumber` requires further processing in `CurlForm`.

one thing that isn't necessary but i could see being useful would be to derive the current version number from the metadata. so then you could make the logic in `getVersionNumber` slice the `v0` off of either `getCurrentVersion` (or whatever) if `getVersion` returns "current" or the `v0.0.0` version returned by `getVersion` (confusing

another general question i have is about how we manage state in the swagger docs. i'm wondering if, now that we're keeping track of versions in our main redux store, could we retrieve those values from the main store in our swagger docs? i'm curious what the current benefits of managing this state in the swagger store, and those might mean that we wouldn't want to make that type of change. on the other hand, maybe consolidate this state might reduce some of the complexity at play here. i'd imagine that that would be an area for future improvement, unless it's intuitive for you to tie into solving this problem, but i'd love to learn more about how this works.

whuang85:
> @whuang85 cool, thanks, i see how that happened. i think the first thing i'd say is that this won't work for APIs where the current version is higher than 0. so i think that that probably needs to be fixed as part of this PR; that was a fair default to get this feature up and running, but it's not going to work now that API versions are more complex. so i think the goal would be to construct this `v0` (or `v1`, `v2`, ...) in the derived variable from `getVersionNumber`. the derived values from `getVersion` and `getVersionNumber` are a little too similar, and the derived variable from `getVersionNumber` requires further processing in `CurlForm`.
> 
> one thing that isn't necessary but i could see being useful would be to derive the current version number from the metadata. so then you could make the logic in `getVersionNumber` slice the `v0` off of either `getCurrentVersion` (or whatever) if `getVersion` returns "current" or the `v0.0.0` version returned by `getVersion` (confusing
> 
> another general question i have is about how we manage state in the swagger docs. i'm wondering if, now that we're keeping track of versions in our main redux store, could we retrieve those values from the main store in our swagger docs? i'm curious what the current benefits of managing this state in the swagger store, and those might mean that we wouldn't want to make that type of change. on the other hand, maybe consolidate this state might reduce some of the complexity at play here. i'd imagine that that would be an area for future improvement, unless it's intuitive for you to tie into solving this problem, but i'd love to learn more about how this works.

@mjlumetta I haven't actually touched the current `CurlForm` which is defaulting to `0` if a blank/undefined/null version is in the swagger redux store.  What I think I can do is change my new derived variable to be responsible to return the `v#` so that `CurlForm` doesn't have to do the deriving anymore.  Not sure if that is what you were saying.

For managing state in the swagger docs, I'm not exactly sure how the SwaggerPlugin works but I feel like they create an internal store for itself to do what it needs to do.  I'm not sure how we can use our own redux store in place of the SwaggerPlugin's redux store.

mjlumetta:
> @mjlumetta I haven't actually touched the current CurlForm which is defaulting to 0 if a blank/undefined/null version is in the swagger redux store. What I think I can do is change my new derived variable to be responsible to return the v# so that CurlForm doesn't have to do the deriving anymore. Not sure if that is what you were saying.

^ that is exactly what i'm suggesting. that solves the problem, with the assumption that the derivation defaults to the current version and not v0.

> For managing state in the swagger docs, I'm not exactly sure how the SwaggerPlugin works but I feel like they create an internal store for itself to do what it needs to do. I'm not sure how we can use our own redux store in place of the SwaggerPlugin's redux store.

i wouldn't worry about this then. what i'm asking here is whether anything in that store is used by swagger internally or whether it's exclusively consumed by our components. another way of phrasing this question is whether (a) using the plugin store provides any benefits/swagger magic we'd like to take advantage of, or (b) whether it's more of a convenience mechanism for developing self-contained plugins on top of swagger. if it's the latter, then it doesn't really make any difference whether we get it from the plugin store or the main app store, and it would be nice to connect to the main store because it would cut out intermediate steps. in that case, the plugin store would seem more useful for purposes like developing a package that exports one or more swagger plugins, where the plugin code and the application code are separate and independent.

whuang85:
> i wouldn't worry about this then. what i'm asking here is whether anything in that store is used by swagger internally or whether it's exclusively consumed by our components. another way of phrasing this question is whether (a) using the plugin store provides any benefits/swagger magic we'd like to take advantage of, or (b) whether it's more of a convenience mechanism for developing self-contained plugins on top of swagger. if it's the latter, then it doesn't really make any difference whether we get it from the plugin store or the main app store, and it would be nice to connect to the main store because it would cut out intermediate steps. in that case, the plugin store would seem more useful for purposes like developing a package that exports one or more swagger plugins, where the plugin code and the application code are separate and independent.

With the SwaggerPlugin, it does use it's store for things it needs and I'm pretty sure our custom  components that get rendered in the SwaggerUI component require that we communicate with it through the SwaggerPlugin store.  According to this: https://swagger.io/docs/open-source-tools/swagger-ui/customization/overview/ it looks like we can only pass in actions and reducers but I'm not sure how we can do a `connect` to our store.

mjlumetta:
> our custom components that get rendered in the SwaggerUI component require that we communicate with it through the SwaggerPlugin store.

that's the part that i'm proposing we should look into changing - i'm not sure that it actually makes this code simpler. i don't see why we couldn't do a regular old `connect(mapStateToProps, mapDispatchToProps)(CurlForm)` for all of the swagger components, although maybe i'm missing some relevant details. if that's possible, we would still need the `system` object for some things, like `system.spec()`, but we can move actions, reducers, and selectors out of the SwaggerPlugin context. anyway, a conversation for a later time.

mjlumetta:
@whuang85 that's an argument in my mind for using the app store over the plugin store. it's very easy and comprehensible to figure this out in the app store, but the plugin store adds complexity because we have to keep the two stores in sync. you could theoretically use an additional selector in the plugin store, which would simplify `CurlForm` and avoid passing any more data manually, but it wouldn't eliminate the problem of keeping the stores in sync and therefore is less desirable imo. let me know what you're thinking. if it would be easier to discuss synchronously i can be available this afternoon or monday.

src/containers/documentation/swaggerPlugins/VersionSelector.ts
```diff
@@ -1,7 +1,12 @@
+import { createSelector } from 'reselect';
+
 export const VersionSelector = {
   selectors: {
     apiMetadata: (state: any) => state.get('apiMetadata'),
     apiName: (state: any) => state.get('apiName'),
     apiVersion: (state: any) => state.get('apiVersion'),
+    majorVersion: createSelector((state: any) => state.get('apiVersion'), (apiVersion: string) => {
+      return apiVersion ? apiVersion.substring(0, 1) : '0';
```
mjlumetta: is 0 still a safe default here? if not can we make a follow-up ticket for fixing this?

src/containers/documentation/swaggerPlugins/CurlForm.tsx
```diff
@@ -96,13 +96,7 @@ export class CurlForm extends React.Component<ICurlFormProps, ICurlFormState> {
...
       spec.host = this.state.env ? `${this.state.env}-${spec.host}` : spec.host;
     } else {
-      let version = this.props.system.versionSelectors.apiVersion();
-      if (!version) {
-        version = '0';
-      } else {
-        version = version.substring(0, 1);
-      }
-
+      let version = this.props.system.versionSelectors.majorVersion();
```
mjlumetta: can be `const` but not particularly urgent

PR #297, opened 2019-12-19
Redirect unknown apis to explorer

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/3704

If the category key in a developer portal URL is unknown the user will be redirected back to the `explore` page. `/explore/invalid` -> `/explore`

### DIFF
--- a/src/containers/documentation/DocumentationRoot.tsx
+++ b/src/containers/documentation/DocumentationRoot.tsx
@@ -2,10 +2,10 @@ import * as React from 'react';
 
 import classNames from 'classnames';
 import { Flag } from 'flag';
-import { RouteComponentProps } from 'react-router';
+import { Redirect, RouteComponentProps } from 'react-router';
 import { Route, Switch } from 'react-router-dom';
 
-import { getApiCategoryOrder, getApiDefinitions } from '../../apiDefs/query';
+import { getApiCategoryOrder, getApiDefinitions, lookupApiCategory } from '../../apiDefs/query';
 import { IApiCategory, IApiDescription } from '../../apiDefs/schema';
 import SideNav, { SideNavEntry } from '../../components/SideNav';
 import { IApiNameParam } from '../../types';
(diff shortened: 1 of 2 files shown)


src/containers/documentation/DocumentationRoot.tsx
```diff
@@ -93,30 +93,39 @@ function ExploreSideNav() {
 
 export default class DocumentationRoot extends React.Component<RouteComponentProps<IApiNameParam>, {}> {
   public render() {
+    const { apiCategoryKey } = this.props.match.params;
+    const validCategory = !apiCategoryKey || lookupApiCategory(apiCategoryKey) != null;
```
dvogel: This null handling feels strange to me. It seems like the left side of the condition is meant only to avoid sending a null value into `lookupApiCategory`. On my first reading it seemed like `validCategory` meant "it's cool to have a null category" but that makes no sense on it's face. The naming and the null handling together seem disconnected from the fact that the `apiCategoryKey` value  will be ignored if it's null because it will hit the `/explore/` route anyway. Maybe this is the same semantics and clearer?

```
const shouldRouteCategory = apiCategoryKey && lookupApiCategory(apiCategoryKey) != null
```

jasalisbury:
> Could you add a test (unit or e2e, no pref from me) demonstrating this redirect?

There is an e2e test redirecting from `explore/invalid` to `explore`. Do you think I should add more redirecting from `explore/invalid/docs/....` to `explore`?

dvogel:
> > Could you add a test (unit or e2e, no pref from me) demonstrating this redirect?
> 
> There is an e2e test redirecting from `explore/invalid` to `explore`. Do you think I should add more redirecting from `explore/invalid/docs/....` to `explore`?

I wouldn't be opposed to adding that second test. However my original request was based on user error on my part. I'm still so used to tests appearing after code in PRs (due to conventions in past jobs) that I failed to scroll back up to see the test you added.

PR #298, opened 2019-12-26
Use git LFS for visual regression tests.

dvogel:
Github seems to make natively tracked files and LFS tracked files appear the same in the web UI. Here's some evident that these images really are tracked by LFS in this branch:

```
aws✖ dvogel@broom:~/Projects/developer-portal (dv-3459-git-lfs)> git lfs ls-files | head
f808e8d41e * test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-apply-properly-1-snap.png
7357bd62f6 * test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-apply-properly-2-snap.png
763d020772 * test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-apply-properly-3-snap.png
55cd25e833 * test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-appeals-properly-1-snap.png
0ddc810d08 * test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-appeals-properly-2-snap.png
8a8f73baba * test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-appeals-properly-3-snap.png
2c9c932ebc * test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-benefits-properly-1-snap.png
41254b64dd * test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-benefits-properly-2-snap.png
b380e1967b * test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-benefits-properly-3-snap.png
b380e1967b * test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-benefits-properly-4-snap.png
aws✖ dvogel@broom:~/Projects/developer-portal (dv-3459-git-lfs)> git ls-files | grep test/image_snapshots | head
test/image_snapshots/.keep
test/image_snapshots/image_diffs/.keep
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-apply-properly-1-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-apply-properly-2-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-apply-properly-3-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-appeals-properly-1-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-appeals-properly-2-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-appeals-properly-3-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-benefits-properly-1-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-explore-benefits-docs-benefits-properly-2-snap.png
```

I used the `git lfs install --local` method, which seems to make git auto pull LFS files and thus we don't need to make `Jenkinsfile` do it. I haven't been able to verify that explicitly yet but this test run passed: http://jenkins.vfs.va.gov/job/testing/job/developer-portal/job/dv-3459-git-lfs/4/console

I took this opportunity to move the image snapshots to a separate folder that is easier to remember and type. The previous folder location and naming was accepted from `jest-image-snapshot`. This change should help make it easier to determine whether a given test run is pre-LFS or post-LFS. I tried also using a custom image diff dir but that requires a newer version of `jest-image-snapshot` which runs into a type specification incompatibility between a bunch of our libraries that I figured would be best resolved in a separate PR.


dvogel:
> I'm also convinced these files are in git-lfs now tada. UI shows that they're significantly smaller and mentions that they're in git-lfs for me:
> 
> Screen Shot 2019-12-27 at 11 40 28 AM
> 
> The AC in [department-of-veterans-affairs/vets-contrib#3459](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3459) include `Previously committed __image_snapshots__ are migrated to git-lfs`. Does that require rewriting history (and therefore will happen outside of this PR) or am I misinterpreting the requirement?

I was interpreting that AC to mean the existing paths would use LFS going forward -- no need to re-write history since that would be bad form. If the repo size issue was truly untenable I'd be open to re-writing the history. On balance, dealing with the bloat from our past commits seems worth maintaining consistency between commit SHAs and existing releases.

mjlumetta:
> I used the `git lfs install --local` method, which seems to make git auto pull LFS files and thus we don't need to make `Jenkinsfile` do it.

@dvogel could you please explain this further? i'm not exactly following what you mean.

>> The AC in [department-of-veterans-affairs/vets-contrib#3459](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3459) include `Previously committed __image_snapshots__ are migrated to git-lfs`. Does that require rewriting history (and therefore will happen outside of this PR) or am I misinterpreting the requirement?
>
> I was interpreting that AC to mean the existing paths would use LFS going forward -- no need to re-write history since that would be bad form. If the repo size issue was truly untenable I'd be open to re-writing the history. On balance, dealing with the bloat from our past commits seems worth maintaining consistency between commit SHAs and existing releases.

@kalilsn's initial interpretation of that AC was correct according to the ticket and the spike. from the spike by @jasalisbury:

> Remove large files from git. Use `git lfs migrate` to convert previously committed large files https://github.com/git-lfs/git-lfs/wiki/Tutorial#migrating-existing-repository-data-to-lfs. Note this might require a force push because `git lfs migrate` rewrites history.

not offering an opinion on what to do here, because i do see the point of what Drew is saying, and i would probably lean pretty hard in that direction (i guess that is an opinion, but whatever). but i'm 99% sure that the finished version of the spike and the groomed version of the ticket are suggesting using `git lfs migrate` for the purpose of rewriting that history. (as an aside, the Git LFS tutorial appears to suggest not rewriting history on the current remote repository in the [`git lfs migrate` section](https://github.com/git-lfs/git-lfs/wiki/Tutorial#using-git-lfs-migrate): "Push the converted repository as a new repository...".) so i would say we should discuss, given the confusion. i'm def fine with Drew's conclusion, though; it seems way, way easier and would prevent confusion down the road if we don't have that consistency.

dvogel:
> not offering an opinion on what to do here, because i do see the point of what Drew is saying, and i would probably lean pretty hard in that direction (i guess that is an opinion, but whatever). but i'm 99% sure that the finished version of the spike and the groomed version of the ticket are suggesting using `git lfs migrate` for the purpose of rewriting that history. (as an aside, the Git LFS tutorial appears to suggest not rewriting history on the current remote repository in the [`git lfs migrate` section](https://github.com/git-lfs/git-lfs/wiki/Tutorial#using-git-lfs-migrate): "Push the converted repository as a new repository...".) so i would say we should discuss, given the confusion. i'm def fine with Drew's conclusion, though; it seems way, way easier and would prevent confusion down the road if we don't have that consistency.

I'm open to re-writing the history but I'm hoping we can discuss that outside this PR. Trying to re-write history and get to this state with the same PR makes my head spin.

dvogel:
> > I used the `git lfs install --local` method, which seems to make git auto pull LFS files and thus we don't need to make `Jenkinsfile` do it.
> 
> @dvogel could you please explain this further? i'm not exactly following what you mean.

Sure. I was referring to this AC:

> Developer portal build runs git lfs pull in Jenkins after checking out source from git

I figured some reviewers might be looking for a change to `Jenkinsfile` so I was linking to that successful build to explain why there wasn't such a change in the PR, saving folks from wondering if I forgot to commit those changes.

mjlumetta:
@dvogel re: `git lfs migrate` and rewriting history - i agree with that, i wouldn't expect it to be in this PR. meant to clarify what that AC referred to and not to suggest changes to this PR.

with respect to the Jenkinsfile question, i should've been clearer. i was wondering where `git lfs install --local` was used? i didn't see anything in the PR that i would've associated with that other than including it in the description.

dvogel:
> @dvogel re: `git lfs migrate` and rewriting history - i agree with that, i wouldn't expect it to be in this PR. meant to clarify what that AC referred to and not to suggest changes to this PR.
> 
> with respect to the Jenkinsfile question, i should've been clearer. i was wondering where `git lfs install --local` was used? i didn't see anything in the PR that i would've associated with that other than including it in the description.

Thanks for asking again. I see what you mean now. Running that command modified my `.git/config` file as such:

```
[filter "lfs"]
	process = git-lfs filter-process
	required = true
	clean = git-lfs clean -- %f
	smudge = git-lfs smudge -- %f
```

Since that's not committed to the repo, I have no idea how the tests succeeded in the CI job. I'm digging into that further now.

mjlumetta:
@dvogel i'm pretty sure i know why. i previously picked up this ticket and installed Git LFS on Jenkins instances (documented in [this comment](https://github.com/department-of-veterans-affairs/vets-contrib/issues/3459#issuecomment-554047344)). i take it that, either way, the `checkout scm` command in the Jenkinsfile didn't need to be modified in order for Git LFS to work, but i would assume that's why.

also, just an admin q - the ticket looks like it's still in Ready?

dvogel:
> ```
> [filter "lfs"]
> 	process = git-lfs filter-process
> 	required = true
> 	clean = git-lfs clean -- %f
> 	smudge = git-lfs smudge -- %f
> ```
> 
> Since that's not committed to the repo, I have no idea how the tests succeeded in the CI job. I'm digging into that further now.

Once I had a chance to look more closely at this, I see what's happening. Each of those entries in the local git config only affects modifications to the repo. Since the CI job doesn't modify the repo, it doesn't need that config for correct behavior. The reason the CI job was succeeding is because the [ansible build task that installs LFS on the build machines enables it globally](https://github.com/department-of-veterans-affairs/devops/blob/master/ansible/build/roles/common/jenkins-swarm/tasks/main.yml#L65). I originally thought the local installation would cause allow people to avoid a global user install before working on the developer-portal but that's not the case. As such, I've added a note to the docs imploring people to install it.

I've updated this PR to make the git checkout step more explicit about the LFS requirement (as specified in Josh's spike report). The LFS commands now appear in the log. I think the original CI job was succeeding despite the LFS commands not appearing in the log because the global LFS install was causing the commands to run transparently. IOW only the `.gitattributes` in this commit is required to activate LFS during the CI job. So the extra groovy code is probably not necessary. I decided to keep it for the git clean behaviors I added.

One fun trick I learned while investigating this:
```
ls -1sh test/image_snapshots/ | grep visual-regression-test-ts-visual-regression-test-renders-the-header-properly
 28K visual-regression-test-ts-visual-regression-test-renders-the-header-properly-1-snap.png
 28K visual-regression-test-ts-visual-regression-test-renders-the-header-properly-2-snap.png
 12K visual-regression-test-ts-visual-regression-test-renders-the-header-properly-3-snap.png
```
That shows the files after I ran `git lfs pull` and here's the command showing that those full file sizes aren't really in the git repo (the key is the 130 -- which is the byte size of the object):
```
git ls-tree origin/dv-3459-git-lfs -rl | grep visual-regression-test-ts-visual-regression-test-renders-the-header-properly
100644 blob 45f5d804f09d0d40e5d35f6982829c3f46e66813     130	test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-the-header-properly-1-snap.png
100644 blob 2370b233536753ff96a0d8d66c19248a050d4051     130	test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-the-header-properly-2-snap.png
100644 blob f4ec21e6ee1ebf1c29ee166c90c6fe0fcb04ca32     130	test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-the-header-properly-3-snap.png
```

PR #306, opened 2020-01-10
Wh/apply page veteran confirmation

whuang85:
closes https://github.com/department-of-veterans-affairs/vets-contrib/issues/3660


src/containers/ApplyForm.tsx
```diff
@@ -103,8 +107,9 @@ const formFieldsToFragments = {
   communityCare: 'community_care',
   facilities: 'facilities',
   health: 'fhir',
+  vaConfirmation: 'veteran_confirmation',
   vaForms: 'vaForms',
-  verification: ['veteran_confirmation', 'service_history', 'disability_rating'],
+  verification: ['veteran_verification'],
```
edmkitty: This probably doesn't need to be an array

src/reducers/index.spec.ts
```diff
@@ -15,6 +15,7 @@ const app: IApplication = {
       communityCare: false,
       facilities: false,
       health: false,
+      vaConfirmation: false,
```
edmkitty: Maybe this variable should just be `confirmation`

src/reducers/index.ts
```diff
@@ -80,6 +81,9 @@ export function applicationInput(
     case constants.TOGGLE_COMMUNITY_CARE_CHECKED:
       const communityCare = !inputs.apis.communityCare;
       return { ...inputs, apis: { ...inputs.apis, communityCare } };
+    case constants.TOGGLE_VA_CONFIRMATION:
```
edmkitty: This constant should probably be something like `TOGGLE_CONFIRMATION_CHECKED` for consistency

src/apiDefs/query.test.ts
```diff
@@ -67,7 +67,7 @@ describe('query module', () => {
       const verificationApi = lookupApiCategory('verification');
       expect(verificationApi).not.toBeNull();
       expect(healthApi!.apis.length).toBeGreaterThanOrEqual(4);
-      expect(verificationApi!.apiKey).toBe(false);
+      expect(verificationApi!.apiKey).toBe(true);
```
edmkitty: Does this mean the verification api doesn't use an api key? I'm not sure what the `!` does there

whuang85: they have this weird mechanism to allow each api in the category to control if it's oauth or not but apiKey at the category level has to be true for it to work.

src/reducers/index.spec.ts
rtravitz: I think that's probably true based on where I was headed with the lambda. Does `confirmation` read okay?

nfasulo: Agreed. I prefer `veteranConfirmation`, but confirmation matches `verification`.

src/apiDefs/query.test.ts
nfasulo: @edmkitty The `!` after an object is called the non-null assertion operator in TypeScript. `verificationApi!` is the developer to saying that `verificationApi` will not be `null` or `undefined` so TypeScript can stop complaining.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

src/apiDefs/data/categories.ts
```diff
@@ -59,7 +59,7 @@ const apiDefinitions: IApiCategories = {
     properName: 'VA Form API',
   },
   verification: {
-    apiKey: false,
+    apiKey: true,
```
mjlumetta: this is confusing to me (mostly because it's already badly designed). this is used to control the display of the Authorization docs in each category. it used to be the sole determinant of it until Claims v1 introduced a split in the Benefits APIs category. it was actually always that way in the Verification category, prior to these changes, because of Address Validation, but that was internal only, so probably not that big of a deal. 

the root causes of this problem are, imo

* Authorization docs don't really belong to any one category, so it's somewhat unnecessary that they be nested in that way. 
* regardless of what UX is preferred wrt to the previous point, having the authorization mechanism determined by category isn't really accurate.

i think it would be good to determine this exclusively on the individual API level, and then the category switch can be `apis.any(api => api.oauth)`. thoughts on this approach?

src/containers/ApplyForm.tsx
```diff
@@ -202,6 +207,17 @@ class ApplyForm extends React.Component<IApplyProps> {
...
+              <div className="form-checkbox">
+                <input
+                  type="checkbox"
+                  id="confirmation"
+                  name="confirmation"
+                  checked={apis.confirmation}
+                  onChange={props.toggleConfirmation}
+                />
+                <label htmlFor="confirmation">VA Veteran Confirmation API</label>
+              </div>
```
mjlumetta: would you mind refactoring this list of checkboxes to be based on an object and then mapped to this general markup? it seems like they all follow the following pattern:

```
<div className="form-checkbox">
  <input
    type="checkbox"
    id={apiId}
    name={apiId}
    checked={apis[apiId]}
    onChange={changeHandler}
  />
  <label htmlFor={apiId}>{apiLabel}</label>
</div>
```

so can abstract this into a functional component or something and make an object to configure each specific API, if you're up for refactoring that. (sorry, i feel like i'm always asking you to refactor these annoying things on the site...  😬 )

src/types/constants.ts
```diff
@@ -58,4 +58,7 @@ export type SET_REQUESTED_API_VERSION = typeof SET_REQUESTED_API_VERSION;
 export const SET_INITIAL_VERSIONING = 'SET_INITIAL_VERSIONING';
 export type SET_INITIAL_VERSIONING = typeof SET_INITIAL_VERSIONING;
 
-export const CURRENT_VERSION_IDENTIFIER = 'current';
\ No newline at end of file
+export const CURRENT_VERSION_IDENTIFIER = 'current';
+
+export const TOGGLE_CONFIRMATION_CHECKED = 'TOGGLE_CONFIRMATION_CHECKED';
+export type TOGGLE_CONFIRMATION_CHECKED = typeof TOGGLE_CONFIRMATION_CHECKED;
```
mjlumetta: would you mind moving this up to be next to the other API types?

src/apiDefs/data/categories.ts
whuang85: that's what I was originally going to do until I saw the oauthOverride thing.. So this is probably what you are saying but I'm thinking of removing apiKey from `IApiCategories` and then in the `query.ts` change the `includesOauthAPI` to this:

```
function includesOauthAPI(apiList: string[]): boolean {
  return apisFor(apiList).any(api => api.oAuth);
}
```

mjlumetta: that would be amazing! that's exactly what i was thinking. 

here are the places i think depend on `apiKey` elsewhere:

* `ExploreSideNav` in `DocumentationRoot`
* `Routes`
* `AuthorizationDocs`
* `ApplySuccess`

thanks so much, Will

src/containers/ApplyForm.tsx
whuang85: @mjlumetta I refactored the checkbox info into an array of objects (one for oauth and one for apikey) and created a reusuable stateless component inside the ApplyForm but I don't know how I feel about the solution since it seems like I'm adding more code for not that big a difference.

src/apiDefs/query.test.ts
```diff
@@ -52,22 +53,18 @@ describe('query module', () => {
       const benefitsApi = lookupApiCategory('benefits');
       expect(benefitsApi).not.toBeNull();
       expect(benefitsApi!.apis.length).toBeGreaterThanOrEqual(4);
-      expect(benefitsApi!.apiKey).toBe(true);
```
edmkitty: Add expectations for oauth to be false

src/apiDefs/schema.ts
```diff
@@ -32,13 +32,12 @@ export interface IApiDescription {
   readonly description: string;
   readonly enabledByDefault: boolean;
   readonly vaInternalOnly: boolean;
-  readonly oAuth?: boolean;
```
edmkitty: What did the `?` do here?

whuang85: It made it optional but I need to change it to be non optional. Good catch

src/apiDefs/schema.ts
```diff
@@ -32,13 +32,12 @@ export interface IApiDescription {
...
+  readonly oAuth: boolean;
   readonly deprecated?: boolean | moment.Moment;
   readonly deprecationContent?: React.StatelessComponent;
 }
 
 export interface IApiCategory {
-  readonly apiKey: boolean;
```
edmkitty: Do we need to add an entry for `oAuth` here?

whuang85: No because we want oauth to be at the api/endpoint level and not the category level

src/containers/ApplyForm.tsx
```diff
@@ -34,6 +36,13 @@ interface IApplyProps extends IApplication {
   updateOrganization: (value: IErrorableInput) => void;
 }
 
+interface IFormCheckboxProps {
```
edmkitty: What is this? A checkbox?

src/containers/ApplySuccess.tsx
```diff
@@ -78,7 +67,9 @@ function OAuthCredentialsNotice({ clientID, clientSecret, email, selectedApis }
 
 function ApiKeyNotice({ token, email, selectedApis } : IApiKeyNoticeProps) {
   const apiDefs = getApiDefinitions();
-  const apiListSnippet = apisToEnglishList(selectedApis.filter((k) => apiDefs[formFieldsToCategories[k]].apiKey));
+  const apiIds = selectedApis.map(api => FORM_FIELDS_TO_URL_FRAGMENTS[api]);
+  const categoryKeys = new Set(Object.keys(apiDefs).filter(cat => apiDefs[cat].apis.filter(api => apiIds.includes(api.urlFragment) && !api.oAuth).length > 0));
```
edmkitty: seems like a pretty complex line. Maybe we can move this logic to a function?

PR #312, opened 2020-02-12
onboard LGY API

edmangimelli:
resolves https://github.com/department-of-veterans-affairs/vets-contrib/issues/4163

WIP


edmangimelli:
made bad assumption about UX decision (thought we were moving LGY out of Benefits)

PR #329, opened 2020-04-10
remove the period on the homepage VA forms placard

charleystran:
This PR fixes an an inconsistent punctuation in the VA forms placard on the homepage


rtravitz:
@charleystran @bastosmichael Is this something you would still like in? Cleaning up old PRs.

PR #335, opened 2020-04-23
Add PKCE OAuth flow option to apply page

nfasulo:
This PR adds a radio button group to the apply page so that users can select whether to use the OAuth authentication code flow (which all consumers had previously used) or the new PKCE flow.

This work goes along with https://github.com/department-of-veterans-affairs/developer-portal-backend/pull/36.

For https://vajira.max.gov/browse/API-205


nfasulo:
I need modify the new form property name so that the backend gets what it expects.

src/containers/ApplyForm.tsx
```diff
@@ -337,15 +324,37 @@ class ApplyForm extends React.Component<IApplyProps> {
...
       const oAuthRedirectURI = this.props.inputs.oAuthRedirectURI;
       return (
-        <ErrorableTextInput
-          errorMessage={this.props.inputs.oAuthRedirectURI.validation}
-          label="OAuth Redirect URI"
-          field={oAuthRedirectURI}
-          onValueChange={this.props.updateOAuthRedirectURI}
-          required={true}
-        />
+        <div>
```
kalilsn, line 330: This could be a `<React.Fragment>` to avoid the unnecessary `<div>`

nfasulo:
> One question: `Errorable` in `ErrorableRadioButton` is because that's all formation gives us? Having a hard time finding what the error case would be, or where that would get updated.

@rtravitz you are correct: [formation-react](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/visual-design/components/errorableradiobuttons/) only includes errorable versions of input components. I agree there isn't an error case to implement for this.

To add validation to a field you create a validation function to `/utils/validators.ts` which you then wrap around the `newValue` passed to the update action for the field. When the validation fails it sets the value of the `errorMessage` property of the component (`errorMessage={null}` isn't required for this component, but I left it in to match the other input components). A good example of validation can be found in `updateApplicationOAuthRedirectURI`.

PR #338, opened 2020-04-27
Add tests of cached application state

nfasulo:
Added tests for the store to verify that the apply page is only using cached sessionStorage data when appropriate. As the function `loadApplicationState` is not exported I chose to dynamically import the store in each test to reinitialize it after loading the appropriate value into sessionStorage.

### DIFF
--- a/src/store.test.ts
+++ b/src/store.test.ts
@@ -1,9 +1,8 @@
 import { initialApplicationState } from './reducers';
 
-// sessionStorage.getItem('state') is called when the store is imported at application start up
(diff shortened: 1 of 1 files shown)


nfasulo:
@rtravitz `jsdom` does indeed include `sessionStorage`: https://github.com/jsdom/jsdom/blob/e07aac9e66a77e3cd0cd0cf1e413a4989bb9e1af/lib/jsdom/browser/Window.js#L186-L201

PR #339, opened 2020-04-28
Added Facilities release note, additional services and operating status

spetrosky-va:
Added Facilities API release note to document the new features for Facilities API:  the additional services of Podiatry and Nutrition, and the addition of operating status (normal, notice, limited, closed) and additional information flags.


jblefkove-va:
Users may be confused that the most recent `View code change(s)` links no longer go to `vets-api`. Should we also include a note about the extraction to a standalone implementation?

PR #344, opened 2020-05-20
Allow mobile nav menu to scroll for zoomed-in screens

mjlumetta:
completes [API-937](https://vajira.max.gov/browse/API-937).

allows the mobile menu, which also appears on desktop for zoomed-in screens, to scroll when the screen is zoomed in at 400% or greater to comply with WCAG standards. 

* WCAG 2.1 [Success Criterion 1.4.10: Reflow](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) requires that the content be presented in such a way that the user can zoom in at 400% or greater without loss of content and without requiring scrolling in two dimensions. technically, the standard requires these conditions be met for a viewport equivalent to a width of 320 CSS pixels and a height of CSS pixels - in other words, a viewport that is 1280 pixels wide and 1024 pixels high at 100% zoom zoomed in to 400%.

screenshot of the problem before fix:

ZoomedMobileMenu

also renamed some things in `Header`, where `NavBar` is consumed, because the names were annoying/inaccurate.

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -12,15 +12,15 @@ import Search from './Search';
 import './Header.scss';
 import TestingNotice from './TestingNotice';
 
-interface INavBarState {
-  menuVisible: boolean;
+interface IHeaderState {
+  mobileNavVisible: boolean;
 }
 
-export default class Header extends React.Component<{}, INavBarState> {
+export default class Header extends React.Component<{}, IHeaderState> {
   constructor(props: {}) {
     super(props);
     this.state = {
-      menuVisible: false,
+      mobileNavVisible: false,
     };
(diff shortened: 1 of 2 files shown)


mjlumetta:
@rtravitz here are the reproduction steps to test:

* zoom in the browser to 200% or greater. (i am seeing the zoom max out at 300% in firefox and safari, but in chrome i can get up to 500%, without dev tools.)
* open the nav menu and verify that it has its own scrollbar, separate from the main scrollbar.
* click in the nav menu and verify that you can scroll vertically using the arrow keys.

so yeah, i'm not sure what exactly is the cause of that. i'll share a screen recording in slack.

PR #345, opened 2020-05-20
Remove `dev` and `staging` From Curl Generator

edmangimelli:
resolves https://vajira.max.gov/browse/API-960


nfasulo:
Running Jenkins again to generate review instances.

PR #346, opened 2020-05-21
Move OAuth flow links out of radio button group and refactor Apply

mjlumetta:
this PR has two parts:

1. extracting the helper links for the OAuth flows outside of the radio button labels:

OAuthFlowHelperText936

2. refactoring the Apply page

specific parts of the refactor:

* moving Apply-related components into a new directory at `src/containers/developer`.
* using a generic action to toggle APIs on and off instead of a unique action for each API.
* extracting Apply actions into a separate module.
* extracting new components for the developer info, selected APIs, and OAuth info from the big Apply form.

if this PR has too much refactoring in it, i can break it up into smaller chunks.


src/reducers/index.ts
```diff
@@ -53,6 +53,17 @@ export const initialApplicationState: IApplication = {
...
+const applyApis: string[] = [
+  'benefits',
+  'facilities',
+  'vaForms',
+  'confirmation',
+  'claims',
+  'health',
+  'communityCare',
+  'verification',
+];
```
nfasulo, line 65: What about just using the property names from `FORM_FIELDS_TO_URL_FRAGMENTS` in `constants.ts`? Then we wouldn't need an extra arrays here or in the test file that could get out of sync.

mjlumetta: yup, great idea, will do that

mjlumetta:
> My one quibble is with naming the directory developer. Granted I don't much like the name application better. What about apply? Or going long with developerApplication? Not important and I don't mind being outvoted.

@nfasulo fair question. i chose "developer" with the idea that in the future there will be non-Apply pages that relate to the developer (i.e., when we have an authenticated experience). i think either "developerApplication" or "apply" would be fine too, with a preference for "apply" because of length. on a more general note, i would want to eventually eliminate use of "application" for Apply-related code in favor of either "developer application" or "apply" since "application" is ambiguous with application in the sense of a software program.

nfasulo:
@mjlumetta I fully agree that `application` is too confusing a name on its own. I have the same complaint with it in the backend where it refers to the Okta application.

PR #347, opened 2020-05-22
Update formation-react to get better focus in Modal 

mjlumetta:
completes [API-170](https://vajira.max.gov/browse/API-170)

previously, we were using the `focusSelector` prop of the `Modal` component incorrectly. that prop is intended to describe _all_ focusable elements in the modal, but we were using it as if it were setting the _initial_ focus. to solve that issue, i removed the `focusSelector` prop.

* i'm thinking about opening a PR to `formation-react` to enable that prop, because i don't think that starting with focus on the close button is very intuitive. maybe i'm wrong about that.

i also upgraded `formation-react` to 5.2.6 because the `Modal` component has had some pretty significant focus management improvements since last summer.

testing instructions:

* tab forward and backward through the crisis line modal. verify that focus lands on the close button, each of the links in the 4 list items, and the VeteransCrisisLine.net link at the bottom.
* ensure that focus never leaves the modal when tabbing through it either forwards or backwards. tab forward through the bottom of the modal and verify that focus returns to the close button. tab backward through the top of the modal and verify that focus returns to the VeteransCrisisLine.net link at the bottom.
* ensure that no matter how you tab through the modal or how you close the modal (escape vs close button), focus always returns to the crisis line button in the site header.
* to test that `formation-react` hasn't changed anything since upgrading, look at the Support Contact Us page and the Apply page. besides `Modal`, most of the components we use from `formation-react` are form controls, and the exception is `AlertBox`, which we only use to display errors with form submission.


mjlumetta:
@rtravitz can we check in to clarify that at some point? it's looking fine for me in all 3 browsers. we definitely do need to support all major browsers. some things to check that could be interfering (although not sure why they'd work in Chrome but not both others):

* double-check you did an `npm install` on this branch
* make sure you have [keyboard accessibility](https://dequeuniversity.com/mac/keyboard-access-mac) enabled for Mac (it's weird, don't ask me why)

PR #349, opened 2020-05-28
Add proposed codeowners

rtravitz:
This PR sets up Interrobang as codeowners for the developer-portal repository with one exception: everyone in `lighthouse-apis` would own the `src/content` directory where release notes, news, and documentation live.

This configuration will create noise for members of `lighthouse-apis`, many of who will not care about documentation changes. If we want everyone to be able to approve their own documentation PRs until UX takes over review in the future, this is the only way. Codeowners does not currently support a negation syntax where we could exclude requiring Interrobang for review of `src/content`. 

Alternatively, Interrobang could serve as codeowners of the entire repository and rubberstamp the currently infrequent documentation changes.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,3 @@
 # Interrobang manages the structural code for developer-portal.
 * @department-of-veterans-affairs/lighthouse-interrobang
 
-# Everyone owns review of the content until UX takes over.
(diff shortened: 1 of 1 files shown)


rtravitz:
Going with the second option of making @department-of-veterans-affairs/lighthouse-interrobang owners of the whole repository. We'll extract `src/content` to another joint review team at some point in the future.

PR #357, opened 2020-06-12
Refactor release notes to be by API and to have tighter structure

mjlumetta:
this PR completes the refactors called for in [API-1083](https://vajira.max.gov/browse/API-1083). it fixes the following issues:

* moves release notes to the API level and makes them required. requiring them ensures that the anchors in the cards at the top of the page and the side nav links actually correspond to something. moving release notes to the API level also has the benefit of making the files smaller and more digestible.
* extracts more of the release note structure into React, which means that there's less burden on the people editing release notes to maintain anchors and such.
* changes API-level headings to `h2`s and note-level headings to `h3`s. for accessibility reasons, there should only be one `h1` per page.

including @spetrosky-va and @careyotto for Health API content and general UX, respectively.


src/apiDefs/deprecated.test.ts
```diff
@@ -1,5 +1,8 @@
 import 'jest';
 import * as moment from 'moment';
+
+// we just need a Markdown component for  our test IApiDescription
```
rtravitz, line 4: Can you elaborate on this comment?

mjlumetta: the schema now requires `releaseNotes` in the `IApiDescription`, so it won't compile without a component here (and this isn't a TSX file, so it's not possible to inline a fake one)

careyotto:
@mjlumetta I'd like to revisit the design. These changes - specifically 1) making the font size of the date smaller and 2) Repeating the API name twice in a row when there is only one API in that category - places too much visual emphasis on the API name rather than the date. This makes it harder to grok at first glance that you're on the release notes section of the website rather than the API documentation section.

Also, there's some naming weirdness around Facilities API vs VA Facilities API (for example)

I think this will be easier to resolve with a call and screenshare than with ticket writing. Let's plan to chat next week.

PR #360, opened 2020-06-23
remove appeals api feature flag

charleystran:
The PR removes the feature flag so that the Appeals API tab and Decision Reviews sub section appear on the production developer portal.


rtravitz:
Are the diffs okay on the snapshots? Can approve if they're as expected and they get updated.

charleystran:
@mjlumetta correct because they are internal only. for the typo issue, I can fix in a follow up PR where we move the current appeals status from the benefits tab to the appeals tab. If you are ok with that

src/content/apiDocs/appeals/appealsReleaseNotes.mdx
```diff
@@ -1,2 +1,2 @@
-### Month 99, 2099
-(v1) These are dummy release notes.  Link to PRs here!
+### June 06, 2020
```
mjlumetta, line 1: should this be "6" instead of "06"? (not a blocking question for me)

charleystran: just following the 99, but I can change it if you want. Not a huge deal. I would defer to your teams opinion as you are the owners of the portal

PR #365, opened 2020-06-30
Deprecate Urgent Care

mjlumetta:
completes [API-281](https://vajira.max.gov/browse/API-281). content is from [API-833](https://vajira.max.gov/browse/API-833). see [this Google doc](https://docs.google.com/document/d/10vaULW8R_cIHNg9XuA56cl4DVgD8woe8-uVJhP83IOk/edit#) for finalized copy.

note that this PR is premised on a deprecation date of 7/13/20 and a deactivation date of 7/20/20.

### Urgent Care Deprecation

there should not be any visible changes to the application by default. to test the notices, you must run the application locally and manually edit the deprecation/deactivation dates.

for those who may not have run the application locally, you can run the application with the following steps:

1. navigate to the root of the developer portal repo on your machine.
2. run `npm install`.
3. run `npm start`. the portal should open up on `localhost`, with a default port of 3000, so you can navigate to `http://localhost:3000`. the portal should automatically open in the browser when you run `npm start`.

to test the notices, go to `src/apiDefs/data/health.ts`. find the section for Urgent Care; it is the second object in the `healthApis` array. locate the `deactivationInfo` field; it's the first field in the definition object.

to test deprecation, set `deactivationInfo.deprecationDate` to a date in the past. navigate to `/explore/health/docs/urgent_care`. check that the deprecation alert box shows up at the top of the page with the expected message and that the documentation is still available underneath. the Urgent Care API should still be navigable in the side nav and on the Health category page. check that the Urgent Care release notes are still available at `/release-notes/health` - in the side nav, in the cards at the top of the page, and the notes themselves in the page.

to test deactivation, set `deactivationInfo.deactivationDate` to a date in the past. ensure that the Urgent Care API is no longer available in the side nav or on the Health category page. then navigate to `/explore/health/docs/urgent_care` and check that the expected deactivation notice is at the top of the page, with no documentation below. for release notes, navigate to `release-notes/health` and check that the Urgent Care release notes _are not_ available at `/release-notes/health` but _are_ available at `/release-notes/deactivated`. the Deactivated APIs page should include subnavs in the side nav, card links, and release notes content.

### Argonaut Regression Testing

verify that the changes hold up for the Argonaut API. mainly this means verifying that the deactivation notice is still up at `/explore/health/docs/argonaut`, but you can optionally manually set the deactivation date to a date in the future and test that the deprecation notice works as expected. that's not really necessary, though.

because the release notes for deactivated APIs are now available, you should see the Argonaut ("Veterans Health API (Legacy)") release notes at `/release-notes/deactivated` in all the same places as the deactivated Urgent Care notes (side nav, cards, release notes themselves). for product/UX folks, if there are updates you would like to make to those release notes since they haven't ever been displayed previously, please note that.

### Tabbed APIs Intro Blurb

i did a small refactor to move the text outside of the tabbed section on the FHIR page from the category level to the API level where it logically belongs. to test, go to the FHIR page and check that the text between the heading and the tabbed documentation is (a) present and (b) the same as it is in prod.

### FHIR R4 Intro

the FHIR R4 intro paragraph should no longer be present. this paragraph:

> This CoverageEligibilityResponse resource is compliant with FHIR version R4 standards; more information regarding VA’s implementation of this resource can be found in the Urgent Care API.

should no longer be present at the top of the FHIR R4 tab.

### Release Notes

release notes for deactivated APIs are now available at `/release-notes/deactivated`. check that this page mimics release notes pages for true categories, that deactivated APIs are all available on this page, and that deactivated APIs are _not_ available on their category's page.


src/apiDefs/data/health.ts
```diff
@@ -93,10 +74,29 @@ const healthApis : IApiDescription[] = [
...
+    ],
+    enabledByDefault: true,
+    name: 'Urgent Care Eligibility API (FHIR)',
+    oAuth: true,
+    releaseNotes: UrgentCareReleaseNotes,
+    removalInfo: {
+      deprecationContent: UrgentCareDeprecationNotice,
+      deprecationDate: moment('13 Jul 2020 00:00 EDT'),
+      removalContent: UrgentCareRemovalNotice,
+      removalDate: moment('20 Jul 2020 00:00 EDT'),
```
rtravitz, line 92: This is nice and understandable 👍

src/apiDefs/data/health.ts
```diff
@@ -106,6 +106,13 @@ const healthApis : IApiDescription[] = [
     name: 'Veterans Health API (Legacy)',
     oAuth: true,
     releaseNotes: ArgonautReleaseNotes,
+    removalInfo: {
+      deprecationContent: HealthArgonautDeprecationNotice,
+      // see the RFC 2822 date format section here: https://momentjs.com/docs/#/parsing/string-format/
+      deprecationDate: moment('15 Sep 2019 00:00 EDT'),
+      removalContent: HealthArgonautRemovalNotice,
+      removalDate: moment('01 Oct 2019 00:00 EDT'),
+    },
```
rtravitz, line 115: Are we ever going to totally remove fully removed APIs from the code and send the old routes to a default page? There will be extra noise in the code as more APIs get deprecated in the future if we leave all of the old ones after removal.

src/containers/documentation/ApiPage.tsx
```diff
@@ -13,13 +13,23 @@ import { IApiNameParam } from '../../types';
...
+  const isDeprecated = isApiDeprecated(api);
+  const isRemoved = isApiRemoved(api);
+
+  if (!isDeprecated && !isRemoved) {
+    return null;
+  }
+
+  const content = isRemoved 
+    ? api.removalInfo!.removalContent 
+    : api.removalInfo!.deprecationContent;
```
rtravitz, line 25: Having trouble understanding the `!` after `removalInfo`. What's that syntax?

src/apiDefs/data/health.ts
mjlumetta: what i'm anticipating is that once we move to a database this will no longer be an issue. it will add some bloat in the meantime, but my thought is that it's part of the larger problem of making API definitions scaleable

src/containers/documentation/ApiPage.tsx
mjlumetta: it's Typescript syntax to tell the compiler that `removalInfo` is non-null. the calls to `isApiRemoved` and `isApiDeprecated` confirm that `removalInfo` is defined, but the compiler can't infer that. here's [a StackOverflow post](https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci) and [Typescript docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) explaining it.

i prefer not to use it in general when possible, so if you have suggestions for how to massage this function to eliminate it, by all means please make them. but it is useful when used sparingly (here's [an example](https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/containers/documentation/ApiDocumentation.tsx#L60) of where we already use it).

rtravitz: TIL!

rtravitz:
> Visual regression testing failed. Review these diffs and then [update the snapshots](https://github.com/department-of-veterans-affairs/developer-portal/blob/master/docs/testing.md#visual-regression-testing).

ohno dot gif. Is this the flapping regression issue, or something else, I wonder?

mjlumetta:
> ohno dot gif. Is this the flapping regression issue, or something else, I wonder?

i flipped the order of Urgent Care and FHIR, so the Health category page looks different. maybe i should flip that back

mjlumetta:
> I just have a question about whether we should be using the word "deprecated" or "retired."

@nfasulo asked @darduda about that today and confirmed that the terminology is "deprecated" and "deactivated".

nfasulo:
Perfect because GitHub appears to have eaten my comment.

mjlumetta:
@nfasulo @rtravitz this is the new bit:

### Release Notes

release notes for deactivated APIs are now available at `/release-notes/deactivated`. check that this page mimics release notes pages for true categories, that deactivated APIs are all available on this page, and that deactivated APIs are _not_ available on their category's page.

PR #367, opened 2020-07-01
Add PR template and move CODEOWNERS to github directory

rtravitz:
This PR adds a PR template and moves the `CODEOWNERS` file into the `.github` directory. A PR template will help contributors to the repository understand development guidelines as the program grows.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
+### Description
+<!-- 
+What is your PR doing and why? Please link a ticket and any other relevant
+relations like Slack threads or Sentry issues.
+
+Make sure you have read the development guide: https://github.com/department-of-veterans-affairs/developer-portal/blob/master/docs/development.md
```
mjlumetta, line 6: i would also include here that more specific guides exist in https://github.com/department-of-veterans-affairs/developer-portal/blob/master/docs/guides. currently this is only CSS, but i don't want to overload `development.md` with everything we want to cover.

related: should we just remove the `guides` directory and flatten it out into `docs`? might make it easier to find.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+-->
+
+### Process
+<!-- 
+All new components should have associated unit tests, at minimum. If your PR is modifying any
+existing component with little or no testing can you improve the testing in that corner of the codebase?
+
+Is your change something that will cause an update in the documentation?
+
+If neither of these checks are relevant to your PR you can delete them. But for the testig box specifically
```
mjlumetta, line 16: "testing" typo

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+What is your PR doing and why? Please link a ticket and any other relevant
+relations like Slack threads or Sentry issues.
+
+Make sure you have read the development guide: https://github.com/department-of-veterans-affairs/developer-portal/blob/master/docs/development.md
+-->
+
+### Process
+<!-- 
+All new components should have associated unit tests, at minimum. If your PR is modifying any
+existing component with little or no testing can you improve the testing in that corner of the codebase?
```
mjlumetta, line 12: i would also add something about manual testing here? 

also, this is definitely necessary, but i feel like it isn't very useful without more detailed testing guidelines. we've got those tickets coming up soon, though

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+
+Is your change something that will cause an update in the documentation?
+
+If neither of these checks are relevant to your PR you can delete them. But for the testig box specifically
+it may be helpful to leave an explanation about *why*.
+-->
+- [ ] Tests added or updated for change
+- [ ] Docs updated
```
mjlumetta, line 20: i would also add something about accessibility here. i sort of doubt that anyone will carefully think that through when opening a PR, but if we then hit them with feedback about making components accessible, it shouldn't be a surprise

.github/PULL_REQUEST_TEMPLATE.md
rtravitz: TESTIG

(will fix 👍)

.github/PULL_REQUEST_TEMPLATE.md
rtravitz: The guides are linked from `development.md`. I think `development.md` being a place that links out to other guides is a nice spot—it could also have that "I want to" table at some point.

.github/PULL_REQUEST_TEMPLATE.md
rtravitz: Can you say more about manual testing?

And I agree that it will be better with clearer guidelines. But making the expectation "you will add tests for things you touch" explicit is a positive change, IMO, even in cases where they may need to ask what kinds of tests to add because we haven't made that information clear enough yet.

.github/PULL_REQUEST_TEMPLATE.md
rtravitz: This is an excellent point. And probably warrants an accessibility guide at some point, too!

.github/PULL_REQUEST_TEMPLATE.md
mjlumetta: okay! sounds good to me

.github/PULL_REQUEST_TEMPLATE.md
mjlumetta: absolutely. it would be great to write one of those whenever it's convenient

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+
+Make sure you have read the development guide: https://github.com/department-of-veterans-affairs/developer-portal/blob/master/docs/development.md
+-->
+
+### Process
+<!-- 
+All new components should have associated unit tests, at minimum. If your PR is modifying any
+existing component with little or no testing can you improve the testing in that corner of the codebase?
+
+Is your change something that will cause an update in the documentation?
```
mjlumetta, line 14: might be good to clarify that this is referring to docs for working with the developer portal and not the API docs hosted on the portal, because i think some people will associate "documentation" with "API documentation" when they read this. for a lot of people on the program, that's their primary concern, if not their only concern, related to the portal.

.github/PULL_REQUEST_TEMPLATE.md
mjlumetta: i guess i would want to avoid giving the impression that manual testing is unimportant by omission. i'm sure that most people will do some manual testing, but it might be good to encourage them to make sure that testing is thorough, and asking them to write out manual testing steps might push them in that direction. it's easy to spot-check changes and miss a potential issue.

examples:
* real one: Dash added Appeals to the portal, but they didn't look at the release notes section, so Appeals had a dummy release note dated "Month 99, 2099".
* hypothetical: testing on a desktop browser but not testing on mobile/with a narrower viewport.

another angle: talking about unit tests, or any automated tests, is targeting a developer audience. there are still a fair number of people interacting with the portal repo who are making purely content changes because we have so much content defined in the repo. we should probably have the template take those PRs into account. we could also potentially even explicitly call out those different audiences here, since they're likely to have different processes. 

for example - and this isn't exactly manual testing - should copy changes be approved _before_ opening a PR? dunno, but maybe we could suggest that copy has been reviewed by at least one other person before opening the PR, or something like that.

this comment got way longer than i intended, but hope it's useful in some way.

rtravitz: It's definitely useful. It had me thinking about multiple templates, one of which would be used for `src/content`. But there are still some Typescript files that may need testing in that directory. Chewing on it for a bit...

.github/PULL_REQUEST_TEMPLATE.md
rtravitz: Realizing when passing back through that I'm unsure what this checkbox would say before we actually have guidelines. The code goes through some `axe` checks automatically. I'm inclined to not add this for now, understanding that this template will continue to evolve.

.github/PULL_REQUEST_TEMPLATE.md
rtravitz: Good call—clarified!

.github/PULL_REQUEST_TEMPLATE.md
rtravitz: I think I keep coming back to manual checks being a symptom of the larger testing problem. Promoting manual checks means that we don't have good enough e2e tests for structural changes. And was missing a release note indicative of not reviewing the snapshot diffs?

mjlumetta: i think that's totally sufficient. for the Dash release note thing, we don't have visual regression tests for individual API release note pages; it's tough to say what the best way to catch that would be now (maybe just "don't add dummy content"). for the rest, we can think about it and deal with those concerns later.

.github/PULL_REQUEST_TEMPLATE.md
mjlumetta: i understand that it would be tough to recommend concrete steps at this point. i would say we can do something generic like "Accessibility concerns have been considered and addressed"

PR #371, opened 2020-07-02
Reorganize package.json

rtravitz:
### Description

This PR supports [API-1322](https://vajira.max.gov/browse/API-1322). 

#### Notable changes
- Moves packages that are only used to build and test the application down to the `devDependencies` section of `package.json`. 
- Moves the `scripts` section to the top to match the default location after an `npm init`. 
- Moves babel configuration to its own file to make it easier to see at a glance which babel plugins are used when running a global search. The `package.json` will show up for plugins regardless of whether the package is used as part of the build or not.

#### What this PR does NOT do
- Add or remove any packages or scripts. Substantive changes will come in future PRs. This PR only exists to organize and take stock of which dependencies the app is currently using.

### Process

No changes are made to the functionality, and no tests need to be added or changed as a result.

- [x] Docs updated
There are no obvious documentation changes to make.

There are no accessibility concerns with these changes.


package.json
```diff
@@ -85,80 +59,106 @@
...
     "jest": "^25.5.4",
+    "jest-axe": "^3.2.0",
     "jest-environment-puppeteer": "^4.1.1",
     "jest-fetch-mock": "^3.0.3",
     "jest-image-snapshot": "^3.1.0",
     "jest-junit-reporter": "^1.1.0",
+    "jest-puppeteer": "^4.1.1",
+    "js-yaml-loader": "^1.2.2",
     "lodash.noop": "^3.0.1",
+    "markdown-component-loader": "github:adhocteam/markdown-component-loader#dv-3146-from110",
```
mjlumetta, line 129: we should uhh figure out why this is and try to do it in a sustainable way. not sure if that fits in this ticket or not, but it shouldn't hold up this PR.

rtravitz: Oh, there's a ticket for it 😏 @nfasulo

nfasulo: :fork_and_knife: I'm not seeing a corresponding issue in Jira but here is the original ticket for this kind of cleanup https://github.com/department-of-veterans-affairs/vets-contrib/issues/1277.

I found PR Drew created to update the original package: https://github.com/ticky/markdown-component-loader/pull/179/files

rtravitz: Forked libraries is the one I meant, of course!

PR #375, opened 2020-07-07
Change Google Search Console owner

nfasulo:
### Description

This PR removes Jeff Dunn and adds myself as an owner of the Google Search Console for the developer portal. The Google Search Console allows us to look at the searches performed by search.usa.gov through through the search bar on the developer portal. Google verifies ownership by checking an html file that is included in builds of the developer portal. More info on this verification can be found [here](https://support.google.com/webmasters/answer/9008080#html_verification).

Jeff has also added @charleystran and @bastosmichael as owners. Their verifications can be added in the same way.

PR that added Jeff: https://github.com/department-of-veterans-affairs/developer-portal/pull/71

### DIFF
--- a/config/webpackDevServer.config.js
+++ b/config/webpackDevServer.config.js
@@ -87,7 +87,7 @@ module.exports = function(proxy, allowedHost) {
       // folder will receive the generated index.html contents.
       rewrites: [
         { from: /favicon.ico/, to: '/favicon.ico' },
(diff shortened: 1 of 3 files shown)


nfasulo:
>That hash is generated in the console?
Yes, the console generates the HTML file for you. Once it is deployed I will then have to go back to the console to force verification. At that point I should be able to add you both as owners for redundancy.

PR #382, opened 2020-07-16
Group related checkboxes in fieldset [API-155]

gileswells:
### Description
Group related checkboxes into a fieldset and update the aria-label to allow for assistive devices to better read the section of the form.
https://vajira.max.gov/browse/API-155

### Process


gileswells:
@rtravitz best thing about Prettier is that it just does the linting for me.

PR #383, opened 2020-07-16
Updates FHIR and Urgent Care release notes to indicate deprecation Urgent Care APIs

bschofield-va:
# Merge after July 20 10:30 E but before 10:30 E July 21

### Description

Updates release notes for UC and FHIR apis.
https://vajira.max.gov/browse/API-1083

### Process

- [ ] Docs updated


rtravitz:
One note: the dev portal is on a daily deploy to production, so if we want this to appear on the 21st we should either merge it before 10:30am Eastern on the 21st or after 10:30am on the 20th. Or we could run an off-cycle deploy, but I'd rather not over some release notes.

src/content/apiDocs/health/urgentCareReleaseNotes.mdx
```diff
@@ -1,3 +1,6 @@
+### July 21, 2020
+The Urgent Care Eligibility API is deprecated and scheduled for deactivation in the 3rd quarter of 2020.
```
mjlumetta, line 2: shouldn't this release note be dated July 13? not sure when you'll be removing the endpoints themselves, but the announcement for deprecation went up last Monday

rtravitz:
One thing that happened was that the timer for deactivating the API kicked in and changed the end result versus what the screenshots showed on master. 
visual-regression-test-ts-visual-regression-test-renders-explore-health-properly-1-diff

src/content/apiDocs/health/fhirAPIReleaseNotes.mdx
```diff
@@ -1,3 +1,7 @@
+### July 21, 2020
+The CoverageEligibilityResponse resource for Urgent Care is deprecated and scheduled for deactivation in the 3rd quarter of 2020.
+We are working on new FHIR R4 resources and will add those in the 3rd quarter of 2020.
```
mjlumetta, line 3: this is the only release note for this API because the resource isn't going away for a while longer, right? or does it need to have both the deprecation and deactivation notes?

PR #386, opened 2020-07-22
Api 1491 add revoke grants documentation

lukemarkwordtlibertyits:


src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -150,6 +150,25 @@ token={your refresh token}&token_type_hint=refresh_token
 
 Clients may only revoke their own tokens.
 
+#### Revoking Grants
+
+A user will only be prompted to consent to a particular client to use their data once; however, it is possible to revoke these `grants` for a specified user and client in the Sandbox and Development environments with the following endpoint:
```
beaugrantham-va, line 155: There is only sandbox and prod from an external developer perspective... I would drop "and Development" here.

src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -150,6 +150,25 @@ token={your refresh token}&token_type_hint=refresh_token
...
+
+{
+	"client_id": {client_id}
+	"email": {test account email}
+}
+```
+ 
+Both the `email` address of the user and the `client_id` of the application must be passed into the body of the request. In the case of a bad request, an error response will be returned with a description of the error.
+
+**Note:** This endpoint is not available in the Production environment and excludes identity provider grants.
```
dbrown-va, line 170: We may want to add a bit more detail on what the identity provider grants are.  Something like "These grants are specifically provided to our authorization server for identity verification only".  Even then it may cause some confusion since these grants are once per test user instead of consumer, so most Sandbox consumers may never see those grants.

dbrown-va:
> Looks good to me. Out of curiosity, who from UX approved this? I'm wondering if we need to help them out being able to approve in GitHub so we have a record of it.

Darcy did via a JIRA subtask plus in Slack.

Agreed that consolidating these reviews to GitHub may be beneficial, especially since that bot generates a deployed version to check.

PR #387, opened 2020-07-23
add onClick handler to NavBar to close after selection is made

stephennovosel:
### Description

PR for [API-175](https://vajira.max.gov/browse/API-175). Closes nav bar when any item has been clicked.

This PR maps the existing onClose event to individual menu item components' onClick handler.

### Process

- [ ] Docs updated

### DIFF
--- a/src/components/NavBar.tsx
+++ b/src/components/NavBar.tsx
@@ -30,7 +30,11 @@ interface INavBarState {
   visibleSubNavs: IVisibleSubNavState;
 }
 
-function DocumentationSubNav() {
+interface IDocuemntationSubNavProps {
+  onClose: () => void;
+}
+
+function DocumentationSubNav(props: IDocuemntationSubNavProps) {
   const apiDefs = getApiDefinitions();
   const apiCategoryOrder = getApiCategoryOrder();
 
@@ -51,15 +55,15 @@ function DocumentationSubNav() {
   return (
     <ul className={classNames('va-api-sub-nav', 'vads-u-margin-y--0', 'vads-u-padding-left--0')}>
       <li key="all" className={itemStyles}>
-        <NavLink exact={true} to="/explore" className={linkStyles} activeClassName="va-api-active-sub-nav">
+        <NavLink onClick={props.onClose} exact={true} to="/explore" className={linkStyles} activeClassName="va-api-active-sub-nav">
           Overview
         </NavLink>
       </li>
       {apiCategoryOrder.map(apiKey => {
         return (
-          <Flag name={`categories.${apiKey}`} key={apiKey}>
+          <Flag  name={`categories.${apiKey}`} key={apiKey}>
             <li className={itemStyles}>
-              <NavLink to={`/explore/${apiKey}`} className={linkStyles} activeClassName="vads-u-font-weight--bold">
+              <NavLink to={`/explore/${apiKey}`} onClick={props.onClose} className={linkStyles} activeClassName="vads-u-font-weight--bold">
                 {apiDefs[apiKey].name}
               </NavLink>
             </li>
@@ -79,7 +83,7 @@ const navItemStyles = (isFirstChild: boolean = false) => classNames(
   'medium-screen:vads-u-padding-y--0',
   !isFirstChild && ['vads-u-border-top--1px', 'vads-u-border-color--gray-lighter', 'medium-screen:vads-u-border-top--0'],
   );
-  
+
 const navLinkStyles = classNames(
   'vads-u-color--gray-dark',
   'vads-u-display--block',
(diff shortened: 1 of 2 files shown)


src/components/NavBar.tsx
```diff
@@ -30,7 +30,11 @@ interface INavBarState {
   visibleSubNavs: IVisibleSubNavState;
 }
 
-function DocumentationSubNav() {
+interface IDocuemntationSubNavProps {
+  onClose: () => void;
```
mjlumetta, line 34: i would suggest naming this something else, like `on[Mobile]NavClose` or `close[Mobile]Nav`. `onClose` makes sense for `NavBar`, although it's somewhat ambiguous there. for `NavBar`, it could mean either "do this thing when the nav closes itself" or "call this when the nav is supposed to close so that the parent component actually closes it", and in context, it's the latter. i don't think it makes as much sense when it's carried down to a specific part of the nav contents. as an event name, `onMobileNavClose` would emphasize the fact that this is a function to be called when the nav is supposed to close, and as an action name, `closeNav` would emphasize that the click event will trigger the nav's close. open to suggestion on this one, but would prefer steering away from `onClose`.

src/components/MainNavItem.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+  targetUrl: '/mock',
+};
+
+describe('MainNavItem', () => {
+  it('should render the MainNavItem', () => {
+    const Component = shallow(<MainNavItem {...mockProps}>Mock</MainNavItem>);
+    expect(Component).toBeTruthy();
+  });
+
+  it('if mobile, should call the onClick prop when its NavLink child has been been clicked', () => {
```
mjlumetta, line 25: would suggest wrapping this in a `describe()` block for mobile tests, maybe? i think we would expect behavior to roughly be in the categories of mobile, desktop, or both, so that could establish a pattern for additional tests in the future

src/components/MainNavItem.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    // this test is applicable only to mobile
+    const MobileComponentMock = (
+      <Router>
+        <MainNavItem
+          excludeLargeScreen={true}
```
mjlumetta, line 30: couple of things that i would suggest considering:

1. is it worth it to add a test for the desktop version to ensure that `onClick` is _not_ called when you click the desktop version?
2. should we render the nav item with both the large screen and small screen props and use more specific selectors? that would be closer to the use of the component in the app. i don't think we ever actually use `<MainNavItem excludeLargeScreen={true} />`.

i think this coverage is good, so these are more of a nice to have. up to you when you feel it's ready to merge on this front.

src/components/MainNavItem.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    onMouseEnter: jest.fn(),
+    onMouseLeave: jest.fn(),
+  },
+  onClick: jest.fn(),
+  targetUrl: '/mock',
+};
+
+describe('MainNavItem', () => {
+  it('should render the MainNavItem', () => {
+    const Component = shallow(<MainNavItem {...mockProps}>Mock</MainNavItem>);
```
mjlumetta, line 21: any reason not to use `mount` here? might as well test the whole tree from this point, esp since it's not very large anyway and this test is just checking for thrown errors, basically. but this kind of raises a question as we go forward with improving our test coverage: do we prefer shallow or full DOM rendering when it's ambiguous as to which is better in that case? gonna tag in @rtravitz @nfasulo and @gileswells since this is sort of a philosophy question

src/components/MainNavItem.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+          excludeLargeScreen={true}
+          {...mockProps}
+        >
+          Mock
+        </MainNavItem>
+      </Router>
+    );
+    const wrapper = mount(MobileComponentMock);
+    wrapper.find("NavLink").simulate('click');
+    expect(mockProps.onClick).toHaveBeenCalled();
```
mjlumetta, line 39: call `unmount()` when finished?

src/components/MainNavItem.test.tsx
stephennovosel: I think for purely testing that the component itself is rendered and truthy, we won't need to mount the entire Component tree. In theory, if each component level tests itself for truthness, we wouldn't need to fully mount the component within the test as we will know that the child will also be truthful (assuming tests pass). Additionally, we wouldn't want a test to fail on this component because of an error thrown somewhere down the tree, as that error should fail a test specific to the failing component. For component unit testing, i would say that we should stick to shallow rendering unless otherwise needed, unless someone else has some suggestions @rtravitz @nfasulo @gileswells

src/components/MainNavItem.test.tsx
stephennovosel: Technically the MainNavItem never unmounts even though the nav bar will close, so we won't need to run unmount here. As a side note, we will really only need to use unmount from enzyme if we are testing some side effect of the unmount lifecycle

src/components/MainNavItem.test.tsx
nfasulo: My inclination would be to use `shallow` where we can and leave mount for test where we need to verify child elements.

PR #390, opened 2020-07-27
API-1467 Add markdown-it-attrs and override duplicate IDs

gileswells:
### Description
Updated the various release note pages to not have overlapping ids. `markdown-it-anchor` is placing them automatically and will typically de-duplicate but because we have multiple markdown files on a single page it doesn't de-dupe across the different files causing the issue. `markdown-it-attrs` allows us to set the id manually for items that would typically conflict.

Also updated the few dates that weren't following the same pattern as others. (2 digit day vs 1 digit and removed ordinal suffixes)

### Process


gileswells:
@nfasulo docs updated with note about IDs

config/webpack.config.prod.js
```diff
@@ -310,11 +306,12 @@ module.exports = (envName) => {
                   options: {
                     enabledMarkdownItRules: ['smartquotes', 'table'],
                     markdownItPlugins: [
+                      [require('markdown-it-attrs')],
```
mjlumetta, line 309: would think that we'd probably want to add an attribute restriction to "id" as documented in the [Security](https://github.com/arve0/markdown-it-attrs#security) section of the docs. so `{ allowedAttributes: ['id'] }`. probably in dev too but prod would be good

docs/release-notes.md
```diff
@@ -30,6 +30,9 @@ Example:
 All release notes in the file should be separated by a horizontal rule (`---`). There
 shouldn't be one after the final note.
 
+If there are multiple release notes on the same day that appear on the same release notes page (even if they span multiple
+markdown files) you must add a custom ID to one to prevent duplicate IDs. Example `{#february-13%2C-2019-2}`
```
mjlumetta, line 34: i'd suggest a few additional things:

1. make the id format a little more human-readable. it's not really necessary to include the URL-encoded comma, and it could be beneficial to replace "-2" with "-[apiIdentifier]". that way, people adding release notes don't need to worry about how many other ids are on the page, just that there are any.
2. include the format with placeholders, like `{#[date]-[apiIdentifier]}`. mostly that's to emphasize that the "#" is the id symbol.
3. could be useful to direct people to install the aXe plugin ([Chrome](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd), [Firefox](https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/)) and run it after adding a release note. that way, they don't have to check manually, and it also forces them to know how to use the aXe extension as an added bonus.

this is pretty small, these are not super essential but could be useful enhancements

src/content/apiDocs/benefits/benefitsIntakeReleaseNotes.mdx
```diff
@@ -35,12 +35,12 @@
 
 ---
 
-### June 3, 2019
+### June 3, 2019 {#june-3%2C-2019-2}
```
mjlumetta, line 38: same thing from above about making the id more human-friendly, if you want to go with that, for all release notes with ids

gileswells:
@mjlumetta Addressed your PR comments.

config/webpack.config.dev.js
```diff
@@ -252,11 +252,30 @@ module.exports = {
...
+                      require('markdown-it-attrs'),
+                      {
+                        allowedAttributes: ['id'],
+                      },
+                    ],
+                    [
+                      require('markdown-it-anchor'),
+                      {
+                        level: 2,
+                        slugify: s =>
```
mjlumetta, line 265: nice, good call on the custom function

PR #393, opened 2020-08-03
API-135 Added aria-labels to API and Bearer Token fields

gileswells:
### Description
Added `aria-label` to both the API key and Bearer Token fields for better s

### Process


src/containers/documentation/swaggerPlugins/CurlForm.tsx
```diff
@@ -206,6 +206,7 @@ export class CurlForm extends React.Component<ICurlFormProps, ICurlFormState> {
           <h3> API Key: </h3>
           <div>
             <input
+              aria-label="Enter API Key"
```
mjlumetta, line 209: would prefer `aria-labelledby` referencing the h3 above, so that the visual and auditory experiences are linked together

PR #394, opened 2020-08-03
API-1322 upgrade various core packages and fix breaking changes

stephennovosel:
### Description

Upgrades to various packages within the developer-portal and fixes to breaking changes associated to packages

### Requested feedback

Fixes #


tsconfig.json
```diff
@@ -9,7 +9,6 @@
     "allowJs": true,
     "jsx": "react",
     "moduleResolution": "node",
-    "rootDir": "src",
```
mjlumetta, line 12: why was this removed?

stephennovosel:
@mjlumetta good catches, let me look over this and add those changes to the branch.

tsconfig.json
stephennovosel: The jest upgrade started throwing an error for referring to modules used outside of the rootDir scope. Not 100% sure why, but this was able to remove that error without breaking anything else

rtravitz:
PR is closing to perform upgrades separately and 🤞 reduce the complexity of each individual PR.

PR #396, opened 2020-08-04
Update README

ryanlink:
Update `dev-api.vets.gov` to `sandbox-api.va.gov` in README .env.local

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,8 @@ npm install
 Next, create a `.env.local` file with the following contents (don't worry about what these mean yet):
 ```
 PORT=3001
-REACT_APP_VETSGOV_SWAGGER_API=https://dev-api.vets.gov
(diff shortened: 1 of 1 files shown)


charleystran:
@mjlumetta You are correct on this, but I do believe a change is in order to move dev-api.va.gov over dev-api.vets.gov. I think there may have been some confusion between portal developers and api consumers.

mjlumetta:
@charleystran definitely, would be good to get that updated. thanks for making that change.

ryanlink:
> can you share a little bit of what issues or concerns motivated these changes? i don't think we want to make these specific changes, but want to make sure we're addressing the root problem. 🙂

sure!  an Oddball employee was onboarding and got a CORS error running the dev portal locally.  he was using what the README currently says for his .env.local

@charleystran says it should be dev-api.va.gov.  "Given these settings are actually for development and not for API testing dev is the right use case"

PR #397, opened 2020-08-04
API-1790 Add deactivated API category for urgent care API

stephennovosel:
### Description

https://vajira.max.gov/browse/API-1322

adds a deactivated APIs category to the documentation and release notes pages with information on relevant APIs.  

### Process


src/content/apiDocs/health/urgentCareDeactivationNotice.mdx
```diff
@@ -1,3 +1,4 @@
-Our Urgent Care Eligibility API was deactivated on July 20, 2020. The API and related 
-documentation is no longer accessible. For more information, [contact us](/support/contact-us) 
-or visit our [Google Developer Forum](https://groups.google.com/forum/m/#!forum/va-lighthouse).
+The Urgent Care Eligibility API was deactivated on July 10, 2020 and is no
+longer accessible. We are working on new FHIR R4 resources, which will be
+added to the API [Documentation](/explore) page in the 3rd
+quarter of 2020.
```
mjlumetta, line 4: why was this changed?

src/content/apiDocs/deactivated/index.ts
```diff
@@ -0,0 +1,14 @@
+import { IApiCategoryContent } from '../../../apiDefs/schema';
+import DeactivatedIntro from './DeactivatedIntro';
+import deactivatedOverview from './deactivatedOverview.mdx';
+
+const deactivatedContent: IApiCategoryContent = {
+  intro: DeactivatedIntro,
```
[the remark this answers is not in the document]
stephennovosel: makes sense to me. I opted for a component because of the design which had the VA design component "AlertBox". Do you know of a way to achieve that with markdown?

src/content/apiDocs/health/urgentCareDeactivationNotice.mdx
stephennovosel: I changed this to match the text in the design. We can change it back to the original but I wasn't sure if the text was on the design for a specific reason.

src/containers/releaseNotes/CategoryReleaseNotesPage.tsx
```diff
@@ -29,18 +51,22 @@ export default class CategoryReleaseNotesPage extends React.Component<
...
 > {
   public render() {
-    const apiDefs = getApiDefinitions();
     const { apiCategoryKey } = this.props.match.params;
-    const { apis } = apiDefs[apiCategoryKey];
+
+    const isDeactivatedPage = apiCategoryKey === 'deactivated';
+
+    const apiDefs = isDeactivatedPage ? getDeactivatedCategory() : getApiDefinitions();
+    const { apis, name: categoryName } = apiDefs[apiCategoryKey];
```
mjlumetta, line 59: going to offer an ~~ opinionated ~~ take about the pattern above: deactivated APIs should be its own component. i understand the impulse to reuse the same component, but sometimes it's possible to err too far on the side of following Don't Repeat Yourself. adding a special case to an existing component can sometimes make the component harder to understand, and it's harder to extend if you get other special cases. 

instead, you can take smaller pieces that you're more confident are shared and have them accept more data as props. then, you can write two container components, one for a regular category and one for deactivated APIs. each of those would depend on one or more shared components that would accept things that fork based on whether the category is real or deactivated as props. so for example, you'd probably want to extract the following pieces as props in this case:

* `categoryKey`
* `categoryDefinition`
  * `apis`
  * `categoryName`
* `flagName`

aaand that's it, i'm pretty sure? anyway, you'd end up with "two smaller top-level, opinionated components that share one smaller, unopinionated component" instead of "one big component that is opinionated about its contents".

if you'd like, i can share resources from where i learned about this guideline. it kind of blew my mind when i learned it - my reaction was pretty much "that makes a ton of sense" followed by "wow, i do that all the time". that may not be a universal experience, but i think it applies in this case.

very long explanation here. in any case, let me know if you'd like to talk this through synchronously, i know that was a lot.

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -54,11 +58,19 @@ function SideNavCategoryEntry(apiCategoryKey: string, apiCategory: IApiCategory)
     });
   };
 
+  const sideNavEntry = (
+    <SideNavEntry to={`/release-notes/${apiCategoryKey}`} name={apiCategory.name}>
+      {subNavLinks()}
+    </SideNavEntry>
+  );
+
+  if (apiCategoryKey === 'deactivated') {
```
mjlumetta, line 67: similarly to my comments in `CategoryReleaseNotesPage`, i would recommend not adding a special case and inlining this below where it's instantiated for deactivated APIs.

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -43,7 +47,7 @@ function SideNavApiEntry(api: IApiDescription) {
   );
 }
 
-function SideNavCategoryEntry(apiCategoryKey: string, apiCategory: IApiCategory) {
+function SideNavCategoryEntry(apiCategoryKey: string, apiCategory: IApiBaseCategory) {
   const subNavLinks = () => {
     return apiCategory.apis.map(api => {
       if (apiCategory.apis.length > 1) {
```
mjlumetta, line 53: given my comment below suggesting not to reuse `SideNavCategoryEntry` for deactivated APIs, i'd also like to complain that this function not written by Steve feels way more complicated than it needs to be. i'd suggest replacing it with something like the following both places it needs to be used:

```
{apis.length > 1 && apis.map(api => SideNavApiEntry(api))}
```

Steve, if you can clean that up while you're in here, that'd be awesome.

src/content/apiDocs/deactivated/DeactivatedIntro.test.tsx
```diff
@@ -0,0 +1,14 @@
+import AlertBox from '@department-of-veterans-affairs/formation-react/AlertBox';
```
mjlumetta, line 1: to remove still

src/content/apiDocs/deactivated/DeactivatedIntro.tsx
```diff
@@ -0,0 +1,13 @@
+import AlertBox from '@department-of-veterans-affairs/formation-react/AlertBox';
+import * as React from 'react';
+
+const DeactivatedIntro = () => {
```
mjlumetta, line 4: need to move to `src/containers`

src/content/apiDocs/deactivated/index.ts
```diff
@@ -0,0 +1,14 @@
+import { IApiCategoryContent } from '../../../apiDefs/schema';
```
mjlumetta, line 1: to remove

src/content/apiDocs/health/urgentCareDeactivationNotice.mdx
mjlumetta: yeah, it should be changed back to the original. that may have been confusing, sorry about that

src/containers/releaseNotes/CategoryReleaseNotesPage.tsx
```diff
@@ -12,10 +14,30 @@ import PageHeader from '../../components/PageHeader';
...
+      );
+    }
+    return null;
+  };
+
+  const flagName = apiCategoryKey === 'deactivated' ? `deactivated_apis.${api.urlFragment}` : `hosted_apis.${api.urlFragment}`;
   return (
-    <Flag name={`hosted_apis.${api.urlFragment}`}>
+    <Flag name={flagName}>
+      {renderDeactivatedNotice()}
```
mjlumetta, line 40: this should go inside the div under the h2. i think this is the only visual issue left to fix

rtravitz:
@mjlumetta We figured out @stephennovosel's issues with `git lfs` and the snapshots. But unclear whether it will be workable for this PR. Just a heads up that this may need to close and open a new one if this one stays polluted with changing every snapshot.

stephennovosel:
I will resolve all of the comments and once we feel that it is ready to merge I will close this PR and open a new one so that we can correctly update the snapshots so that Mike won't have to go through it again.

src/containers/releaseNotes/CategoryReleaseNotesPage.tsx
```diff
@@ -68,14 +94,20 @@ export default class CategoryReleaseNotesPage extends React.Component<
...
-          {apis.map(api => (
-            <ApiReleaseNote key={api.urlFragment} api={api} />
+          {apis.map((api: IApiDescription) => (
+            <React.Fragment>
+              <ApiReleaseNote
+                apiCategoryKey={apiCategoryKey}
+                key={api.urlFragment}
+                api={api}
+              />
+            </React.Fragment>
```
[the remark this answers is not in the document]
stephennovosel: no this was left from when I was calling a component from within the Fragment but has since been removed.

src/containers/releaseNotes/CategoryReleaseNotesPage.tsx
stephennovosel: agreed that we can absolutely abstract some stuff out of this component and it would end up making things a lot more easily testable. in either case i think we should think about how we enter these queries into react context. if we call the queries from within the component then i'm not sure it is possible to test, however if they were connected to redux actions the query function it self would be a prop which we would be able to test is called during certain points.

stephennovosel: after looking through the component a bit more i think this would be a good opportunity to create a custom hook for this as opposed to a second component.

We could create a hook called `useGetFeatureFlagName` or something of that nature which will be called from within the `CategoryReleaseNotesPage` component. The hook itself would (upon the component mount) get the route params and run some logic based off of that `apiCategoryKey` route. The benefit of this would be that other portions of the app could re-use the same hook without having to create a separate component.

We could even take it a step further in with the hook not only read the URL for the key, but also return the query data based off of the route, and this functionality would no longer exist in any component and instead be reused in any component that essentially is querying the apiDefs based on the route.

PR #398, opened 2020-08-06
Api 1469 update prompt documentation

lukemarkwordtlibertyits:
### Description
Updated Authorization Docs to mention the support of the `consent` and `none` prompts. Also, changed the PKCE authorization code grant section to discourage the use of the `offline_access` scope and use silent authentication instead.

### Requested feedback

Concerns about readability. New silent authentication section.


src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -264,39 +264,21 @@ Pragma: no-cache
   "access_token": "SlAV32hkKG",
   "expires_in": 3600,
   "refresh_token": "8xLOxBtZp8",
-  "scope": "openid profile email offline_access",
+  "scope": "openid profile email",
+  "prompt": "none",
+  "patient": "1558538470",
```
beaugrantham-va, line 269: This attribute would not be in the response for the example request above. The example request should be updated to include the `launch/patient` scope or this attribute should be removed from the response.

src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -252,7 +252,7 @@ grant_type=authorization_code&code=z92dapo5&state=af0ifjsldkj&redirect_uri=<your
 ```
 (⚠️Be sure to replace `<yourRedirectURL>` with the redirect URL that you provided during registration)
 
-The authorization server will respond with an access token. If you requested the `offline_access` scope, you will also receive a `refresh_token`. The response will look like this:
+The authorization server will respond with an access token. Since the PKCE flow is typically used by applications that cannot safely store information, using the `offline_access` scope is not recommended. Silent authentication can be used to retrieve a new `access_token` safely. A new authentication request can be made using the `none` prompt like this:
```
beaugrantham-va, line 255: This last sentence references a _request_ but what follows is a _response_... I think this is confusing.

src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -213,7 +213,7 @@ with the following query parameters:
...
-| `prompt`                | Optional     | Supported prompts: `login`. If specified, the user will be forced to provide credentials regardless of session state. If omitted, an existing active session with the identity provider may not require the user to provide credentials.|
+| `prompt`                | Optional     | Supported prompts: `login`, `consent` and `none`.<p>If `login` is specified, the user will be forced to provide credentials regardless of session state. If omitted, an existing active session with the identity provider may not require the user to provide credentials.</p><p>Without the `consent` prompt, the user will only be asked to consent to claims once. If `consent` is specified, the user will be asked to consent to their claims being used regardless of prior consent.</p><p>If `none` is specified, an application will attempt an authorization request without user interaction. When the session is invalid or there are claims the user has not consented to, one of the following errors will be thrown: `login_required` or `consent_required`.</p> |
```
beaugrantham-va, line 216: > Without the `consent` prompt, the user will only be asked to consent to claims once.
I would remove this since it's restated in the positive with the following sentence.

beaugrantham-va, line 216: I would replace your usage of _claims_ with _scopes_ since that's what we reference in the rest of this document, and because _claims_ has a different meaning when it comes to JWTs.

src/content/apiDocs/oauthTechnical.mdx
beaugrantham-va, line 255: I'm interested in reading more about `offline_access` being discouraged with PKCE. Can you link a spec that references this so I can understand it more thoroughly?

src/content/apiDocs/oauthTechnical.mdx
beaugrantham-va, line 216: > If omitted, an existing active session with the identity provider may not require the user to provide credentials.

I would move this nearer to the end, and standalone, since it's referencing the omission of any of the three valid `prompt`s.

src/content/apiDocs/oauthTechnical.mdx
lukemarkwordtlibertyits: Not sure why I added prompt or patient. I removed them.

src/content/apiDocs/oauthTechnical.mdx
lukemarkwordtlibertyits: I reverted this section back to its previous state, as we have not decided whether it is worth advising for silent authentication yet.

src/content/apiDocs/oauthTechnical.mdx
lukemarkwordtlibertyits: Done

lukemarkwordtlibertyits: done

PR #399, opened 2020-08-07
API-1322 Upgrade react and react-dom

stephennovosel:
### Description

upgrade react and react-dom together to latest version. the benefit of this change is to utilize react hooks within the code base.


rtravitz:
@stephennovosel Can you provide a list of what changed / what to look out for in the intervening versions?

PR #401, opened 2020-08-10
API-1790 Add deactivated API section in release notes

stephennovosel:
### Description

Adds deactivated API option to release notes side nav and associated content.


src/apiDefs/deprecated.ts
```diff
@@ -31,3 +31,12 @@ export const getDeactivatedFlags = () => {
     return flags;
   }, {});
 };
+
+export const getDeactivatedCategory = () => ({
+  deactivated: {
```
mjlumetta, line 36: i don't think that including `deactivated` as a key is necessary here; i think that this function can return the current value of the `deactivated` key. i understand what you're going for in terms of backward compatibility with the real categories, but i think there's value in recognizing that deactivated APIs are their own special case and treating it as such. Deactivated APIs _is not_ truly an API category, and it doesn't necessarily have to conform to the data structures used by the real categories if it's not naturally a good fit.

also, along related lines, would love to see the type on this function to understand the relationship with real categories.

src/containers/releaseNotes/CategoryReleaseNotesCardSection.tsx
```diff
@@ -0,0 +1,52 @@
...
+import CardLink from '../../components/CardLink';
+import OnlyTags from '../../components/OnlyTags';
+
+// tslint:disable-next-line:interface-name
+interface CategoryReleaseNotesCardSectionProps {
+  apiFlagName: string;
+  apiDefinition: BaseAPICategory;
+}
+
+const CategoryReleaseNotesCardSection = (props: CategoryReleaseNotesCardSectionProps) => {
```
mjlumetta, line 16: same comment as above on `ApiReleaseNotes` about whether this can be inlined in the `CategoryReleaseNotesPage` file. up to you

mjlumetta: both for brevity and clarity, i'd like to remove the "Category" from this variable name. since it's used for things that aren't true categories (even if we refer to them as "categories" or "synthetic categories"), it might be useful to disambiguate going forward. but also, more importantly, it's really long. maybe should lint for long variable names.

src/containers/releaseNotes/CategoryReleaseNotesList.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import classNames from 'classnames';
+
+// tslint:disable-next-line:interface-name
+interface CategoryReleaseNotesListProps {
+  apiFlagName: string;
+  apiDefinition: BaseAPICategory;
+}
+
+const CategoryReleaseNotesList = (props: CategoryReleaseNotesListProps) => {
```
mjlumetta, line 15: same thing as above on `ApiReleaseNotes` and `CategoryReleaseNotesCardsSection` about moving this into `CategoryReleaseNotesPage`, except i feel more strongly about it here because this is so brief. could easily be inlined where it's used without a separate component, even.

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -43,22 +47,24 @@ function SideNavApiEntry(api: IApiDescription) {
...
+    apiCategory.apis.length > 1 && apiCategory.apis.map(api => SideNavApiEntry(api))
+  );
+
+  const sideNavEntry = (
+    <SideNavEntry to={`/release-notes/${apiCategoryKey}`} name={apiCategory.name}>
+      {subNavLinks()}
+    </SideNavEntry>
+  );
+
+  if (apiCategoryKey === 'deactivated') {
```
mjlumetta, line 61: same thing as in `CategoryReleaseNotesPage` about adding special cases. i think that the markup represented by `sideNavEntry` at this point is brief enough to inline where it's used.

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -43,22 +47,24 @@ function SideNavApiEntry(api: IApiDescription) {
...
-    return apiCategory.apis.map(api => {
-      if (apiCategory.apis.length > 1) {
-        return SideNavApiEntry(api);
-      } else {
-        return null;
-      }
-    });
-  };
+function SideNavCategoryEntry(apiCategoryKey: string, apiCategory: BaseAPICategory) {
+  const subNavLinks = () => (
```
mjlumetta, line 51: this would be pretty easy to inline where it's used - would like to do that while we're. not sure why this was split off to begin with, but not necessary at this point. (it was probably me)

src/hooks/useApiCategoryKey.tsx
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useState } from 'react';
+
+import { getDeactivatedCategory } from '../apiDefs/deprecated';
+import { getApiDefinitions } from '../apiDefs/query';
+
+// ideally we would use the useHistory/useLocation hook provided
+// in react-router-dom v5+
+const useApiCategoryKey = (matchParams: string) => {
```
mjlumetta, line 8: see `CategoryReleaseNotesPage` for thoughts on this hook. tldr is that i think there's a simpler way to achieve the same outcome. this does achieve the desired effect, but i'm not sure it's the ideal implementation.

src/containers/releaseNotes/CategoryReleaseNotesList.tsx
mjlumetta: same thing about naming as `CategoryReleaseNotesCardsSection`, except i feel less strongly about it here because it doesn't have the length issue. but consistency could be nice.

src/containers/releaseNotes/CategoryReleaseNotesCardSection.tsx
```diff
@@ -0,0 +1,52 @@
...
+                <CardLink
+                  name={name}
+                  subhead={
+                    vaInternalOnly || trustedPartnerOnly ? (
+                      <OnlyTags {...{ vaInternalOnly, trustedPartnerOnly }} />
+                    ) : (
+                      undefined
+                    )
+                  }
+                  url={`/release-notes/${name}#${dashUrlFragment}`}
```
mjlumetta, line 38: should `apiCategoryKey` instead of `name` here

src/containers/releaseNotes/ApiReleaseNote.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { IApiDescription } from '../../apiDefs/schema';
+
+const ApiReleaseNote = ({ api, flagName }: { api: IApiDescription, flagName: string }) => {
+  const dashUrlFragment = api.urlFragment.replace('_', '-');
+
+  return (
+    <Flag name={`${flagName}.${api.urlFragment}`}>
+      <div id={dashUrlFragment}>
+        <h2>{api.name}</h2>
+        {api.deactivationInfo && isApiDeactivated(api) && (
```
mjlumetta, line 17:
```suggestion
        {isApiDeactivated(api) && (
```

src/containers/releaseNotes/CategoryReleaseNotesPage.tsx
```diff
@@ -1,84 +1,57 @@
...
-          <div className={defaultFlexContainer()}>{apiCards}</div>
-        </div>
-      );
-    }
+const CategoryReleaseNotesPageContent = ({ apiCategory, apiFlagName } : { apiCategory: BaseAPICategory, apiFlagName: string }) => {
 
+  if (apiCategory && apiFlagName !== '') {
     return (
-      <section role="region" aria-labelledby={`${apiCategoryKey}-release-notes`}>
+      <section role="region" aria-labelledby={`${apiCategory.name}-release-notes`}>
```
mjlumetta, line 17:
```suggestion
      <section role="region" aria-labelledby={`${apiCategoryKey}-release-notes`}>
```

src/containers/releaseNotes/CategoryReleaseNotesPage.tsx
```diff
@@ -1,84 +1,57 @@
...
         <PageHeader
-          halo={apiDefs[apiCategoryKey].name}
+          halo={apiCategory.name}
           header="Release Notes"
-          id={`${apiCategoryKey}-release-notes`}
+          id={`${apiCategory.name}-release-notes`}
```
mjlumetta, line 21:
```suggestion
          id={`${apiCategoryKey}-release-notes`}
```

src/containers/releaseNotes/CategoryReleaseNotesPage.tsx
```diff
@@ -1,84 +1,57 @@
...
-        );
-      });
-
-      cardSection = (
-        <div role="navigation" aria-labelledby={`${apiCategoryKey}-release-notes`}>
-          <div className={defaultFlexContainer()}>{apiCards}</div>
-        </div>
-      );
-    }
+const CategoryReleaseNotesPageContent = ({ apiCategory, apiFlagName } : { apiCategory: BaseAPICategory, apiFlagName: string }) => {
```
mjlumetta, line 13: would prefer not to extend the name - "CategoryReleaseNotesPage" or even "ReleaseNotesPage" would be better. "PageContent" also refers to the component responsible for rendering `<main />`.

src/containers/releaseNotes/CategoryReleaseNotesPage.tsx
```diff
@@ -1,84 +1,57 @@
...
+    <CategoryReleaseNotesPageContent
+      apiCategory={categoryDefinition}
+      apiFlagName='hosted_apis'
+    />
+  );
+};
+
+export const DeactivatedReleaseNotesPage = (props: RouteComponentProps<IApiNameParam>) => {
+  const apiCategoryKey = 'deactivated';
+  const categoryDefinition = getDeactivatedCategory()[apiCategoryKey];
```
mjlumetta, line 50:
```suggestion
  const categoryDefinition = getDeactivatedCategory();
```

rewrite `getDeactivatedCategory()` to return the "category" definition. "deactivated" isn't really a category key, and it's not clear what other keys would go in that returned object in the future (plus, not future-proofing can be a good guideline sometimes).

src/containers/releaseNotes/CategoryReleaseNotesPage.tsx
```diff
@@ -1,84 +1,57 @@
...
-          ))}
-        </div>
       </section>
     );
   }
-}
+  return null;
+};
+
+export const ActiveCategoryReleaseNotesPage = (props: RouteComponentProps<IApiNameParam>) => {
```
mjlumetta, line 37:
```suggestion
export const CategoryReleaseNotesPage = (props: RouteComponentProps<IApiNameParam>) => {
```

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -75,14 +85,21 @@ export class ReleaseNotes extends React.Component<RouteComponentProps<IApiNamePa
...
+                {SideNavCategoryEntry('deactivated', deactivatedApis)}
               </SideNav>
               <div className={classNames('vads-l-col--12', 'medium-screen:vads-l-col--8')}>
-                <Route exact={true} path="/release-notes/" component={ReleaseNotesOverview} />
-                <Route
-                  exact={true}
-                  path="/release-notes/:apiCategoryKey"
-                  component={CategoryReleaseNotesPage}
-                />
+                <Switch>
```
mjlumetta, line 91: shouldn't need to use the switch if you use `exact={true}`

src/hooks/useApiCategoryKey.tsx
mjlumetta: can delete this file

src/containers/releaseNotes/CategoryReleaseNotesPage.tsx
```diff
@@ -1,84 +1,57 @@
...
+export const DeactivatedReleaseNotesPage = (props: RouteComponentProps<IApiNameParam>) => {
+  const apiCategoryKey = 'deactivated';
+  const categoryDefinition = getDeactivatedCategory()[apiCategoryKey];
+  return (
+    <CategoryReleaseNotesPageContent
+      apiCategory={categoryDefinition}
+      apiFlagName='deactivated_apis'
+    />
+  );
+};
```
mjlumetta, line 57: great encapsulation here! love the final result 🙌

src/containers/releaseNotes/ApiReleaseNote.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+};
+
+const mockProps = {
+  api: { ...mockApi },
+  flagName: 'mock_apis',
+};
+
+describe('ApiReleaseNote', () => {
+  it('should render correctly', () => {
+    const wrapper = shallow(<ApiReleaseNote {...mockProps} />);
```
mjlumetta, line 34: prefer `mount()` to shallow: https://kentcdodds.com/blog/why-i-never-use-shallow-rendering. same for other tests

mjlumetta:
closing in favor of a new PR since we need to target `API-1895/upgrade-typescript` now.

PR #404, opened 2020-08-11
Api 1741 add aud support

lukemarkwordtlibertyits:
### Description
Added support in the oauthTechnical documentation for the `aud` parameter

### Requested feedback
Grammar, wording, etc ...


src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -39,6 +39,7 @@ with the following query parameters:
 | `redirect_uri`  | **Required** | The URL you supplied. The user will be redirected to this URL after authorizing your application. |
 | `response_type` | **Required** | Supported response types: `code` |
 | `state`         | **Required** | Ensures authorization flow integrity. See [security](#security-considerations). |
+| `aud`           | **Required** | The FHIR resource server an application is making API calls to. See [security](#security-considerations). |
```
beaugrantham-va, line 42: I would drop _FHIR_ since this documentation is also applicable to non-FHIR APIs.

beaugrantham-va, line 42: I think we also need to include the expected value(s). For now it should should be the default `aud` and once we complete the re-org it will be the API specific `aud`s.

src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -373,6 +377,10 @@ The [OpenID Connect documentation](https://openid.net/specs/openid-connect-core-
 
 Your application must validate JWT signatures. This allows your application to verify that the provided JWT originates from our authorization servers and prevents your application from accepting a JWT with claims that are attempting to impersonate one of your users.
 
+### aud
```
beaugrantham-va, line 380: I would move this up above _JWT Signature validation_, along side the other security related parameters (_state_, _nonce_).

lukemarkwordtlibertyits: Done

src/content/apiDocs/oauthTechnical.mdx
lukemarkwordtlibertyits: Removed the FHIR specification and added `api://default` as the accepted `aud` value

lukemarkwordtlibertyits: Removed the FHIR specification and added `api://default` as the accepted `aud` value

beaugrantham-va:
The changes look good.

After consideration let's hold on this update and coordinate with the other authz changes.

PR #405, opened 2020-08-12
API-172 Managed Focus on page navigation

gileswells:
### Description
When navigating to pages and #hashed links internal to a page the browser's focus is now being set to the top most header of that page / part of the page as needed to direct screen readers to the content they're looking for.

https://vajira.max.gov/browse/API-172
https://vajira.max.gov/browse/API-173
https://vajira.max.gov/browse/API-174

### Process
For the functional components there is a new hooks.ts file that I created for new React hooks.

- [ ] Docs updated

### Requested feedback
Looking for feedback on how the hooks are being implemented and also if there's any additional changes / optimizations that can be seen for the class components.


gileswells:
Very out of date and going a different direction with this one

PR #407, opened 2020-08-25
Upgrade typescript and react-redux and switch to connected-react-router

mjlumetta:
### Description

makes progress on [API-1895](https://vajira.max.gov/browse/API-1895) and [API-1896](https://vajira.max.gov/browse/API-1896).

this PR starts with Typescript, but you know how upgrading packages go. it's like playing dominoes. one thing leads to another.

changes:

- upgrades `typescript` to 3.9.7
- upgrades `react-redux` to 7.2.1
- replaces `react-router-redux` with `connected-react-router` 6.8.0

what this PR does not do:

- upgrading to Typescript v4
- upgrading `history` (API-1896)

helps unblock [API-1790](https://vajira.max.gov/browse/API-1790) and [API-1317](https://vajira.max.gov/browse/API-1317).

### Process

  * this is going to unblock some tests. really would like to have better tests while making this change, but hopefully what we've got works, plus manual testing
- [ ] Docs updated
  *. no UI changes

### Requested feedback

did it... did it break?


src/containers/RoutedContent.tsx
```diff
@@ -1,32 +1,20 @@
...
-}
-
-const mapStateToProps = (state: IRootState) => {
-  return {
-    ...state.routing,
-  };
-};
-
-class RoutedContent extends React.Component<IRoutedContentProps, { }> {
+export default class RoutedContent extends React.Component {
```
mjlumetta, line 10: probably just gonna get rid of this file while i'm in here because it doesn't really serve a purpose

src/containers/apply/ApplyForm.tsx
```diff
@@ -185,15 +183,20 @@ class ApplyForm extends React.Component<IApplyProps> {
...
+      this.anyApiSelected() &&
+      termsOfService &&
+      applicationTypeComplete &&
+      redirectURIComplete
     );
   }
 }
 
-export default connect(
+export default connect<IApplication, ApplyDispatchProps>(
```
mjlumetta, line 199: this sort of `connect()` refactoring to make these types explicit might not be strictly necessary anymore. i did that to fix a compile error before i upgraded `react-redux`. adding the types didn't fix the error, and upgrading `react-redux` did, so maybe could do without these types.

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -13,13 +13,18 @@ import { history } from '../../store';
 
 import '../../../node_modules/react-tabs/style/react-tabs.scss';
 
-interface IApiDocumentationProps {
+// tslint:disable:interface-name
```
mjlumetta, line 16: we decided we didn't like the I prefix and wanted to standardize on not that in the frontend and the backend

src/store.ts
```diff
@@ -51,13 +50,12 @@ const store = createStore(
   combineReducers<IRootState>({
     apiVersioning,
     application,
-    routing,
+    router: connectRouter(history),
```
mjlumetta, line 53: `react-router-redux` -> `connected-react-router`. similar but not the same syntax

tsconfig.json
```diff
@@ -21,13 +21,15 @@
...
     "scripts",
     "acceptance-tests",
     "webpack",
     "jest",
     "src/setupTests.ts",
     "src/e2eHelpers.ts",
     "*.jest.config.js",
-    "jest-puppeteer.config.js"
+    "jest-puppeteer.config.js",
+    "SitemapBuilderWebpackPlugin.js"
```
mjlumetta, line 33: something about the typescript upgrade made it mad about the `require('./config/paths')` in the sitemap plugin

package.json
```diff
@@ -28,10 +28,12 @@
     "@fortawesome/react-fontawesome": "^0.1.0",
     "@sentry/browser": "^4.6.4",
     "classnames": "^2.2.6",
+    "connected-react-router": "^6.8.0",
     "flag": "^3.0.0-1",
     "get-video-id": "^3.1.4",
     "highlight.js": "^9.16.1",
     "history": "^4.7.2",
+    "hoist-non-react-statics": "^3.3.2",
```
mjlumetta, line 36: this came out of `react-redux`: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33690. nasty dependency issue, didn't really dig into why it happens, but this fixed it

package.json
```diff
@@ -45,16 +47,16 @@
...
     "react-router": "^4.2.0",
     "react-router-dom": "^4.2.2",
     "react-router-hash-link": "^1.2.1",
-    "react-router-redux": "5.0.0-alpha.9",
     "react-router-sitemap": "git+https://github.com/nfasulo/react-router-sitemap.git#6060031e64211392e2070da260e45b29972bd194",
     "react-scroll": "^1.7.16",
     "react-tabs": "^3.0.0",
     "redux": "4.0.0",
     "redux-thunk": "^2.3.0",
+    "seamless-immutable": "^7.1.4",
```
rtravitz, line 59: What's the purpose of this addition? I'm not seeing it referenced anywhere else, and there's no comment attached like with `hoist-non-react-statics`.

src/containers/RoutedContent.tsx
rtravitz: If it doesn't have a purpose then agree—just drop it.

src/containers/apply/ApplyForm.tsx
rtravitz: Not sure I'm clear on why this was a problem / is no longer a problem.

mjlumetta: i don't know whether or not it's strictly necessary. given that that's the case, i'll remove them and if it doesn't break, i'll leave them out because unnecessary refactoring extends the scope of the PR, like you said. otherwise i can give a clearer explanation.

package.json
mjlumetta: right, my bad, should've left one. it's a peer dependency of `connected-react-router`: https://github.com/supasate/connected-react-router/blob/master/package.json#L41

src/components/PageContent.tsx
```diff
@@ -3,18 +3,18 @@ import { RouteComponentProps } from 'react-router';
 import { SiteRoutes } from '../Routes';
 
 export default class PageContent extends React.Component<RouteComponentProps, {}> {
-  private mainRef: React.RefObject<HTMLMainElement> = React.createRef<HTMLMainElement>();
+  private mainRef: React.RefObject<HTMLElement> = React.createRef<HTMLElement>();
```
mjlumetta, line 6: `HTMLMainElement` is no longer supported in Typescript: https://github.com/microsoft/TypeScript/issues/32042

tsconfig.json
gileswells: Is there a gain to creating a ticket to update the sitemap plugin to allow this to be removed from the exclude list or is there not enough benefit?

mjlumetta: since it's plain Javascript, i don't really see any reason to include it... not really sure what about the upgrade made this get picked up now but not before, but i'm pretty comfortable with it excluding it. guessing that it made `require`s within scripts get picked up? idk.

mjlumetta:
> I am not a fan of this change. As it stands we'd need to add a disable on a per file basis and my understanding is that IInterfaceName is a pretty well established naming convention. My vote would be no and if I missed that discussion that's my bad.
>
> If we're moving to not having it then at minimum it needs to be added to the (https://github.com/department-of-veterans-affairs/developer-portal/blob/master/tslint.json)[tslint.json] config so that it's picked up by IDEs and doesn't throw errors all over the place. I'd also ideally want to see a ticket to just handle this change repo-wide so it's handled once and it's over with. On top of that a custom rule that prevents an interface name from starting with II would be a nice to have so people don't accidentally add it in the future when joining the project.

yeah, so, this was a decision we made a few months ago - it was a month or two before you started, so this is very much not your bad. but we decided that we wanted to have style consistency across the backend and the frontend, and we landed on no prefix when we converted the backend to a Node app and were thinking through those conventions. i very much do not care which convention we follow, as long as there is one, and i find our previous goal of consistency across those projects valuable. so basically, one of those has to change, and the decision we landed on was to change the frontend. i am okay with revisiting it for a limited period only if that's okay with others, but i would like to settle it after that.

since we have to convert the whole repo, i'm making those changes as i touch things and disabling on a per-file basis so as to make those changes incrementally. personally, i'm okay with the suboptimal split for a while until we get a critical mass of the no-prefix ones and can clean the whole thing up; however, if you'd prefer not to have a mixture and create a ticket for it so we can do it in one big swoop instead, i totally understand that too. 

so yeah, that's the background - let me know how you'd like to proceed based on that. my main interest here is settling on one answer going forward.

gileswells:
Having a singular standard on naming conventions is worth while and seeing how this has been in transition for a while I don't think it would be in our best interest to throw a wrench into that. I would prefer to see having the tslint updated vs the per file override and since that change shouldn't hurt anything I don't see a real reason to put it off. The custom rule idea to prevent interfaces from starting with `II` could come in a later ticket / PR.

mjlumetta:
@gileswells gotcha, i think that should be good. couple small clarifying questions, to make sure i'm understanding you correctly: when you say "update the tslint" so we can get rid of the disable statements, is the update you're thinking of "turn the `interface-name` rule off"? and second, what do you mean by the custom rule part?

gileswells:
@mjlumetta It's more of a nice to have if there's a way to add a custom rule to the tslint ruleset so that we don't have anyone that's used to using a prefix we can head those off with a rule to alert them to the naming convention. https://palantir.github.io/tslint/develop/custom-rules/

Basically a custom tslint rule that would prevent interfaces from starting with === `II`. Not something blocking this PR at all but a nice to have for the future.

gileswells:
@mjlumetta and yes so we can remove the file by file rule disables. I know they don't technically hurt anything but no reason to keep it if we are changing the global config is one spot.

mjlumetta:
@gileswells that all sounds good to me. had not thought of just taking out the rule before, thanks for pointing that out. will make the changes and ping you for what i hope can be a quick re-review once it's complete.

re: the custom rule - i think we might not have to create a custom rule? i think we can probably do it with the `interface-name` rule by setting `interface-name: [true, 'never-prefix']`. [the docs](https://palantir.github.io/tslint/rules/interface-name/) seem to suggest that's an option

PR #412, opened 2020-08-28
API-2138 - Secondary Skip Nav within the main content element

gileswells:
### Description
This PR allows for a secondary accessibility skip link that will allow users to skip over the side navigation after navigating to a new page and focus on the `h1` of the page directly.

### Process
- [ ] Docs updated

### Requested feedback
Looking for feedback on the implementation of the refs in particular. To do this the "React way" we'd be passing the heading ref from the `PageContent` component through to each component. To do this as the previous skip nav item was done we could do it with just a `document.querySelector('h1')` as seen [here](https://github.com/department-of-veterans-affairs/developer-portal/blob/677fb407958c0607d302aad31dde94111626f3d5/src/components/Header.tsx#L111).


src/components/Header.scss
```diff
@@ -19,8 +19,8 @@ a.va-api-skipnav {
   background-color: #f9f9f9;
   position: absolute;
   z-index: 100;
-  top: -100px;
-  
+  top: -1000px;
```
mjlumetta, line 22: guessing this is because the shared class requires the distance to be increased. making this number really big (magnitude-wise, anyway) is a totally valid option here, but another option would be making it go left.

also, tiny style thing, but not sure if we want the gold outline on the top?

mjlumetta:
ok, gonna spitball some thoughts re: massaging the refs solution (although the same stuff could apply to the other approach, too, but my main interest would be in cutting down the manual ref-passing). 

so, our current component hierarchy around this level looks like this:

```
PageContent
|_ main
    |_ SiteRoutes // every page, in the non-literal sense
        |_ Route
            |_ News
```

what if we inverted the hierarchy so that the `Page`/`PageContent`/whatever we call this component was below `SiteRoutes` and its individual `Route`? and in particular, what if we made `Page` a wrapper for a component that conceptually represents a page - that is, the top-level React component under `<main />`? so something like this:

```
SiteRoutes // every page, in the non-literal sense
|_ Route
    |_ Page // wrapper for news with some stuff in here, including the secondary skip nav
        |_ main
            |_ News
```

summarizing some of the aspects of this hypothetical approach:

* for each top-level component, the module would export `Page(News)`. this pattern would require us to explicitly mark those top-level components as such. (sidebar: i'm really not sure about the "page" language in the single-page app context, but i don't know a great replacement... kinda thinking of it as a virtual page. thoughts?)
* `Page` would now be at the level of this individual page rather than something that renders all of the pages by route. we could configure things differently by page; for example, we could parameterize whether or not the secondary skip nav is included, which i think we do want because not all pages have a side nav. for example, we could say `Page(ApplyForm, { sideNav: false })` or something along those lines.
* `Page` would handle location changes. for each individual `Page` component, the scope of URL changes would be smaller, but there would still be differences. ex: `/release-notes` vs `/release-notes/health`.

for the secondary skip nav specifically, the implications would be that `Page` contains the secondary skip nav.

thoughts on this approach? other alternatives we might consider?

src/components/PageContent.tsx
```diff
@@ -1,27 +1,47 @@
...
+          Skip side navigation
+        </HashLink>
+        <SiteRoutes forwardedRef={this.pageHeadingRef} />
       </main>
     );
   }
+
+  private handleSkipNavClick = () => {
+    if (this.pageHeadingRef && this.pageHeadingRef.current) {
+      this.pageHeadingRef.current.focus();
```
mjlumetta, line 42: how does this event handler interact with the `focus()` event in the `componentDidUpdate()` lifecycle method?

gileswells:
@mjlumetta The more I work at testing `forwardRef` the more and more it looks like something that I don't think is really beneficial to us. We could use it but then we'd be needing to wrap each and every component down to the `PageHeader` component with `React.forwardRef()` and it seems like much more headache than it's worth. In the end, based on this POC and the work toward evaluating `forwardRef` I think we should just stick with the setup we've been using for the main skip nav and target the `h1` tag with a `document.querySelector('h1')`. The overhead of everything else just seems way over the top.

mjlumetta:
@gileswells that reasoning sounds great to me. the explanation makes sense, and it does seem way simpler.

gileswells:
Resolved the comments that were either addressed or outdated due to going with the querySelector method.

src/components/PageHeader.scss
```diff
@@ -1,3 +1,6 @@
 .header-halo {
   text-transform: uppercase;
 }
+h1[tabindex]:focus {
```
mjlumetta, line 4: i think that this and the similar `#main` rule in `PageContent.scss` probably point to something in the global styles being too aggressive in terms of setting the outline. looks like we set this focus outline for `*:focus`, which seems really aggressive to me. i would say elements that aren't natively focusable should not get the outline, and that's the spirit of the rule here.

looking in the dev tools, looks like there's also a VADS rule for `[tabindex]:focus` to the same effect, which i'm guessing is why `[tabindex]` is on both rules. i would think that we could probably go so far as to say that we can override that rule. probably should check that this doesn't break natively focusable elements that mightly mistakenly have a `tabindex`, but think it's okay.

thoughts?

mjlumetta (commented):
so, i think this is mostly good based on the design we talked about. one issue is that we need to restrict the skip nav to pages with a side nav. 

there's another core issue that i didn't think of during our design discussions and that is also do to my taking a shortcut on the primary skip nav 😬 so we may need to take another look at how to accomplish this in a way that emulates the native behavior we're try to implement.

src/components/PageContent.tsx
```diff
@@ -20,8 +23,20 @@ export default class PageContent extends React.Component<RouteComponentProps, {}
   public render() {
     return (
       <main id="main" ref={this.mainRef} tabIndex={-1}>
+        <Link
+          to={window.location.pathname + window.location.search}
```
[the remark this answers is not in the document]
gileswells: I was going to use `HashLink` but in the end I went with `Link` because there wasn't a standard hash to attach to. The `h1` doesn't have an ID on most pages and for the pages it does it's not uniform. If we wanted to update that to something like `#page-title` across the board then I'd be ok with that and then `HashLink` is absolutely the way to go. It doesn't really seem like the IDs that are custom do anything special so making it uniform across the board wouldn't break anything.

src/components/PageContent.tsx
```diff
@@ -20,8 +23,20 @@ export default class PageContent extends React.Component<RouteComponentProps, {}
...
+        >
+          Skip side navigation
+        </Link>
         <SiteRoutes />
       </main>
     );
   }
+
+  private handleSkipNavClick = (e: React. MouseEvent<HTMLAnchorElement>) => {
+    e.preventDefault();
```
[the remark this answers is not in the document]
gileswells: If we handle the above ID change on the h1 then we solve this as you say by removing the click handler entirely.

src/content/termsOfService.mdx
```diff
@@ -1,4 +1,10 @@
-# VA API Terms of Service 
+---
+imports:
+  PageHeader: ../components/PageHeader.tsx
+---
+
+<PageHeader header="VA API Terms of Service" />
```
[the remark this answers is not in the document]
gileswells: Partially yes and the other part is for uniformity across the board. I'd much prefer to have a single place to modify page headers and know that it's a uniform change.

src/components/PageContent.scss
```diff
@@ -0,0 +1,3 @@
+#main[tabindex]:focus {
+  outline: none;
```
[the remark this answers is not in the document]
gileswells: Outline and border are different and it's specifically designed to be used for focusing elements. So there's no outline off-screen or anywhere because there's no focus to trigger it before there's focus.

src/components/PageHeader.scss
gileswells: I don't like the idea of wholesale overwriting that `[tabindex]:focus` that already exists because there's lots of potential legitimate uses. I'd much prefer to selectively override it as we find issues like the `h1` on the page and the `main#main` content wrapper.

src/components/PageContent.tsx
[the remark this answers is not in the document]
mjlumetta: yeah, so, for the first part about the hash, i think we do need to have a meaningful link in the href, or else it shouldn't be a link. while an href to the current page is not technically violating the "link without an href" rule, i don't know if it's in the spirit of semantic HTML? if we truly want the click handler to be solely responsible for setting focus, we should probably use a button instead. (never heard of a skip nav _button_, but i think that should be okay, right? can do some quick research to see if there's any reason it's usually a link and not a button, but like, i don't see a problem with it.)

with that said, i think that something like `#page-title` would probably be fine. i was thinking the opposite, but now i think it's probably okay. the main thing that the header id is used for is having an `aria-labelledby` on the content region or whatever (this pattern is on the API and category documentation pages and the category release notes page, as well as the quickstart page). i don't think there's any reason that has to be unique to the page, so we could set the id statically in `PageHeader`. the only thing i would say is that `PageHeader` should export that id as a constant so we don't have to use a magic string all over the place.

_with all that said,_ i think that if we absolutely needed to have an anchor link to a custom header id, we could do that. disclaimer up front that i absolutely _do not_ think we should do this, because one of the other two solutions should work and so i'm pretty sure this would be overkill. but if we were going to do it, we could try something like

1. making sure both the secondary skip nav link and the h1 are wrapped in a `Route` and generating a unique id based on the route. then we could use that same generation function if we needed to use it somewhere - for example, maybe we want a "Back to Top" link for keyboard focus at the bottom of the page.
2. attempting to create a higher-order component that wraps pages with a side nav so that they have some kind of shared structure that includes the secondary skip nav. then this higher-order component can be responsible for knowing that id. i could see this kind of refactor being useful down the road for some purpose, but again, overkill for our purposes here. the one exception i can mayyybe see that sort of shared structure helping with is the whole "only include the secondary skip nav on pages with a side nav" problem.

so yeah... something from the first two paragraphs might actually be useful, and the

src/components/PageContent.tsx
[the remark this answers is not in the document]
mjlumetta: sadly, that's not actually the case. there's an explanation of why that's the case here: https://github.com/department-of-veterans-affairs/developer-portal/blob/master/docs/issues.md#focus

src/content/termsOfService.mdx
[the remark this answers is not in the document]
mjlumetta: okay, cool. i could go either way on this one, so sounds good by me

src/components/PageContent.scss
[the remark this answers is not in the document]
mjlumetta: sorry, don't think i follow there? 

another way of putting this question would be "what does this rule accomplish?" `main` should be receiving focus prior to your changes, so i don't get why we need to have this rule now. i can see that there's some kind of outline if i toggle the rule off, but i don't know how it got there.

here's what i see in prod:

Screen Shot 2020-09-09 at 5 11 41 PM

and on your branch:

Screen Shot 2020-09-09 at 5 12 28 PM

src/components/PageHeader.scss
mjlumetta: what legitimate uses are you thinking of?

gileswells: Form field inputs for one. Can't think of another legit usage but the form fields seem large enough to me.

mjlumetta: can you explain the use case for `tabindex` on a form field? i'm not familiar with a case where a form input might require one.

here's my reasoning for why i don't think we need the broad rule:

1. having a `tabindex` implies that an element is not natively focusable. if it were natively focusable, we wouldn't need a `tabindex`.
2. we will never want to put the focus outline on an element that is not natively focusable and only receives focus programmatically. or, more accurately stated, we _do not currently_ want to put the focus outline on any element that isn't natively focusable, and we can deal with the exceptions as they arise.

i could definitely be missing some case where that is true, in which case, happy to be proven wrong and would like to understand that case.

gileswells: tabindex is more than just giving focus to a non natively focusable item. It allow allows you to order where the focus goes among tabindexed items. So if you have a 2 column form you can force the cursor to go left right down the page even though the DOM order is 2 divs containing many fields making them out of order.

mjlumetta: that's a great point, hadn't been thinking of that. so my follow-up question is, then, do we have any instances of that? or, alternatively, would `input:focus` (or `checkbox:focus`, `select:focus`, etc) be insufficient for some reason?

src/components/PageContent.tsx
```diff
@@ -20,8 +23,20 @@ export default class PageContent extends React.Component<RouteComponentProps, {}
   public render() {
     return (
       <main id="main" ref={this.mainRef} tabIndex={-1}>
+        <Link
+          to={window.location.pathname + window.location.search}
+          className={classNames('va-api-skipnav', 'vads-u-padding-x--2', 'vads-u-padding-y--1')}
+          onClick={this.handleSkipNavClick}
+        >
+          Skip side navigation
+        </Link>
```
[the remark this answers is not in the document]
gileswells: I've updated this to use borders instead of outline since you can't hide top/left on an outline. Also double checked the a11y specs on that and so long as we handle it in some other clear manner this is fine to do.

src/components/PageContent.tsx
```diff
@@ -20,8 +23,20 @@ export default class PageContent extends React.Component<RouteComponentProps, {}
   public render() {
     return (
       <main id="main" ref={this.mainRef} tabIndex={-1}>
+        <Link
```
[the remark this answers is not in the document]
gileswells: What if we add a side nav for all pages? It might just be a set of hash links for parts of the page but having a standard for including a side nav on all pages would make this easier.

src/components/PageContent.scss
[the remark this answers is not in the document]
gileswells: I thought part of the fixes overall were to remove the halo from `#main` when it's focused. This might be more of an API-172 item that snuck it's way in here.

src/components/PageHeader.scss
gileswells: > or, alternatively, would input:focus (or checkbox:focus, select:focus, etc) be insufficient for some reason?

Well there it is. We have those. I can think of no further reasons why objecting to a blanket `tabindex -> outline:none` makes sense. Will implement.

src/components/PageContent.tsx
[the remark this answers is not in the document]
gileswells: `#page-heading` static across the board is the plan and set it as a constant string

src/components/PageContent.tsx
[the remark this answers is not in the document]
mjlumetta: @gileswells came up with the idea during our working session of including the skip nav as the first focusable element within `SideNav`. it's still the first focusable element in the document order under `<main/>`, so it receives focus and becomes visible on the first tab, and it's never present if there is no `SideNav`. we all think this is a great idea (Giles, @ediiotero, and me).

src/components/PageContent.tsx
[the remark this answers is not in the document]
gileswells: This will be resolved by moving the skip nav to the top of the side nav via the other comment thread. Resolving this one.

src/components/SideNav.tsx
```diff
@@ -143,19 +152,26 @@ export default class SideNav extends React.Component<ISideNavProps> {
...
+          >
             {this.props.children}
           </ul>
         </nav>
       </div>
     );
   }
+
+  private focusPageHeader = (e: React.MouseEvent<HTMLAnchorElement>) => {
+    e.preventDefault();
```
gileswells, line 174: @mjlumetta I feel like using a `preventDefault()` here is the right call to make. You mentioned something about history on the previous comment but for me it's moving within the page and not navigating between pages so I don't think that history should be brought into this. Also by preventing the URL change it solves issues with the side nav thinking that the skip nav was the new "current" url and it would cause the current page to lose it's current url styling.

ediiotero (approved):
LGTM, not much to add. Your comment on History for the nav makes sense.

src/components/SideNav.tsx
mjlumetta: i would disagree with that position. i can't find the previous comment, but there are a few ways that i would approach this.

* generic question about `preventDefault`: what behavior are you preventing, and why do you need to prevent it? for me, this doesn't meet the criteria.
* history isn't restricted to navigating between routes; it includes navigation within the page to different anchors.  "moving within the page" qualifies for that navigation and seems like a natural case for an anchor link.
* the history is going to change whether you want it to or not as long as we're using React Router. internally, React Router calls `event.preventDefault()` and then `history.push()`.
* without unnecessarily aggressive `preventDefault`-ing, we wouldn't even need this click handler. my preference would be to use native anchor link behavior, but that is broken whether the element is a `Link` or an `a`.
  - one day, maybe, we'll be able to fix this, but until then we should adhere as closely as possible to native behavior.
* most important: links _should_ change the history. if it shouldn't affect history, then it shouldn't be a link.
  - i asked in the Ad Hoc #accessibility channel about whether a skip nav button was a legitimate alternative to a link, and the consensus was a pretty firm no. on the more general side of the advice i got back, someone said, "One other critical thing to consider: links go places; buttons do things."

the styling question is interesting. i would say that the nav link should have its function handle that. haven't looked at the whole implementation yet so i don't have strong opinions in

gileswells: This, to me, isn't a navigation item. It's managing the focus of the browser and not changing or really even scrolling the page. My reasoning for using it is that it's the most efficient way to block React Router from changing the URL in the browser. Once that happens the side nav will re-render and apply the current url styling to the matching address in the url inclusive of the hash.

According to the styling of the side nav you would expect that `/explore/benefits/docs/claims?version=current#page-header` would be for the overview page for the benefits API at `/explore/benefits`. I haven't found a way to override the styling of the which link is active based on a url without the hash included. The main content skip we have in the header also has this bug. If you're on a page with a side nav you won't get the proper styling on a url like `/explore#main`

As a forward thinking item in API-174, I have no desire to mimic this `preventDefault()` implementation for the oAuth pages or other instances where a hash link would push the user down the page. That, to me, is navigation and would trigger a history item being added. Just moving the focus to the h1 of the page to allow a screen reader to skip the side nav to me doesn't constitute navigating to a portion of the page since the page doesn't move. I recognize this is about as close to splitting hairs as I think we can get. I just don't agree that it's truly navigation of the page since the page doesn't move in reaction to this click.

gileswells:
@mjlumetta changes from working session ^

src/components/SideNav.tsx
```diff
@@ -128,11 +132,16 @@ export default class SideNav extends React.Component<ISideNavProps> {
...
-      )}>
+      <div
+        className={classNames(
+          'vads-l-col--12',
+          'vads-u-padding-right--5',
+          'medium-screen:vads-l-col--4',
+        )}
+      >
+        <HashLink className="va-api-secondary-skip-nav" to="#page-header" onClick={this.focusPageHeader}>
+          Skip Navigation
```
mjlumetta, line 143: might want another clarifying word here, like "skip side navigation" or "skip page navigation"?

src/styles/base.scss
```diff
@@ -83,6 +84,11 @@ a:focus,
   @include focus-outline;
 }
 
+#page-header:focus,
+#main:focus {
+  outline: none;
+}
```
mjlumetta, line 90: i have some updates on this focus discussion that run counter to Trevor's advice, but they're not part of this PR. we'll work that out in 172-174.

mjlumetta (commented):
couple cleanup things but it's ready to go once those have been addressed.

src/components/PageContent.scss
```diff
@@ -0,0 +1,10 @@
...
+}
+
+#main a.va-api-secondary-skipnav:focus {
```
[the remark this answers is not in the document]
gileswells: No need to move at all. This got orphaned when I moved the markup to SideNav.tsx and changed the class name. Whole file is getting axed.

src/components/SideNav.scss
```diff
@@ -6,8 +6,10 @@
 }
 
 li.va-api-sidenav-entry {
-  a:hover {
+  a:hover,
+  a:focus {
     background-color: $color-gray-lighter;
+    outline: none;
```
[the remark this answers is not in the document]
gileswells: Not sure. I might have crossed wires with this on 172. Deleted and will revisit if needed in 172.

src/components/SideNav.scss
```diff
@@ -21,4 +23,21 @@ a.va-api-active-sidenav-link + ul.va-api-sidenav-sub-list {
...
\ No newline at end of file
+}
+
+a.va-api-secondary-skip-nav {
+  position: absolute;
+  top: -1000px;
+
+  &:focus {
+    background: $va-api-color-blue-dark;
+    color: #ffffff !important;
```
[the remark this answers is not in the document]
gileswells: I changed it over to the utility classes. As for putting the rules in the `:focus` version of the element I did that with the thought that if it's not showing up on the page for regular users then why spend the cpu cycles (basically 0 impact) on drawing it off screen if it's not being used anyways.

src/components/SideNav.tsx
gileswells: Added Page

src/components/SideNav.tsx
```diff
@@ -128,11 +132,27 @@ export default class SideNav extends React.Component<ISideNavProps> {
...
+        )}
+      >
+        <HashLink
+          className={classNames(
+            'va-api-secondary-skip-nav',
+            'vads-u-padding--1p5',
+            'vads-u-margin-bottom--5',
+            'vads-u-line-height--3',
+            'vads-u-text-decoration--none',
+            'vads-u-display--block',
```
mjlumetta, line 149:
```suggestion
            'vads-u-display--block',
            'vads-u-color--white',
```

src/components/SideNav.scss
```diff
@@ -21,4 +22,16 @@ a.va-api-active-sidenav-link + ul.va-api-sidenav-sub-list {
...
   display: block;
-}
\ No newline at end of file
+}
+
+a.va-api-secondary-skip-nav {
+  position: absolute;
+  top: -1000px;
+  background: $va-api-color-blue-dark;
+  color: #ffffff !important;
```
mjlumetta, line 31:
```suggestion
```

using the VADS utility for this as well. this block looks great, super readable, thanks for moving that stuff out

mjlumetta (commented):
tiny suggestions. going to approve as soon as i can check the review instance (yes, i know that's kinda lazy)

mjlumetta (approved):
💯 🙌 🚀

PR #413, opened 2020-08-28
Add section for deactivated API release notes

mjlumetta:
### Description

completes [API-1790](https://vajira.max.gov/browse/API-1790). design ticket is [API-1495](https://vajira.max.gov/browse/API-1790).

adds a new section showing release notes for deactivated APIs. tests have been added using React Testing Library in support of [API-1317](https://vajira.max.gov/browse/API-1790).

### Process

  * added tests with React Testing Library
- [ ] Docs updated
  * no structural changes, tightened up an ARIA label or so


src/containers/releaseNotes/ReleaseNotesOverview.tsx
```diff
@@ -13,13 +13,23 @@ export default () => {
...
+          features and repair defects. We avoid doing so whenever possible but occasionally we need
+          to make breaking changes that require developers to modify their existing applications to
+          see the benefits of these features and fixes.
         </p>
         <p>
-          We recommend that developers periodically check this list for announcements of breaking changes and added features. Changes will also be announced via direct email whenever possible to addresses used to obtain developer keys for each API. Please <a href="https://developer.va.gov/support/contact-us">contact us</a> with any questions or to request support.
+          We recommend that developers periodically check this list for announcements of breaking
+          changes and added features. Changes will also be announced via direct email whenever
+          possible to addresses used to obtain developer keys for each API. Please{' '}
+          <a href="/support/contact-us">contact us</a> with any questions or to request support.
```
mjlumetta, line 25: making this `href` relative to the host is the only change in this file.

package.json
```diff
@@ -132,6 +134,7 @@
     "postcss-preset-env": "6.0.6",
     "postcss-safe-parser": "4.0.1",
     "prettier": "^1.16.4",
+    "pretty-format": "^26.4.2",
```
mjlumetta, line 137: issue with requiring `pretty-format` from `@testing-library/react`

src/App.tsx
```diff
@@ -17,9 +17,31 @@ import { history } from './store';
 import 'highlight.js/styles/atom-one-dark-reasonable.css';
 import './styles/base.scss';
 
-class App extends React.Component {
+export function getFlags() {
```
mjlumetta, line 20: make this reusable, esp for tests. can now mock any of the `get*Flags()` functions and `getAllApis()` and get correct flags in a test context. also can now unit-test this function itself more easily, although that is not done at this point.

src/apiDefs/data/categories.ts
```diff
@@ -16,42 +16,36 @@ import verificationApis from './verification';
...
   appeals: {
     apis: appealsApis,
-    buttonText: 'Get Your Key',
     content: appealsContent,
     name: 'Appeals API',
     properName: 'Appeals API',
   },
   benefits: {
     apis: benefitsApis,
-    buttonText: 'Get Your Key',
```
mjlumetta, line 26: been meaning to take this thingy out for a while. i think it used to be on the home page but hasn't been for like a year

src/apiDefs/query.ts
```diff
@@ -20,7 +20,7 @@ const getApiDefinitions = () => apiDefs;
 const getApiCategoryOrder = () => apiCategoryOrder;
 
 const getAllApis = (): IApiDescription[] => {
-  return Object.values(apiDefs).flatMap((category: IApiCategory) => category.apis);
+  return Object.values(getApiDefinitions()).flatMap((category: IApiCategory) => category.apis);
```
mjlumetta, line 23: wanted not to reference `apiDefs` directly in multiple places. need to prepare to use a real database and corresponding data repository layer for these definitions in the future.

src/components/PageHeader.tsx
```diff
@@ -3,32 +3,28 @@ import * as React from 'react';
...
-          <h2 className={classNames('vads-u-font-size--lg')}>
-            {this.props.description}
-          </h2>
-        }
-      </div>
-    );
-  }
-}
\ No newline at end of file
+export default function PageHeader(props: PageHeaderProps) {
```
mjlumetta, line 16: the functional change here, in addition to changing to a function, is adding the `containerId` property above to the outside `div`. it's for the purpose of using in an `aria-labelledby` so the accessible name is "Benefits API Release Notes" instead of "Release Notes". legacy of when we flipped the halo and header text on the Release Notes page.

src/containers/releaseNotes/CategoryReleaseNotes.tsx
```diff
@@ -0,0 +1,136 @@
+import AlertBox from '@department-of-veterans-affairs/formation-react/AlertBox';
```
mjlumetta, line 1: renamed this file to get away from vague use of the language of "page"

src/containers/releaseNotes/CategoryReleaseNotes.tsx
```diff
@@ -0,0 +1,134 @@
...
+interface ReleaseNotesCardLinksProps {
+  categoryKey: string;
+  apiCategory: BaseAPICategory;
+  flagName: string;
+}
+
+const ReleaseNotesCardLinks = (props: ReleaseNotesCardLinksProps) => {
+  const { apiCategory, categoryKey, flagName } = props;
+  const flags: { [apiId: string]: boolean } = getFlags()[flagName];
+  const apis: IApiDescription[] = apiCategory.apis.filter(api => flags[api.urlFragment]);
```
mjlumetta, line 26: we can't use `Flag` for this because we only want to show these card links if there is more than one API that will actually be shown on the page, not more than one API in the category, period. if the category has 4 APIs but 3 of them are disabled or deactivated, we don't want this component to return anything. `getFlags()` is a good substitute for `<Flag />`, now that it's public.

Eallan919:
Looking good! Can we get a space after the "," and contact us? under the relase note itself, please?
Screen Shot 2020-09-02 at 5 16 04 PM

mjlumetta:
@Eallan919 thanks for the catch. i think that whole release note should actually be update so it's one line in the markdown file and wraps properly... hmm

.env.production
```diff
@@ -8,3 +8,4 @@ REACT_APP_TRY_IT_OUT_ENABLED=false
 REACT_APP_SENTRY_DSN=https://4409b200977e4429bb210b601439ef0a@developer.va.gov/sentry/17
 REACT_APP_SENTRY_RELEASE=$SENTRY_RELEASE
 REACT_APP_SHOW_TESTING_NOTICE=false
+REACT_APP_ARGONAUT_API_ENABLED=false
```
charleystran, line 11: Wondering what this change means as it related to this PR, or is this cleaning up a missing setting?

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -2,94 +2,116 @@ import * as React from 'react';
...
-function SideNavApiEntry(api: IApiDescription) {
-  const dashUrlFragment = api.urlFragment.replace('_', '-');
+interface SideNavAPIEntryProps {
+  api: IApiDescription;
+  categoryKey: string;
+}
 
+function SideNavAPIEntry(props: SideNavAPIEntryProps) {
+  const { api, categoryKey } = props;
+  const dashUrlFragment = props.api.urlFragment.replace('_', '-');
```
anthony-kopczyk, line 22: I know this is just updated old code, but I am curious about this line. This seems to be here so we can have react components named with hyphens, and then have urls map easily?

I can't think of a way, but probably worth updating this line in the future to be more self-explanatory without needing to look through the rest of the code to get understanding.

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -2,94 +2,116 @@ import * as React from 'react';
...
-  const subNavLinks = () => {
-    return apiCategory.apis.map(api => {
-      if (apiCategory.apis.length > 1) {
-        return SideNavApiEntry(api);
-      } else {
-        return null;
-      }
-    });
-  };
+interface SideNavCategoryEntryProps {
```
anthony-kopczyk, line 49: I like the code clean up with this.

src/containers/releaseNotes/ReleaseNotesOverview.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+      expect(cardLink.getAttribute('href')).toBe('/release-notes/lotr');
+
+      cardLink = screen.getByRole('link', {
+        name: 'Sports API Learn more about throwing, running, and hitting',
+      });
+      expect(cardLink).toBeInTheDocument();
+      expect(cardLink.getAttribute('href')).toBe('/release-notes/sports');
+    });
+
+    it('does not render a card for a disabled category', () => {
```
anthony-kopczyk, line 64: I like this test. I feel like this kind of check is the kind that would usually get left out. Thanks for adding it!

src/containers/releaseNotes/ReleaseNotesOverview.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+      renderComponent();
+      expect(
+        screen.queryByRole('link', {
+          name: `Sports API ${fakeCategories.sports.content.shortDescription}`,
+        }),
+      ).toBeNull();
+    });
+
+    it('has a card link for deactivated APIs if there are any', () => {
```
anthony-kopczyk, line 88: Maybe rephrase this to 'has a card link for deactivated APIs when one exists'? I was trying to figure out if this test was somehow covering both conditions.

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -2,94 +2,116 @@ import * as React from 'react';
...
+            </SideNav>
+            <div className={classNames('vads-l-col--12', 'medium-screen:vads-l-col--8')}>
+              <Switch>
                 <Route exact={true} path="/release-notes/" component={ReleaseNotesOverview} />
                 <Route
                   exact={true}
-                  path="/release-notes/:apiCategoryKey"
-                  component={CategoryReleaseNotesPage}
+                  path="/release-notes/deactivated"
+                  render={DeactivatedReleaseNotes}
```
anthony-kopczyk, line 105: I am just now looking up the difference between using render and component in a Route path. I'd like a second set of eyes on this to say if it's good. - or if you could explain it for me why we are good to use render on DeactivedReleaseNotes here but use component for CategoryReleaseNotes below.

src/containers/releaseNotes/ReleaseNotes.test.tsx
```diff
@@ -0,0 +1,231 @@
...
+            ),
+          },
+        });
+
+        renderComponent();
+        const sideNav = screen.getByRole('navigation', { name: 'Release Notes Side Nav' });
+        expect(queryByRole(sideNav, 'link', { name: 'Sports API' })).toBeNull();
+      });
+
+      it('has an entry for deactivated APIs', () => {
```
anthony-kopczyk, line 108: Similar to ReleaseNotesOverviewTest this could be reworded to be something like 'has an entry for deactived APIs when there are deactived APIs'

anthony-kopczyk (commented):
This is good. I have a couple questions, probably mostly because I'm new. There is language on two tests that I think could be reworded more explicitly. I'm also unsure about the ReleaseNotes Route, render/component attributes. Someone that knows the app more would be better to look at that specific piece.

I don't see anything that needs changed right away but don't want to approve until I know about that React Route component render/component stuff. - granted, I could probably approve. Even if it should be the other way I don't see it breaking anything.

.env.production
mjlumetta: for other reviewers: the Argonaut API's release notes aren't very meaningful because they're from an earlier iteration of the release notes where we were less consistent about updating them. as a result, when we decided to show release notes for deactivated APIs, UX asked us to hide those, for now at least. you can see the release notes in the source file or in the dev and staging review instances. (i'm choosing to leave it on by default for those other environments - i don't want to set a precedent for getting rid of inconvenient things from deactivated APIs wholesale by marking them disabled in addition to deactivated.)

src/App.tsx
mjlumetta: yeah, i think that's a good question, to which i don't have a good answer right now. would think that moving the data this calculation is based on out of the pseudo-database and into a real external database would influence that decision heavily - maybe pointing to putting something like this in the Redux store? thanks for raising it, though, and think we should keep an eye out going forward.

src/components/PageHeader.tsx
```diff
@@ -3,32 +3,27 @@ import * as React from 'react';
 
 import './PageHeader.scss';
 
-interface IPageHeaderProps {
+interface PageHeaderProps {
```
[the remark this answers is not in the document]
mjlumetta: yeah, definitely - should've left a comment on this, because this question has come up before recently and there's a big difference between a yes and no answer. yes, we decided to go with no prefix around the time the backend got stood up as its own Node app, and we decided to be consistent across the front and back end. apologies for making an assumption here!

src/components/PageHeader.tsx
mjlumetta: yeah, exactly. for most components that have a `PageHeader`, if they're doing an `aria-labelledby`, they want to refer directly to the h1. but, in this case, the halo is actually more descriptive of the specific page, so we need to go up a level to get the category name in there too.

src/containers/releaseNotes/ReleaseNotes.tsx
mjlumetta: yeah, i think that's a good note. i'm not 100% sure why this is the way that it is (unless it's prettier anchors, i guess?). so yeah, i totally agree.

src/containers/releaseNotes/ReleaseNotes.test.tsx
mjlumetta: yup, that's a good point, thanks! will clarify that language

src/containers/releaseNotes/ReleaseNotesOverview.tsx
```diff
@@ -33,6 +45,11 @@ export default () => {
             </Flag>
           );
         })}
+        {deactivatedCategory.apis.length > 0 && (
+          <CardLink name={deactivatedCategory.name} url="/release-notes/deactivated">
+            This is a repository for deactivated APIs and related documentation and release notes.
```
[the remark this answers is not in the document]
mjlumetta: good question - it's the latter. sometimes the release notes link out to PRs and whatnot, but this isn't intended to be in the technical version control sense.

src/containers/releaseNotes/ReleaseNotes.tsx
mjlumetta: wow, great catch. i'm not really sure when/why i made this `render`, it should probably be `component`. i believe they're functionally equivalent in this case, but `render` is more frequently used for inline functions.

anthony-kopczyk: When I was looking it up I saw something about one of them making new instances or something, but I'm not sure specifics or how these components are used so not sure at all. haha.

mjlumetta: yeahhh fair, i think the differences are pretty hard to parse out. i think the idea is that `<Route component={() => <div>hello!</div>} />` will mount a new component each time it gets rendered because of how it treats that function internally (maybe because it's anonymous? i don't know either, would have to research it more to get what's happening there). but if you do `<Route render={() => <div>hello!</div>} />`, React Router treats it differently and therefore handles that case.

anthony-kopczyk (approved):
Looks good. The previous reviews were addressed.

charleystran (approved):
LGTM - Good Job on this!

PR #414, opened 2020-08-31
Replace jest-fetch-mock with Mock Service Worker's Node implementation

mjlumetta:
### Description

in support of [API-1317](https://vajira.max.gov/browse/API-1317).

this test changes the HTTP request mocking for Apply and Contact Us tests to use [Mock Service Worker's](https://mswjs.io/docs/) `setupServer` rather than using `jest-fetch-mock`. the motivation for this change is to mock requests in a way that is agnostic to the specific mechanism for making the HTTP request, as well as being a bit more expressive in terms of what it can offer (mocking multiple endpoints/resources in a more natural way).

### Process

  * the whole ticket is updating tests
- [ ] Docs updated
  * going to do a wrap-up docs PR for API-1317
  * n/a, no UI changes


src/actions/apply.spec.ts
```diff
@@ -51,7 +62,19 @@ describe('submitForm', () => {
...
-    fetchMock.mockReject(new Error('Network Failure'));
+    server.use(
+      rest.post(
+        constants.APPLY_URL,
+        (
+          req: MockedRequest,
+          res: ResponseComposition,
+          context: typeof restContext,
+        ): MockedResponse => {
+          return res(context.status(500, 'KABOOM'));
```
mjlumetta, line 73: the docs say you should be able to omit the status text, but i couldn't get it to work after semi-extensive debugging. extremely annoying but fine for tests.

src/containers/support/SupportContactUsForm.test.tsx
```diff
@@ -1,94 +1,96 @@
...
 
     const submitButton = component.find('.usa-button-primary');
     expect(submitButton.length).toEqual(1);
     expect(submitButton.hasClass('usa-button-disabled')).toBe(false);
-  
+
+    // submitting the form should call the done() callback
     submitButton.simulate('click');
-    
-    process.nextTick(() => {
```
mjlumetta, line 55: this broke once i switched things over to the non-mocked version. unsure why, think it has to do with the Node loop, gonna try to figure it out.

src/actions/apply.spec.ts
anthony-kopczyk: Sounds good, but curious - Is this a typescript typing issue where it doesn't have that parameter as optional or does it fail during runtime?

src/actions/apply.ts
```diff
@@ -97,36 +97,36 @@ const apisToList = (apis: IApiList) => {
...
-        headers: {
-          accept: 'application/json',
-          'content-type': 'application/json',
-        },
-        method: 'POST',
+    const request = new Request(constants.APPLY_URL, {
+      body: JSON.stringify(applicationBody),
+      headers: {
+        accept: 'application/json',
+        'content-type': 'application/json',
```
anthony-kopczyk, line 125: Does this need to be capitalized like in a sent HTTP request - 'Content-Type'?

src/types/constants/index.ts
```diff
@@ -8,6 +8,13 @@ export type SET_INITIAL_VERSIONING = typeof SET_INITIAL_VERSIONING;
 
 export const CURRENT_VERSION_IDENTIFIER = 'current';
 
+export const APPLY_URL = `${
+  process.env.REACT_APP_DEVELOPER_PORTAL_SELF_SERVICE_URL
+}/internal/developer-portal-backend/developer_application`;
+export const CONTACT_US_URL = `${
+  process.env.REACT_APP_DEVELOPER_PORTAL_SELF_SERVICE_URL
+}/internal/developer-portal-backend/developer_application`;
```
anthony-kopczyk, line 16: This looks the same as the apply url. Should it be /apply or something similar?

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -167,31 +184,31 @@ export default class SupportContactUsForm extends React.Component<ISupportContac
...
-            && (!this.state.lastName.validation && this.state.lastName.value)
-            && (!this.state.email.validation && this.state.email.value)
-            && (!this.state.description.validation && this.state.description.value));
+    return !(
+      !this.state.firstName.validation &&
+      this.state.firstName.value &&
+      (!this.state.lastName.validation && this.state.lastName.value) &&
+      (!this.state.email.validation && this.state.email.value) &&
+      (!this.state.description.validation && this.state.description.value)
+    );
```
nathanbwright, line 193: Can this conditional be simplified?

anthony-kopczyk: I'm not thinking of anything for making the content of the method simpler, but maybe rename the method as isFormValid or something to reflect the check itself would help with readability and add enough context to make the content easier to understand?

src/actions/apply.spec.ts
mjlumetta: it fails during runtime. it also uses the default `'OK'` instead of `'Internal Server Error'` or what have you.  i could not figure out why, which is annoying, but i was planning to let this one slide.

src/types/constants/index.ts
mjlumetta: oh shoot good catch, thank you, i should have tested this

src/actions/apply.ts
mjlumetta: so, HTTP headers are [case insensitive](https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive) - which, like, had to look that up because i didn't remember. i'm okay with either as long as they're consistent in this object, but i would default to not changing anything where the changes were all from Prettier.

src/containers/support/SupportContactUsForm.tsx
mjlumetta: okay, fixed this one. used Anthony's suggestion about `isFormValid`, and also wrote a tiny helper for each field.

for the `validation` property, it's a string with an error message. check out `src/utils/validators.ts` and the `validatePresence` and `validateEmail` functions. you can see those used in the `onChange` handlers.

src/actions/apply.ts
anthony-kopczyk: What?! That's insane! I had no idea. haha. Yeah. We can leave as is, especially if the changes came from Prettier.

src/containers/support/SupportContactUsForm.tsx
anthony-kopczyk: The helper really cleans it up as well.

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -166,32 +183,42 @@ export default class SupportContactUsForm extends React.Component<ISupportContac
     };
   }
 
-  private get disabled(): boolean {
-    return !((!this.state.firstName.validation && this.state.firstName.value)
-            && (!this.state.lastName.validation && this.state.lastName.value)
-            && (!this.state.email.validation && this.state.email.value)
-            && (!this.state.description.validation && this.state.description.value));
+  private get isFormValid(): boolean {
+    const validateField = (field: IErrorableInput): boolean => !!field.value && !field.validation;
```
anthony-kopczyk, line 187: Is the !! on purpose? Or maybe remnants of a previous version of the code?

anthony-kopczyk (changes requested):
It all looks good to me except the sections where there is '!!'.

src/containers/support/SupportContactUsForm.tsx
anthony-kopczyk:
```suggestion
    const validateField = (field: IErrorableInput): boolean => field.value && !field.validation;
```

mjlumetta: it changes the type signature of the function from `(field: IErrorableInput) => boolean` to `(field: IErrorableInput) => boolean | string`, so it's a no-go on that. tried that first

anthony-kopczyk: Didn't even think about return type. Pulled it down a fiddled with it. There might be a more verbose way to make it so we don't need  `!!` but it's immediately obvious that the typing isn't going to place. This is good. No changes needed.

anthony-kopczyk (dismissed):

src/containers/support/SupportContactUsForm.tsx
mjlumetta: circling back to this: would you prefer if it were `!!(field.value && !field.validation)`? might be equally confusing, but then you can kinda see that's it's forcing the whole result to a boolean. can push that if there are more changes based on other feedback.

anthony-kopczyk: I think that way might actually be more confusing for me. haha. Because then I'd be looking at the entire output rather than the `field.value` piece.

I did some googling on this, and it seems like the `!!field` approach is pretty standard. I just haven't seen it or forgot I've seen it. I think we can leave it as is and if it comes up again in the future while someone is coding we can revisit the it.

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -166,32 +183,42 @@ export default class SupportContactUsForm extends React.Component<ISupportContac
...
+      validateField(description)
+    );
+
+    // return !(
+    //   !this.state.firstName.validation &&
+    //   this.state.firstName.value &&
+    //   (!this.state.lastName.validation && this.state.lastName.value) &&
+    //   (!this.state.email.validation && this.state.email.value) &&
+    //   (!this.state.description.validation && this.state.description.value)
+    // );
```
nathanbwright, line 203: I'm assuming this is just a left over from a previous iteration, right?

nathanbwright (approved):
Don't have a ton of experience with react but it looks good to me.

PR #416, opened 2020-09-02
Typescript wrangling in preparation for adding Cypress

mjlumetta:
### Description

this PR supports [API-1317](https://vajira.max.gov/browse/API-1317). it is a step to prepare for adding Cypress to by scoping Typescript config more granularly. part of the reason for this is that Jest globals and Chai globals by way of Cypress conflict. there's another way around that, but then we couldn't co-locate our integration tests with our source files, and i'm not a huge fan of that.

making this its own PR so as to keep changes small.

this PR also has some advantages for local development. when running the dev server, you won't get compilation errors for test problems.

### Process

  * n/a - tests should still run, however
- [ ] Docs updated
  * n/a

### Requested feedback

do these changes accomplish what i'm claiming? do they pose other problems i haven't considered or tested for?


rtravitz:
How does the test tsconfig file get applied to the test run?

mjlumetta:
> How does the test tsconfig file get applied to the test run?

via the `ts-jest` global in `*.jest.config.js`. 

for other readers: probably hold off on reviewing this PR for noww. experiencing some issues that i think might be specific to VSCode, but need to understand what's going on better.

mjlumetta:
i've got this on a different branch now. closing and will re-open a new PR when it's ready.

PR #417, opened 2020-09-02
Upgrade Webpack plugins to fix high-severity vulnerabilities

mjlumetta:
### Description

fixes 2 high-severity vulnerabilities:

* fixes the Dependabot alert requiring that `dot-prop` be at least version 4.2.1 by upgrading `optimize-css-assets-webpack-plugin` to 5.0.4 (published Sunday! so we're on top of it, sort of)
* removes `terser-webpack-plugin` and `uglifyjs-webpack-plugin` to fix `serialize-javascript` vulnerability. both of those plugins were unused.

ran `npm run build` and it looks fine, so we'll see what it looks like in the deployed review instances. this is only relevant to the compiled production stylesheets, so testing locally is not meaningful for this change.

ticket is [API-1919](https://vajira.max.gov/browse/API-1919).

### Process

- [ ] Docs updated

n/a to all of the above


ediiotero (approved):
Looks good, nothing to add. Good job!

rtravitz (approved):
Whoops, thought I got all of the unused plugins when I pruned the dependencies recently. LGTM ✨

PR #420, opened 2020-09-03
add a check to the renderSwaggerUI function to make sure the element …

ediiotero:
…is ready before rendering

### Description
https://vajira.max.gov/browse/API-1844

Fixing a bug that is triggered when quickly clicking a swagger ui documentation page to an overview page and back again on a slow internet connection.

### Process
To reproduce bug set Network to:
Download: 200kb, Upload: 40kb, Latency: 1000ms


charleystran (approved):
LGTM - Good work on this

mjlumetta:
@ediiotero can you leave this ticket open? want to test out that async lifecycle hook issue.

mjlumetta:
have verified that async/await in lifecycle methods is okay and left [a comment](https://vajira.max.gov/browse/API-1844?focusedCommentId=324836&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-324836) on the ticket. based on the articles linked in the comment, we should maybe add a try/catch to those lifecycle methods as well in order to catch other errors that may not be due to Swagger UI next time we're in the `SwaggerDocs` component, but this PR solves the problem we wanted to solve.

PR #422, opened 2020-09-11
Api 1978 address validation faq

charleystran:
### Description
This PR adds a Question/Answer for how to gain address to the address validation sandbox API.  Text has been approved in this ticket https://vajira.max.gov/browse/API-225 and the corresponding https://vajira.max.gov/browse/API-1978.


rtravitz:
Showing zero files changed on this one. What am I missing?

Eallan919:
same...I don't see an update with the new copy

src/containers/support/SupportFAQ.tsx
```diff
@@ -75,6 +75,26 @@ const developmentQuestions: ISupportQuestion[] = [
     ),
     question: 'What kind of data can I get from the APIs? Do you have any example scenarios for Health, Benefits, Facilities or Veteran Verification?',
   },
+  {
+    answer: (
+      <p>
+        The Address Validation API is for internal VA use only and is not listed on the developer portal. To begin development in the sandbox environment, request a developer <a href="https://developer.va.gov/apply">API key for the Facilities API</a>. Once finished, send an email to <a href="mailto:api@va.gov?subject=Request%20for%20Sandbox%20Access%20to%20Address%20Validation%20API" target="_BLANK" rel="noopener noreferrer">api@va.gov</a> with the subject line "Request for Sandbox Access to Address Validation API." In the email, include the following:
+
+        <ul>
```
mjlumetta, line 83: nesting this `ul` inside the `p` isn't allowed, apparently. there's a console warning from React:

```
Warning: validateDOMNesting(...): <ul> cannot appear as a descendant of <p>.
```

here's [a StackOverflow post](https://stackoverflow.com/questions/7168723/unordered-list-in-a-paragraph-element) to the same effect. so seems like we'd have to go with paragraph, unordered list, paragraph within a `React.Fragment`. that sound right?

also, would you mind wrapping the text for some of the long lines? we should be able to do that because we're in JSX without breaking it. to be fair, i always forget which contacts you can't wrap in, so maybe that's not correct.

src/containers/support/SupportFAQ.tsx
```diff
@@ -75,6 +75,26 @@ const developmentQuestions: ISupportQuestion[] = [
...
+          <li>Email address used for the sandbox Facilities API key request</li>
+          <li>Name of team/project</li>
+          <li>Consumer name (if different from team/project)</li>
+          <li>Expected call volume (our default rate limit is 60 requests per minute)</li>
+          <li>Desired production date</li>
+          <li>Any IP addresses or subnets outside the VA intranet from which you require developer access (permitted in sandbox only)</li>
```
anthony-kopczyk, line 89: The text _outside the VA internet_ is in italics on the associated Jira ticket.

I'm being a stickler because I'm new. :)

src/containers/support/SupportFAQ.tsx
```diff
@@ -75,6 +75,26 @@ const developmentQuestions: ISupportQuestion[] = [
...
+          <li>When indicating your desired production date, note that we require at least one week's notice before a demo can be scheduled, and that it may take up to a week after the demo to grant you the sandbox key for the Address Validation API. </li>
```
anthony-kopczyk, line 90: The Jira ticket shows this as being a separate paragraph rather than part of the bulleted list.

anthony-kopczyk (changes requested):
Cross referenced with the main Jira ticket. Some formatting needs updated to match, or maybe the Jira ticket's description needs updated.

src/containers/support/SupportFAQ.tsx
mjlumetta, line 90: should this be a list item? looks like it's formatted as a paragraph in the ticket, but maybe that's not right.

src/containers/support/SupportFAQ.tsx
mjlumetta, line 89:
```suggestion
          <li>Any IP addresses or subnets <em>outside the VA intranet</em> from which you require developer access (permitted in sandbox only)</li>
```

looks like that might be italicized?

mjlumetta (changes requested):
couple of copy questions, but the main thing is that it looks like you can't nest a `ul` inside a `p`, so we're probably going to have to make these elements siblings. hopefully this is a quick fix

src/containers/support/SupportFAQ.tsx
charleystran: I wouldnt take the ticket as the markup requirements as it was just written by support over engineers. I think we can get a final review once its up and make adjustments.

anthony-kopczyk (approved):

anthony-kopczyk:
I went ahead and approved my review because Mike's review covers the potential changing points in mine --- although it did now count it as an approval, which I wasn't expecting.

src/containers/support/SupportFAQ.tsx
```diff
@@ -75,6 +75,34 @@ const developmentQuestions: ISupportQuestion[] = [
...
+        <br />
+        When indicating your desired production date, note that we require at least one week's notice before a demo can be scheduled, and that it may take up to a week after the demo to grant you the sandbox key for the Address Validation API.
+        <br />
+        <br />
+
+        We will respond to your request within 2 business days.
+      </p>
+             ),
+    question: 'How do I get sandbox access to the Address Validation API (VA internal only)?',
+  },
```
ediiotero, line 105: Looks good but had a question:
I understand the p tag shouldn't have li children but is you made them div's you should be able to use an unordered list. Just curious if p tags had to be used for every answer or if there is something I'm missing as a pattern or accessibility?

ediiotero (commented):
Just had a question about p tags vs divs and using ul's

src/containers/support/SupportFAQ.tsx
```diff
@@ -75,6 +75,34 @@ const developmentQuestions: ISupportQuestion[] = [
     ),
     question: 'What kind of data can I get from the APIs? Do you have any example scenarios for Health, Benefits, Facilities or Veteran Verification?',
   },
+  {
+    answer: (
+      <p>
+        The Address Validation API is for internal VA use only and is not listed on the developer portal. To begin development in the sandbox environment, request a developer <a href="https://developer.va.gov/apply">API key for the Facilities API</a>. Once finished, send an email to <a href="mailto:api@va.gov?subject=Request%20for%20Sandbox%20Access%20to%20Address%20Validation%20API" target="_BLANK" rel="noopener noreferrer">api@va.gov</a> with the subject line "Request for Sandbox Access to Address Validation API." In the email, include the following:
+
+        <br />
```
mjlumetta, line 83: i was picturing this going the other way in terms of using that list but taking it out of the paragraph and making it next to the paragraphs. so 2 paragraphs and a list nested inside some container. this is the element hierarchy i'm imagining that could fix that constraint and preserve the list:

```
<React.Fragment>
  <p />
  <ul />
  <p />
</React.Fragment>
```

mjlumetta (commented):
had a suggestion about the list/paragraph issue, but the issue's fixed right now. would also want to get a review from UX/product/Ryan Link to give the final approval on the content like you'd talked about earlier.

src/containers/support/SupportFAQ.tsx
charleystran: this is done

src/containers/support/SupportFAQ.tsx
charleystran: I think there is a lot to be refactored here especially as we move to dynamic content.So I dont have a great answer, but you bring up a good point as we iterate on this.

ryanlink (dismissed):
can we somehow bullet the AV list items?  otherwise LGTM

mjlumetta (approved):
💯 🚀

PR #425, opened 2020-09-15
[DO NOT DELETE, FOR API-2350] Typescript wrangling 

mjlumetta:
#### Notes

- i currently can't figure out how to break the `jest-environment-puppeteer` `page` global. want to figure that out.
  * actually, i don't know if i can break any Jest globals. this appears to be a problem i need to resolve.
  * they're not available in source files still, so that's good, at least.
  * the way i would've _thought_ that we could break this stuff would be removing `compilerOptions.types`, so that it falls back to `tsconfig.json`, which has `types: ["node"]` in order to exclude Jest globals.
  * tried removing `compilerOptions.types` and then breaking down the `include` array to have fewer files, because i was able to do this earlier
    - with only `src/**/*.spec.ts` and `src/**/*.test.ts`, running the tests broke with Jest globals undefined.
    - adding `src/**/*.test.tsx` caused the tests to pass. why?
      * causes the `src/**/*.spec.ts` and `src/**/*.test.ts` tests to pass as well!
      * seems to imply that somewhere in `src/**/*.test.tsx` Jest globals get imported directly or indirectly
      * have traced this semi-reliably to the addition of `@testing-library/react` and `@testing-library/jest-dom`. commenting out all the tests in the files importing either of those packages caused it to break, so i think we can attribute it to one of those.
- i can't break the Jest intellisense in VSCode. not sure if this is VSCode specific or not. either way, don't necessarily think this is blocking if the tests themselves blow up when Jest types can't be resolved.
- now i'm having trouble resolving the `src/__mocks__/news.test.yml` file in the News tests.
  * when `src/types/**/*` is not included in `tsconfig.test.json`, the error says "Cannot find module '../__mocks__/news.yml' or its corresponding type declarations." when `src/types/**/*` is included in `tsconfig.test.json`, the error says "Property 'sections' does not exist on type 'typeof import("*.yml")'."
    - this difference seems to indicate that `src/types/**/*` does help it find the module but doesn't understand the type.
    - the reason i initially included `src/types/**/*` is that tests in `src/apiDefs` that included markdown components weren't able to resolve the `MarkdownComponent` type in those tests without that included. the difference between the two type resolutions is that the MDX components are being imported in the source files and the YAML is being imported directly into the test file. the YAML with the real news data in `News.tsx` is fine.
    - why does the module resolve in `News.tsx` and with the old `tsconfig.test.json` but not here?
    - **edit:** it started working now and i'm not sure why. i think it was because i broke `yml.d.ts` and then restored the version from `master`. i'm dumb.
    - **edit #2:** actually, it works fine in the actual Jest run but not in VSCode?
  * solution, do not ask me why this works: adding an `index.d.ts` to `src/__mocks__` and redeclaring the "*.yml" module fixed this problem.
    - hypothesis: when importing in VSCode, the ambient types don't get picked up when the non-typescript module (not `.ts`, `.tsx`, or `.d.ts`, plus `.js` and `.jsx` with `"allowJs": true`) is imported directly into a test file. when MDX components were required in tests, they were never _directly_ imported from a `.mdx` file into a `.test.tsx` file. they were imported into a TS file in the main `tsconfig.json`. but maybe once those ambient typings in `yml.d.ts` are picked up for the main `tsconfig.json`,  they don't get applied to files that aren't included in that project. interesting. confusing.

### Process

- [ ] Docs updated

### Manual Test Plan

some tests may require that the tester alter the source code to introduce mistakes, so you will have to edit the code rather than running it only to test these changes.

#### General Goals for Testing

- need to make sure type errors get raised appropriately
  * why? if a file with real type errors doesn't raise them, then we don't get the benefits of Typescript
  * in other words, if there is not a TS config file covering a given file, it won't ever get type-checked and will get treated as plain Javascript in its execution context (everything is an `any`)
    - ... at least, i'm _pretty sure_ that's what happens. feel free to correct me, would love to learn.
  * want to catch unresolved types rather than have them silently be treated as Javascript
- need to make sure types get resolved correctly in all contexts (local dev server, test, build)
  * should be a matter of running each process and making sure everything compiles
- need to make sure that type imports are scoped correctly
  * Jest globals (`jest`, `describe`, `expect`, `it`, etc) should not be available in source files but should be available in test files
  * type errors in test files do not prevent the dev server from compiling
- need to make sure Intellisense still works for the cases above
  * i'm using VSCode, and testing in other editors/local dev environments would be extremely beneficial

#### Unit Tests

- test that type errors get raised
  * with a completely fake type: add `const unfoundType: FakeTypeIDontExist = 'fake type'` to a test file
  * with a real but incorrectly used type: import any type from the portal and assign a value that does not match the type.
  * for each of those two cases, run the test file and ensure that Jest blows up because it can't resolve the type. no tests should execute.
  * check that Intellisense catches those type errors
  * check that the dev server compiles and works with those type errors in the test file
- test that the tests can run with the types scoped as is with `npm run test:unit`
  * if the type errors get raised appropriately, as illustrated by the previous test cases, then running the tests ensures that all types do get resolved correctly. most significantly, that means Jest globals, which are not in source files, are present in test files.

### Requested feedback

- does it work? please try to break these changes.
- how does it work in your editor/dev environment?


src/actions/apply.spec.ts
```diff
@@ -1,4 +1,4 @@
-import 'jest';
+import 'jest-fetch-mock';
```
mjlumetta, line 1: needed to resolve `fetchMock` as a type. could also achieve this by adding `jest-fetch-mock` to `compilerOptions.types` in `tsconfig.test.json`, i believe. either way, shouldn't matter because `jest-fetch-mock` is going out in #414.

src/App.e2e.ts
```diff
@@ -5,7 +5,6 @@
   better way to ensure the site is rendering correctly, so we may add a monitoring
   specific test in the future.
 */
-import 'jest';
```
mjlumetta, line 8: removing explicit `jest` imports since these files should always execute in the Jest context. can revert these deletes if we want to allow the imports.

src/containers/support/SupportContactUsForm.test.tsx
```diff
@@ -1,94 +1,93 @@
 import { mount } from 'enzyme';
-import 'jest';
+import 'jest-fetch-mock';
```
mjlumetta, line 2: same thing as in `apply.spec.ts` with spying on `global.fetch`

tsconfig.json
```diff
@@ -16,20 +16,18 @@
     "noImplicitAny": true,
     "strictNullChecks": true,
     "suppressImplicitAnyIndexErrors": true,
-    "noUnusedLocals": true
+    "noUnusedLocals": true,
+    "types": ["node"]
```
mjlumetta, line 20: by default, `types` would resolve to any package in `node_modules/@types`, but by explicitly specifying `node` only, we don't get Jest globals anymore. https://www.typescriptlang.org/tsconfig#types

tsconfig.json
```diff
@@ -16,20 +16,18 @@
...
   },
+  "include": ["src/**/*"],
```
mjlumetta, line 22: there's not really any reason to have non `src` files in the Typescript project, so explicitly specify `include`. as a result, we can remove all those non `src` files from `exclude`. (can also do `node_modules`, too, i'm pretty sure, but i didn't because was still testing at the time.)

tsconfig.json
```diff
@@ -16,20 +16,18 @@
...
     "src/e2eHelpers.ts",
-    "*.jest.config.js",
-    "jest-puppeteer.config.js",
-    "SitemapBuilderWebpackPlugin.js"
+    "src/**/*.test.tsx",
+    "src/**/*.spec.ts",
+    "src/**/*.test.ts",
+    "src/**/*.e2e.ts",
+    "src/**/*.accessibility.ts",
+    "src/visualRegressionTest.ts"
```
mjlumetta, line 31: basically, this says to exclude every test file

tsconfig.test.json
```diff
@@ -1,6 +1,15 @@
...
+    "types": ["node", "jest"]
+  },
+  "include": [
+    "setupJest.ts",
+    "src/**/*.spec.ts",
+    "src/**/*.test.ts",
+    "src/**/*.test.tsx",
+    "src/types/**/*"
+  ],
+  "exclude": []
```
mjlumetta, line 14: override `exclude` because pretty much all the files we need in the test project are excluded by `tsconfig.json`

anthony-kopczyk:
@mjlumetta I think at one point there was discussion of closing this PR. This has quite a few merge conflicts on it now so wanted to check before it gets even more.

mjlumetta:
@anthony-kopczyk i think you're right. in spite of the title, i'm definitely not merging this. on the other hand, i do want to have this handy so i can remember what i did when i get back to API-2350, so i think i might try to leave it open if that's alright with you?

anthony-kopczyk:
Sounds good to me!

mjlumetta:
closing this one – it's out of date enough that if someone needs it they can look up the closed version.

PR #427, opened 2020-09-18
Install ESLint and set up initial ESLint config

mjlumetta:
### Description

initial PR to support [API-1323](https://vajira.max.gov/browse/API-1323).

changes:

* install [ESLint](https://eslint.org/docs/user-guide/getting-started). i started with [these migration instructions](https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration).
  - i also installed `@typescript-eslint/parser` and `@typescript-eslint/eslint-plugin`, per those instructions.
* i ran `npx tslint-to-eslint-config`, a utility created for the purpose of this migration. this tool generated an initial `.eslintrc.js` file.
  - based on the recommendation of the tool, i also installed a few other packages to support the generated `.eslintrc.js`.
    * `@typescript-eslint/eslint-plugin-tslint`
    * `eslint-plugin-import`
    * `eslint-plugin-jsdoc`
    * `eslint-plugin-prefer-arrow`
    * `eslint-plugin-react`
  - it's possible that any part of this generated config is not necessary for our project. i will add
* i added a `.eslintignore`. with the VSCode ESLint plugin, rather than assuming that the `parserOptions.project` defined as `tsconfig.json` contains all files to be linted, it tries to lint all TS/JS files that aren't explicitly ignored. don't think this is fundamental to ESLint but seems useful to have configured.
  - if we would like to lint these later, we may want to create another ticket to add a new `eslint` project. it shouldn't go in this ticket, though.
  - i also cut down the TS config project scope by adding an explicit `"include": ["src/**/*"]` to make this easier to reason about.
* i added an `npm run eslint` script to temporarily have that script while not affecting the `npm run lint` script and its usage for CI/dev environment linting while the ESLint migration is not yet complete.

these changes should not affect CI, since the current linting check has not changed. it will not affect local development except perhaps insofar as ESLint and TSLint plugins may attempt to run at the same time. if you run into issues, i recommend disabling any ESLint plugins in this repo when you are not working explicitly with the ESLint changes. obviously this depends heavily on your dev environment setup, so please let me know what does or does not work for you. the most important thing i'm interested in verifying is that these issues don't blow up your environment.

subsequent PRs will fix the ESLint issues as they stand and refine the config.

i will post the results of the `tslint-to-eslint-config` run in a separate comment.

### Process

- [ ] Docs updated

n/a to all of the above

### Requested feedback

* is it acceptable to merge this in without fixing issues or completing replacing TSLint?
* what issues can you identify with your editor and linting plugins, either trying to use TSLint or testing the new ESLint setup?
* what linting rules would you like to see us standardize on and/or lint for (if they haven't been linted well previously)?


package.json
```diff
@@ -8,6 +8,7 @@
     "build": "node scripts/build.js",
     "lint": "./node_modules/.bin/tslint --config tslint.json --project tsconfig.json",
     "lint:ci": "./node_modules/.bin/tslint --config tslint.ci.json --project tsconfig.json -t junit > lint-results.xml",
+    "eslint": "./node_modules/.bin/eslint --ext .ts src",
```
anthony-kopczyk, line 11: Do we want to add eslint:ci within here and reconfigure the pipeline or wait until this has been merged and seems to be good for all developers? and then I guess on top of that, when do we want to remove the old lint commands?

package.json
```diff
@@ -144,7 +153,7 @@
     "ts-jest": "^25.5.1",
     "ts-loader": "^5.3.0",
     "tsconfig-paths-webpack-plugin": "^3.2.0",
-    "tslint": "^5.12.0",
+    "tslint": "^5.20.1",
```
anthony-kopczyk, line 156: Kind of goes with the other comment, of when we'll clean up any remaining tslint things - should this be removed?

tsconfig.json
```diff
@@ -18,18 +18,6 @@
...
-    "webpack",
-    "jest",
-    "src/setupTests.ts",
-    "src/e2eHelpers.ts",
-    "*.jest.config.js",
-    "jest-puppeteer.config.js",
-    "SitemapBuilderWebpackPlugin.js"
-  ]
+  "include": ["src/**/*"],
+  "exclude": ["src/e2eHelpers.ts"]
```
anthony-kopczyk, line 22: Are these changes needed for the eslint configuration/still used?

.eslintrc.js
```diff
@@ -0,0 +1,227 @@
...
+            message:
+              'Avoid using the `Function` type. Prefer a specific function type, like `() => void`.',
+          },
+          Boolean: {
+            message: 'Avoid using the `Boolean` type. Did you mean `boolean`?',
+          },
+          Number: {
+            message: 'Avoid using the `Number` type. Did you mean `number`?',
+          },
+          String: {
```
anthony-kopczyk, line 62: I didn't know there were capitalized versions of these types. Learn something new every day. haha.

anthony-kopczyk (commented):
This looks good from what I can judge (I haven't really used eslint or did much configuration with tslint before). The only question is whether we're taking out tslint things now or waiting. If we're leaving tslint things in here, we should be able to approve this - but I did just get a bunch of linting errors.

I just ran `npm run eslint` and got a ton of lint errors. I think it mostly revolves around not having some rules downloaded, or missing dependencies after running `npm install`. 

```
Warning: React version not specified in eslint-plugin-react settings. See https://github.com/yannickcr/eslint-plugin-react#configuration .

Could not find implementations for the following rules specified in the configuration:
    jsx-no-string-ref
    jsx-self-close
    react-a11y-anchors
    react-a11y-aria-unsupported-elements
    react-a11y-event-has-role
    react-a11y-image-button-has-alt
    react-a11y-img-has-alt
    react-a11y-lang
    react-a11y-meta
    react-a11y-no-onchange
    react-a11y-props
    react-a11y-proptypes
    react-a11y-required
    react-a11y-role
    react-a11y-role-supports-aria-props
    react-a11y-tabindex-no-positive
    react-a11y-titles
```

tsconfig.json
mjlumetta: i don't think that this is necessary for the ESLint changes in particular; i added it while i was trying to resolve an issue that turned out to be resolved with the `.eslintignore`. however, this is something that we should probably be fixing anyway unless we're actually trying to use TS outside of `src`. (note: there are currently two TS files outside of `src`. `setupJest.ts` i think we can get rid of once `enzyme` is dead, and i'm pretty sure `config/jest/imageSnapshot.ts` could be JS - there are no types used.) fixing this removed a linting error from `tsconfig.json` about files being outside of `rootDir`:

```
File '/Users/mikelumetta/developer-portal/setupJest.ts' is not under 'rootDir' '/Users/mikelumetta/developer-portal/src'. 'rootDir' is expected to contain all source files.
```

here's the error i was trying to fix, in case anyone's curious. basically it seems like ESLint will complain if it finds a file that it thinks it should be able to lint (JS, TS, JSX, TSX, i think JSON, etc) outside of `parserOptions.project`. ... so actually, the TS config change was probably _not_ helpful for fixing that issue, but whatever.

```
Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: .eslintrc.js.
The file must be included in at least one of the projects provided.
```

johanna-adhoc:
Hi Mike!

Can we add the react hook rules as part of this PR? [https://reactjs.org/docs/hooks-rules.html](https://reactjs.org/docs/hooks-rules.html)

Maybe as a follow up item. Is there a way that we can have the eslint checks active only for certain folders as part of the build process?

For example the eslint checks will be active for folder src/components/crisisLine and have tslint checks to ignore src/components/crisisLine. So both checks are kinda active but never for the same files.

.eslintrc.js
```diff
@@ -0,0 +1,227 @@
...
+    'one-var': ['error', 'never'],
+    'prefer-arrow/prefer-arrow-functions': 'error',
+    'prefer-const': 'off',
+    'quote-props': 'off',
+    radix: 'error',
+    'react/jsx-boolean-value': 'error',
+    'react/jsx-curly-spacing': 'off',
+    'react/jsx-equals-spacing': 'off',
+    'react/jsx-key': 'error',
+    'react/jsx-no-bind': 'error',
```
johanna-adhoc, line 188: Hi Mike

I noticed this rule and what the eslint docs recommend to pass it. For the case of hooks the docs recommend using  useCallback/useMemo to define the functions but I have read a few articles that mention that passing non binded methods as props is most cases is not a performance deal breaker and the over use of useCallback and useMemo hooks can have a bigger performance impacts in slowness. I am sharing some articles below, let me know your thoughts!.

https://kentcdodds.com/blog/usememo-and-usecallback
https://medium.com/javascript-in-plain-english/react-usememo-and-when-you-should-use-it-e69a106bbb02
https://dmitripavlutin.com/dont-overuse-react-usecallback/

mjlumetta: wow, yeah, thanks for calling this out. those articles were extremely helpful for understanding those two hooks. do you have a position on this either way? my inclination based on those readings would be to turn this rule off. so yeah, thanks for sharing this concern, would definitely not have pulled this out on my own.

johanna-adhoc: yeah, I would rather have this rule off

mjlumetta: cool, sounds good! turned it off

johanna-adhoc (approved):
Approved. But please can you add the eslint react hooks rules before you merge the PR?
Thanks!

package.json
```diff
@@ -108,6 +112,12 @@
     "dotenv-expand": "4.2.0",
     "enzyme": "^3.10.0",
     "enzyme-adapter-react-16": "^1.7.1",
+    "eslint": "^7.9.0",
```
gileswells, line 115: I would suggest also adding `prettier-eslint` via npm to facilitate in tooling to auto-format code.

package.json
gileswells, line 11: I would suggest we add `--ext .tsx` as well so we get more linting coverage.

Included extensions | Errors | Warnings
--- | --- | ---
`.ts` only | 153 | 53
`.ts` + `.tsx` | 737 | 192

```suggestion
    "eslint": "./node_modules/.bin/eslint --ext .ts --ext .tsx src",
```

.eslintrc.js
```diff
@@ -0,0 +1,231 @@
...
+    'eslint-plugin-jsdoc',
+    'eslint-plugin-prefer-arrow',
+    'eslint-plugin-import',
+    'eslint-plugin-react',
+    '@typescript-eslint',
+    '@typescript-eslint/tslint',
+    'react-hooks',
+  ],
+  rules: {
+    '@typescript-eslint/adjacent-overload-signatures': 'error',
```
johanna-adhoc, line 40: For hooks we also need the below rules added:

 "react-hooks/rules-of-hooks": "error", 
 "react-hooks/exhaustive-deps": "warn"

gileswells (changes requested):
Basically the core of my thoughts circle around bringing in tooling to allow for automated linting while people are developing within their IDE of choice. I would like to the `prettier-eslint` and maybe add some info into https://github.com/department-of-veterans-affairs/developer-portal/blob/master/docs/development.md to discuss setting it up in a common IDE like VS Code or others.

package.json
mjlumetta: Giles and i discussed leaving this until a later PR because it's probably better to add autoformatting after the rules are finalized. that said, it's not blocking, especially because others may not have the Prettier/ESLint plugin(s) set up in their environment.

package.json
mjlumetta: fixed. errors more than fill my terminal

gileswells (approved):

PR #428, opened 2020-09-18
Add basic unit tests for ApiDocumentation

mjlumetta:
### Description

completes [API-1317](https://vajira.max.gov/browse/API-1317).

adds unit tests for `ApiDocumentation`. these are testing pretty basic features of our Swagger documentation. it uses the new React Testing Library and MSW Node tools to construct the tests.

### Process

  * PR is only adding tests
- [ ] Docs updated
  * n/a - no UI changes

### Requested feedback

is this a good level of tests to start with? are the tests meaningful?

### DIFF
--- a/src/containers/documentation/ApiDocumentation.test.tsx
+++ b/src/containers/documentation/ApiDocumentation.test.tsx
@@ -1,0 +1,129 @@
+import '@testing-library/jest-dom/extend-expect';
+import { findByRole, fireEvent, getByText, render, screen } from '@testing-library/react';
+import { FlagsProvider } from 'flag';
+import { MockedRequest, rest, restContext } from 'msw';
+import { MockedResponse, ResponseComposition } from 'msw/lib/types/response';
+import { setupServer } from 'msw/node';
+import * as React from 'react';
+import { Provider } from 'react-redux';
+import * as openAPIData from '../../__mocks__/openAPIData.test.json';
+import { IApiDescription as APIDescription } from '../../apiDefs/schema';
+import store, { history } from '../../store';
+import ApiDocumentation from './ApiDocumentation';
+
+const api: APIDescription = {
+  description: "it's a great API!",
+  docSources: [
+    {
+      openApiUrl: 'https://example.com/my/openapi/spec',
+    },
+  ],
+  enabledByDefault: true,
+  name: 'My API',
+  releaseNotes: () => <div>My API's release notes</div>,
+  trustedPartnerOnly: false,
+  urlFragment: 'my_api',
+  vaInternalOnly: false,
+};
+
+const server = setupServer(
+  rest.get(
+    'https://example.com/my/openapi/spec',
+    (req: MockedRequest, res: ResponseComposition, context: typeof restContext): MockedResponse => {
+      return res(context.status(200), context.json(openAPIData));
+    },
+  ),
+);
+
+describe('ApiDocumentation', () => {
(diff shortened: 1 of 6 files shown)


tsconfig.json
```diff
@@ -16,7 +16,8 @@
     "noImplicitAny": true,
     "strictNullChecks": true,
     "suppressImplicitAnyIndexErrors": true,
-    "noUnusedLocals": true
+    "noUnusedLocals": true,
+    "resolveJsonModule": true
```
mjlumetta, line 20: to be able to load the JSON for the mock OpenAPI spec

src/containers/documentation/ApiDocumentation.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+          <ApiDocumentation apiDefinition={api} categoryKey="fake" location={history.location} />
+        </FlagsProvider>
+      </Provider>,
+    );
+  });
+
+  afterEach(() => server.resetHandlers());
+  afterAll(() => server.close());
+
+  it('renders successfully', () => {
```
anthony-kopczyk, line 57: Not sure common practices with jest tests, just getting into it more now. I know we definitely need this to render successfully, but since this is happening in the `beforeEach` is this worth having? I assume if `beforeEach` throws it will stop the test execution? This makes it explicit though and that's good. I'm flexible either way, never thought about this before.

anthony-kopczyk (commented):
I think this is good to go and well implemented but would like someone else to take a look. I'm not that familiar with the front end react tests.

ediiotero (dismissed):
Tests look good, msw also seem interesting.

src/containers/documentation/ApiDocumentation.test.tsx
mjlumetta: yeah, i think you're right that this one doesn't provide much value. gonna remove it.

tslint.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": ["tslint:recommended", "tslint-react", "tslint-config-prettier"],
   "linterOptions": {
-    "exclude": ["config/**/*.js", "node_modules/**/*.ts"]
+    "exclude": ["config/**/*.js", "node_modules/**/*.ts", "src/__mocks__/**/*"]
```
mjlumetta, line 4: the linter lints JSON and like why would we lint these

src/containers/documentation/ApiDocumentation.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+  afterEach(() => server.resetHandlers());
+  afterAll(() => server.close());
+
+  it('renders the OpenAPI URI', async () => {
+    expect(await screen.findByText('https://example.com/my/openapi/spec')).toBeInTheDocument();
+  });
+
+  it('has a section for each tag', async () => {
+    const tags: string[] = ['Places', 'People'];
+    for (const tag of tags) {
```
nathanbwright, line 63: I don't know if this is standard testing practice in front-end tests, but it's been my experience that having loops with assertions in them can lead to frustration in the future... Not necessarily a blocking comment, just a comment. :)

mjlumetta: no, you're right, that's a good note. it should be enough to check 2 tags, and the for block is literally one line. will update.

ediiotero (approved):
LGTM

PR #429, opened 2020-09-18
Fix Dependabot alert for node-forge by reinstalling webpack-dev-server

mjlumetta:
### Description

fixes 1 Dependabot alert and partially but not entirely resolves another.

* fixed, high severity: `node-forge` must be >= 0.10.0
  - used to be 0.9.0
  - reinstalling `webpack-dev-server` with no upgrade fixed it (`node-forge` is a dependency of a dependency of `webpack-dev-server`
* ~mitigated, low severity: `node-fetch` must be >= 2.6.1~
  - ~was 1.7.3 under `swagger-ui` and 2.6.0 under `msw`~
  - ~upgraded `msw` to 0.21.2~
  - ~upgraded `swagger-ui` to 3.34.0~
    * ~this fixed the version of `node-fetch` under `swagger-client`, but it didn't fix another under `react` because `swagger-ui` uses React 15 (yikes)~

**edit:** once again, Swagger UI is terrible, so gonna deal with the low severity `node-fetch` issue in a subsequent PR.

### Process

- [ ] Docs updated

n/a - no functional changes. do manual testing to ensure dev server, unit tests, and Swagger UI haven't broken.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4772,9 +4772,9 @@
       }
     },
     "@types/glob": {
-      "version": "7.1.2",
-      "resolved": "https://registry.npmjs.org/@types/glob/-/glob-7.1.2.tgz",
-      "integrity": "sha512-VgNIkxK+j7Nz5P7jvUZlRvhuPSmsEfS03b0alKcq5V/STUKAa3Plemsn5mrQUO7am6OErJ4rhGEGJbACclrtRA==",
+      "version": "7.1.3",
+      "resolved": "https://registry.npmjs.org/@types/glob/-/glob-7.1.3.tgz",
+      "integrity": "sha512-SEYeGAIQIQX8NN6LDKprLjbrd5dARM5EXsd8GI/A5l0apYI1fGMWgPHSe4ZKL4eozlAyI+doUE9XbYS4xCkQ1w==",
       "dev": true,
       "requires": {
         "@types/minimatch": "*",
@@ -5648,6 +5648,12 @@
         "@babel/runtime-corejs3": "^7.10.2"
       }
     },
+    "arr-diff": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/arr-diff/-/arr-diff-4.0.0.tgz",
+      "integrity": "sha1-1kYQdP6/7HHn4VI1dhoyml3HxSA=",
+      "dev": true
+    },
     "arr-flatten": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/arr-flatten/-/arr-flatten-1.1.0.tgz",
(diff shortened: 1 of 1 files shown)


charleystran (dismissed):
LGTM

rtravitz:
Unclear why this is failing an accessibility test @mjlumetta

mjlumetta:
> Unclear why this is failing an accessibility test @mjlumetta

guarantee it's the Swagger changes. i can look into it and figure out how to resolve it, and if it's a stumbling block, i'll revert that part and fix the high severity vulnerability on its own.

mjlumetta:
updated to remove Swagger UI changes for low severity `node-fetch` issue since those are going to require somewhat more substantive changes.

rtravitz:
@mjlumetta should there be a corresponding `package.json` change? Could just changing the lockfile cause npm to potentially solve to a vulnerable version in the future?

mjlumetta:
@rtravitz no `package.json` change. here's the dependency tree from `npm ls node-forge` before and after the change, with annotations to explain what those packages specify as dependencies.

```
// before
└─┬ webpack-dev-server@3.11.0 (depends on selfsigned ^1.10.7)
  └─┬ selfsigned@1.10.7 (depends on node-forge ^0.9.0)
    └── node-forge@0.9.0

// after
└─┬ webpack-dev-server@3.11.0 (still depends on selfsigned ^1.10.7, last published 5/8/2020)
  └─┬ selfsigned@1.10.8 (now depends on node-forge ^0.10.0, last published 9/15/2020, node-forge update published 9/2/2020)
    └── node-forge@0.10.0
```

so basically, reinstalling `webpack-dev-server` will reinstall its dependencies as well with the highest version possible under the constraints specified by the packages. so in this case, it installs `selfsigned@1.10.8`, which requires `node-forge@^0.10.0`. even if `selfsigned` hadn't released a new version, i would hope that reinstalling `webpack-dev-server` would still upgrade `node-forge` to `0.10.0` because it has `node-forge@^0.9.0` and therefore could depend on `0.10.0`.

> Could just changing the lockfile cause `npm` to potentially solve to a vulnerable version in the future?

not sure i entirely follow here, but i think the answer is "not in a way that is likely to happen". there are two ways i could envision this happening:

1. a sub-dependency of the reinstalled package has a new vulnerability in a newer version that has been upgraded in the reinstall, and the package maintainers haven't released a new version of the package that rolls the sub-dependency back to its latest secure version. in this case, if the vulnerability is a known one, my hope would be that either `npm audit` or Dependabot alerting would catch this. given that `npm audit` is part of our CI pipeline, i hope that we'd be able to catch any of these issues. resolving those vulnerabilities wouldn't really be any different from how we resolve vulnerabilities now, as far as i'd understand it (curious if you disagree).
2. a package maintainer sets their dependency explicitly to a lower version that has a vulnerability. i can't imagine why that would happen, but it's possible.

does that address what you were going for?

rtravitz (approved):
The `npm ls node-forge` answers my question.

PR #433, opened 2020-09-21
CodeBuild Release

duganth-va:
### Description
Adds buildspec-release.yml for automated release jobs.

### Process
- [ ] Docs updated

### Requested feedback
Developer feedback on the release process.


cicd/buildspec-release.yml
```diff
@@ -0,0 +1,67 @@
...
+    commands:
+      - gh config set prompt disabled
+      - gh config set git_protocol ssh
+      - COMMIT_HASH=${CODEBUILD_RESOLVED_SOURCE_VERSION}
+      - echo "Generating release for ${COMMIT_HASH}" 
+  build:
+    commands:
+      # Check CI status of current commit hash
+      - ci_status=$(gh api /repos/department-of-veterans-affairs/developer-portal/commits/${COMMIT_HASH}/status | jq -r .state)
+      - echo ${ci_status}
```
mjlumetta, line 23: tiny detail, but would probably suggest adding a little more on the log message. like "Commit status: $(ci_status)"

cicd/buildspec-release.yml
```diff
@@ -0,0 +1,67 @@
+# This release file uses the new git hub cli tool called gh to automate release and artifact uploade.
+# Learn more here; https://github.com/cli/cli
+version: 0.2
+env:
+  variables:
+    ENVS: "dev staging production"
```
mjlumetta, line 6: how does the dev/staging auto-release work in this case again? i know you've talked about this at some point, but i can't remember how it's supposed to go 😅

mjlumetta (commented):
looks good so far, left a few comments. same as the CI PR, haven't done any extensive testing but will circle back to that later. also, big fan of the `gh` CLI tool 💯

cicd/buildspec-release.yml
duganth-va: more details the better, sounds good.

cicd/buildspec-release.yml
```diff
@@ -0,0 +1,47 @@
...
+    S3_ARCHIVE_BUCKET: "developer-portal-builds-s3-upload"
+  parameter-store:
+    GITHUB_TOKEN: "/dvp/devops/va_bot_github_token"
+    SLACK_WEBHOOK: "/dvp/devops/codebuild_slack_webhook"
+phases:
+  pre_build:
+    commands:
+      - gh config set prompt disabled
+      - gh config set git_protocol ssh
```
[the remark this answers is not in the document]
duganth-va: I like that mantra of over referencing external docs, I'll adopt. Here are the relevant links:
https://github.com/cli/cli
https://cli.github.com/manual/gh_config_set

cicd/buildspec-release.yml
duganth-va: The ENVS variable here is just a list of the assets we will post to the release tag. These assets are pulled from the S3 archive bucket. 

There's been a change in how we are going to operationalize this and worth discussing. 

First I was going to have this release job run like the Jenkins job on a schedule and then trigger the deploy jobs to prod automatically. However, when we move this pipeline Gabe wants to introduce this opportunity to introduce the MR process. This means that prod and sandbox deploys will have to happen under a MR process which I'll link in the readme.  The consequence of this operationally shift would result in changing the file based upon how the team would like to handle release/deploy functionality. It's going to be a discussion we will need to have but either of those scenarios are easy enough to build in CodeBuild. 

The original idea was to mimic Jenkins:
Cron job -> release job -> deploy job to production

Here are the laid out:
Deploy Scenarios
CI on PR -> deploy to review bucket
CI on master -> archive build artifacts for release -> deploy job to dev and staging 
? release
? prod deploy

gileswells (changes requested):
Looks solid but just a couple questions.

cicd/buildspec-release.yml
```diff
@@ -0,0 +1,67 @@
...
+  build:
+    commands:
+      # Check CI status of current commit hash
+      - ci_status=$(gh api /repos/department-of-veterans-affairs/developer-portal/commits/${COMMIT_HASH}/status | jq -r .state)
+      - echo ${ci_status}
+      - |
+        while [[ ${ci_status} == "pending" ]]; do
+          echo "CI still running -- sleep"
+          ci_status=$(gh api /repos/department-of-veterans-affairs/developer-portal/commits/${COMMIT_HASH}/status | jq -r .state)
+          echo ${ci_status}
```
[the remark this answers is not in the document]
duganth-va: Will do!

cicd/buildspec-release.yml
```diff
@@ -0,0 +1,67 @@
...
+          echo "${env} artifact uploaded"
+        done
+      # - | 
+      #   if [[ ${BUILD} == "true" ]]; then
+      #     aws codebuild start-build --project-name devportal-deploy --environment-variables-override name=ENVS,value=dev name=RELEASE,value=${NEW_TAG}
+      #   fi
+  post_build:
+    commands:
+      # Post to Slack if failure.
+      if [[ CODEBUILD_BUILD_SUCCEEDING -eq 0 ]]; then
```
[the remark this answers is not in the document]
duganth-va: Yes I think a release notification makes sense for successes

cicd/buildspec-release.yml
```diff
@@ -0,0 +1,94 @@
...
+          mv "release/developer-portal-${env}.tar.bz2" "release/developer-portal-${env}-${VERSION}.tar.bz2"
+          gh release upload ${NEW_TAG} "release/developer-portal-${env}-${VERSION}.tar.bz2#${NEW_TAG}_${env}.tar.bz2"
+          echo "${env} artifact uploaded"
+        done
+      # This is will kick off a a deployment to dev and staging deployments
+      - | 
+        if [[ ${DEPLOY} == "true" ]]; then
+          echo "Initiating build for ${NEW_TAG}"
+          aws codebuild start-build --project-name developer-portal-deploy --environment-variables-override name=RELEASE,value=${NEW_TAG}
+        fi
```
mjlumetta, line 85: does this block need to be commented out when merged pending #457, similarly to the block for this release job in the CI buildspec?

... and, now that i'm mentioning it, should that block be uncommented in this PR?

cicd/buildspec-release.yml
```diff
@@ -0,0 +1,94 @@
...
+      - NEW_TAG=$(increment.sh ${OLD_TAG}) 
+      - echo "Creating ${NEW_TAG} release"
+      - gh release create ${NEW_TAG}
+      - echo "Uploading artifacts"
+      - VERSION=${NEW_TAG#"cb-developer-portal/v"}
+      # Upload artificats to release after adding version number to file.
+      - |
+        for env in ${ENVS}; do
+          mv "release/developer-portal-${env}.tar.bz2" "release/developer-portal-${env}-${VERSION}.tar.bz2"
+          gh release upload ${NEW_TAG} "release/developer-portal-${env}-${VERSION}.tar.bz2#${NEW_TAG}_${env}.tar.bz2"
```
mjlumetta, line 77: this is perhaps a "i need to read the docs more carefully" comment, but it took me a really long time to figure out what the # was doing in this command. might be worth calling out in a comment or something that the string after the "#" is the display name of the asset. that said, you could dismiss this as me needing to read the whole docs page before trying to understand the thing.

... also makes me wonder why they couldn't make the display name be its own optional arg, but that's neither here nor there

cicd/buildspec-release.yml
```diff
@@ -0,0 +1,94 @@
...
+#  - CodeBuild Environmental Variables - https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html
+#  - CodeBuild Enironment Image - ECS deployer - https://github.com/department-of-veterans-affairs/lighthouse-codebuild-containers/tree/main/ecs-deployer
+#  - AWS CLI 2.0 - https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html
+#  - Git Hub CLI - https://github.com/cli/cli
+# 
+# All Custom scripts are on the CodeBuild Environment container. 
+# https://github.com/department-of-veterans-affairs/lighthouse-codebuild-containers/tree/main/
+#
+# Custom Scripts:
+#  - slackpost.sh
```
mjlumetta, line 16: tiny thing, but do you want to call out `increment.sh` here?

cicd/buildspec-release.yml
```diff
@@ -0,0 +1,94 @@
...
+        for env in ${ENVS}; do
+          mv "release/developer-portal-${env}.tar.bz2" "release/developer-portal-${env}-${VERSION}.tar.bz2"
+          gh release upload ${NEW_TAG} "release/developer-portal-${env}-${VERSION}.tar.bz2#${NEW_TAG}_${env}.tar.bz2"
+          echo "${env} artifact uploaded"
+        done
+      # This is will kick off a a deployment to dev and staging deployments
+      - | 
+        if [[ ${DEPLOY} == "true" ]]; then
+          echo "Initiating build for ${NEW_TAG}"
+          aws codebuild start-build --project-name developer-portal-deploy --environment-variables-override name=RELEASE,value=${NEW_TAG}
```
mjlumetta, line 84: does the `developer-portal-deploy` project still need to be created in Terrraform?

cicd/buildspec-release.yml
duganth-va: I throughly enjoy comments to explain tools in the context in which they are used so this is valuable to me and I'll add the comment.  Especially since the `#` in line 72 is serving a completely different purpose it may cause much confusion. thanks!

cicd/buildspec-release.yml
duganth-va: no all the Terraform infrastructure has been created so i'll comment this out prior to merge! I'll add a comment here as well!

cicd/buildspec-release.yml
duganth-va: Yeah we will uncomment this until we transition to deployments with codebuild.

Yes we should thanks uncomment that in this PR thanks!!

mjlumetta (approved):
💯 ✨

cicd/buildspec-release.yml
```diff
@@ -0,0 +1,94 @@
...
+      # Check that all files exist 
+      - | 
+        for env in ${ENVS}; do
+          if [[ ! -f release/developer-portal-${env}.tar.bz2 ]]; then
+            echo "developer-portal-${env}.tar.bz2 was not found. Aborting release process and sending notification to slack " >&2
+            exit 1
+          fi
+        done
+      # Create new tag
+      - OLD_TAG=$(git tag --sort=-creatordate | grep cb- | head -1)
```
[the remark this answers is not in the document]
mjlumetta: that seems great to me! thanks for the clarification here

cicd/buildspec-release.yml
duganth-va: its undergoing a change right now,  when the new documentation comes out (should be shortly) I'll add to docs and update in slack.

gileswells (approved):
LGTM

rtravitz (dismissed):
Makes sense ✨

cicd/buildspec-ci.yml
```diff
@@ -139,10 +139,10 @@ phases:
...
       # Call the release job if this is running on master branch.
       - phase=release
-      # - |
-      #   if [[ ${BRANCH} == 'master' ]]; then
-      #     aws codebuild start-build --project-name developer-portal-release --source-version ${CODEBUILD_RESOLVED_SOURCE_VERSION}
-      #   fi
+      - |
+        if [[ ${BRANCH} == 'master' ]]; then
+          aws codebuild start-build --project-name developer-portal-release --source-version ${CODEBUILD_RESOLVED_SOURCE_VERSION}
+        fi
```
anthony-kopczyk, line 145: Nothing to change, but are we going to be renaming branches to `main`? This will require an update at that point. Might be good to point in a jira ticket now for it or added to another jira ticket if there isn't already one.

cicd/buildspec-release.yml
anthony-kopczyk: I like the comment, and I don't think a change is needed. For readability I think it might be more clear if the comment is moved to within the for loop, so it's closer to that line. Even seeing you were going to add a comment, I still didn't realize it was there until I looked around for it more. That's not necessary though before merging, and maybe there's a reason not to move it into the for loop for similar readability reasons - would kind of clutter the for loop.

anthony-kopczyk (commented):
This looks good to me. Most of the questions I had were already answered on previous reviews and discussions.

@gileswells I'm good with approval on this PR, like the other one, as long as you are too.

gileswells (approved):
LGTM

rtravitz (approved):

PR #435, opened 2020-09-21
Upgrade msw to 0.21.2 and swagger-ui to 3.31.1 to remediate node-fetch vulnerability

mjlumetta:
### Description

upgrades `swagger-ui` to 3.31.1 and `msw` to 0.21.2 in order to remediate the low severity `node-fetch` vulnerability that raised a Dependabot alert. part of [API-2578](https://vajira.max.gov/browse/API-2578). 

note that it was _not_ possible to entirely resolve the alert, because `swagger-ui` depends on React 15, which depends on `node-fetch` 1.7.3.

i upgraded `swagger-ui` to 3.31.1 instead of the latest version because at a certain point they screwed something up that caused the UI to render an empty link. the commit is [here](https://github.com/swagger-api/swagger-ui/commit/d9f5691f657dc64842b892255fe4a67cb6246f1b#diff-79a71326f6eb6a6ada641d99a59cd3b5R151). changing the conditional from `externalDocsUrl` to `externalDocs` turns out to be wrong because the `externalDocs` object is always truthy (i think because of how those selectors are written). anyway....

diff of `npm ls node-fetch`:

```
# before
├─┬ msw@0.20.5
│ └── node-fetch@2.6.0
└─┬ swagger-ui@3.24.0
  ├─┬ react@15.6.2
  │ └─┬ fbjs@0.8.17
  │   └─┬ isomorphic-fetch@2.2.1
  │     └── node-fetch@1.7.3  deduped
  └─┬ swagger-client@3.9.5
    └─┬ cross-fetch@0.0.8
      └── node-fetch@1.7.3

# after
├─┬ msw@0.21.2
│ └── node-fetch@2.6.1
└─┬ swagger-ui@3.31.1
  ├─┬ react@15.6.2
  │ └─┬ fbjs@0.8.17
  │   └─┬ isomorphic-fetch@2.2.1
  │     └── node-fetch@1.7.3    # BAD STUFF
  └─┬ swagger-client@3.10.12
    └─┬ cross-fetch@3.0.6
      └── node-fetch@2.6.1
```

### Process

  * n/a, should only catch regressions
- [ ] Docs updated
  * n/a
  * aXe tests continue to pass, and we expect nothing more than that from Swagger UI

### DIFF
--- a/src/styles/swagger.scss
+++ b/src/styles/swagger.scss
@@ -126,7 +126,12 @@
     border-radius: 4px;
   }
 
-  svg {
-    display: none;
+  svg.svg-assets {
+    // for some reason the SVG assets component is like ~300x100 pixels
+    // swagger-ui source appears to set these styles on the React element, but they
+    // didn't make it to the browser
+    position: absolute;
+    width: 0;
+    height: 0;
   }
 }
--- a/src/styles/swagger.scss
+++ b/src/styles/swagger.scss
@@ -125,4 +125,8 @@
     background: #41444e;
     border-radius: 4px;
   }
+
+  svg {
+    display: none;
+  }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -2213,23 +2213,23 @@
       }
     },
     "@babel/runtime-corejs2": {
-      "version": "7.6.3",
-      "resolved": "https://registry.npmjs.org/@babel/runtime-corejs2/-/runtime-corejs2-7.6.3.tgz",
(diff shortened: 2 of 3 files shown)


src/styles/swagger.scss
```diff
@@ -125,4 +125,8 @@
     background: #41444e;
     border-radius: 4px;
   }
+
+  svg {
```
mjlumetta, line 129: for some reason, this svg is something like 300x100 pixels

charleystran (approved):
:shipit:

nathanbwright (approved):

PR #436, opened 2020-09-23
API-174 Handle focus on tier-3 side nav links

gileswells:
### Description
This will focus the targeted element when clicking on a hash link in tier 3 items in the SideNav. This is most 

### Process
- [ ] Docs updated

### Requested feedback
The ticket specifies this only for tier-3 items but the tier-1 links on `/news` do not get their focus handled. Would we like to extend that request to include these tier-1 items and make the focus handling agnostic to side nav depth?


gileswells:
It looks like my VS Code added a fair bit of markdown normalization to the files I touched. The markdown files just really had `tabIndex` added to them and a clarification as to why on the `issues.md` file.

src/components/SideNav.tsx
```diff
@@ -93,6 +93,9 @@ export class SideNavEntry extends React.Component<ISideNavEntryProps> {
             'vads-u-border-left--5px': subNavLevel === 0,
           })}
           isActive={this.navHashLinkIsActive}
+          onClick={(e: React.MouseEvent<HTMLAnchorElement>) => {
+            this.onHashAnchorClick(e, subNavLevel);
+          }}
```
ediiotero, line 98: remove subNavLevel so it applies to all anchor links (from convo in working session)

src/components/SideNav.tsx
```diff
@@ -111,6 +114,15 @@ export class SideNavEntry extends React.Component<ISideNavEntryProps> {
...
   }
+  private onHashAnchorClick = (e: React.MouseEvent<HTMLAnchorElement>, subNavLevel: number) => {
+    if (subNavLevel === 2) {
+      const id: string = e.currentTarget.href.split('#')?.[1];
+      if (id) {
+        document.getElementById(id)?.focus();
+      }
+    }
+  }
+
```
ediiotero, line 125: change to be ignostic focus handler for hash links

ediiotero (commented):
Paired on review during working session with @mjlumetta , @johanna-adhoc , and @anthony-kopczyk .
@gileswells met with us after to talk about our comments

src/components/SideNav.tsx
```diff
@@ -111,6 +114,15 @@ export class SideNavEntry extends React.Component<ISideNavEntryProps> {
       </li>
     );
   }
+  private onHashAnchorClick = (e: React.MouseEvent<HTMLAnchorElement>, subNavLevel: number) => {
+    if (subNavLevel === 2) {
+      const id: string = e.currentTarget.href.split('#')?.[1];
```
mjlumetta, line 119: alternative suggestion from our pairing session (non-blocking, either are acceptable):

```
const linkURL = new URL(e.currentTarget.href);
if (linkURL.hash) {
  const id = linkURL.hash.slice(1);
  document.getElementBy(id)?.focus();
}
```

your call, either is fine. you may want to double-check that we have `URL()` polyfilled, because i think IE might not fully support it? but pretty sure that's what `url-polyfill` is.

https://developer.mozilla.org/en-US/docs/Web/API/URL

mjlumetta (commented):
added a couple comments from our pairing session. looking good overall. biggest thing is making the click handler more generic so that News and Release Notes also benefit from this change.

src/components/SideNav.tsx
gileswells: I think I'd just prefer leave the core of this one as is. No particular reason beyond it's readable to most everyone without needing to know how to use `new URL()` (which in this case is arguably very easy).

src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -3,30 +3,30 @@ imports:
...
-* **A Redirect URL** - that will receive your OpenID Connect token
-* **Application name** - The name and/or URL of your application that identifies it on authorization screens.
-* **Type of Application** - Mobile app, Single page web application, or a Server-based application
-* **Scopes** - Data you'd like to access with your app (see [#scopes](#scopes))
+- **A Redirect URL** - that will receive your OpenID Connect token
+- **Application name** - The name and/or URL of your application that identifies it on authorization screens.
+- **Type of Application** - Mobile app, Single page web application, or a Server-based application
+- **Scopes** - Data you'd like to access with your app (see [#scopes](#scopes))
 
 ## Building OpenID Connect Applications
```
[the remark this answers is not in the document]
gileswells: done

gileswells:
There was some discussion on where to put the new utility method for handling focus [in slack](https://lighthouseva.slack.com/archives/CAZF880HW/p1601057954031400) and I didn't generate any strong objections to it's current location in the `utils` folder. 

Should have all the needed updates per the working session completed and ready for review/merge.

src/utils/clickHandlers.ts
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
+
+export const onHashAnchorClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
```
mjlumetta, line 3: maybe add a comment explaining why this helper function is necessary? function itself looks perfect

mjlumetta (commented):
yeah, so everything looks great, and it works great on the Authorization page. that said, it doesn't work on the News and Release Notes pages, for reasons i could not immediately figure out. if you happen to know, i'd say fix it and i'll re-review, and otherwise, maybe we can talk through it.

gileswells:
@mjlumetta news and release notes were messed up. Those are fixed now.

src/containers/News.tsx
```diff
@@ -111,10 +119,11 @@ export default function News() {
                   <section
                     aria-label={section.title}
                     key={section.id}
-                    id={section.id}
                     className="vads-u-margin-bottom--4"
                   >
-                    <h2>{section.title}</h2>
+                    <h2 id={section.id} tabIndex={-1}>
+                      {section.title}
+                    </h2>
```
anthony-kopczyk, line 126: I can't help but wonder which way is better on this. If it makes sense for the section itself to have the tabIndex as -1. but then there would be redundant content when screen readers go to the heading tag. Can you maybe walk me through this?

anthony-kopczyk: I do like, this makes the test a little easier to read.

anthony-kopczyk (commented):
I want to approve, but I feel like I should wait until I have an answer to my question on the tabIndex on the News.tsx.

Should be good to approve after that has a response of some kind.

src/containers/News.tsx
gileswells: So with the ID / tabIndex on a section then it's got the potential for a screen reader to attempt to read the entire section vs just the header. Also in the ticket it's asking for the focus to target the headings vs the whole container so that also facilitated the change here.

anthony-kopczyk (approved):
Looks good to go! Everything existing has been addressed, and I don't see anything else.

PR #437, opened 2020-09-23
Api 2500: Hooks migration CategoryReleaseNotes, ServersContainer, OperationTag

anthony-kopczyk:
### Description
This updates the given React components to use React Hooks. Note that the CategoryReleaseNotes was already using React hooks, so there were no changes to make. Also these files are used for SwaggerUI, so as far as testing, I need to investigate.

### Process
- [x] Looked into converting the files to .tsx: Was unable to convert the OpreationTag.jsx because it's dependencies don't have types defined
- [x] Added unit tests if it makes sense: I'm going to make the argument that it doesn't make sense to add unit tests to these. OperationTag is still in .js and ServersContainer is hidden on the page, so it doesn't really server function at this time.

### Requested feedback
- I set default properties on the OperationTag based on the old code. I'm not really sure the right way to do it. I know there are a few different ways, but not sure if there's a *best* way.


src/containers/documentation/swaggerPlugins/ServersContainer.tsx
```diff
@@ -0,0 +1,54 @@
...
+    selectedServer: () => any,
+  };
+  oas3Actions: {
+    setServerVariableValue: () => any,
+    setSelectedServer: () => any,
+  };
+  getComponent: (componentName: string) => typeof React.Component;
+}
+
+export default function ServersContainer(props: ServersContainerProps): JSX.Element {
```
mjlumetta, line 20: so, since we hide this, i would suggest that we resolve this issue a bit differently. i think we can get rid of this component entirely and the corresponding `Servers` component. steps to do that:

1. in `src/containers/documentation/swaggerPlugins/index.tsx`, replace the `ServersContainer` key under `components` with `ServerContainers: () => null`. see here: https://github.com/swagger-api/swagger-ui/blob/master/docs/customization/plugin-api.md#components
  1. optionally, you can also do that to the `Servers` component, or you can just delete that line since `Servers` is a child of `ServersContainer`.
2. remove the block that says `display: none` for the schemes container in `StyleOverride.scss`.
3. delete `ServersContainer` and `Servers`.

does that work? would be great to delete this code 😄

src/containers/documentation/swaggerPlugins/OperationTag.jsx
```diff
@@ -7,102 +7,98 @@ import { sanitizeUrl } from "@braintree/sanitize-url";
...
+      </h3>
+
+      <Collapse isOpened={showTag}>
+        {children}
+      </Collapse>
+    </div>
+  );
 }
+
+OperationTag.propTypes = {
```
mjlumetta, line 95: there are some different props (additional, not changing these) in the [current version of this component](https://github.com/swagger-api/swagger-ui/blob/master/src/core/components/operation-tag.jsx). although, tbf, we have 3.31.1 now, which is not latest. but regardless, this component has probably changed since it was copied and modified. 

anyway, all this to say that we probably _should not_ attempt to change those things, or at least, i'm not inclined to. but worth pointing out for awareness.

anthony-kopczyk:
I'm not sure why the changes are showing the ServersContainer.jsx as deleted. I uses `git mv` and in the individual commits it's not deleted - maybe it's just how git is displaying the PR diff?

mjlumetta:
> I'm not sure why the changes are showing the ServersContainer.jsx as deleted. I uses git mv and in the individual commits it's not deleted - maybe it's just how git is displaying the PR diff?

@anthony-kopczyk is that not because you made it TSX instead of JSX?

src/containers/documentation/swaggerPlugins/OperationTag.jsx
```diff
@@ -7,102 +7,98 @@ import { sanitizeUrl } from "@braintree/sanitize-url";
...
+  const {
+    deepLinking,
+    docExpansion
+  } = getConfigs();
+
+  const isDeepLinkingEnabled = deepLinking && deepLinking !== "false";
+
+  const Collapse = getComponent("Collapse");
+  const Markdown = getComponent("Markdown");
+  const DeepLink = getComponent("DeepLink");
```
mjlumetta, line 34: we currently standardize on single quotes, fyi, but we don't have the TSLint rule for it, soooo 🤷 i wouldn't worry about fixing it here, but if you'd like to, that'd be cool. will fix as a part of the ESLint implementation

mjlumetta (commented):
yeah, looks good overall! mainly have a suggestion related to ditching ServersContainer in a different way

src/containers/documentation/swaggerPlugins/ServersContainer.tsx
anthony-kopczyk: I can look into this.

src/containers/documentation/swaggerPlugins/OperationTag.jsx
anthony-kopczyk: I'll go ahead and switch them over since I'm going to look at those other changes anyway.

src/containers/documentation/swaggerPlugins/OperationTag.jsx
anthony-kopczyk: Thanks. Wasn't aware of that.  We'll need to keep that in mind on all of the swagger plugins as we do this... I'm wondering if we should actually leave these swagger related files as classes, so that they're closer to the original, for updating as we move along?

anthony-kopczyk:
> > I'm not sure why the changes are showing the ServersContainer.jsx as deleted. I uses git mv and in the individual commits it's not deleted - maybe it's just how git is displaying the PR diff?
> 
> @anthony-kopczyk is that not because you made it TSX instead of JSX?

For some reason the overall diff doesn't show the rename. The individual commit shows it as a rename though. I feel like this is a github bug or missing feature in the PR files changed tab.

https://github.com/department-of-veterans-affairs/developer-portal/pull/437/commits/58ac6a045da27995efd493fbded622b9d318899f

src/containers/documentation/swaggerPlugins/ServersContainer.tsx
anthony-kopczyk: Well after taking those out, this PR got small. haha!

anthony-kopczyk: I didn't remove the override in StylesOverride.scss. Basically, there's a div that's still there and without the override it creates unnatural looking blank space on the screen.

anthony-kopczyk:
Reverted the changes to OperationTag. Since it's originally from Swagger apparently there's some issue with React 16. And also we probably want to keep these closer to the original versions, for easier upgrades in the future.

johanna-adhoc (approved):
LGTM 👍 . 
OperationTag will not be migrated to hooks because of the swaggerUI dependency constraint.

src/containers/documentation/swaggerPlugins/ServersContainer.tsx
mjlumetta: oh yeah, i see what you're saying. yeah, this PR got really small, i'm loving it

mjlumetta (approved):
big fan of this here PR

PR #438, opened 2020-09-23
API-2498: Hooks migration header component

johanna-adhoc:
### Description
Migration of component logic from classes to hooks.

### Process
- [N/A] Docs updated

### Requested feedback
This PR introduces a component folder structure and a barrel file at the root of components folder.
The idea is that by the end of the hooks migration all independent components will be located in its own folder that will only contain files related to the component. 

The barrel file will handle the export of all components folders for a cleaner import syntax.
Example:

 import { Header, CrisisLine, Footer } from './components';

Please let me know your thoughts about this! 😃


anthony-kopczyk (commented):
I think these are good, but want to have someone else review since I'm not as familiar with React Hooks.

src/components/header/Header.tsx
```diff
@@ -0,0 +1,123 @@
...
+            </button>
+          </div>
+        </div>
+        <NavBar isMobileMenuVisible={mobileNavVisible} onMobileNavClose={toggleMenuVisible} />
+      </header>
+    </>
+  );
+};
+
+Header.propTypes = {};
```
mjlumetta, line 121: one thing i've never really thought carefully about or been clear on is how to use `propTypes` in Typescript since prop types can be declared like they can't in Javascript. wouldn't be surprised if you'd noticed this already, but very few components here have `propTypes` defined (if any?). i would guess that the benefits supplied by `propTypes` would be runtime checks and more complete use of React patterns, i guess? if you have thoughts or answers on this question, would love to get more thoughts/perspective on this 🙂 unless there's a definitive answer to that, though, i'm pro explicit inclusion of `propTypes`.

mjlumetta (approved):
looks great to me 💅 🎉  extremely excited for this hooks conversion. left one comment about a small adjustment to some stuff in the old component that doesn't really make sense and one philosophical React/TS question.

src/components/header/Header.tsx
johanna-adhoc: We came across this same question. And found below answer

_"Typescript and PropTypes serve different purposes. Typescript validates types at compile time, whereas PropTypes are checked at runtime.

Typescript is useful when you are writing code: it will warn you if you pass an argument of the wrong type to your React components, give you autocomplete for function calls, etc.

PropTypes are useful when you test how the components interact with external data, for example when you load JSON from an API. PropTypes will help you debug (when in React's Development mode) why your component is failing by printing helpful messages"_

https://stackoverflow.com/questions/41746028/proptypes-in-a-typescript-react-application

As you mention using propTypes in typescript is not super common so we are following this article to do it https://fettblog.eu/typescript-react/prop-types/

Let me know your thoughts

src/components/header/Header.tsx
```diff
@@ -0,0 +1,123 @@
...
+          <div
+            className={classNames(
+              'va-api-logo',
+              'vads-u-margin-left--2',
+              'medium-screen:vads-u-margin-left--0',
+            )}
+          >
+            <Link
+              to="/"
+              title="Digital VA home page"
```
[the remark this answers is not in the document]
johanna-adhoc: Sure, I can make the change

src/components/header/Header.tsx
mjlumetta: sounds good to me - the runtime vs compile time distinction makes sense. thanks for the answer and the resources!

src/components/header/Header.tsx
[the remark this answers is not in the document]
johanna-adhoc: done :)

src/components/header/Header.tsx
```diff
@@ -0,0 +1,123 @@
...
+const handleSkipNavClick = () => {
+  const mainElement: HTMLElement | null = document.querySelector('main');
+  if (mainElement) {
+    mainElement.focus();
+  }
+};
+
+const Header = (): JSX.Element => {
+  /**
+   * TOGGLE MENU VISIBLE
```
anthony-kopczyk, line 32: With grouping everything based on functionality, part of me wants to encapsulate the grouped functions into named blocks or something to make them separated without the need of comments.

anthony-kopczyk (approved):
Looks great. The questions Mike had brought up were all addressed, and I don't see anything else.

src/components/header/Header.tsx
anthony-kopczyk: Or I guess, one, _kind of_ resolution is if we have something that is large enough that we need a bunch (like 5 or more) different comment blocks, maybe we can take that as the sign we need to break up the component into multiple.

johanna-adhoc: yeah, I was thinking how that would look like for a functionality that depends on useEffect and/or useRef, which need to be declared at the root 🤔 .

Maybe in future PRs if you see a use case for the named blocks you can try to implement it and see if it could work and makes sense.

johanna-adhoc: > Or I guess, one, _kind of_ resolution is if we have something that is large enough that we need a bunch (like 5 or more) different comment blocks, maybe we can take that as the sign we need to break up the component into multiple.

That makes a lot of sense. I have seen in past project that if a component has multiple util/helper methods those are placed in a separated file.

mjlumetta (approved):
re-approving 💯 ↪️ ↩️

src/components/header/Header.tsx
anthony-kopczyk: Sounds great. I'll try something out on a future PR.

PR #439, opened 2020-09-24
API-2499: Hook migration ReleaseNotes, Support, SupportFAQ

anthony-kopczyk:
### Description
This PR is for converting the three components to react hooks
- ReleaseNotes
- Support
- SupportFAQ

The only one that needed updated was SupportFAQ.

### Process
- [N/A] Docs updated


src/containers/support/SupportFAQ.tsx
```diff
@@ -138,29 +138,27 @@ interface ISupportQuestion {
 }
 
 const SupportQuestions = (props: ISupportQuestionsProps) => {
-  const content: IPanelContent[] = props.questions.map((q: ISupportQuestion) => {
-    return {
+  const content: IPanelContent[] = props.questions.map((q: ISupportQuestion) => (
```
anthony-kopczyk, line 141: If the function is just returning a value we can encapsulate it in parentheses. eslint wants this and I think it turns out nice because it gets rid of return statements.

Inline arrow function: `const getLength = (param) => param.length()`
Other:
```
const getObject = (vals) => (
  {
    thing: vals[1],
    thing1: vals[2],
  }
);

File: src/containers/support/SupportFAQ.tsx [hunk 2]
```diff
@@ -138,29 +138,27 @@ interface ISupportQuestion {
...
-        <div>
-          <SupportQuestions title="General" questions={generalQuestions}/>
-          <SupportQuestions title="Development" questions={developmentQuestions}/>
-          <SupportQuestions title="Troubleshooting/Support" questions={supportQuestions}/>
-        </div>
-      </section>
-    );
-  }
-}
+const SupportFAQ: () => JSX.Element = () => (
```
anthony-kopczyk, line 153: eslint was throwing fits when I had this as `export default function SupportFAQ` because it prefers arrow functions.
In order to export the arrow function, because js doesn't like `export default const SupportFAQ` I had to put that delcaration on its own line and then export it.

anthony-kopczyk: Also not sure about return type as JSX.Element.

johanna-adhoc: The arrow function for sure we need. Once eslint gets enabled as the syntax checker the enforcing of arrow functions will kick in, so we should already start writing methods like that from now on. The exporting at the bottom is perfect, that is a very common way to do it.

"export default SupportFAQ;"

As for the return type JSX.Element I don't think is necessary. That is my personal preference to declare a return type if the method doesn't return void, but I don't think this will break eslint.

src/containers/support/SupportFAQ.tsx
johanna-adhoc: yep! this is the way to go

johanna-adhoc (approved):
👍

anthony-kopczyk:
Waiting to merge - because eslint has some issues with quotes and strings within the JSX, Want to get that conversation resolved first.

mjlumetta (approved):

PR #440, opened 2020-09-24
Upgrade Jest and related testing tools packages

mjlumetta:
### Description

upgrades a number of Jest-related packages to the latest version, or the latest possible version where i ran into issues (`jest-image-snapshot`). completes [API-1894](https://vajira.max.gov/browse/API-1894).

changed dependencies:

* `jest` itself: 25.5.4 -> 26.4.2
* Jest transpilation
  * `babel-jest`: 25.5.1 -> 26.3.0
  * `ts-jest`: 25.5.1 -> 26.4.0
* Jest Puppeteer environment
  * `puppeteer`: 1.15.0 -> 2.1.1
    * the latest version of Puppeteer is 5.3.1, but Jest Puppeteer has a peer dependency: `jest-puppeteer@4.4.0 requires a peer of puppeteer@>= 1.5.0 < 3`
  * `jest-puppeteer`: 4.1.1 -> 4.4.0
  * `jest-environment-puppeteer`: 4.1.1 -> 4.4.0
  * `expect-puppeteer`: 4.1.1 -> 4.4.0
* `jest-image-snapshot` for visual regression: 3.1.0 -> 4.0.2
  * could not upgrade to latest (4.2.0) because of a dependency getting loaded as its browser version rather than its Node version
* accessibility testing
  * `axe-core`: 3.3.1 _with no explicit dependency_ -> 4.0.2
  * `jest-axe`: 3.2.0 -> 4.0.0
* DefinitelyTyped packages in support of the above
  * `@types/jest`: 25.2.3 -> 26.0.14
  * `@types/jest-axe`: 2.2.1 -> 3.5.0
  * `@types/jest-image-snapshot`: 3.1.0 -> 4.1.0
  * `@types/puppeteer`: 1.11.1 -> 2.1.4

### Process

  - changes packages in test framework, so automated testing isn't possible
- [N/A] Docs updated

### Requested feedback

please try to break anything you can! 

are the pixel-fudging visual regression failures okay to update?

also, if anyone knows why `ssim.js` as a dependency of `jest-image-snapshot >= 4.1.0` is resolving to its browser version and not its Node version, please let me know.


mjlumetta:
i'm trying to see what the visual regression tests look like because they are failing intermittently on my machine 😞

package.json
```diff
@@ -135,13 +135,13 @@
     "postcss-safe-parser": "4.0.1",
     "prettier": "^1.16.4",
     "pretty-format": "^26.4.2",
-    "puppeteer": "^1.15.0",
+    "puppeteer": "^1.20.0",
```
mjlumetta, line 138: `puppeteer` has a lot later versions, but `jest-puppeteer` requires some v1 version.

package.json
```diff
@@ -145,13 +146,13 @@
     "postcss-safe-parser": "4.0.1",
     "prettier": "^1.16.4",
     "pretty-format": "^26.4.2",
-    "puppeteer": "^1.15.0",
+    "puppeteer": "^2.1.1",
```
mjlumetta, line 149: latest is 5.3.1, but `jest-puppeteer` has a peer dependency: `jest-puppeteer@4.4.0 requires a peer of puppeteer@>= 1.5.0 < 3`

package.json
```diff
@@ -118,18 +119,18 @@
...
     "html-loader": "^0.5.5",
     "html-webpack-plugin": "4.0.0-beta.2",
-    "jest": "^25.5.4",
-    "jest-axe": "^3.2.0",
-    "jest-environment-puppeteer": "^4.1.1",
-    "jest-image-snapshot": "^3.1.0",
+    "jest": "^26.4.2",
+    "jest-axe": "^4.0.0",
+    "jest-environment-puppeteer": "^4.4.0",
+    "jest-image-snapshot": "~4.0.2",
```
mjlumetta, line 131: latest is 4.2.0, but versions >= 4.1.0 use `ssim.js` as a dependency. when i run the visual regression tests, Jest tries to load `ssim.js` as its browser version, leading to a "window is not defined" error in the test. i'm not really sure why this would be. the Jest Puppeteer environment _creates_ a browser context in the Chromium browser it launches, but it doesn't _run in_ that browser context; it runs in the Jest Node process from which it launches that browser.

src/base.accessibility.ts
```diff
@@ -12,6 +12,25 @@ describe('Accessibility tests', () => {
 
       await page.goto(`${puppeteerHost}${path}`, { waitUntil: 'networkidle0' });
       await page.addScriptTag({ path: require.resolve('axe-core') });
+
+      // THIS IS BAD. code highlighting on the Authorization page has several color contrast
+      // issues. upgrading axe-core from 3.3.1 to 4.0.2 and jest-axe from 3.2.0 to 4.0.0 introduced
+      // stricter rules, which is good, but we can't resolve this issue without UX input.
+      if (path === '/explore/health/docs/authorization') {
```
mjlumetta, line 19: this speaks for itself, but i'm following up with UX.

src/components/CodeWrapper.tsx
```diff
@@ -9,7 +9,8 @@ interface CodeWrapperProps {
 export default class CodeWrapper extends React.Component<CodeWrapperProps, {}> {
   public render() {
     return (
-      <div className="code-wrapper">
+      // future improvement: add role="region" and add appropriate aria-labels
+      <div className="code-wrapper" tabIndex={0}>
```
mjlumetta, line 13: makes all code blocks keyboard focusable. i think this is okay, but the question we should ask here is whether it will ever be a net negative to have a code block within the keyboard focus order.

ediiotero (commented):
This looks good. I ran all the tests, only things failing were the e2e ci, and accessibility ci all for the same reason, timeout of 30000 ms exceeded. Not sure but maybe its my local?
You know this already but there are a few failing tests when I run the visual regression test, the size wXh is off by a few pixels but only on the height. Looked at the docs regarding SSIM. There is a recommendation to not test pixel to pixel more on that [here](https://github.com/americanexpress/jest-image-snapshot#ssim-performance-considerations). There are also options you can pass to .toMatchImageSnapshot to set a threshold for SSIM.  I didn't run into a "window is not defined" error while running the tests.
Hope some of this helps.

mjlumetta:
thanks for the review @ediiotero. i think the 30000ms timeout is probably local - i get that sometimes too. not really sure why, might need to check on that.

> There is a recommendation to not test pixel to pixel...

can you say more about that? i see that the docs suggest that SSIM has some advantages over pixelmatch. is that what you're referring to? in any case, we cannot use SSIM because SSIM is the package that caused the "window is not defined" error, so i upgraded to a version that does not have it (4.0.2 vs 4.1.0+). would still love to pair on this piece if you're available, and if we do that, that'll enable use of SSIM in the future, should we choose to use it.

src/base.accessibility.ts
gileswells: So long as a ticket is made / has been made for this as a follow up then I'm on board with delaying this.

gileswells (approved):
@mjlumetta I had no issues with timeout and do believe that it was based on his machine and not an issue with the tests itself.

src/components/CodeWrapper.tsx
gileswells: I think this is a good idea. I mean if it doesn't work for a specific user then they can just tab to the next one but for those that want this then it's definitely a good add in my book.

gileswells (approved):
Continues to look good with the last commits

mjlumetta:
@ediiotero and i paired briefly on upgrading `jest-image-snapshot` to latest and didn't make any significant progress. Eddie is going to spend a little more time on it tomorrow, but we agreed we'd do it in a follow-up PR if he's successful and that this PR can be merged now.

src/base.accessibility.ts
mjlumetta: already in progress https://vajira.max.gov/browse/API-2670

PR #441, opened 2020-09-25
API-2495: Hooks migration SupportContactUs component

johanna-adhoc:
### Story [API-2495](https://vajira.max.gov/browse/API-2495)
Migration of SupportContactUs component logic from classes to hooks.

Note: VersionSelect Component cannot be migrated due to dependencies limitations.

### Process
- [N/A] Docs updated

### Requested feedback
More an FYI. SwaggerUI library has some issues with React version 16+ and does not support hooks implementation. Components rendered inside the SwaggerUI Object , like VersionSelect, cannot be migrated to hooks due to the library limitation.


src/containers/support/SupportContactUs.tsx
```diff
@@ -1,57 +1,51 @@
...
-    if (this.state.sent) {
-      return <SupportConfirmation />;
-    } else {
-      return (
+  /**
+   * RENDER
+   */
+  return (
+    <>
+      {sent && <SupportConfirmation />}
```
anthony-kopczyk, line 39: I've seen this before for rendering, and I'm not sold one way or the other. Can you sell me on this particular method?

anthony-kopczyk:
> More an FYI. SwaggerUI library has some issues with React version 16+ and does not support hooks implementation. Components rendered inside the SwaggerUI Object , like VersionSelect, cannot be migrated to hooks due to the library limitation.

This kind of came up in another PR with OperationTag. Wondering if we should leave the swagger plugins as classes to leave them more closely resembling the original files for future updates.

That other PR: https://github.com/department-of-veterans-affairs/developer-portal/pull/437

src/containers/support/SupportContactUs.tsx
johanna-adhoc: I like to have just one return statement because if the files grows in size too much the logic that returns the component content is still centralized and easy to find, but that is totally a personal preference.

johanna-adhoc:
> > More an FYI. SwaggerUI library has some issues with React version 16+ and does not support hooks implementation. Components rendered inside the SwaggerUI Object , like VersionSelect, cannot be migrated to hooks due to the library limitation.
> 
> This kind of came up in another PR with OperationTag. Wondering if we should leave the swagger plugins as classes to leave them more closely resembling the original files for future updates.
> 
> That other PR: #437

Yes, if the component is going to be rendered inside the swaggerUI there is no choice but to leave it as a class. If we know that that is the case for all components within src/documentation/swaggerPlugins then we can say that that folder will not get migrated.

anthony-kopczyk (approved):
Everything looks good to me!

src/containers/support/SupportContactUs.tsx
```diff
@@ -1,57 +1,51 @@
...
-  sent: boolean;
-}
-
-export default class SupportContactUs extends React.Component<{}, ISupportContactUsState> {
-  constructor(props: {}) {
-    super(props);
-    this.state = {
-      sent: false,
-    };
+const SupportContactUs = (): JSX.Element => {
```
mjlumetta, line 27: when we do add our future ESLint linting, would we prefer to standardize on arrow functions or the classic `function()` declaration? or remain agnostic?

mjlumetta (approved):
looks good! 💅 🚀 

Swagger 😡 🙈

src/containers/support/SupportContactUs.tsx
gileswells: I'm on board with standardizing on arrow functions.

johanna-adhoc: I personally prefer arrow functions, and I thought the eslint config was already set to enforce them with "prefer-arrow/prefer-arrow-functions" rule 🤔 

However it is I vote to use arrow functions as the standard.

anthony-kopczyk: I'm also in agreement

PR #442, opened 2020-09-25
Api 2496: Hook migration for ApiPage, VeteransCrisisLine

anthony-kopczyk:
### Description
API-2496: https://vajira.max.gov/browse/API-2496

Updated the two components to use React Hooks
- ApiPage
- VeteransCrisisLine

Note that we plan to update ApiPage to use hooks for obtaining it's routing properties so I left its PropTypes as an empty object, since it won't be taking any in the future. https://vajira.max.gov/browse/API-2650

### Process
- [N/A] Docs updated

### Requested feedback
- I created a custom hook for the modal on VeteransCrisisLine. Wondering if we should make an effects folder to pull any custom effects out for reuse later on - or at least to clean up the files. Any thoughts on that? If people agree I can go ahead and do that under this PR.

### DIFF
--- a/src/containers/documentation/ApiPage.tsx
+++ b/src/containers/documentation/ApiPage.tsx
@@ -73,4 +73,6 @@ const ApiPage = (props: RouteComponentProps<IApiNameParam>): JSX.Element => {
   );
 };
 
+ApiPage.propTypes = {};
+
 export default ApiPage;
--- a/src/containers/documentation/ApiPage.tsx
+++ b/src/containers/documentation/ApiPage.tsx
@@ -21,13 +21,9 @@ const DeactivationMessage = ({ api }: { api: IApiDescription }) => {
     return null;
   }
 
-  if (!api.deactivationInfo) {
-    return null;
-  }
-
   const content = isDeactivated
-    ? api.deactivationInfo.deactivationContent
-    : api.deactivationInfo.deprecationContent;
+    ? (api.deactivationInfo?.deactivationContent || (() => 'Deactived API'))
+    : (api.deactivationInfo?.deprecationContent || (() => 'Deprecated API'));
   return (
     <div className={classNames('usa-alert', 'usa-alert-info', 'va-api-deprecation-alert')}>
       <div className={classNames('usa-alert-body')}>{content({})}</div>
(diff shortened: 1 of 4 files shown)


src/containers/documentation/ApiPage.tsx
```diff
@@ -21,47 +21,56 @@ const DeactivationMessage = ({ api }: { api: IApiDescription }) => {
...
 
+  if (!api.deactivationInfo) {
+    return null;
+  }
+
   const content = isDeactivated
-    ? api.deactivationInfo!.deactivationContent
-    : api.deactivationInfo!.deprecationContent;
+    ? api.deactivationInfo.deactivationContent
+    : api.deactivationInfo.deprecationContent;
```
anthony-kopczyk, line 30: For eslint I tried removing the non-null assertions, and I agree they should be removed - but, I'm not sure how we should handle these cases or what we should return if they would be null. Any feedback on this is greatly appreciated!

src/containers/documentation/ApiPage.tsx
```diff
@@ -21,47 +21,56 @@ const DeactivationMessage = ({ api }: { api: IApiDescription }) => {
     return null;
   }
 
+  if (!api.deactivationInfo) {
+    return null;
+  }
+
```
anthony-kopczyk, line 27: Added for eslint, since it doesn't like non-null assertions.

src/containers/documentation/ApiPage.tsx
```diff
@@ -21,47 +21,56 @@ const DeactivationMessage = ({ api }: { api: IApiDescription }) => {
...
+  const api = getApi(params.apiName);
+  if (api === null) {
+    return <ExplorePage />;
   }
-}
+
+  const category = lookupApiCategory(params.apiCategoryKey);
+  if (category === null) {
+    return <ExplorePage />;
+  }
```
anthony-kopczyk, line 57: Added this for the same issue as above with non-null assertions. If category would be null, I don't know if we should return the explore page, if we should throw an error, or if we should set a default value and return the regular page.

johanna-adhoc (approved):
LGTM 💯

PR #443, opened 2020-09-25
Api 1402/log validation errors

nwalker-va:
### Description
This PR is for [API-1402](https://vajira.max.gov/browse/API-1402). It will log specific validation errors in Sentry that occur from filling out forms such as the "Contact Us" form and "Developer Application" form so that it is easier to identify and debug specific reasons why an error might have occurred. 

### Process
- [ ] Docs updated


src/actions/apply.spec.ts
```diff
@@ -88,13 +88,39 @@ describe('submitForm', () => {
...
+    const getState = jest.fn();
+    getState.mockReturnValueOnce(appState);
+    await actions.submitForm()(dispatch, getState, undefined);
+    expect(dispatch).toBeCalledWith({
+      type: constants.SUBMIT_APPLICATION_BEGIN,
+    });
+    expect(dispatch).toBeCalledWith({
+      status: 'email must be valid email',
+      type: constants.SUBMIT_APPLICATION_ERROR,
+    });
```
mjlumetta, line 114: do we want to add an assertion about Sentry capturing the exception? would that be hard to mock?

src/actions/apply.ts
```diff
@@ -129,13 +129,16 @@ export const submitForm: ActionCreator<SubmitFormThunk> = () => {
...
-        if (!response.ok) {
+        if (!response.ok && response.status !== 400) {
           throw Error(response.statusText);
         }
         return response;
       })
       .then(response => response.json())
       .then(json => {
+        if (json.errors) {
+          throw Error(json.errors.join(', '));
```
mjlumetta, line 140: one thing i noticed while testing this is that the catch block for this promise chain _appears_ to be trying to log a warning instead of an error, but the result in Sentry is an error. it's possible that i haven't looked far enough into the docs to understand what's happening here, or that the original author didn't accomplish what they thought they did. i think an error is more appropriate there, but should we try to clean that up while we're working with Sentry error handling in this API request?

mjlumetta (commented):
looks good!! left a couple of questions. 

my main question, outside of those small inline questions, would be about Contact Us. the description mentions it, but it's not covered in this PR. was that supposed to be included in this PR, or will that be done separately?

src/actions/apply.spec.ts
```diff
@@ -88,13 +88,39 @@ describe('submitForm', () => {
     });
   });
 
-  it('dispatches error events when fetch returns non-200', async () => {
+  it('dispatches error events when fetch returns validation errors with status 400', async () => {
     server.use(
       rest.post(
         constants.APPLY_URL,
         (req: MockedRequest, res: ResponseComposition, context: typeof restContext) => {
           return res(
```
rtravitz, line 96: There's an eslint fix that might be nice to clean up here around implicit arrow function returns, even though it's not enforced yet.

src/actions/apply.spec.ts
rtravitz: Agree that we probably want to assert that given the acceptance criteria of the ticket. It should be possible to mock because of how `Sentry` is imported.

rtravitz (commented):
Looks good overall! Same questions as @mjlumetta mostly:
1. Did you intend to also handle the contact form with this PR?
2. There are some ESLint errors in both the test and the apply function. While we're not enforcing those, would it be good to fix them up as we go along?

src/actions/apply.ts
```diff
@@ -129,13 +129,16 @@ export const submitForm: ActionCreator<SubmitFormThunk> = () => {
 
     return fetch(request)
       .then(response => {
-        if (!response.ok) {
+        if (!response.ok && response.status !== 400) {
```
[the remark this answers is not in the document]
nwalker-va: Yes, great suggestion! Being a newcomer myself, that would be super helpful :)

src/actions/apply.spec.ts
nwalker-va: Will look into it

nwalker-va:
@rtravitz @mjlumetta Thanks for looking over my PR! In regards to the question concerning the Contact Us form, the validation errors are handled the same way and run through the same function as the developer application, but you did help me realize that I should add a test for that on the contact form. And maybe that is what you were referring to. I shall add that to this PR.

nwalker-va:
@rtravitz I'll also look into the ESLint errors. Thanks for pointing it out

rtravitz:
> @rtravitz @mjlumetta Thanks for looking over my PR! In regards to the question concerning the Contact Us form, the validation errors are handled the same way and run through the same function as the developer application, but you did help me realize that I should add a test for that on the contact form. And maybe that is what you were referring to. I shall add that to this PR.

I might be misunderstanding, but I don't actually think that, that's true. The [contact us form submits in its own function here](https://github.com/department-of-veterans-affairs/developer-portal/blob/b5e45d6bcd8d362c7a5ba44a37842046db77d32e/src/containers/support/SupportContactUsForm.tsx#L205-L220), which handles the error directly in the component without ever sending to Sentry. Is there some other automatic capture happening that grabs the backend error properly?

nwalker-va:
@rtravitz @mjlumetta Ah, you are correct Ryan! My bad. I got mixed up because the developer application and contact us form both use the same validation middleware on the backend where the validation errors are being generated. I forgot this is in the front end that we are logging the sentry errors, so I've got to log it in both places. Thank you!

src/actions/apply.ts
nwalker-va: I am new to Sentry and I'm not sure what the original author intended, but I did see this: 
Screen Shot 2020-09-29 at 9 55 18 AM

nwalker-va: The level is set to warning, but should it be error? Is this what you are referring to? @mjlumetta

src/actions/apply.spec.ts
nwalker-va: Done!

src/actions/apply.spec.ts
nwalker-va: Fixed up some of the lint errors in this file :D

src/actions/apply.ts
mjlumetta: okay, i think i get what's happening here. i was getting confused by the big "Error" header in the top left. seems like the `setLevel('warning')` determines the label you highlighted and using `captureException()` determines the Error event type (if that's what it is)? there are alternative methods for `captureEvent()` and `captureMessage()`, but i guess this seems like the right one. an error with a warning label is a bit confusing to me, but i guess i get it if it's because it's not an error-error? like it means that the server handled the bad request correctly... i guess the issue would be that client-side validation didn't work, but that could probably be worked around by somebody messing with the browser.

really long way of saying no changes here 😅

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -212,9 +213,25 @@ export default class SupportContactUsForm extends React.Component<
...
+      // The developer-portal-backend sends a 400 status, along with an array of validation error strings, when validation errors are present on the form.
+      if (!response.ok && response.status !== 400) {
+        throw Error(response.statusText);
+      }
+      return response;
+    })
+    .then(response => response.json())
+    .then(json => {
+      if (json.errors) {
+        throw Error(json.errors.join(', '));
```
mjlumetta, line 227: would suggest one very small change here and in the equivalent apply promise chain to distinguish them from each other. could we add some prefix to the front of the error message to indicate where the errors come from, like "Validation errors in Contact Us form: "? i guess the URL will be captured as well, and if that's sufficient for other people, i'm cool with that. but this could be helpful in terms of making the source more visible.

mjlumetta (commented):
one tiny change request for tweaking the error message to be different across the two forms, but overall looks great! 💅

src/containers/support/SupportContactUsForm.tsx
nwalker-va: I like the suggestion. Went ahead and made the change.

src/actions/apply.ts
rtravitz: +1 to Mike's evaluation. A bit confusing, but it makes sense for the captured literal `Error` to be labeled as a warning that our frontend validation is out of sync with our backend validations. No changes!

rtravitz (commented):
There's a `SupportContactUsForm.test.tsx` that we likely want a test in, too.

src/containers/support/SupportContactUsForm.test.tsx
```diff
@@ -93,4 +97,43 @@ describe('SupportContactUsForm', () => {
...
+      firstName: {value: 'Spongebob', dirty: true},
+      lastName: {value: 'Squarepants', dirty: true},
+      organization: {value: 'Krusty Krab', dirty: true},
+    });
+    component.update();
+
+    const form = component.find('Form');
+    form.find('.usa-button-primary').simulate('click');
+
+    formSubmissionSpy.mockImplementationOnce(() => {
```
mjlumetta, line 118: i'm a bit confused by defining the mock implementation after the submit button is clicked, as well as including the server mock and an assertion in the implementation function. i think this is causing the test to pass without asserting the things we'd like it to.

looking closer, i think that we probably don't want to mock `formSubmission` in this case since it includes the code under test (spying is still good if we'd like). i _think_ that if you take the stuff in the mock function outside into the body of the test function, with the server mock above the submit click and the assertions where they are now after it, then it would work.

mjlumetta (changes requested):
think there's an issue with the test setup that causes the assertion not to execute as expected. let me know if you're seeing something different or would like to talk it through more or if the comment's not clear.

src/containers/support/SupportContactUsForm.test.tsx
nwalker-va: I agree. I had originally set this up the way that you mentioned, but had the hardest time getting it to pass and wasn't able to. Eventually I ended up here, but you're correct. I should not mock formSubmission. I will refactor this to test what we actually want. I'll have to play around with it some more though because it doesn't pass if I take it out. I'll figure out why though.

nwalker-va: Fixed!

mjlumetta (approved):
looks good!! 🙌 🚀

rtravitz (approved):
A little questionable to use non-LOTR test data... (nice work!) :shipit:

PR #444, opened 2020-09-28
fix side skip nav hover bug

ediiotero:
### Description

Ticket: [API-2593](https://vajira.max.gov/browse/API-2593)

Bugfix for the hover on the skip nav above the sid nav.

On hover, it displayed a white box with a yellow border, making the text and background white. I made CSS changes to fall in line with the styling of the side nav. It now has a gray background and dark text similar to the hover of the side nav.

### Requested feedback

It's a small change but not sure it UX needs input on this or not.


gileswells (dismissed):
Looks solid to me. I was considering just leaving it blue even on hover since it's a keyboard nav thing but I prefer the standard grey hover state.

src/components/SideNav.scss
```diff
@@ -28,9 +28,19 @@ a.va-api-secondary-skip-nav {
...
   &:focus {
     position: relative;
     top: auto;
   }
+
+  &:visited {
+    color: $color-white;
+  }
+
+  &:hover {
```
mjlumetta, line 42: the hover style should really match the focus style, imo. like @gileswells said, the purpose of the link is for keyboard navigation, so this should be a rare thing. there are basically two reasons i would say the styles shouldn't change:

1. hovering + focusing would actually _lose_ emphasis if focus is dark blue but hover + focus is gray. emphasis is particularly important for this design of the skip nav because we want it to be extremely clear that it is not part of the side nav, since they are separate but adjacent.
2. there's no separate meaning conveyed by hovering, and therefore the styles should not change.

thoughts on this?

src/components/SideNav.tsx
```diff
@@ -147,7 +147,6 @@ export default class SideNav extends React.Component<ISideNavProps> {
             'vads-u-line-height--3',
             'vads-u-text-decoration--none',
             'vads-u-display--block',
-            'vads-u-color--white',
```
mjlumetta, line 150: would restore this in conjunction with the changes suggested above

src/components/SideNav.scss
```diff
@@ -28,9 +28,19 @@ a.va-api-secondary-skip-nav {
...
   background: $va-api-color-blue-dark;
+  color: $color-white;    
 
   &:focus {
     position: relative;
     top: auto;
   }
+
+  &:visited {
+    color: $color-white;
```
mjlumetta, line 39: what made this change necessary? losing `!important`?

note for the future: while i think we may want to add the VADS utility back and `!important` is bad anyway, if you'd like to use a VADS style sometimes but not always, you can use SASS's `@extend` directive. not a necessity, but it's a handy trick to be aware of. https://sass-lang.com/documentation/at-rules/extend

mjlumetta (changes requested):
would suggest changing the approach to fixing this bug. i think we want to cancel out the changes from hover rather than introducing a new style to fix this bug.

src/components/SideNav.scss
ediiotero: The visited link was purple when you went back to it. 
Removed and working as it was before the change

src/components/SideNav.tsx
ediiotero: done

src/components/SideNav.scss
ediiotero: keeping the background blue makes sense, as it is not part of the side nav.
- Changed to the blue background with white text

gileswells (approved):
Looks great with the updates to have it match focus.

nathanbwright (approved):
Looks good to me!

mjlumetta (approved):
looks great! thanks for bringing it in line with the focus state 🙂 🎉

PR #445, opened 2020-09-28
API-2535: Hooks Migration App and AuthorizationCard

johanna-adhoc:
### Story [API-2535](https://vajira.max.gov/browse/API-2535)
Migration of App and  AuthorizationCard components logic from classes to hooks.

### Process
- [N/A] Docs updated

### Requested feedback
All feedback welcome.


src/components/authorizationCard/AuthorizationCard.tsx
```diff
@@ -0,0 +1,21 @@
+import * as PropTypes from 'prop-types';
```
mjlumetta, line 1: can you share a bit about the motivation for nesting this component in its own directory? off the top of my head, i'm not sure what else would belong here with this file, if we're aiming to group related components in shared directories. i could see this fitting into a "navigation" directory, or maybe "authorization" or "documentation", or something like that, if we wanted to group it with others. thoughts?

src/components/authorizationCard/AuthorizationCard.tsx
```diff
@@ -0,0 +1,21 @@
...
+import * as React from 'react';
+
+import CardLink from '../CardLink';
+
+const AuthorizationCardProps = {
+  categoryKey: PropTypes.string.isRequired,
+};
+
+type IAuthorizationCardProps = PropTypes.InferProps<typeof AuthorizationCardProps>;
```
mjlumetta, line 10: one thing that we decided not too long ago but haven't yet standardized on everywhere was eliminating the `I` prefix from interface names. mostly the prefix _is_ present on most interfaces in this repo still, but we want to move away from it. 

not really sure how to do that with the `propTypes` object + the Typescript type (reallyyy love the `PropTypes.InferProps<>` though, that's awesome). i guess renaming the object is an option (`authorizationCardProps`? `componentProps`? dunno), or we could inline `PropTypes.InferProps<...>` in the function definition. i guess we can't reorder them, though.

other question: is including the prefix on types a convention? not familiar with that, but was curious if you had any background to share there. would also like to hear your thought, or anyone else's, on types vs interfaces in general. (there are more interfaces than types in this project, i think, but there's no real pattern or consistency.)

mjlumetta (commented):
couple of stylistic comments but nothing major. looks great, wanting to get your thoughts on those questions before approving.

src/components/authorizationCard/AuthorizationCard.tsx
johanna-adhoc: The other file that will be place in this directory is authorizationCard.test.tsx. So the structure that I was following is just to create a folder per component so the tsx, scss and test.tsx files live in the component folder (following what I did with the header component.)

I was not aiming to create folder buckets with categories because to be honest I am not sure what categories would work to distribute all these components.

src/components/authorizationCard/AuthorizationCard.tsx
johanna-adhoc: Not really a convention, I just tried to use the same name that the component was already using. I can rename IAuthorizationCardProps => AuthorizationCardPropTypes if that makes more sense.

mjlumetta: yeah, i think that'd be good. if this works for you, it would make more sense to me to use "AuthorizationCardPropTypes" for "AuthorizationCard.propTypes" and "AuthorizationCardProps" for "typeof props". as long as we're getting away from the prefix, though, it'll work for me.

src/components/authorizationCard/AuthorizationCard.tsx
mjlumetta: ahhh okay, i see. i'm not really sure either how to group the components, so it makes sense to me not to try to create those groupings. given that, for this case of components with no logical siblings, i would lean toward keeping the components flattened at `src/components`, if you're open to it. i'm okay with either - it's kind of a "6 on one hand, half dozen on the other" situation imo.

src/App.tsx
```diff
@@ -37,33 +37,27 @@ export function getFlags() {
...
   };
-}
+};
 
-export default class App extends React.Component {
-  public render() {
-    const appFlags = getFlags();
-    // the double flex container only exists and is flexed to
-    // address a bug in IE11 where min-height is only respected
-    // if the parent of a flex container is also a flex container.
```
anthony-kopczyk, line 47: It might be worth leaving thing comment in the code. If we ever do a styling overhaul I could see us reintroducing this bug without realizing it.

src/components/authorizationCard/AuthorizationCard.tsx
anthony-kopczyk: *Keep in mind I don't have as much react experience, and I don't really know the common practices on structuring component folders*

We'll be scaling the app up as we add components for accounts, so I'm not sure how well a flattened `src/components` folder is going to work out in a few months. That makes me want to move the components and their other files into the folders like this, but, at that point we'll probably need to figure out logical groupings anyway.

I'll go ahead and preemptively shout out that I don't like making `common` folders because I feel like that makes things more confusing for new developers. `Common` folders don't seem to be updated to remove old, unused components, or people will put things in there that aren't actually common and that ruins the whole intent behind them.

All that said, either way works for me.

anthony-kopczyk (commented):
Looks good. I think this is ready to approve once Mike's questions are addressed, whether that means making or not making code changes.

src/components/authorizationCard/AuthorizationCard.tsx
johanna-adhoc: My personal preference (as you might have already noticed) is to keep the idea of the component folder. Reasons are that the amount of files in components folder is already big and the content is so long. I foresee that once we start implementing the new modules/pages it will continue to grow and because each component has an average of 2.5 files (min 2 files once all components have proper unit testing) , adding the component folder will cut the content in half. That is the reason I started to create theses folders. 

To Anthony's point I also agree that if we defined bucket folders wrong that would make the whole thing more confusing so those are maybe not a good idea.

However if we decided to drop the components folder then I should remove the header folder as well as  we should be consistent for all components.

Let me know your thoughts.

src/components/authorizationCard/AuthorizationCard.tsx
johanna-adhoc: ok! I renamed these

src/App.tsx
johanna-adhoc: yeah, that is a good point. I added it back

src/components/authorizationCard/AuthorizationCard.tsx
mjlumetta: yeah... i guess i might've misunderstood the intent behind the `header` directory because i was thinking of how the `crisisLine` directory works, where there are multiple components that form a more complex component. so, in my head, i was thinking that other components like `Banner`, `NavBar`, etc could also go there. i would agree that `common` directories are not particularly useful, and i wouldn't want to keep this flat forever because of the new components that we'll introduce as you build out accounts and then as Team Okapi and Team Koala build out consumer and provider features enabled by accounts. i guess i was thinking of building out more groups like the `header` directory as i was picturing it, but i'm cool with the 1 directory: 1 component structure too now that i understand it better. so no changes required.

mjlumetta (approved):
🥇 🚀

anthony-kopczyk (approved):
Looks good!

PR #446, opened 2020-09-29
API-1792: 404 page design

gileswells:
### Description
https://vajira.max.gov/browse/API-1792
Updates the 404 page to be more pleasing to the eye and also more functional in the means of providing popular links.

### Process
- [ ] Docs updated

### Requested feedback
I set the headers to each section as links. It was unclear in the screenshots if that was required. I feel this is the right call but if someone objects please let me know.


gileswells:
@brentschneider See screenshots above for what it looks like at various break points. The desktop one includes the hover state on the Release Notes header. Also I updated the headings and links on the second grouping to match what's current on the site vs what was in the designs. If this needs to be reverted back please let me know.

localhost_3000_cookie (1)
localhost_3000_cookie (2)
localhost_3000_cookie (3)

src/containers/NotFound.scss
```diff
@@ -0,0 +1,42 @@
+@import '../styles/variables';
+@import '~@department-of-veterans-affairs/formation/sass/full';
+
+.vaapi-not-found-header {
+  background: #c4c4c4;
+
+  h1,
+  p {
+    color: #132f52;
```
ediiotero, line 9: Not sure what direction to go with this. There seem to be a lot of hardcoded colors in our CSS so I guess this is fine, though I would lean toward vars for theme colors. For this blue maybe $va-api-color-blue-dark would be a better choice. Again, not sure just kind of asking a question so I have a better idea of how to approach this when I am working on CSS.

src/containers/NotFound.tsx
```diff
@@ -1,11 +1,143 @@
...
+              'small-screen:vads-l-col--12',
+              'medium-screen:vads-u-margin-y--0',
+              'vads-u-margin-bottom--2',
+              'medium-screen:vads-u-order--last',
+              'vads-u-order--first',
+            )}
+          >
+            <img
+              className="vads-u-width--auto"
+              src={require('../assets/404.svg')}
```
ediiotero, line 62: if we want to continue with the pattern established in the app, maybe move the svg to an import and using that in your src={ }

ediiotero (commented):
This all looks good to me just had a couple of questions

brentschneider:
This PR passes from a design perspective.

Excellent work on implementing the layout @gileswells!
Nice touch converting the group headings to links.

src/containers/NotFound.scss
gileswells: Spot on with the color catch. Not sure where I was sourcing that blue from but it didn't match an existing color. After re-sampling I found the right variable to go with to match the 404 in the SVG.

src/containers/NotFound.tsx
gileswells: Just how I've done it in the past in other companies. I've updated this to fit the pattern here.

gileswells:
@brentschneider there were some minor color updates to match the supplied art better. Could you please 👍  if the colors are good to go? No other layout changes in this update. (h1 and p tag in the header are darker/match the 404 graphic blue color and the section header also is updated to match the 404 graphic's blue color)

image

ediiotero (approved):
looks good 👍

PR #449, opened 2020-10-01
Api 2543: Hook migration for PageContent

anthony-kopczyk:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-2543

This upgrades the PageContent component to use react hooks.

This component uses Location, but until we upgrade react-router and react-router-dom we can't use those associated react hooks. We'll have to revisit this component to update it to use the routing hooks in another PR after the upgrades.

### Process
- [N/A] Docs updated

### Requested feedback
I tried to exactly mimic the previous functionality where it only scrolls when the component updates, not on the initial render. I can change it to also scroll on initial render, but that causes an issue in the unit tests. I'd like feedback on that.


src/components/PageContent.tsx
```diff
@@ -2,26 +2,38 @@ import * as React from 'react';
 import { RouteComponentProps } from 'react-router';
 import { SiteRoutes } from '../Routes';
 
-export default class PageContent extends React.Component<RouteComponentProps, {}> {
-  private mainRef: React.RefObject<HTMLElement> = React.createRef<HTMLElement>();
+const focusAndScroll = (elementToFocus?: HTMLElement | null) => {
```
gileswells, line 5: Is there a reason that you're made this optional and have a null option available? From what I can tell there's no instance where the PageContent component is rendered that the `<main>` element isn't included.

gileswells (changes requested):

src/components/PageContent.tsx
```diff
@@ -2,26 +2,38 @@ import * as React from 'react';
...
+};
+
+const PageContent = (props: RouteComponentProps): JSX.Element => {
+  const mainRef = React.useRef<HTMLElement>(null);
+  const prevLocationRef = React.useRef<string | null>(null);
+
+  const { location } = props;
+
+  React.useEffect(() => {
+    const currentPath = location.pathname && location.hash;
```
mjlumetta, line 20: i'm a bit confused by how this works. wouldn't this always be the value of the `hash`?

src/components/PageContent.tsx
```diff
@@ -2,26 +2,38 @@ import * as React from 'react';
...
+    const prevLocation: string | null = prevLocationRef.current;
```
mjlumetta, line 21: haven't entirely wrapped my head around this one yet, but is this ref entirely necessary? i wonder if it could be eliminated by using more precise dependencies for `useEffect` - so `location.pathname`, basically. talking through this...

- the old condition was "do not execute if the old path and new path are the same and there is a hash". negated, it's "do execute if the old path and the new path are different or if they are the same but there is no hash". so the old behavior _almost_ forks on whether or not `location.pathname` has changed.
- the hash piece, when there is a hash, means that if we've navigated to an anchor link in the same page, we don't want to set the focus on the main content. including it, as opposed to leaving the condition as `props.location.pathname !== prevProps.location.pathname`, means that the focus + scroll _should_ happen if we are on the same page but navigate from a URL for the page with a hash anchor to one without one. my question for you and other reviewers is, is that exception meaningful, or should we stop worrying about the hash entirely and say this effect always happens if the `pathname` changes?

on the other hand, maybe we do need the ref to distinguish between the initial render vs subsequent renders... meh. but we could cut out the rest of the conditional where the explicit path check happens.

src/components/PageContent.tsx
```diff
@@ -2,26 +2,38 @@ import * as React from 'react';
...
-        <SiteRoutes />
-      </main>
-    );
-  }
-}
+    prevLocationRef.current = currentPath;
+  }, [location]);
+
+  return (
+    <main id="main" ref={mainRef} tabIndex={-1}>
```
mjlumetta, line 31: could conceivably get rid of the `mainRef` ref if we wanted and replace it with `document.getElementById('main')?.focus()` or `document.querySelector('main')?.focus()`. we've used that more vanilla JS approach elsewhere for skip nav links, although those had to jump between multiple components, so it was more necessary. worth considering but no changes necessary

src/components/PageContent.tsx
```diff
@@ -2,26 +2,38 @@ import * as React from 'react';
...
+  if (elementToFocus) {
+    elementToFocus.focus();
+  }
 
-  public componentDidUpdate(prevProps: RouteComponentProps) {
-    const { location } = this.props;
-    if (prevProps.location.pathname === location.pathname && location.hash) {
-      return;
-    }
+  window.scrollTo(0, 0);
```
mjlumetta, line 10: should we actually be scrolling to the top of the page or to the top of main? i don't think it super matters here given the size of the header, but wanted to ask

src/components/PageContent.tsx
mjlumetta: tested and confirmed. this value doesn't ever reflect the path, and it causes `focusAndScroll` to execute in every render (unless i guess you go to the same anchor? or maybe the same anchor on a different page? didn't try these). the focus and scroll for same page anchor links still works though, it appears; guessing that's because those are implemented via click handlers farther on? unsure.

mjlumetta (changes requested):
there's an issue with the conditional checking that should be fixed. other than that, got some general refactor questions about how this component works, but nothing that needs to be addressed.

i think running `useEffect` equivalently on the initial render and subsequent renders is part of the idea behind the hook, right? or what you're encouraged to do. anyway, i think using the ref is a fine solution for that. or maybe even a boolean for that if we don't need the value? dunno.

src/components/PageContent.tsx
anthony-kopczyk: I'll look into it. I think React was complaining that it was possibly not initialized, but that could have been due to another bug.

src/components/PageContent.tsx
anthony-kopczyk: I don't know. This is from the previous code too. As a user I generally prefer that it would scroll to the top when navigating if it's effectively loading a different page. Maybe we need to reevaluate this code completely? I really don't know why it's here, but assumed I'd leave everything functionally equivalent.

src/components/PageContent.tsx
anthony-kopczyk: Yes. I had the same question. This is a direct copy from the preexisting code. I didn't look into it specifically but assumed it was due to compatibility with an older browser or something that might not have hash defined. Will look into this further.

src/components/PageContent.tsx
mjlumetta: i think it works. i agree that you're right not to change it, and i believe that is the intended functionality. was confusing myself a bit with the focus and scrolling (since those are really different things - scroll to top of the body, focus on main).

src/components/PageContent.tsx
mjlumetta: i think there's an order of operations thing going on here. the old condition is `prevProps.location.pathname === location.pathname && location.hash`. in that conditional, the `===` gets evaluated first and then the `&&`. so the condition is the same as `(prevProps.location.pathname === location.pathname) && location.hash`.

anthony-kopczyk: Ooh. Good eye! Thanks!

src/components/PageContent.tsx
anthony-kopczyk: For this case I'll leave it with the ref for now. My thinking is to use hooks wherever it is easy-ish. But, definitely will keep that in mind for other situations like the skip nav links.

src/components/PageContent.tsx
anthony-kopczyk: I'll reevaluate the need for the logic. My intent was to keep everything that was originally there, but the logic doesn't make a whole lot of sense to me either. - and then it's a little bit more confusing because we have other links that scroll down for hashlinks that basically overrides this behavior.

src/components/PageContent.tsx
anthony-kopczyk: Should be fixed in the latest commit. I think I got confused and thought mainRef had a chance to be undefined, which really doesn't make any sense now that I look at it again.

src/components/PageContent.tsx
anthony-kopczyk: I think we do want the distinction. If you navigate into API documentation and want to back to the top of the page using the left side nav, it doesn't navigate up without the `scroll and focus when there is no hash`

anthony-kopczyk: I think the issues with this are likely resolved in the latest commit. I restructured to use two effects, one for when the pathname changes and one for when the hash changes.

anthony-kopczyk:
I added a fix for ModalController to this. I can move it into a separate PR if we think this one might take more before being able to approve.

src/components/PageContent.tsx
```diff
@@ -2,26 +2,37 @@ import * as React from 'react';
...
-    if (this.mainRef && this.mainRef.current) {
-      this.mainRef.current.focus();
+  React.useEffect(() => {
+    // If the hash was removed, then we want to scroll back to the top of the page
+    if (location.hash === '') {
+      focusAndScroll(mainRef.current);
     }
-    window.scrollTo(0, 0);
-  }
+  }, [location.hash]);
```
anthony-kopczyk, line 27: These effects are going to cause unnecessary re-renders right? I'm not sure how to really resolve that right now, but I think this matches the previous existing functionality.

src/components/PageContent.tsx
anthony-kopczyk: I went ahead and restructured back to a single effect with fixing the condition.

I think the separate effects may have worked but for the sake of getting this PR through I switched back to make the review easier.

I tried to break the logic into a When/If.
- When the location changes, if the path changes we focus and scroll.
- When the location changes, if there is no hash we focus and scroll.

src/components/PageContent.tsx
```diff
@@ -2,26 +2,40 @@ import * as React from 'react';
...
-  public componentDidUpdate(prevProps: RouteComponentProps) {
-    const { location } = this.props;
-    if (prevProps.location.pathname === location.pathname && location.hash) {
+    if (prevLocation === location.pathname && location.hash) {
       return;
     }
 
-    if (this.mainRef && this.mainRef.current) {
-      this.mainRef.current.focus();
+    if (prevLocationRef.current) {
```
johanna-adhoc, line 26: The PR looks good. The only thing left to change is that this condition should be checking mainRef object instead of prevLocationRef. This condition seems to be there just to make sure the DOM element exists before attempting to focus it, but because the method focusAndScroll already does that, this condition can be totally removed and the method called directly.

anthony-kopczyk: Oh. My intent behind this was to keep it from triggering on the initial component load. Though, I suppose it shouldn't matter if this kicks off at that point. I'll take that out.

anthony-kopczyk: Should be resolved now. Removing that caused an issue with the tests. When the App test tried to render, it runs `focusAndScroll` and `window.scrollTo` wasn't defined. I set it as a jest mock for now. I'm not sure where we'd want to put that kind of thing or maybe there's a package somewhere that we could use instead.

src/hooks/ModalController.tsx
```diff
@@ -2,18 +2,26 @@ import * as React from 'react';
...
 
   React.useEffect(() => {
     const handleEscape = (event: KeyboardEvent) => {
-      if (event.key === 'Escape' && modalVisible) {
+      if (event.key === 'Escape' && modalVisibleRef.current) {
         setModalVisible(false);
       }
     };
-    
+
```
anthony-kopczyk, line 24: Wondering if I should have kept this as it was previously and just added modalVisible as a dependency to `useEffect`. But then that would require redefining that function every time modalVisible updates, and that doesn't seem right to me. But, with using the ref, I have the overhead of adding a ref and also creating the extra setter. I'm wondering what is more efficient.

When modalVisible changes it has to re-render anyway, so. I guess I'm going to be looking into some react docs. If anyone has thoughts, please comment!

TS000: I like the idea of adding modalVisible to useEffect.

johanna-adhoc (dismissed):
LGTM

johanna-adhoc (dismissed):
💯

setupJest.ts
```diff
@@ -1,5 +1,9 @@
+import 'jest';
```
mjlumetta, line 1: is `jest` not globally defined here?

src/components/pageContent/PageContent.tsx
```diff
@@ -0,0 +1,37 @@
...
+    elementToFocus.focus();
+  }
+
+  window.scrollTo(0, 0);
+};
+
+const PageContent = (props: RouteComponentProps): JSX.Element => {
+
+  const mainRef = React.useRef<HTMLElement>(null);
+  const prevLocationRef = React.useRef<string | null>(null);
```
mjlumetta, line 16: would suggest naming this `prevPathRef` if you make another commit but def don't make another commit for that (motivation is specifically from just having refactored some path/hash logic in the side nav)

src/hooks/ModalController.tsx
mjlumetta: did you happen to find any references informing the ref decision? would love to learn from those and/or have something linked in a comment, if there was anything external informing this decision

src/components/pageContent/PageContent.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  const { location } = props;
+
+  React.useEffect(() => {
+    
+    if (prevLocationRef.current !== location.pathname || location.hash) {
```
mjlumetta, line 22: i think we also still need a `prevLocationRef.current != null` check here. focus skips the header on initial page load, which isn't what we want, i don't think.

mjlumetta (changes requested):
focus is getting set on main on the initial page load, and i think we want that to still be set to body. can be fixed with a check on that ref, i believe. other than that it looks ready.

src/components/pageContent/PageContent.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+import { RouteComponentProps } from 'react-router';
+import { SiteRoutes } from '../../Routes';
+
+const focusAndScroll = (elementToFocus: HTMLElement | null) => {
```
gileswells, line 5: I cannot find a way that that you'd need null as an option. There is no condition on `<main>` so it would never be null.

src/components/pageContent/PageContent.tsx
```diff
@@ -0,0 +1,37 @@
...
+  if (elementToFocus) {
```
gileswells, line 6: Without the above null option there is no need for this conditional.

gileswells (changes requested):
Still can slim down the `focusAndScroll` params in my view.

src/components/pageContent/PageContent.tsx
anthony-kopczyk: I agree that I don't think it can actually be null where it's called, but when I try to remove `| null` from the method, I get lint errors in my IDE that `mainRef.current` could be null.

src/components/pageContent/PageContent.tsx
anthony-kopczyk: Shoot. I knew that I had that there for a reason!

src/hooks/ModalController.tsx
anthony-kopczyk: I didn't find anything on specifics of using an effect or a ref. But I am definitely debating the efficiency of one or the other. I need to get deep into react with when it renders and those kinds of things.

I think, unless people have strong feelings, I'm going to leave it as it is now, and might revisit it later.

src/components/PageContent.tsx
anthony-kopczyk: Mike pointed out that without this the header doesn't get focus on initial page load. So we have to have something checking if `prevRef.current` is null to determine if it's initial page load or not. I added this back in in the latest commit.

setupJest.ts
anthony-kopczyk: It is defined globally. I ended up reverting the changes to this file. With the PageContent effect only happening when it's not an initial page load, the tests don't trigger `scrollTo` at the moment.

src/components/pageContent/PageContent.tsx
anthony-kopczyk: Should be fixed in the latest commit.

src/components/pageContent/PageContent.tsx
anthony-kopczyk: I went ahead and made this change. The ref is storing pathname not the entire location object as well, so this is cleaner.

src/components/pageContent/PageContent.tsx
anthony-kopczyk: Should be fixed in the latest commit.

src/hooks/ModalController.tsx
anthony-kopczyk: If we use effect, then we need to pass in the `modalVisible` field as a dependency. So every time it toggles the modal will need a re-render (which I guess it would anyway?). But then with the ref, we don't need to pass that as a dependency, so it doesn't need to redefine the effect on re-renders - but the ref is an extra variable to consider. The pros/cons seem about even to me at the moment.

johanna-adhoc (approved):

src/components/pageContent/PageContent.tsx
gileswells: `const mainRef = React.useRef() as React.MutableRefObject<HTMLElement>;` That would resolve the issue with the | null but in the grand scheme of things that's literally what I just found Googling on SO so I'm not sure this would the "right" answer as much as it is an answer.

If you wish to use it or not I will leave entirely up to you. I no longer have a strong opinion either way at this stage.

gileswells (dismissed):
LGTM :shipit:

mjlumetta (dismissed):
✨ 🚀

johanna-adhoc (approved):

TS000 (approved):

PR #452, opened 2020-10-01
Fix ESLint errors in src/components and add ESLint stage to Jenkins CI job

mjlumetta:
### Description

in support of [API-1323](https://vajira.max.gov/browse/API-1323). does 3 main things:

1. fixes all ESLint errors in `src/components`.
2. changes the `npm run eslint` script to only lint `src/components`.
3. adds an ESLint step to the Jenkins CI job so that new code can't cause lint errors in files that have already been fixed. here's [an example failure](http://jenkins.vfs.va.gov/blue/organizations/jenkins/testing%2Fdeveloper-portal/detail/API-1323%2Feslint-components/5/pipeline).

this PR was a little larger in scope than i had hoped. going to try to keep subsequent PRs a bit more manageable. 

### Process

  * failing ESLint checks for files/directories that are supposed to be clean
- [n/a] Docs updated

### Requested feedback


mjlumetta:
linking to [an ESLint CI failure](http://jenkins.vfs.va.gov/blue/organizations/jenkins/testing%2Fdeveloper-portal/detail/API-1323%2Feslint-components/5/pipeline) for my future self

src/components/EmbeddedYoutubeVideo.tsx
```diff
@@ -34,7 +37,15 @@ export default function EmbeddedYoutubeVideo({url, title} : IEmbeddedYoutubeVide
   const embedUrl = `${YOUTUBE_BASE}${videoInfo.id}`;
   return (
     <div className={wrapperStyles}>
-      <iframe className={iframeStyles} title={title} src={embedUrl} frameBorder="0" allowFullScreen={true} />
+      <iframe
+        className={iframeStyles}
+        title={title}
+        src={embedUrl}
+        frameBorder="0"
+        allowFullScreen
```
mjlumetta, line 45: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md

if we want to flip this the other way to require `{true}` or just turn this rule off, that's fine with me.

src/components/Form.tsx
```diff
@@ -1,63 +1,77 @@
...
 
     return (
       <form className={this.props.className}>
         {this.props.children}
         <ProgressButton
-          buttonText={this.state.sending ? "Sending..." : "Submit"}
+          buttonText={this.state.sending ? 'Sending...' : 'Submit'}
           disabled={this.props.disabled}
+          /* eslint-disable @typescript-eslint/unbound-method
+             --- will be fixed when this component is converted to a function with hooks */
```
mjlumetta, line 47: if we'd rather put the function bind in the `render()` body, that's cool too. the rule's not smart enough to detect being bound in the constructor, apparently

src/components/Form.tsx
```diff
@@ -1,63 +1,77 @@
...
 
   private submittingForm(): void {
-    this.setState({sending: true}, async () => {
-      try {
-        await this.props.onSubmit();
-        this.setState({sending: false}, () => this.props.onSuccess());
-      } catch(e) {
-        this.setState({sending: false, error: true});
-      }
+    this.setState({ sending: true }, () => {
```
mjlumetta, line 66: i don't think this callback was supposed to be async, really? i don't think it had adverse affects because `setState()` doesn't handle the promise, but now we move the unhandled promise inside this function so it's solely within our code.

src/components/GroupedAccordions.tsx
```diff
@@ -4,56 +4,67 @@ import * as React from 'react';
...
-  private panelRefs: Array<React.RefObject<React.Component<{}, ICollapsiblePanelStates>>>;
+/* eslint-disable @typescript-eslint/indent */
+export default class GroupedAccordions extends React.Component<
+  GroupedAccordionsProps,
+  GroupedAccordionsStates
+> {
+  private panelRefs: Array<
+    React.RefObject<React.Component<Record<string, unknown>, CollapsiblePanelStates>>
+  >;
+  /* eslint-enable @typescript-eslint/indent */
```
mjlumetta, line 37: this rule is kind of broken, apparently. see `.eslintrc.js` for the issue link. if we'd like to put that link in more places, sounds good to me.

src/components/HoverImage.tsx
```diff
@@ -1,44 +1,50 @@
...
     this.state = {
       src: props.imagePath,
     };
-    this.onMouseEnter = this.onMouseEnter.bind(this);
-    this.onMouseLeave = this.onMouseLeave.bind(this);
+
+    /* eslint-disable no-underscore-dangle -- temporary until we can replace with hooks */
+    this.onMouseEnter = this._onMouseEnter.bind(this) as () => void;
+    this.onMouseLeave = this._onMouseLeave.bind(this) as () => void;
+    /* eslint-enable no-underscore-dangle */
```
mjlumetta, line 25: didn't really know what to do with this one. suggestions welcome

src/components/OnlyTags.tsx
```diff
@@ -1,11 +1,24 @@
...
 import * as React from 'react';
-
 import TrustedPartnerOnlyTag from './TrustedPartnerOnlyTag';
 import VAInternalOnlyTag from './VAInternalOnlyTag';
 
-const OnlyTags = ({vaInternalOnly, trustedPartnerOnly}:{vaInternalOnly: boolean, trustedPartnerOnly: boolean}) : JSX.Element => <>
-  {vaInternalOnly ? <VAInternalOnlyTag /> : null}
-  {trustedPartnerOnly ? <TrustedPartnerOnlyTag /> : null}
-</>;
+const OnlyTagsPropTypes = {
```
mjlumetta, line 6: i was trying to lint for the presence of `propTypes`, but the `eslint-plugin-react` rule for this, [`react/prop-types`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md), will consider the Typescript-typed props as meeting this rule. unfortunate - maybe we can circle back later.

> **Note:** You can provide types in runtime types using `PropTypes` and/or statically using TypeScript or Flow. This rule will validate your prop types regardless of how you define them.

src/components/SideNav.tsx
```diff
@@ -1,26 +1,26 @@
...
+  subNavLevel?: number;
 }
 
 // Constructs a NavHashLink in the sidebar that also takes into account the
 // hash when determining if it's active
-export class SideNavEntry extends React.Component<ISideNavEntryProps> {
-  public static defaultProps = {
-    subNavLevel: 0,
-  };
+const SideNavEntry: React.FunctionComponent<SideNavEntryProps> = (props: SideNavEntryProps) => {
```
mjlumetta, line 21: i rewrote this as a function to get past the "one class per file" rule

src/components/crisisLine/CrisisPanelInfo.tsx
```diff
@@ -1,30 +1,43 @@
...
+const CrisisPanelInfoPropTypes = {
+  children: PropTypes.node.isRequired,
+  id: PropTypes.string,
+  target: PropTypes.string.isRequired,
+};
 
-interface ICrisisPanelInfoProps {
-  target: string;
+interface CrisisPanelInfoProps extends PropTypes.InferProps<typeof CrisisPanelInfoPropTypes> {
+  // left icon off because it was difficult to find the right PropTypes type
```
mjlumetta, line 14: suggestions welcome here

tslint.json
```diff
@@ -45,6 +45,7 @@
     "react-a11y-role-supports-aria-props": {},
     "react-a11y-tabindex-no-positive": {},
     "react-a11y-titles": {},
+    "jsx-boolean-value": false,
```
mjlumetta, line 48: conflicts with `react/jsx-boolean-value`

src/components/EmbeddedYoutubeVideo.tsx
ediiotero: I'm ok with this as is

src/components/GroupedAccordions.tsx
ediiotero: I think a comment or a link would be a good idea

ediiotero (approved):
this all looks great to me

mjlumetta:
closing in favor of #499, #501, and a PR to be named later.

PR #453, opened 2020-10-02
API-2349/update-typescript-to-v4

gileswells:
### Description
Upgrades Typescript from 3.9.7 to 4.0.3 which is the latest stable release. There is 4.1 in beta currently but stuck with the stable release for all the good reasons.

### Process
- [ ] Docs updated

### Requested feedback

### DIFF
--- a/src/base.accessibility.ts
+++ b/src/base.accessibility.ts
@@ -10,7 +10,7 @@ describe('Accessibility tests', () => {
         page.on('request', mockSwagger);
       }
 
-      await page.goto(`${puppeteerHost}${path}`, { waitUntil: 'networkidle0' });
+      await page.goto(`${puppeteerHost}${path}`, { waitUntil: 'networkidle2' });
       await page.addScriptTag({ path: require.resolve('axe-core') });
 
       // THIS IS BAD. code highlighting on the Authorization page has several color contrast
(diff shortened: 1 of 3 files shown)


src/base.accessibility.ts
```diff
@@ -10,7 +10,7 @@ describe('Accessibility tests', () => {
         page.on('request', mockSwagger);
       }
 
-      await page.goto(`${puppeteerHost}${path}`, { waitUntil: 'networkidle0' });
+      await page.goto(`${puppeteerHost}${path}`, { waitUntil: 'networkidle2' });
```
gileswells, line 13: Consider navigation to be finished when there are no more than 2 network connections for at least 500ms. networkidle0 required 0 network connections within 500ms. On my local this change slimmed down the test by 10 seconds overall with no false failures.

gileswells: Previous build failed at the accessibility test with `76.987 s` duration and the current build succeeded with `77.055 s` duration with this change in place. 🤷‍♂️

rtravitz: Nice optimization! Maybe throw a comment above about the subtleties of `networkidle2` you mentioned here?

rtravitz (approved):
It passes CI and deploys working review bundles. That's good in my book. 🚀

PR #458, opened 2020-10-05
Fix ESLint checks for Support components

mjlumetta:
### Description

in support of [API-1323](https://vajira.max.gov/browse/API-1323). fixes ESLint errors in `src/containers/support`.

see #452 for the opening PR for configuring ESLint. that PR also includes adding ESLint to the Jenkins CI job. i can add that here as well if this one ends up going first for whatever reason, but going to default to letting that PR get merged first.

### Process

- [n/a] Docs updated
  - no UI changes

### Requested feedback


src/containers/support/SupportFAQ.tsx
```diff
@@ -33,75 +43,101 @@ const developmentQuestions: ISupportQuestion[] = [
...
     ),
     question: 'Can I start using the API as soon as I sign up?',
   },
   {
-    answer: 
-      <p> Yes, we have implemented basic rate limiting of 60 requests per minute. If you exceed this quota, your request will return a 429 status code. You may petition for increased rate limits by emailing api@va.gov and requests will be decided on a case-by-case basis.
-      </p>,
+    answer: (
+      <p>
+        {' '}
```
gileswells, line 87: Extra space should be removed

src/containers/support/SupportFAQ.tsx
```diff
@@ -8,23 +8,33 @@ const generalQuestions: ISupportQuestion[] = [
     answer: (
       <p>
         No - this is a support page for software developers utilizing the Veterans Affairs (VA)
-        Application Programming Interface (API). If you are a veteran seeking assistance, please visit
-        the <a href="http://www.va.gov">US Department of Veterans Affairs website</a> to access and
-        manage your VA benefits and health care. There are also helpful reference links and Q&amp;A at
-        the VA Inquiry Routing &amp; Information System <a href="https://iris.custhelp.va.gov/app/answers/list">(IRIS)</a>.
+        Application Programming Interface (API). If you are a veteran seeking assistance, please
+        visit the <a href="http://www.va.gov">US Department of Veterans Affairs website</a> to
```
gileswells, line 12: Linking off site should include a `target="blank"`

src/containers/support/SupportFAQ.tsx
```diff
@@ -8,23 +8,33 @@ const generalQuestions: ISupportQuestion[] = [
...
+        access and manage your VA benefits and health care. There are also helpful reference links
+        and Q&amp;A at the VA Inquiry Routing &amp; Information System{' '}
+        <a href="https://iris.custhelp.va.gov/app/answers/list">(IRIS)</a>.
```
gileswells, line 15: Linking off site should include a `target="blank"`

src/containers/support/SupportFAQ.tsx
```diff
@@ -33,75 +43,101 @@ const developmentQuestions: ISupportQuestion[] = [
...
+      'What kind of data can I get from the APIs? Do you have any example scenarios for Health, Benefits, Facilities or Veteran Verification?',
   },
   {
     answer: (
       <React.Fragment>
         <p>
-          The Address Validation API is for internal VA use only and is not listed on the developer portal. To begin development in the sandbox environment, request a developer <a href="https://developer.va.gov/apply">API key for the Facilities API</a>. Once finished, send an email to <a href="mailto:api@va.gov?subject=Request%20for%20Sandbox%20Access%20to%20Address%20Validation%20API" target="_BLANK" rel="noopener noreferrer">api@va.gov</a> with the subject line "Request for Sandbox Access to Address Validation API." In the email, include the following:
+          The Address Validation API is for internal VA use only and is not listed on the developer
+          portal. To begin development in the sandbox environment, request a developer{' '}
+          <a href="https://developer.va.gov/apply">API key for the Facilities API</a>. Once
```
gileswells, line 112: This should use a `<Link>` tag for an internal link

src/containers/support/SupportOverview.tsx
```diff
@@ -15,19 +15,19 @@ interface ISupportOverviewProps {
...
-              {section.description}
-            </CardLink>
-          );
-        })}
-      </div>
-    </section>
-  );
-}
+const SupportOverview: React.FunctionComponent<ISupportOverviewProps> = (
+  props: ISupportOverviewProps,
```
gileswells, line 19: Shouldn't we update this `I` named interface since we're touching it?

src/containers/support/SupportOverview.tsx
```diff
@@ -15,19 +15,19 @@ interface ISupportOverviewProps {
...
+): JSX.Element => (
+  <section role="region" aria-labelledby={PAGE_HEADER_ID}>
+    <PageHeader {...headerProps} />
+    <div className={defaultFlexContainer()}>
+      {props.sections.map((section: ISection) => (
```
gileswells, line 24: Another `I` named interface that could be updated

tslint.json
```diff
@@ -47,6 +47,7 @@
     "react-a11y-titles": {},
     "jsx-no-lambda": {
       "severity": "off"
-    }
+    },
+    "jsx-boolean-value": false
```
gileswells, line 51: I like this change. I am assuming this will only apply to a parameter that would have been set to `attr={true}` and not something like `attr={foo === bar}` correct?

gileswells (changes requested):

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -26,11 +26,13 @@ interface ISupportContactUsFormProps {
...
 }
 
+/* eslint-disable @typescript-eslint/indent */
 export default class SupportContactUsForm extends React.Component<
   ISupportContactUsFormProps,
   ISupportContactUsFormState
 > {
-  constructor(props: ISupportContactUsFormProps) {
+  /* eslint-enable @typescript-eslint/indent */
+  public constructor(props: ISupportContactUsFormProps) {
```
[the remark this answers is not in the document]
mjlumetta: yes, good point. i put that rule in #452 but not here. will add.

tslint.json
mjlumetta: yeah. this specifically isn't making the change - the TSLint rule conflicts with the ESLint rule, so this is shutting the TSLint rule off. but yes, i believe anything that's calculated at runtime shouldn't be covered by the rule.

you can see the TSLint `jsx-boolean-value` rule in [`tslint-react`](https://github.com/palantir/tslint-react#rules) and the ESLint rule `jsx-boolean-value` in [`eslint-plugin-react`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md). the TSLint rule defaults to "always" (as in always specify `attr={true}`) and the ESLint rule defaults to "never". so yeah, we'll go with the "never" default, unless anyone feels strongly to the contrary. glad you like it, anyway.

src/containers/support/SupportFAQ.tsx
mjlumetta: good catch! fixed

src/containers/support/SupportFAQ.tsx
mjlumetta: good catch, thanks! also updating to be an internal URL

src/containers/support/SupportFAQ.tsx
mjlumetta: i would prefer not to do this as a blanket rule for the general kinds of reasons articulated in [this article](https://css-tricks.com/use-target_blank/) - mainly, that the default behavior is usually appropriate, and that users can choose to override that default behavior with a modifier key if they choose. if you or others feel strongly about it, though, i can change it.

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -13,7 +13,7 @@ import { validateEmail, validatePresence } from '../../utils/validators';
 
 import './SupportContactUsForm.scss';
 
-interface ISupportContactUsFormState {
+interface ContactUsFormState {
```
gileswells, line 16: Why did you drop `Support` off the front of the interface name? I'd prefer we stick with the pattern that is `ComponentName(State|Props)`

src/containers/support/SupportFAQ.tsx
gileswells: Based on my history on some seriously competitive SEO ranking sites this was true there. That being said doing more research with your link and others I would tend to agree that as a blanket rule it's not the greatest. I concede by position. 👍

src/containers/support/SupportFAQ.tsx
```diff
@@ -3,162 +3,193 @@ import { Link } from 'react-router-dom';
...
 
-const SupportQuestions = (props: ISupportQuestionsProps) => {
-  const content: IPanelContent[] = props.questions.map((q: ISupportQuestion) => (
-    {
-      body: q.answer,
-      title: q.question,
-    }
-  ));
+const SupportQuestions = (props: SupportQuestionsProps) => {
+  const content: IPanelContent[] = props.questions.map((q: SupportQuestion) => ({
```
gileswells, line 176: I missed this `I` named interface in my previous review.

gileswells (changes requested):
Couple small naming comments but overall looks great.

src/containers/support/SupportContactUsForm.tsx
gileswells, line 16:
```suggestion
interface SupportContactUsFormState {
```

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -22,15 +22,17 @@ interface ISupportContactUsFormState {
   organization: IErrorableInput;
 }
 
-interface ISupportContactUsFormProps {
+interface ContactUsFormProps {
```
gileswells, line 25:
```suggestion
interface SupportContactUsFormProps {
```

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -22,15 +22,17 @@ interface ISupportContactUsFormState {
...
   onSuccess: () => void;
 }
 
+/* eslint-disable @typescript-eslint/indent */
 export default class SupportContactUsForm extends React.Component<
-  ISupportContactUsFormProps,
-  ISupportContactUsFormState
+  ContactUsFormProps,
```
gileswells, line 31:
```suggestion
  SupportContactUsFormProps,
```

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -22,15 +22,17 @@ interface ISupportContactUsFormState {
...
+  ContactUsFormState
```
gileswells, line 32:
```suggestion
  SupportContactUsFormState
```

gileswells (changes requested):
Add back Support prefix

src/containers/support/SupportFAQ.tsx
mjlumetta: fixed. i think it didn't lint for this since it was defined in a different file? anyway, no longer a problem

src/components/GroupedAccordions.tsx
```diff
@@ -24,7 +24,10 @@ interface ICollapsiblePanelStates {
   open: boolean;
 }
 
-export default class GroupedAccordions extends React.Component<IGroupedAccordionsProps, IGroupedAccordionsStates> {
+export default class GroupedAccordions extends React.Component<
+  IGroupedAccordionsProps,
```
gileswells, line 28: Just leave a mental note for these two. Not worth dealing with now. Rather get it merged.

gileswells (approved):

gileswells (approved):

src/components/GroupedAccordions.tsx
mjlumetta: yeah... would rather deal with it in-scope, should be on #452. merging now.

PR #459, opened 2020-10-05
Fix ESLint check in API definitions

mjlumetta:
### Description

supports [API-1323](https://vajira.max.gov/browse/API-1323). adds ESLint checks to all files in `src/apiDefs`. follows #452 and #458.

### Process

  - will be added to Jenkins CI pipeline, but support for that is in #452 
- [n/a] Docs updated
  - no UI changes

### Requested feedback

_specific to this PR:_ what is our preferred approach to asserting that the environment variable for Swagger host is defined?

general:


src/apiDefs/data/appeals.ts
```diff
@@ -1,13 +1,20 @@
-import { 
+import {
   AppealsStatusReleaseNotes,
   DecisionReviewReleaseNotes,
 } from '../../content/apiDocs/appeals';
 import { IApiDescription } from '../schema';
 
-const swaggerHost : string = process.env.REACT_APP_VETSGOV_SWAGGER_API!;
-const appealsApis : IApiDescription[] = [
+if (!process.env.REACT_APP_VETSGOV_SWAGGER_API) {
```
gileswells, line 7: Is there a better way to handle this vs including it on multiple files across the app? Can we promote this to a higher point so we only have to do it once? Or at least move it to maybe a file in the utils directory.

gileswells (commented):
Overall looks good. Just the question about the environment variable.

Also if you want to drop that timeout into this PR I can approve it quickly if need be.

src/apiDefs/data/appeals.ts
mjlumetta: yeah, good question. was thinking more about this and i think it'd be okay to disable the rule here. thoughts on that? another option could be `VAR || ''`, which would cause requests to fail but is roughly equivalent to how this works at runtime now.

gileswells: I think `VAR || ''` would be my vote. I know before I added that variable to my `.env.local` the app just worked except for the swagger docs and it showed an error in the console. I'd prefer to keep that functionality or if it did change that there would be a visual error on the page so that someone expecting the page to render doesn't sit there without realizing that something went wrong.

mjlumetta: that is a great point on the whole app still working. i'll go with that solution, in any case.

gileswells (approved):
Looks good to me. 

My only change, which I am totally ambivalent to and can be skipped if you disagree, might be to move the repeated `const swaggerHost: string = process.env.REACT_APP_VETSGOV_SWAGGER_API || '';` into `src/types/constants/index.ts` but that's just because I have a dislike of repeated code.

Apply the updated change or not, either way works for me.

gileswells (dismissed):

gileswells (approved):
Looks solid to me. Provided Jenkins has no issues neither do I.

gileswells (approved):

PR #460, opened 2020-10-05
Api 2523

TS000:
### Description
https://vajira.max.gov/browse/API-2523
Moved Class and Function based React components to modern Hooks. 

### Process
Process was simple as the components were pretty shallow in their functionality. 
ExtendedLayout.tsx was left as it is since Swagger doesn't support Hooks (yet). 

### Requested feedback
This is my first commit, please let me know if I need to adjust how I'm pushing commits.


TS000:
I think I need to split the branch into separate commits

PR #462, opened 2020-10-05
2527/hooks migration

TS000:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-2527

Moved the Class component to modern React Hooks component: 
* ApiKeyAuth: ✔️   
* CategoryPage: ✖️ 
* AuthorizationDocs: ✖️ 

Both CategoryPage and AuthorizationDocs use a React Router version (v4) that is not compatible with React Hooks so no changes were made to those files.

### Process

- [N/A] Docs updated

### Requested feedback
Is there a source of truth for ReactRouter of which features work and which don't with Hooks? I'd like feedback on if I approached CategoryPage and AuthorizationDocs correctly.


anthony-kopczyk (changes requested):
Think this may be as a result of our conversation earlier when I said about not switching some files to hooks that are dependent.

What you have looks good. While we can't get them switched over to use hooks, we can go ahead and switch over components to be functional even if we need to still pass in routing properties as props instead of using hooks.

Can you make changes to the other two files, even thought we'll still be passing in the routing as props?

TS000:
All set @anthony-kopczyk. I agree converting them to functional is the way to get them closer to hooks without breaking React Router.

anthony-kopczyk (changes requested):
Similar to the other PR. We've been adding return types to the functions `JSX.Element`. For the eslinting rules we'll eventually enable, we also want to enforce arrow hooks as long as possible.

src/containers/Home.tsx
```diff
@@ -14,103 +14,93 @@ const rightColumnClasses = classNames('medium-screen:vads-l-col--8', 'small-scre
...
-          <div className={leftColumnClasses}>
-            <img className={imageClasses} src={imageSrc} alt="" role="presentation" />
-          </div>
-          <div className={rightColumnClasses}>
-            <h2>
-              {title}
-            </h2>
-              {children}
-          </div>
+const HomeSection = ({ariaLabel, imageSrc, title, children} : {ariaLabel: string; imageSrc: string; title: string; children: React.ReactNode}): JSX.Element => (
```
anthony-kopczyk, line 17: I think this is only used here, so doesn't need done. But this would be a little cleaner if the type was pulled out into an interface. Again, I think this is the only place that it's used, so probably not necessary. Or, thoughts on doing it this way?

```
const HomeSection = (props: {
  ariaLabel: string;
  imageSrc: string;
  title: string;
  children: React.ReactNode}
): JSX.Element => (
```

This way we'd need to access everything via `props.prop` or have a separate line
`const { ariaLabel, imageSrc, title, children } = props;`

Not sure best way to write it.

src/containers/Home.tsx
```diff
@@ -14,103 +14,93 @@ const rightColumnClasses = classNames('medium-screen:vads-l-col--8', 'small-scre
...
+  <section role="region" aria-label={ariaLabel}>
+    <div className={flexContainer}>
+      <div className="vads-l-row">
+        <div className={leftColumnClasses}>
+          <img className={imageClasses} src={imageSrc} alt="" role="presentation" />
+        </div>
+        <div className={rightColumnClasses}>
+          <h2>
+            {title}
+          </h2>
```
anthony-kopczyk, line 27: Thoughts on changing this to `<h2>{ title }</h2>`? This is another, I'm not sure which way would be better. haha! Maybe we already have a lint rule defined for this.

src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -10,24 +11,24 @@ import ApiKeyAuth from './ApiKeyAuth';
...
-        return (
-          <div className="va-api-authorization-docs">
-            <PageHeader halo={category.name} header="Authorization" />
-            <OAuth />
-          </div>
-        );
-      } else {
-        return (<ApiKeyAuth apiCategoryKey={apiCategoryKey} />);
-      }
+export function AuthorizationDocs({ match }: RouteComponentProps<IApiNameParam>): JSX.Element {
```
anthony-kopczyk, line 14: Our eslint rules (there's a plugin on vscode that you should be able to install that will highlight things), prefer arrow functions. Can you update this to an arrow function?

src/containers/documentation/CategoryPage.tsx
```diff
@@ -13,61 +14,61 @@ import { defaultFlexContainer } from '../../styles/vadsUtils';
...
 
-export default class CategoryPage extends React.Component<RouteComponentProps<IApiNameParam>, {}> {
-  public render() {
-    const { apiCategoryKey } = this.props.match.params;
-    const {
-      apis,
-      name: categoryName,
-      content: { intro, overview },
-    } = getApiDefinitions()[apiCategoryKey];
+export default function CategoryPage({ match }: RouteComponentProps<IApiNameParam>): JSX.Element {
```
anthony-kopczyk, line 17: Same as the other file. Can you update this to an arrow function?

anthony-kopczyk (changes requested):
There are two files that we should switch to arrow functions, and then a few things that can address but don't necessarily need to address.

anthony-kopczyk (approved):
I scanned through again and don't see anything else. Everything else that's been mentioned has been addressed. Good to go!

PR #463, opened 2020-10-05
Api 2541: Hook migrations for ApiDocumentation

anthony-kopczyk:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-2541

Converts the APIDocumentation component to use react hooks

### Process
- [N/A] Docs updated

### Requested feedback

There was some old code I ended up removing because it seems non-functional. It's intention was to set tab index, but that seems to be set just by setting the api version.


TS000 (approved):
Nice use of useEffect!

anthony-kopczyk:
@TS000 I just noticed I have some return types not declared. I'll probably do another commit to add those.

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -54,6 +54,43 @@ const ApiDocumentation = (props: IApiDocumentationProps): JSX.Element => {
...
+
+  React.useEffect(() => {
+    if (
+      location.pathname !== prevLocation?.pathname ||
+      location.search !== prevLocation?.search
+    ) {
+      setTabIndexFromQueryParams();
+    }
+  }, [location.pathname, location.search, setTabIndexFromQueryParams, prevLocation]);
+
```
anthony-kopczyk, line 79: Turns out this code is actually used in some specific apis - notably the health FHIR page. It has some tabs that you can load directly into. I originally didn't realize that the tabs for `tabIndex` were different than the regular swagger documentation.

johanna-adhoc:
The handling of the API version updates in the query param could be refactored as shown below. Please let me know your thoughts. 

  /*
   * API Version
   */
  const dispatch = useDispatch();
  const queryParams = new URLSearchParams(location.search ?? undefined);
  const apiVersion =  queryParams.get('version');
  
  React.useEffect(()=>{
    console.log('gets in effect to change apiVersion');
    dispatch(actions.setRequstedApiVersion(apiVersion));
  },[dispatch, apiVersion, location.pathname]);

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -1,134 +1,117 @@
...
+  // Get doc source keys
+  const hasKey = (source: IApiDocSource) => !!source.key;
+  const tabKeys = docSources
+    .filter(hasKey)
+    .map(source => source.key?.toLowerCase() || '');
+
+  // Return tab index
+  const sourceTabIndex = tabKeys.findIndex(sourceKey => sourceKey === queryStringTab);
+  return sourceTabIndex === -1 ? 0 : sourceTabIndex;
```
anthony-kopczyk, line 46: Generally I encapsulate the inline if statements with parentheses because I think it makes it more readable. I can make the commit but curious on other's thoughts on this/if people feel one way or the other.

src/hooks/Previous.tsx
```diff
@@ -0,0 +1,14 @@
...
+  const theRef = React.useRef<T | null>(props ?? null);
+
+  React.useEffect(() => {
+    theRef.current = props ?? null;
+  }, [props]);
+
+  return theRef.current;
+};
+
+export default usePrevious;
```
anthony-kopczyk, line 14: This file is now unused. I'm not sure if we should keep it. The current implementation of ModalController (not in this branch) could use this for instance. I'm not sure if there are other places that might use it. That said, this code isn't that complicated and we could probably recreate in 5-10 minutes if we did need it again (Tons of people on the internet have similar code for this as well, so it's not like we even need to make it without having external resources to read).

anthony-kopczyk: Generally I'd say delete, but curious on others thoughts on this as well.

src/apiDefs/schema.ts
```diff
@@ -16,6 +16,14 @@ export interface IApiCategoryContent {
...
 }
 
+export const ApiCategoryContentPropType = PropTypes.shape({
+  intro: PropTypes.any.isRequired,
+  overview: PropTypes.any.isRequired,
+  placardText: PropTypes.string.isRequired,
+  quickstart: PropTypes.any,
+  shortDescription: PropTypes.string.isRequired,
+});
+
```
anthony-kopczyk, line 26: These propTypes aren't 100%, and I didn't use infer props partially because of this. I can try updating this prop types but I left a lot of these propTypes as `any`. As far as typescript is concerned we still have those typing, but runtime checks aren't as strict with these. I'll start looking into making these more specific, but if others think we can wait and maybe make another ticket for this. I'll hold off. Just let me know!

src/hooks/Previous.tsx
johanna-adhoc: I think is fine to leave it. It seems the need to be able to see previous props could arise again but I would all add test coverage to it.

johanna-adhoc (commented):
The new implementation is so clean! 🥇

src/hooks/Previous.tsx
anthony-kopczyk: I added a test. The only decent way I found was to create a component that uses the hook and test that component. I got that from the react docs.

I also ended up kind of structuring the hooks folder like we have done with components. Since the hooks folder is starting off still I thought this would be a good time to start using that implementation. (use index.ts and import from it)

> If you need to test a custom Hook, you can do so by creating a component in your test, and using your Hook from it. Then you can test the component you wrote.

https://reactjs.org/docs/hooks-faq.html

src/apiDefs/schema.ts
```diff
@@ -16,6 +16,14 @@ export interface APICategoryContent {
   readonly quickstart?: React.FunctionComponent;
 }
 
+export const ApiCategoryContentPropType = PropTypes.shape({
```
anthony-kopczyk, line 19: @mjlumetta I saw in the PR for eslint that the camel case acronyms were switched to all capital letters. Is that a standard we're using now? I looked at the eslint PR but didn't notice until merging main into this branch.

johanna-adhoc (approved):

PR #467, opened 2020-10-06
[STAGED] CodeBuild CI Transition

duganth-va:
### Description
This PR is staged to support https://vajira.max.gov/browse/API-2617 and switch from Jenkins CI pipeline as the `required` check and transition to the CodeBuild Pipeline. It will also serve as a place for requests for changes or issues that developers notice during the initial parallel period.

### Process

- [ ] Docs updated


duganth-va:
There was an issue  found during @mjlumetta's testing where a review deployment was completed on a bad commit due to the logic on [line 59](https://github.com/department-of-veterans-affairs/developer-portal/blob/1135233930af3eef764170b2284bfd3f8946cde2/cicd/buildspec-ci.yml#L59).

Seems there are two options, 
1. deploy if CI status is good from commit hash 
2. rerun tests on PR creation.

Thoughts?

PR #468, opened 2020-10-07
Upgrade react-router and react-router-dom to 5.2.0

mjlumetta:
### Description

supports [API-1899](https://vajira.max.gov/browse/API-1899).

does some package upgrades: 
- `react-router` from 4.3.1 to 5.2.0
- `react-router-dom` from 4.3.1 to 5.2.0
- `@types/react-router` from 4.4.3 to 5.1.8
- `@types/react-router-dom` from 4.3.1 to 5.1.6
- `react-router-hash-link` from 1.2.1 to 2.1.0

as a result of the React Router upgrade, path matching in nav links changed, so i updated the `SideNavEntry` component accordingly. while i was touching that component, i refactored its tests and the `navHashLinkIsActive` function (a) to be more readable and (b) to allow an "exact" match when there's a hash that represents shared site anchor (so `#main` and `#page-header`.
- to achieve the second outcome, i added a `sharedAnchors` prop. ideally, we'll never need to explicitly set this prop and can always rely on the default prop value, but who knows. seeking feedback on this approach.

also adds an explicit dev dependency on JSDOM. sitemap generation has been broken since Jest got upgraded to v26 (`7715d0d0050b8889ce5658ddadb5e69f63355f72`) because the only JSDOM in `node_modules` was a dependency of Jest and it went from v15 => v16. there was a breaking change in v16, so i added the explicit dependency and used the new JSDOM VM API.

### Process

  - refactored and updated tests in `SideNav.test.tsx`
- [ ] Docs updated

comprehensively tested the active nav link highlighting manually. reviewers should probably try this out themselves.

### Requested feedback

- is the refactoring for `SideNavEntry`, both the source code and tests, useful? what could be improved?


src/components/MainNavItem.tsx
```diff
@@ -1,10 +1,11 @@
 import classNames from 'classnames';
 import * as React from 'react';
+import { match as Match } from 'react-router';
 import { NavLink } from 'react-router-dom';
 import { desktopOnly, mobileOnly } from '../styles/vadsUtils';
 
 export interface ILargeScreenNavItemProps {
-  isActive: (match: {}) => boolean;
+  isActive: (match: Match | null) => boolean;
```
mjlumetta, line 8: this is the functional change here

src/components/NavBar.tsx
```diff
@@ -230,16 +293,16 @@ export default class NavBar extends React.Component<INavBarProps, INavBarState>
...
-  }
+  };
 
   private toggleDefaultNavLink = (useDefault: boolean) => {
     this.setState({ useDefaultNavLink: useDefault });
-  }
+  };
 
-  private checkActiveNavLink = (match: {}) => {
+  private checkActiveNavLink = (match: Match | null) => {
```
mjlumetta, line 302: this is the functional change here

SitemapBuilderWebpackPlugin.js
```diff
@@ -1,36 +1,38 @@
...
 const vm = require('vm');
 const fs = require('fs');
 
 class SitemapBuilderPlugin {
   constructor(options) {
     if (typeof options === 'undefined' || !options.routesFile) {
-      console.error('To use SitemapBuilderPlugin you must provide an options object with routesFile as a property')
+      console.error(
+        'To use SitemapBuilderPlugin you must provide an options object with routesFile as a property',
+      );
```
anthony-kopczyk, line 35: Was this a result of prettier column size? It looks odd to me, but if it's column size related then I'm cool with it.

SitemapBuilderWebpackPlugin.js
```diff
@@ -1,36 +1,38 @@
...
+ *     3. Save the compiled bundle into memory and execute the script
+ *        using a Node.js virtual machine with implementations of web standards (notably DOM and Window)
+ *     4. use the resulting exported function `getSitemapData` to build a sitemap with react-router-sitemap
+ *     5. add `sitemap.xml` to the original webpack compilation's assets so that it is output like any other file in the build bundle
+ *   You can pass this plugin the option verbose: true to receive messages as it moves through this flow.
+ */
 
 const webpack = require('webpack');
-const MemoryFileSystem = require('memory-fs')
+const MemoryFileSystem = require('memory-fs');
```
anthony-kopczyk, line 25: It's cool we're using this. I used this package once on a project I never fully got off the ground to write some files to end users without needing to put them on disk. Nice to see the package gets used.

SitemapBuilderWebpackPlugin.js
```diff
@@ -58,8 +60,8 @@ class SitemapBuilderPlugin {
         config.optimization = { minimize: false };
 
         this.compileRoutes(config)
-          .then((mfs) => this.executeRoutes(mfs))
-          .then((sitemapConfig) => this.buildSitemap(sitemapConfig, compilation))
+          .then(mfs => this.executeRoutes(mfs))
+          .then(sitemapConfig => this.buildSitemap(sitemapConfig, compilation))
           .then(() => this.finish(callback))
           .catch(err => this.finishWithError(err, compilation, callback));
```
anthony-kopczyk, line 66: I haven't looked at all. Guessing we can't change this to async/await? Not sure what version this js file is running.

SitemapBuilderWebpackPlugin.js
```diff
@@ -111,13 +114,11 @@ class SitemapBuilderPlugin {
...
-        .filterPaths(sitemapConfig.pathFilter)
-        .applyParams(sitemapConfig.paramsConfig)
-        .build(prodURL)
-        .save(path.join(paths.appBuild, 'sitemap.xml'))
-    ).sitemaps[0].cache;
+    const sitemap = new Sitemap(sitemapConfig.topLevelRoutes())
+      .filterPaths(sitemapConfig.pathFilter)
+      .applyParams(sitemapConfig.paramsConfig)
+      .build(prodURL)
+      .save(path.join(paths.appBuild, 'sitemap.xml')).sitemaps[0].cache;
```
anthony-kopczyk, line 121: I like the old way this was written with the extra set of parentheses. It makes it more obvious that this was being set to the cache and kind of separates it from the filter/applyParams/build/save methods.

src/components/SideNav.tsx
```diff
@@ -27,48 +29,49 @@ export class SideNavEntry extends React.Component<ISideNavEntryProps> {
...
+      const hashMatch: boolean = !location.hash || this.props.sharedAnchors.includes(location.hash);
+      return !!pathMatch && hashMatch;
+    } else {
+      // default partial matching. since the nav link doesn't have a hash, partial matching 
+      // works whether or not the location has a hash.
       return !!pathMatch;
     }
-    return false;
-  }; // tslint:disable-line: semicolon
+  };
```
anthony-kopczyk, line 70: Good comments!

anthony-kopczyk (commented):
Everything I looked at so far seems good. I didn't get a look at the test file though.

SitemapBuilderWebpackPlugin.js
```diff
@@ -97,9 +99,10 @@ class SitemapBuilderPlugin {
       const jsdom = require('jsdom');
       const { JSDOM } = jsdom;
       const dom = new JSDOM(``, { runScripts: 'outside-only' });
-      const sitemapConfig = dom.runVMScript(script).sitemapConfig();
+      const vmContext = dom.getInternalVMContext();
```
mjlumetta, line 102: new API for this in JSDOM 16: https://github.com/jsdom/jsdom#interfacing-with-the-nodejs-vm-module-using-getinternalvmcontext

SitemapBuilderWebpackPlugin.js
mjlumetta: yes i'm pretty sure that was it. we do need to do some standardization on our prettier setup so we don't constantly get these noisy changes (that said, this one would've happened anyway because this file rarely gets updated)

SitemapBuilderWebpackPlugin.js
mjlumetta: all credit goes to @nfasulo!!

SitemapBuilderWebpackPlugin.js
mjlumetta: how about this rewrite:

```
const todoNameThis = new Sitemap(sitemapConfig.topLevelRoutes())
      .filterPaths(sitemapConfig.pathFilter)
      .applyParams(sitemapConfig.paramsConfig)
      .build(prodURL)
      .save(path.join(paths.appBuild, 'sitemap.xml'));
const sitemap = todoNameThis.sitemaps[0].cache;
```

SitemapBuilderWebpackPlugin.js
mjlumetta: that's a good question, i'm not sure. in general, i'd prefer not to refactor this too much because i'm not super familiar with how the webpack plugin works. i can if you'd like me to look into it, though.

src/components/NavBar.tsx
```diff
@@ -142,16 +162,27 @@ export default class NavBar extends React.Component<INavBarProps, INavBarState>
...
-                src={closeButton} alt="Close button"
-                className={classNames('vads-u-color--gray-dark', 'vads-u-max-width--none')} />
+                src={closeButton}
+                alt="Close button"
+                className={classNames('vads-u-color--gray-dark', 'vads-u-max-width--none')}
+              />
             </button>
-            <Search inMenu={true} className={classNames('vads-u-margin-y--2', 'vads-u-padding-y--0', 'vads-u-width--full')} />
+            <Search
+              inMenu={true}
```
gileswells, line 171: is this an tslint requirement till the eslint stuff gets merged and chalks this up as a fault?

gileswells (changes requested):

src/components/NavBar.tsx
mjlumetta: yeah, exactly. i don't really want to mess with the ESLint stuff until the PR for each file/directory/whatever is merged, and then `npm run eslint` should enforce that for everything that's been covered so far. this is on #452, so it should be fixed soon, i'll just have to deal with the many, many merge conflicts

package-lock.json
```diff
@@ -12853,7 +12841,7 @@
       "dependencies": {
         "combined-stream": {
           "version": "1.0.6",
-          "resolved": "https://registry.npmjs.org/combined-stream/-/combined-stream-1.0.6.tgz",
+          "resolved": "http://registry.npmjs.org/combined-stream/-/combined-stream-1.0.6.tgz",
```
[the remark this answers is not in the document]
mjlumetta: i took a look and it and i cannot figure out why. i found [this thread](https://registry.npmjs.org/) from 2ish years ago about basically this issue that says it was a bug in `npm` related to misconfig on the part of the package auathors, but it also says it was fixed? also, apparently NPM uses HTTPS regardless of what `package-lock.json` says? confusing. anyway, in my environment, the result of `npm config get registry` is `https://registry.npmjs.org/`, and my `npm` is version 6.14.8. i deleted my `node_modules`, checked out `package-lock.json` from master, cleared the cache, and reinstalled, and it still had some `http`s in there.

i can keep looking into this if you want. if it's okay with you, though, i'd say that this kind of churn is acceptable when we're actually updating `package.json`, although not otherwise. one thing i've learned recently that i've found useful is `npm install --no-save` for when i need to download already installed packages, which installs packages without updating `package.json` or the lockfile. that prevents those noisy lockfile-only commits. thoughts on this? will keep investigating if you'd like to try to resolve it. let me know and we can go from there.

gileswells (approved):
Looks good to go. Ideally we'd get a higher version that 5.2.0 but there hasn't been even a comment on my issue asking for a new release. https://github.com/ReactTraining/react-router/issues/7647

PR #469, opened 2020-10-07
API-2544: Use React hooks for GroupedAccordions component

anthony-kopczyk:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-2544

Migrates the GroupedAccordions component to use hooks

### Process
- [N/A] Docs updated

### Requested feedback

The CollapsiblePanel component comes from the VA formation-react repo. It doesn't use hooks so we have to use `setState` on those components, which also means that we need to use refs to access those objects and change their state. I am specifically curious about feedback on that implementation of refs and if there's a better way to do it. Thought about maybe encapsulating things within subcomponents, but thought that might be just as complicated.


src/components/groupedAccordions/GroupedAccordions.tsx
```diff
@@ -0,0 +1,90 @@
...
+          <CollapsiblePanel ref={panelRef} panelName={c.title} startOpen={allExpanded} key={index}>
+            {c.body}
+          </CollapsiblePanel>
+        );
+      })}
+    </section>
+  );
+};
+
+export { GroupedAccordions };
```
alexandervalencia, line 90: Is there a reason we're no longer exporting as default here?

Also, this file is missing a newline at the end.

src/components/groupedAccordions/GroupedAccordions.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export interface PanelContent {
+  readonly body: string | JSX.Element;
+  readonly title: string;
+}
+
+interface IGroupedAccordionsProps {
+  readonly panelContents: PanelContent[];
+  readonly title: string;
+}
```
alexandervalencia, line 17: According to this Slack thread (https://lighthouseva.slack.com/archives/CAZF880HW/p1602185426189800) it looks like we can drop the `I` prefix on interfaces.

alexandervalencia (approved):
Hey @anthony-kopczyk I just had a few questions and small observation, up to you if you want to make the changes. Otherwise looks good.

src/components/groupedAccordions/GroupedAccordions.tsx
anthony-kopczyk: If we export this way then we can import using `import { GroupedAccordion } from './file-name';`

It's a little cleaner and makes us be more explicit in our imports. If we use default exports we don't really know for sure what we're importing unless we go look at the file.

Exporting as default we could do `import flyingBuffalos from './GroupedAccordions';` and then if the export changes it won't be obvious to developers.

In addition, with that `components/index.ts` file, it lets us do something like
```
import {
  GroupedAccordions,
  SomeOtherComponent,
  AnotherComponent,
} from '../components';
```
That's nice for readability. If we did default I think we could pull it off but `components/index.ts` would be more complicated I think.

src/components/groupedAccordions/GroupedAccordions.tsx
anthony-kopczyk: Fixed in the latest commits

alexandervalencia (approved):
LGTM 🚀

PR #473, opened 2020-10-08
API-2497: Hooks Migration Navbar & HoverImage Components

alexandervalencia:
### Story [API-2497](https://vajira.max.gov/browse/API-2497)

Migration of NavBar and HoverImage components from classes to hooks/arrow functions.

### Process

- [N/A] Docs updated

### Requested feedback

All feedback welcome.


src/components/NavBar/NavBar.tsx
```diff
@@ -109,199 +100,186 @@ const navLinkStyles = classNames(
...
-    this.state = {
-      useDefaultNavLink: true,
-      visibleSubNavs: {
-        documentation: false,
-      },
-    };
-  }
+const NavBar = (props: INavBarProps): JSX.Element => {
+  const [useDefaultNavLink, setUseDefaultNavLink] = React.useState(true);
+  const [visibleSubNavs, setVisibleSubNavs] = React.useState({ documentation: false });
```
anthony-kopczyk, line 105: I know this was preexisting functionality and was maybe intended for scalability, but it's weird to me we have `visibleSubNavs` but it only contains data about `documentation`. Makes me think we should only have `documentationVisibility` and `setDocumentationVisibility`. But, I really don't think it's worth changing. Just a comment more than anything.

src/components/NavBar/NavBar.test.tsx
```diff
@@ -1,21 +1,40 @@
-import { shallow } from 'enzyme';
+import '@testing-library/jest-dom/extend-expect';
```
anthony-kopczyk, line 1: I haven't gotten a chance to get familiar with the react testing library things, so if somebody can please review this for me!

I have no comments other than it looks good.

src/components/hoverImage/HoverImage.tsx
```diff
@@ -23,8 +23,8 @@ export default class HoverImage extends React.Component<IHoverImageProps, IHover
     const { src } = this.state;
     return (
       <img
-        alt=''
-        role='presentation'
+        alt=""
```
anthony-kopczyk: I know  this is from old code too, but I wonder if we should even be specifying an `alt` value if it's going to be blank. Maybe there's some guidance in 508 compliance on this.

src/containers/News.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 
 import CardLink from '../components/CardLink';
 import EmbeddedYoutubeVideo from '../components/EmbeddedYoutubeVideo';
-import HoverImage from '../components/HoverImage';
+import HoverImage from '../components/hoverImage/HoverImage';
```
anthony-kopczyk, line 6: Can you add the hover image to the `components/index.ts` file so we can use `import { HoverImage } from '../components';`? There should be a few other components in there to use as an example.

anthony-kopczyk: Same goes for the other components that have been updated if they aren't there as well.

anthony-kopczyk (changes requested):
Looks good overall. We can add the components to `components/index.ts`. Other than that though, I don't have any comments besides "It's probably a good idea to get somebody else to look over the test files"

src/components/hoverImage/HoverImage.tsx
alexandervalencia: A11y rules say an alt should always be present, even if it's blank. Ref: https://www.w3.org/WAI/tutorials/images/decorative/

src/containers/News.tsx
alexandervalencia: Will do! 👍

src/components/header/Header.tsx
```diff
@@ -7,7 +7,7 @@ import { defaultFlexContainer, desktopOnly, mobileOnly } from '../../styles/vads
 import { onHashAnchorClick } from '../../utils/clickHandlers';
 import Banner from '../Banner';
 import VeteransCrisisLine from '../crisisLine/VeteransCrisisLine';
-import NavBar from '../NavBar';
+import NavBar from '../navBar/NavBar';
```
gileswells, line 10: This needs to be updated as it doesn't match the case of the path.
```suggestion
import NavBar from '../NavBar/NavBar';
```

src/containers/News.tsx
gileswells: This also needs to have the properly cased `HoverImage` on the import

src/components/hoverImage/HoverImage.tsx
```diff
@@ -0,0 +1,22 @@
...
+  imagePath: string;
+  hoverImagePath: string;
+}
+
+const HoverImage = (props: IHoverImageProps): JSX.Element => {
+  const [src, setSrc] = React.useState(props.imagePath);
+
+  return (
+    <img
+      alt=""
```
gileswells, line 13: We should be passing the alt in from the props so that it can be set. It can be optional if we want but it should be able to be set. Leaving it as always blank is not good accessibility.

gileswells (changes requested):
Overall I think the conversion to hooks looks to be on the right path but I'd like to have a larger discussion about `HoverImage` as a whole in the near future. There's a simple requested change for now to improve it for accessibility reasons included in my review.

src/components/header/Header.tsx
alexandervalencia: Good catch, somehow the folder name didn't properly update to `navBar`

src/components/hoverImage/HoverImage.tsx
alexandervalencia: While I agree with you, that wouldn't be considered part of this migration process. We're not trying to update/change any functionality, just conversion from class to function based component.

src/components/navBar/NavBar.test.tsx
```diff
@@ -0,0 +1,40 @@
+import '@testing-library/jest-dom/extend-expect';
+import { cleanup, render } from '@testing-library/react';
+import 'jest';
+import { noop } from 'lodash';
```
anthony-kopczyk, line 4: Not sure what others think on this. With ES6 we can replace a lot of lodash functionality. Is this preexisting? If it is we don't need to worry about it, but we could replace noop with () => undefined to go pure ES6.

anthony-kopczyk (approved):
Looks like the requested changes were addressed.

johanna-adhoc (approved):
LGTM 🚀

gileswells (approved):
LGTM

johanna-adhoc (approved):
LGTM

gileswells (approved):
LGTM

PR #474, opened 2020-10-09
Define specific Swagger UI types and fix ESLint errors in CurlForm

mjlumetta:
### Description

fixes ESLint errors in `CurlForm` in support of [API-1323](https://vajira.max.gov/browse/API-1323).

- adds types to `swagger-ui.d.ts`
- refactoring `CurlForm` a bit as part of removing ESLint errors
- adding `CurlForm` tests
  - these use real OpenAPI Specs i've downloaded to `src/__mocks__/openAPIData`. i didn't want to go to the length of defining fake data to cover all the cases. seeking feedback on this decision.
- sets `esModuleInterop: true` so that CommonJS modules (defined with `exports = myModuleExport`) can be imported as a default import (`import myModuleExport from 'my-module'` instead of `import * as myModuleExport from 'my-module'`).
  - https://stackoverflow.com/questions/56238356/understanding-esmoduleinterop-in-tsconfig-file 
  - https://stackoverflow.com/questions/49256040/a-namespace-style-import-cannot-be-called-or-constructed-and-will-cause-a-failu
  - https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html

### Process

  - added unit tests for `CurlForm`
- [ ] Docs updated
  - labeled environment select, no other changes

### Requested feedback

- is refactoring the `CurlForm` tests so they're faster a prerequisite for merging this PR?
- is it alright to use real OAS data downloaded to JSON files in `src/__mocks__`?


src/__mocks__/fakeCategories.tsx
```diff
@@ -1,4 +1,4 @@
-import * as moment from 'moment';
+import moment from 'moment';
 import * as React from 'react';
```
anthony-kopczyk, line 2: Do these changes mean we could use `import React from 'react'`?

src/containers/documentation/swaggerPlugins/CurlForm.test.tsx
```diff
@@ -0,0 +1,481 @@
...
+  render,
+  screen,
+} from '@testing-library/react';
+import SwaggerUI from 'swagger-ui';
+import * as React from 'react';
+import * as decisionReviews from '../../../__mocks__/openAPIData/decisionReviews.test.json';
+import * as fhirR4 from '../../../__mocks__/openAPIData/fhirR4.test.json';
+import { SwaggerPlugins, System } from './index';
+
+// some tests in this file are long-running because of the expense of rendering Swagger UI,
```
anthony-kopczyk, line 24: Hmm. is this why the `ApiDocumentation` test sometimes fails on my laptop if I have a bunch of things running?

src/containers/documentation/swaggerPlugins/CurlForm.test.tsx
```diff
@@ -0,0 +1,481 @@
...
+// so we double the timeout to 10s from the default 5s.
+jest.setTimeout(10000);
+
+const expandOperation = (operationTag: string, description: string): HTMLElement => {
+  const operationTagHeader = screen.getByRole('heading', { name: operationTag });
+  expect(operationTagHeader).toBeInTheDocument();
+  expect(operationTagHeader.nextElementSibling).not.toBeNull();
```
anthony-kopczyk, line 31: Can we use `.toBeTruthy()` here?

I read this a week or two ago and they touch on some jest `expect` conditions that I don't fully understand. But if it came back undefined (can it?) this wouldn't evaluate properly.
https://github.com/adobe/brackets/issues/5479

src/containers/documentation/swaggerPlugins/CurlForm.test.tsx
```diff
@@ -0,0 +1,481 @@
...
+
+const normalizeCurlText = (original: string): string => original.trim().replace(/\s+/g, ' ');
+const testCurlText = async (
+  expectedCurl: string,
+  operationContainer: HTMLElement,
+): Promise<void> => {
+  // assumption: there is exactly one expanded operation and exactly one curl form visible
+  const curlHeading = await findByRole(operationContainer, 'heading', { name: 'Generated Curl' });
+  expect(curlHeading).toBeInTheDocument();
+  expect(curlHeading.nextElementSibling).not.toBeNull();
```
anthony-kopczyk, line 57: Same thing with `.not.toBeNull()`

src/containers/documentation/swaggerPlugins/CurlForm.test.tsx
```diff
@@ -0,0 +1,481 @@
...
+      renderDecisionReviews();
+      operationContainer = expandOperation('Higher-Level Reviews', createHLRDesc);
+    });
+
+    it('renders the request body inputs', async () => {
+      const bodyHeading = await findByRole(operationContainer, 'heading', {
+        name: 'Request Body:',
+      });
+      expect(bodyHeading).toBeInTheDocument();
+      expect(bodyHeading.nextElementSibling).not.toBeNull();
```
anthony-kopczyk, line 420: `.not.toBeNull`

src/containers/documentation/swaggerPlugins/CurlForm.test.tsx
```diff
@@ -0,0 +1,481 @@
...
+  "included": [
+    "contestableIssue1",
+    "contestableIssue2"
+  ]
+}'`;
+
+      await testCurlText(expectedCurl, operationContainer);
+    });
+  });
+});
```
anthony-kopczyk, line 481: I think this is good? It's a long file. In the future, might be beneficial, even if we leave CurlForm.tsx intact as is, to split this test file out into multiple chunks.

src/containers/documentation/swaggerPlugins/CurlForm.tsx
```diff
@@ -1,118 +1,140 @@
+import { faCopy } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
anthony-kopczyk, line 2: Never heard of fortAwesome. Had to do a few takes to see that fontawesome is from fortawesome. 👍

src/containers/documentation/swaggerPlugins/CurlForm.tsx
```diff
@@ -1,118 +1,140 @@
...
+import { 
+  OpenAPISpec,
+  OpenAPISpecV2,
+  OpenAPISpecV3,
+  Operation,
+  Parameter,
+  Schema,
+  Server,
+  SwaggerSpecObject, 
+} from 'swagger-ui';
```
anthony-kopczyk, line 15: I like the goal of adding these types.

src/types/swagger-ui.d.ts
```diff
@@ -1 +1,91 @@
-declare module 'swagger-ui';
+/**
```
anthony-kopczyk, line 1: Already said elsewhere, but I like this.

src/types/index.ts
```diff
@@ -9,9 +9,23 @@ export interface IApiNameParam {
...
+  healthcheck: string;
+}
+
+export interface APIMetadata {
+  meta: {
+    versions: VersionMetadata[];
+  };
+}
+
 export interface IApiVersioning {
```
anthony-kopczyk, line 26: Change this to `ApiVersioning`?

anthony-kopczyk: Or maybe not. This might be large enough scope to do in a different PR.

anthony-kopczyk (commented):
Looks good to me, but probably good to get some more sets of eyes on this.

src/components/GroupedAccordions.tsx
```diff
@@ -1,5 +1,5 @@
 import CollapsiblePanel from '@department-of-veterans-affairs/formation-react/CollapsiblePanel';
-import * as classNames from 'classnames';
+import classNames from 'classnames';
```
anthony-kopczyk, line 2: You might have to manually remake this change due to my moving this file to a folder the wrong way.

https://lighthouseva.slack.com/archives/CAZF880HW/p1602262623210500

src/containers/documentation/swaggerPlugins/CurlForm.test.tsx
mjlumetta: that's a good question... probably, yeah?

src/containers/documentation/swaggerPlugins/CurlForm.tsx
mjlumetta: yeah, it's good to have them 😌 and we kind of need them for these ESLint checks. now if only Swagger UI would actually define their own types...

src/types/index.ts
mjlumetta: yes, if we start fixing every file that gets touched by fixing another file, it'll snowball pretty quick. i had a branch where i tried to go further on the Swagger stuff and it ended up being a colossal mess

src/containers/documentation/swaggerPlugins/CurlForm.test.tsx
mjlumetta: yeah... this is a good point, i think, but i also think this is an acceptable case for using `.not.toBeNull()`. the reason i think that it's acceptable here is that the signature of `el.nextElementSibling` (and `previousElementSibling`, etc) is `HTMLElement | null`. therefore, if it's somehow undefined, i'd be okay with it blowing up on the next line because that isn't a failure we could predict. lmk what you think.

src/__mocks__/fakeCategories.tsx
mjlumetta: umm yeah i think we can, but i'd prefer not to change it here, if that's alright. it's only _necessary_ to change the import syntax if the export is callable or otherwise not an object. we can change that going forward, but i would prefer to keep the footprint of the PR from expanding unless it needs to.

src/containers/documentation/swaggerPlugins/CurlForm.tsx
```diff
@@ -142,6 +165,16 @@ export class CurlForm extends React.Component<ICurlFormProps, ICurlFormState> {
         : this.state.bearerToken;
       options.securities = {
         authorized: {
+          // support multiple means of passing the bearer token. this is mostly due to swagger-client
+          // not being particularly sophisticated on this front.
+          // Bearer auth security (Claims): https://swagger.io/docs/specification/authentication/bearer-authentication/
+          // OAuth 2.0 security (Health): https://swagger.io/docs/specification/authentication/oauth2/
+          // https://github.com/swagger-api/swagger-js/blob/master/src/execute/oas3/build-request.js#L78
+          OauthFlow: token ? {
```
mjlumetta, line 173: try FHIR on this branch vs FHIR in production

src/containers/documentation/swaggerPlugins/CurlForm.test.tsx
anthony-kopczyk: With the current code, we'd miss the case where it's somehow undefined. `toBeTruthy` will make sure it's both not null and it's not undefined.

If we trust the library not to return undefined I'm on board leaving this as is.

anthony-kopczyk: Oh I see. Yeah. The next line should prevent `undefined` from being an issue. Yeah. I'm on board leaving this as is.

mjlumetta: okay, cool! in terms of the trust that this returns null, i would say i have a high degree of trust. if it were a library we had installed from NPM, no matter how broadly used, i would say that's a good reason for distrust there and for using `toBeTruthy()` - but since we're dealing with the core DOM APIs, i do feel pretty comfortable asserting null.

anthony-kopczyk (dismissed):
This looks good to me.

src/containers/documentation/swaggerPlugins/CurlForm.tsx
```diff
@@ -210,11 +250,11 @@ export class CurlForm extends React.Component<ICurlFormProps, ICurlFormState> {
               aria-label="Enter API Key"
               value={this.state.apiKey}
               onChange={e => {
-                this.handleInputChange('apiKey', e.target.value);
+                this.setState({ apiKey: e.target.value });
               }}
             />
             <small>
-              Don't have an API Key? <a href="/apply"> Get One </a>
+              Don&apos;t have an API Key? <a href="/apply"> Get One </a>
```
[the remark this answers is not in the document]
anthony-kopczyk: I don't have answers to the two questions in the original requested review, so might be worth getting input from someone those before merging.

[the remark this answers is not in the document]
anthony-kopczyk: I guess, to spew some thoughts

> is refactoring the CurlForm tests so they're faster a prerequisite for merging this PR?

I don't think we need to, as long as it can pass in jenkins. Maybe we can put in a jira ticket for it if we know this is something we'll want to do

> is it alright to use real OAS data downloaded to JSON files in src/__mocks__?

I want to say 'no', but my mind is telling me to follow that up with 'idea'.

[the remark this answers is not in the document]
mjlumetta: various responses:

- yeah, the `&apos;` was an ESLint rule check that i am fine with enforcing, but also fine with turning off if people would prefer. 
- i don't think there is any documentation to update, no. but thanks for bringing that up!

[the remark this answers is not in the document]
anthony-kopczyk: I was trying to be clever and say 'no idea' without writing it out that way. :)

But my gut does say no, just because I don't know what all information is really in there and on precedence of the question being thought, that makes me think you may have been thinking no as well.

Like you suggested, the best way is to probably use what's there (because we want it to be similar to the real structure) but renaming things to be obviously fake. I don't know the details on the information, how sensitive it is, and such.

src/actions/index.ts
```diff
@@ -10,7 +11,7 @@ export interface ISetRequestedApiVersion extends Action {
 
 export interface ISetInitialVersioning extends Action {
```
gileswells, line 12: I named interface

src/actions/index.ts
```diff
@@ -21,7 +22,10 @@ export const setRequstedApiVersion: ActionCreator<ISetRequestedApiVersion> = (ve
   };
 };
 
-export const setInitialVersioning: ActionCreator<ISetInitialVersioning> = (docUrl: string, metadata: any) => {
+export const setInitialVersioning: ActionCreator<ISetInitialVersioning> = (
+  docUrl: string,
+  metadata: any,
```
gileswells, line 27: Shouldn't this now be `APIMetadata`?

src/containers/documentation/ApiDocumentation.test.tsx
```diff
@@ -6,7 +6,7 @@ import { MockedResponse, ResponseComposition } from 'msw/lib/types/response';
 import { setupServer } from 'msw/node';
 import * as React from 'react';
 import { Provider } from 'react-redux';
-import * as openAPIData from '../../__mocks__/openAPIData.test.json';
+import * as openAPIData from '../../__mocks__/openAPIData/openAPIData.test.json';
```
gileswells, line 9: Is this able to rename to just `import openAPIData` as well or is that just for JS files?

src/containers/documentation/swaggerPlugins/CurlForm.tsx
```diff
@@ -210,11 +250,11 @@ export class CurlForm extends React.Component<ICurlFormProps, ICurlFormState> {
               aria-label="Enter API Key"
               value={this.state.apiKey}
               onChange={e => {
-                this.handleInputChange('apiKey', e.target.value);
+                this.setState({ apiKey: e.target.value });
```
gileswells, line 253: Why did you opt to change up the `handleInputChange()` method? Just trying to get your thinking behind this.

gileswells (changes requested):
Some questions on thought process and a sneaky interface name to be renamed.

src/actions/index.ts
mjlumetta: was planning not to fix in this PR to avoid snowballing. at some point, i'll open a PR that adds `src/actions` to `npm run eslint`, at which point it will have to be fixed. is that alright with you?

src/actions/index.ts
mjlumetta: yes, it should. i may have left it as not that to avoid snowballing changes - if i can make it the right type without snowballing, i'll make the change. otherwise, though, would like to leave it as is. i think that regardless of this PR i will be doing some pretty substantial refactoring around the version metadata actions and reducers in subsequent PRs.

src/containers/documentation/ApiDocumentation.test.tsx
mjlumetta: i think we can do that, but it doesn't matter as much for objects. want me to make the change? i was planning not to make changes for object imports in this PR.

src/actions/index.ts
gileswells: Seems logical to me. Much prefer a lack of snowballs.

src/containers/documentation/ApiDocumentation.test.tsx
gileswells: I can take it or leave it.

src/containers/documentation/swaggerPlugins/CurlForm.tsx
mjlumetta: the rationale i was working from goes like this:

1. we want to remove the arbitrary inputs (parameters, request body) from the top-level state so that the state can be properly typed. in other words, we don't want to assign `this.state.paramName = e.target.value`, which is what `handleInputChange()` previously did. 
2. to solve (1), i created `this.state.paramValues`. i updated `handleInputChange()` to assign input params to `state.paramValues`.
3. however, we still want to have known input values - API key, bearer token, and environment - at the top-level state instead of with arbitrary keys. therefore, `handleInputChange()` doesn't work for these inputs, so i replaced the event handler here with a direct call to `setState()`. my thought was that it's brief enough so as not to add length to the code, and it actually removes a bit of indirection while still satisfying all the constraints.

does this help?

gileswells: That makes sense. There's a small part of me that wonders if it would have been better to add a condition into `handleInputChange()` that looks for those 3 known fields and applies them to the state differently but that adds complexity that's not needed. LGTM

gileswells (approved):

PR #475, opened 2020-10-09
Fix ESLint errors in Release Notes

mjlumetta:
### Description

fixes ESLint errors in 

also turns on ESLint rule [`eol-last`](https://eslint.org/docs/rules/eol-last) per @anthony-kopczyk's request. https://lighthouseva.slack.com/archives/CAZF880HW/p1602258250200600

### Process

  - n/a, using tests for regression
- [n/a] Docs updated


src/containers/releaseNotes/CategoryReleaseNotes.test.tsx
```diff
@@ -26,26 +26,26 @@ describe('ReleaseNotesCollection', () => {
   });
 
   describe('CategoryReleaseNotes', () => {
-    function renderComponent(route: string = '/release-notes/lotr') {
-      cleanup();
+    const renderComponent = async (route = '/release-notes/lotr') => {
+      await cleanup(); // clean up beforeEach render if we're testing a different page
```
mjlumetta, line 30: `cleanup()` actually returns a promise, which i did not know previously

.eslintrc.js
```diff
@@ -137,7 +137,7 @@ module.exports = {
     complexity: 'off',
     'constructor-super': 'error',
     curly: 'error',
-    'eol-last': 'off',
+    'eol-last': 2,
```
anthony-kopczyk, line 140: When I looked this up it looked like it had string options like "always" and "unix". What does this 2 mean?

src/containers/releaseNotes/CategoryReleaseNotes.tsx
```diff
@@ -16,7 +16,7 @@ import { APINameParam } from '../../types';
 interface ReleaseNotesCardLinksProps {
   categoryKey: string;
   apiCategory: BaseAPICategory;
-  flagName: string;
+  flagName: 'enabled' | 'hosted_apis';
```
anthony-kopczyk, line 19: That's so cool you can type things this way instead of making an enum!

src/containers/releaseNotes/CategoryReleaseNotes.tsx
```diff
@@ -88,34 +88,32 @@ interface ReleaseNotesCollectionProps {
...
-      />
-      <div className={classNames('vads-u-width--full', 'vads-u-margin-top--4')}>
-        {props.apiCategory.apis.map((api: APIDescription) => (
-          <APIReleaseNote flagName={props.apiFlagName} key={api.urlFragment} api={api} />
-        ))}
-      </div>
-    </section>
-  );
-};
+const ReleaseNotesCollection = (props: ReleaseNotesCollectionProps) => (
```
anthony-kopczyk, line 91: Two things.

1. Return type

2. I'm not sure if we'd want to do this, but since I saw the cool typing thing above, you can destructure the props inside of the constructor if you wanted.

Not sure if this is exact syntax but:
```
const ReleaseNotesCollection = ({ categoryKey, apiCategory, flagName}: ReleaseNotesCollectionProps): JSX.Element => (
  <div>My Element</div>
);
```

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -35,7 +35,7 @@ function SideNavAPIEntry(props: SideNavAPIEntryProps) {
           {(api.vaInternalOnly && api.trustedPartnerOnly && <br />) || null}
           {api.trustedPartnerOnly && (
             <span>
-              <small>Internal VA use only.{/*Trusted Partner use only.*/}</small>
+              <small>Internal VA use only.{/* Trusted Partner use only. */}</small>
```
anthony-kopczyk, line 38: Is this comment displayed? We have the if condition above. Not sure if that's needed. Doesn't need changed in this PR since preexisting but a thought/question.

anthony-kopczyk (dismissed):
Mostly just saw some missing return types, but one of them was fixed with a latest change to the main branch, so maybe they're already resolved elsewhere.

.eslintrc.js
mjlumetta: oh, the 2 means 'error' with the default value of 'always'. i can set it to the string 'error' instead so it's easier to read, don't know why i did it this way. https://eslint.org/docs/user-guide/configuring#configuring-rules

src/containers/releaseNotes/CategoryReleaseNotes.tsx
mjlumetta: yeah!! i mean, it basically is an enum, but built in

src/containers/releaseNotes/CategoryReleaseNotes.tsx
mjlumetta: 1. good call.
2. yeah, we can do that for sure if you'd prefer. i kind of like breaking it up, though, because keeping the `props` param name and keeping the function signature shorter feels slightly easier to read to me. that said, it's sort of a 6 on one hand, half dozen on the other situation, so if you prefer to destructure the props in the function signature, i'll make that change.

src/containers/releaseNotes/ReleaseNotes.tsx
mjlumetta: yeah, so, the idea here is that there is one API, Loan Guaranty, that is technically "trusted partner only". however, when it went to production, the decision was made to display it the same as "VA internal only" (i think there's only one "trusted partner" that's pretty close to the VA?). it's possible that in the future they could be folded together (Loan Guaranty becomes the same as internal only, and trusted partners have access to internal only APIs on a case-by-case basis), or the UI could change to say "trusted partner only" - which is why this comment is included, to have a placeholder for the possibility that that decision might happen. could go either way. does that help clear things up?

nathanbwright (approved):
Looks good to me!

src/containers/releaseNotes/CategoryReleaseNotes.tsx
anthony-kopczyk: 2. We can leave it as is.

src/containers/releaseNotes/ReleaseNotes.tsx
anthony-kopczyk: Yeah. That makes sense. Thanks!

anthony-kopczyk (approved):
Latest looks good to me!

nathanbwright (approved):

PR #476, opened 2020-10-09
Upgrade flag module to 4.4.0 to resolve componentWillReceiveProps deprecation warning

mjlumetta:
### Description

https://vajira.max.gov/browse/API-2475

upgrades `flag` to 4.4.0 in order to partially resolve a warning about the deprecation of the `componentWillReceiveProps` lifecycle hook.

old warning:

```
Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.

- Move data fetching code or side effects to componentDidUpdate.
- If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
- Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. - In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

Please update the following components: FlagsProvider, Route, Router, Switch
```

#### `flag` API changes

https://www.npmjs.com/package/flag

* rather than exporting `Flag` and `FlagsProvider` directly, `flag` exports a typed `createFlags` function that returns those two components, as well as corresponding hooks.
* the `name` prop is now an array of keys rather than a single `key1.key2.key3` string.
* `flag` has two builtin hooks called `useFlag` and `useFlags`, although i'm not refactoring anything to add them.

### Process

  - used regression tests
- [ ] Docs updated

manually checked by disabling APIs and making sure they don't show up in Documentation, Release Notes, or Home. i recommend testing with one API that has other APIs in its category and one that is the only API in its category, so that you can test category flagging as well. i used Claims and Facilities.


nathanbwright:
Looks good to me just reading through it and running the unit tests (and verifying that there are no more deprecation warnings).

While I was reading through the PR I was looking at the way we reference relative file paths and thought there has to be a better way. After digging I found this interesting article (the main site seems to be down at the moment, so I've included a wayback machine link): https://web.archive.org/web/20160302053107/http://allanhortle.com/2015/02/04/store-your-client-code-in-node_modules.html

It definitely not something that I'm proposing for this PR, it's just something interesting to keep in mind as the project grows. Maybe this is common knowledge but it's a TIL moment for me. 😄

ediiotero (approved):
Ran the tests and tested manually, all looks good, no warnings

mjlumetta:
> While I was reading through the PR I was looking at the way we reference relative file paths and thought there has to be a better way. After digging I found this interesting article (the main site seems to be down at the moment, so I've included a wayback machine link): https://web.archive.org/web/20160302053107/http://allanhortle.com/2015/02/04/store-your-client-code-in-node_modules.html

@nathanbwright i'm not sure that this specific suggested approach of creating artificial `node_modules` directories is the right way to solve that problem, but there are other ways. i'm pretty sure you can set up the module resolution so that you can do something like `import module from 'src/module'` from anywhere in the project. we can take a look at how to do that in the future for sure if the relative imports are getting too clunky.

PR #484, opened 2020-10-14
API-2650: Use route hooks for ApiPage

anthony-kopczyk:
### Description
Updates the ApiPage component to use route hooks

### Process
- [N/A] Docs updated

### Requested feedback

Not requested feedback but a note.

This component uses the location hook only to inject `location` to a child component. That child component is from swagger UI, and we are not converting those to use hooks at the moment. At the moment, that work isn't planned to be done so we''ll need to continue the use of location in this component for now.


TS000 (approved):
Looks great! Love how much these hooks are cleaning things up!

mjlumetta (approved):

johanna-adhoc (approved):
💯

src/containers/documentation/ApiPage.tsx
```diff
@@ -1,7 +1,6 @@
 import classNames from 'classnames';
-import { Flag } from 'flag';
 import { useLocation, useParams } from 'react-router-dom';
-
+import * as React from 'react';
```
TS000, line 3: I really like having the react import at the top and in then in this case following by the router hooks

anthony-kopczyk: Went ahead and moved this this morning.

TS000 (approved):

PR #487, opened 2020-10-14
API-2943: Add ModalController test

anthony-kopczyk:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-2943

Adds test coverage for the modal controller hook

### Process
- [N/A] Docs updated

### Requested feedback
I tried using the `renderHook`  function from the testing library but ran into issues with eslint claiming I was using an `any` type, but I wasn't. If anybody has experience with the renderHook and wants to take a spin at it, feel free. For now I followed the same pattern using on the `Previous` hook.

Also, while not directly related, the hook files are named right now without the `use` prefix. When I started making them I thought this would be fine, but without `use` it doesn't look quite right. Curious thoughts on this. If there's a consensus I can put in a jira ticket for it.


package.json
```diff
@@ -70,10 +70,9 @@
     "@babel/plugin-proposal-class-properties": "^7.4.0",
     "@babel/preset-env": "^7.4.3",
     "@babel/preset-react": "^7.0.0",
-    "@testing-library/dom": "^7.24.5",
```
anthony-kopczyk, line 73: We don't need this in package.json ( I think I added it in another PR by accident ).

TS000 (dismissed):

anthony-kopczyk:
I was unable to get the modal controller test working with `renderHook`. I could not find a way to simulate a 'escape' key press on the element using both `fireEvent` and `userEvent`. If anyone has ideas I can try switching it over again but no luck so far. If the react-hooks testing library gave access to the html element it creates I think it would work, but it hides access.

I went ahead, for my own sanity, and implemented it on the `usePrevious` hook, just to make sure I wasn't completely off base, and was able to implement that without issue.

src/hooks/previous/Previous.tsx
```diff
@@ -1,11 +1,11 @@
 import * as React from 'react';
 
 // Use generics so that we can use this for any type
-const usePrevious = <T extends unknown>(props?: T): T | null => {
+const usePrevious = <T extends unknown>(props: T): T | null => {
```
TS000, line 4: Super slick! Def using this in the future.

TS000 (approved):

PR #488, opened 2020-10-14
API-2908: update category page to use route hooks

TS000:
### Description
Replaced React Router 4 prop based params with React Router 5 useParams 🎸 

- [ n/a ] Docs updated

### Requested feedback
All feedback appreciated.


src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable prefer-arrow/prefer-arrow-functions */
```
anthony-kopczyk, line 1: I haven't looked at the entire file. Is this still needed?

src/containers/documentation/CategoryPage.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import { RouteComponentProps } from 'react-router';
+import { Flag } from 'flag';
```
anthony-kopczyk, line 2: There's a duplicate `import { Flag }` but from a different file on line 9. I think line 9 is correct and this one can be removed but not certain.

anthony-kopczyk (changes requested):
1. I think maybe this was based off of the branch for updating authorizationDocs. Shouldn't be a big deal because that PR is ready to be merged already. But probably doin't want to merge this until that one is merged first
2. There's a duplicate import it looks like in `CategoryPage`
3. There's a line in `AuthorizationDocs` that might be able to be removed? Doesn't need to be in this PR though.

src/containers/documentation/AuthorizationDocs.tsx
TS000: I'm confused as to why this file was part of my PR.

anthony-kopczyk (approved):
LGTM

PR #490, opened 2020-10-15
Hotfix/secondary skip nav timing issue

gileswells:
### Description
Top level documentation pages (eg. /explore/benefits) would not properly take focus when using the skip nav. This resolves that issue.

### Process
- [ ] Docs updated


src/utils/clickHandlers.ts
```diff
@@ -4,6 +4,8 @@ export const onHashAnchorClick = (e: React.MouseEvent<HTMLAnchorElement>) => {
   // This provides for focus to be applied when clicking a /hash#link
   const id: string = e.currentTarget.href.split('#')?.[1];
   if (id) {
-    document.getElementById(id)?.focus();
+    setTimeout(() => {
+      document.getElementById(id)!.focus();
```
anthony-kopczyk, line 8: Is the change from `?` to `!` purposeful? I don't think the eslinter likes `!`

anthony-kopczyk (commented):
Just one question.

src/utils/clickHandlers.ts
gileswells: That was a goof from my first attempts to fix it. That's changed back now.

anthony-kopczyk (approved):
Looks good to me!

PR #311, opened 2020-02-10
Sandbox Migration Alert

bastosmichael:
### Descrpription

Started with ticket https://github.com/department-of-veterans-affairs/vets-contrib/issues/4043

Now points to ticket https://vasdvp.atlassian.net/secure/RapidBoard.jspa?rapidView=140&projectKey=DASH&modal=detail&selectedIssue=DASH-95&quickFilter=393&assignee=5e386d8f6061160ca042d953

Screenshot from 2020-02-10 13-57-36

Screenshot from 2020-02-27 13-28-02
Screenshot from 2020-02-27 13-28-10
Screenshot from 2020-02-27 13-28-17
Screenshot from 2020-02-27 13-28-24
Screenshot from 2020-02-27 13-28-32
Screenshot from 2020-02-27 13-28-38
Screenshot from 2020-02-27 13-28-46

Feature flag applies to individual API pages only, we do not have currently a feature flag for individual page elements on the platform.

### Acceptance Criteria
- Notice for sandbox landing page is created.

PR #492, opened 2020-10-15
API-1686: 404s on wildcard routes

gileswells:
### Description
https://vajira.max.gov/browse/API-1686
Provide a better design for endpoint urls that do not exist over the current solution. The new implementation (seen at /explore/benefits/docs/cookie and other paths) also pulls in the list of available APIs under that category to list. Also the list will only show APIs that are currently active or depreciated but will skip any urls that have already been deactivated.

### Process
- [ ] Docs updated

### Requested feedback
Is there a desire to attempt to search via the url fragment to see if there's an API under a different category that matches that fragment and load that as a suggested link as well? Currently this would not be possible due to (https://vajira.max.gov/browse/API-2991) but it's something that could be done in the future once API-2991 is closed out.


src/styles/base.scss
```diff
@@ -94,11 +94,3 @@ a:focus,
 input:focus::placeholder {
   color: $color-gray;
 }
-
-.usa-alert::before {
```
gileswells, line 98: This was previously used on the homepage for a sandbox migration alert in #311 but is not in use anymore and messes with the expected layout from the VA Design System.

src/containers/documentation/EndpointNotFoundPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <div role="region" aria-labelledby={PAGE_HEADER_ID}>
+      <AlertBox
+        className="vads-u-margin-top--0 va-api-alert-box"
+        headline="Page not found."
+        content="Try using the links below or the search bar to find your way forward."
+        status="warning"
+      />
+      <PageHeader header={category!.name} />
```
anthony-kopczyk, line 27: eslint will complain about the `!`. Is this meant to be a `?`. Or if it doesn't allow undefined values you can do something like `category?.name ?? null`. If it doesn't allow null values then we'll probably need to return a default.

src/Routes.tsx
```diff
@@ -50,7 +50,6 @@ export function SiteRoutes() {
       <Route path="/beta" component={BetaPage} />
       <Route path="/beta-success" component={BetaSuccess} />
       <Route path="/explore/:apiCategoryKey?" component={DocumentationRoot} />
-      <Route exact={true} path="/explore/:apiCategoryKey/docs/:apiName" />
```
anthony-kopczyk, line 53: I think I might be missing something. Why is this route gone? Does it get handled under the DocumentationRoot component?

anthony-kopczyk (changes requested):
I think really there's just the update for using the react router hook `useParams` and updating the props that are currently receiving the params as props to no longer receive them.

alexandervalencia (approved):
LGTM, just had a small question 🚀

src/containers/documentation/EndpointNotFoundPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+          ))
+        }
+      </ul>
+    </div>
+  );
+};
+
+EndpointNotFoundPage.propTypes = {};
+
+export default EndpointNotFoundPage;
```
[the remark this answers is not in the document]
TS000: Should we make a new ticket to convert this to hooks and merge with the current fixes?

src/Routes.tsx
gileswells: It is handled under the DocumentRoot component. Since the route above doesn't have exact included it won't ever get to this route since the one above is in essence a "starts with" path.

src/containers/documentation/EndpointNotFoundPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { lookupApiCategory } from '../../apiDefs/query';
+import PageHeader from '../../components/PageHeader';
+import { APINameParam } from '../../types';
+import { PAGE_HEADER_ID } from '../../types/constants';
+import { Link } from 'react-router-dom';
+import { APIDescription } from 'src/apiDefs/schema';
+import { now } from 'lodash';
+
+const EndpointNotFoundPage = (props: RouteComponentProps<APINameParam>): JSX.Element => {
+  const { params } = props.match;
```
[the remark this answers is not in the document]
gileswells: I hadn't thought of that BUT at the same time including the other url param allows us to match other APIs if we want to explore the option of searching for other APIs I put in the feedback section of the PR comment.

src/containers/documentation/EndpointNotFoundPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+        className="vads-u-margin-top--0 va-api-alert-box"
+        headline="Page not found."
+        content="Try using the links below or the search bar to find your way forward."
+        status="warning"
+      />
+      <PageHeader header={category!.name} />
+      <ul>
+        {category?.apis
+          .filter((item: APIDescription) => {
+            return !item.deactivationInfo || item.deactivationInfo.deactivationDate.isAfter(now());
```
mjlumetta, line 31: what's the motivation for using lodash's `now()` over `Date.now()`?

src/containers/documentation/EndpointNotFoundPage.tsx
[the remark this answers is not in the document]
mjlumetta: since this is a new component, would love to get whatever hooks changes we need in here. as far as the export pattern goes, i like the new pattern in `src/components`, but i think maybe we should work toward greater standardization (if that's what we're shooting for) from a more holistic code organization perspective. like, in the case of components, the new pattern supports being able to import any component from a single barrel file in `src/components`. personally, i'd be interested in revisiting the whole containers/components organization, bc i don't think it's particularly meaningful right now.

src/containers/documentation/EndpointNotFoundPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+              </Link>
+            </li>
+          ))
+        }
+      </ul>
+    </div>
+  );
+};
+
+EndpointNotFoundPage.propTypes = {};
```
mjlumetta, line 46: currently, we do have the `match` prop... should that go in here?

src/containers/documentation/EndpointNotFoundPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { lookupApiCategory } from '../../apiDefs/query';
+import PageHeader from '../../components/PageHeader';
+import { APINameParam } from '../../types';
+import { PAGE_HEADER_ID } from '../../types/constants';
+import { Link } from 'react-router-dom';
+import { APIDescription } from 'src/apiDefs/schema';
+import { now } from 'lodash';
+
+const EndpointNotFoundPage = (props: RouteComponentProps<APINameParam>): JSX.Element => {
```
mjlumetta, line 14: i'd prefer to name this component `APINotFoundPage` instead of `EndpointNotFoundPage`, to reflect the domain entity we can't find instead of the route (especially since the route isn't technically an endpoint, bc SPA?), if others are amenable to that.

src/containers/documentation/Documentation.scss
```diff
@@ -9,7 +9,10 @@
     }
   }
 
-  .va-api-deprecation-alert {
+  .va-api-alert-box {
+    h3 {
+      margin: 0!important;
```
mjlumetta, line 14: truly annoying to have to use `!important`. should there be another space here? 

also, if this class is more broad than a deprecation alert, should it be rolled up outside of documentation? would default to no change, but figured i'd ask.

mjlumetta (commented):
what you have looks really good - have a few questions, cleanup comments, etc, but nothing too significant there. i do have questions about some stuff that isn't covered yet.

category mismatches: we [redirect back to the documentation root](https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/containers/documentation/DocumentationRoot.tsx#L135) if we can't find a category. that seems kind of jarring to me. can we cover routes like `/explore/fake` and `/explore/fake/docs/cookie` here?

for release notes, it straight-up breaks with an error. Koala is adding error boundaries for the React app as a whole, but we should handle `/release-notes/fake` gracefully too.

src/containers/documentation/EndpointNotFoundPage.tsx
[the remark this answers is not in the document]
gileswells: I don't have a strong opinion on this item either way right now but something to consider in the future.

src/containers/documentation/EndpointNotFoundPage.tsx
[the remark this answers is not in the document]
gileswells: I like how much cleaner it looks now.

src/containers/documentation/EndpointNotFoundPage.tsx
gileswells: `<PageHeader />` being that it's the `h1` of the page doesn't really lend itself to having a header default value since it's visible to the users. I just changed it to omit if there isn't a category name even though I can't see an instance where that would be logically possible.

src/containers/documentation/EndpointNotFoundPage.tsx
gileswells: There wasn't any motivation. I can change it.

src/containers/documentation/EndpointNotFoundPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import AlertBox from '@department-of-veterans-affairs/formation-react/AlertBox';
+import { RouteComponentProps } from 'react-router';
+
+import { lookupApiCategory } from '../../apiDefs/query';
+import PageHeader from '../../components/PageHeader';
+import { APINameParam } from '../../types';
+import { PAGE_HEADER_ID } from '../../types/constants';
+import { Link } from 'react-router-dom';
+import { APIDescription } from 'src/apiDefs/schema';
```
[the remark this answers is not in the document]
gileswells: Should be. vscode was set to import absolute. Finally googled to find the setting to force relative pathing by default.

src/containers/documentation/Documentation.scss
gileswells: There's no need to leave a space before `!important` and I've seen that there are some formatters that will remove it.

src/containers/documentation/EndpointNotFoundPage.tsx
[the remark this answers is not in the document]
anthony-kopczyk: @mjlumetta Possibly. I don't know much about `connected-react-router` to give a 'yes' or 'no', but from the looks of it we don't need to connect to map the location data to a component. I'm not sure about the actual redux actions, if we need pieces from the library for those.

anthony-kopczyk (commented):
Looks good to me with the latest changes. I'll let Mike approve once he gets a chance to review again.

src/containers/documentation/EndpointNotFoundPage.tsx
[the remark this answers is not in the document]
mjlumetta: hmm, i should probably do that too...

src/containers/documentation/Documentation.scss
mjlumetta: interesting! learn something new every day

src/containers/documentation/EndpointNotFoundPage.tsx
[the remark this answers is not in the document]
mjlumetta: we don't use the Redux actions at all. i'm going to make a ticket for removing `connected-react-router` - i'm pretty sure `SwaggerDocs` is the only component using it.

anthony-kopczyk (approved):

src/containers/releaseNotes/CategoryReleaseNotes.tsx
```diff
@@ -115,13 +115,17 @@ const ReleaseNotesCollection = (props: ReleaseNotesCollectionProps) => (
...
+  if (categoryDefinition != null) {
+    return (
+      <ReleaseNotesCollection
+        categoryKey={apiCategoryKey}
+        apiCategory={categoryDefinition}
+        apiFlagName="hosted_apis"
+      />
+    );
+  } else {
+    return <Redirect to="/release-notes" />;
```
mjlumetta, line 127: do we have UX input on this choice? the reason i'm asking is that i'm not sure that this is the behavior we want. if you go to a URL that you expect to work - let's say i want to go to "/release-notes/claims" - and it redirects you to "/release-notes", you have no idea what happened except that it won't let you go to the page you want to go to. there may be accessibility concerns for this behavior as well - not sure, would have to research. that said, i'm fine with this behavior for now if we create a follow-up ticket to research that behavior (is it okay?) and/or design a new experience.

mjlumetta (dismissed):
this PR looks good to go. i do have a request for either a follow-up PR or a follow-up ticket to complete the spirit of the original ticket - your call on which one to do.

src/containers/releaseNotes/CategoryReleaseNotes.tsx
gileswells: I think a toast message makes sense in this instance. That should also apply to `/explore/fakeCategory` because there we also just silently redirect to `/explore` but that would be something for a future ticket. I'll write it up.

mjlumetta: yup, exactly - would like that to apply to the docs pages as well. sounds on point to me. thanks Giles!!

mjlumetta (commented):
re-approving 💯

mjlumetta (approved):
actually re-approving 🤦 💯

PR #493, opened 2020-10-16
API-2909: Use route hooks on PageContent

anthony-kopczyk:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-2909

PageContent should now be using a location hook instead of receiving it as a prop.

### Process
- [N/A] Docs updated

### Requested feedback
Nothing specific


TS000 (approved):
Looks great!

PR #494, opened 2020-10-16
API 2911: Update QuickStart page to use Route hooks

TS000:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-2911

QuickStart page should now be using a param hook instead of receiving it as a prop.

- [ n/a ] Docs updated

### Requested feedback
All feedback is welcome


anthony-kopczyk (approved):
LGTM! These react router hook PRs are so so nice!

PR #495, opened 2020-10-16
API 2912:  Update Category Release Notes to use Route Hooks

TS000:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-2912
Category Release Notes pages should now be using a param hook instead of receiving it as a prop.

- [ n/a ] Docs updated

### Requested feedback
I have a feeling there might be a useLocation use but couldn't 100% nail down if one was needed.

### DIFF
--- a/src/containers/releaseNotes/CategoryReleaseNotes.tsx
+++ b/src/containers/releaseNotes/CategoryReleaseNotes.tsx
@@ -1,7 +1,7 @@
 import AlertBox from '@department-of-veterans-affairs/formation-react/AlertBox';
 import classNames from 'classnames';
 import * as React from 'react';
-import { RouteComponentProps } from 'react-router';
+import { useParams } from 'react-router';
 
 import { getDeactivatedCategory, isApiDeactivated } from '../../apiDefs/deprecated';
 import { getApiDefinitions } from '../../apiDefs/query';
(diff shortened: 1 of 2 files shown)


src/containers/releaseNotes/CategoryReleaseNotes.tsx
```diff
@@ -88,35 +88,33 @@ interface ReleaseNotesCollectionProps {
...
-      />
-      <div className={classNames('vads-u-width--full', 'vads-u-margin-top--4')}>
-        {props.apiCategory.apis.map((api: APIDescription) => (
-          <APIReleaseNote flagName={props.apiFlagName} key={api.urlFragment} api={api} />
-        ))}
-      </div>
-    </section>
-  );
-};
+const ReleaseNotesCollection = (props: ReleaseNotesCollectionProps) => (
```
anthony-kopczyk, line 91: The return type is missing on here.

anthony-kopczyk: Well, actually... this would be inferred right? Maybe we don't need to put it?

src/containers/releaseNotes/CategoryReleaseNotes.tsx
```diff
@@ -127,13 +125,11 @@ export const CategoryReleaseNotes = (props: RouteComponentProps<APINameParam>) =
...
-  return (
-    <ReleaseNotesCollection
-      categoryKey="deactivated"
-      apiCategory={getDeactivatedCategory()}
-      apiFlagName="enabled"
-      alertText="This is a repository for deactivated APIs and related documentation and release notes."
-    />
-  );
-};
+export const DeactivatedReleaseNotes = () => (
```
anthony-kopczyk, line 128: Same thing as before. Return type isn't defined, but should it be? It will be inferred here, right?

anthony-kopczyk (commented):
Looks good to me but not sure about whether or not we should have return types on two of the functions.

src/containers/releaseNotes/CategoryReleaseNotes.tsx
TS000: I do believe it is inferred

anthony-kopczyk (approved):
Looks good to me. Everything has been addressed.

PR #496, opened 2020-10-16
Api 2900:  Create Fetch Wrapper

TS000:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-2900

Async fetch wrapper to act as a middleware provider to the various fetch calls used in the Developer Portal.
I still need to write a test.
- [ ] Docs updated

### Requested feedback
All feedback welcome.


src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+    
+  catch (error) {
+    Sentry.withScope(scope => {
+      scope.setLevel(Sentry.Severity.fromString('warning'));
+      Sentry.captureException(error);
+    });
+    return dispatch(submitFormError(error.message));
+  }
+};
```
anthony-kopczyk, line 27: For readability we could probably write this error handling as a function that returns a promise and use it more like a direct middleware,

All this said, it's probably worth figuring out types, if we can general things moreso. Having request, errorFlag, and dispatch as Records/basically any type will make the API harder to use.

*Note: This code almost definitely doesn't run.*

I wrote some of this with promises and not async/await, but wanted to get general code concept across.

```
// Maybe find a way in the future to let middleware take the original request too?
interface Middldeware {
  (response: Response): Promise<Response>
}

// This wouldn't need to return a promise, but to make it function as a middleware we'd want it to look async
// There might be a better way to write these as async even though they aren't?
const errorHandlingMiddleware: Middleware = (response: Response): Promise<Response> => {
  return new Promise((resolve, reject) => {
    if (!response.ok && response.status !== 400) {
      throw Error(response.statusText);
    }
    resolve()
  })
}

const jsonMiddleware: Middleware = (response: Response): Promise<Response> => {
  const json = response.json()
   if (json.errors) {
      throw Error(`${errorFlag}: ${json.errors.join(', ')}`);
    }

  return response;
}

export const fetchWrap = async (params) => {
  // I like async/await but with the middleware thing this is one case that it looks really clean.
 // Maybe there's a direct await way to write it that's this clean that might help
  await fetch(request)
    .then(errorHandlingMiddleware)
    .then(jsonMiddleware)
    .then(tokenClientMiddleware)
    .catch(sentryErrorHandler);
};
```

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,29 @@
+import * as Sentry from "@sentry/browser";
+
+export const fetchWrap = async <T extends Record<string, unknown> | any>(request: Record<string, unknown>, errorFlag: Record<string, unknown>, dispatch: Record<string, unknown> ): Promise<T> => {
```
anthony-kopczyk: If this is generic to any type, can we do `async <T>` without the `extends`. That should let it work for anything.

For the request, errorFlag, dispatch types. Are those coming from fetch? Does fetch have types we can take? Or is there a @types/fetch node package?

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (json.errors) {
+      throw Error(`${errorFlag}: ${json.errors.join(', ')}`);
+    }
+    if (json.token || json.clientID) {
+      const result = dispatch(submitFormSuccess(json.token, json.clientID, json.clientSecret));
+      history.push('/applied');
+      return result;
+    } else {
+      return dispatch(submitFormError(json.errorMessage));
+    }
```
anthony-kopczyk, line 19: This section seems really specific. Is this for testing purposes? I'm not sure what dispatch is. If that's the 'redux' dispatch, we can probably keep that outside of the wrapper.

anthony-kopczyk (commented):
Looks good so far! I have a couple questions, and some ideas that aren't likely necessary at this time.

src/utils/fetchWrap.ts
TS000: "dispatch" comes from a param in a parent function from src/actions/apply.ts. This is where the wrapper will be used so my starting place was to grab the fetch from apply and turn its needed data into a reusable function.

src/utils/fetchWrap.ts
TS000: For your example async/await could look like this: 

```
export const fetchWrap = async (params) => {
  try { 
     await fetch(request)
     await errorHandlingMiddleware()
     await jsonMiddleware()
     await tokenClientMiddleware()
   }
   catch(error){
       sentryErrorHandler(error)
    };
};

Comment (anthony-kopczyk) [reply src/utils/fetchWrap.ts:24]: Oh. Yes! Because they all update the request. So we can pass that same object into them and it will get updated. Derp. Pass by reference and pointers are a thing.

File: src/utils/fetchWrap.ts (hunk 3 above)
Comment (anthony-kopczyk) [reply src/utils/fetchWrap.ts]: Okay. That sounds good. I'm confusing myself on the scope of this now and want to make sure we're on the same page.

Is the purpose of this wrapper to make like a global try/catch, or to try and catch errors with the actual request/response?

If it's just to try and catch errors (like 400s/500s), then I don't think we'll want the json middleware and token client middleware in this file. Those would be specific to `apply.ts` along with dispatch. Is that your intent? Trying to figure out if those being here are a result of it still being a draft or if it's going to be like this in the final version.

If it's more of a global try/catch, let me know. I'll need to think about that.

Comment (anthony-kopczyk) [reply src/utils/fetchWrap.ts]: You already updated the code! I didn't look at it closely before replying. This is already resolved. Thanks!

File: src/utils/fetchWrap.ts [hunk 4]
```diff
@@ -0,0 +1,24 @@
...
+export const fetchWrap = async (request: Request, errorFlag: string) => {
+  try {
+    const response = await fetch(request);
+    if (!response.ok && response.status !== 300 || 400 || 500) {
+      throw Error(response.statusText);
+    }
+    const json = await response.json() as Record<string, unknown>;
+    if (json.errors) {
+      throw Error(`${errorFlag}: ${json.errors.join(', ')}`);
+    }
```
anthony-kopczyk, line 19: Are we absolutely always receiving json from the server? I'm not sure if we want the json conversion in the wrapper in case we enable users to download files of some kind or text responses.

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,24 @@
...
+  } catch (error) {
+    errorFunc(error);
```
anthony-kopczyk, line 21: In the case of errors, should we send the response back still to expose it for additional handling?

```
// Not sure if this is valid
throw new Error(response, "message")
// or maybe put the params in an object for retrieval?
throw new Error({ response, "message" })
```

Or something like that. Then in a file like `apply.ts`

```
try {
  fetchWrap(request)
} catch (error: FetchWrapError) {
  // Do additional logging or processing after sentry logs - or maybe before sentry logs so that
  // if it's handled we can choose to ignore the sentry logs.
  // There should be a `finally` we can use
}

File: src/utils/fetchWrap.ts (hunk 4 above)
Comment (TS000) [reply src/utils/fetchWrap.ts:19]: Thats a great point. Should we simply return the response or provide logic to format different formats?

File: src/utils/fetchWrap.ts (hunk 5 above)
Comment (TS000) [reply src/utils/fetchWrap.ts:21]: I dont see how returning the original error would hurt. Sounds good to me.

File: src/utils/fetchWrap.ts (hunk 4 above)
Comment (anthony-kopczyk) [reply src/utils/fetchWrap.ts:19]: I think for now return the response directly. If we add the format logic later in a config, we can look at it as more of a helper. We could define the json middleware here and export it for use in other components that want it, so they don't need to rewrite the error handling.

> provide logic to format different formats?

I didn't think about that. We could pass in an optional config object (or I guess just a format object) with properties including formatting and default to json maybe, since I'm guessing we'll almost always be getting json.

```
interface Config {
  format: 'json' | 'text' | 'something else'
}
```

This is a case I'm hesitant to add the config object because I'm afraid we'll decide in 3 months it's more cumbersome with the config than without. This is a good candidate to get more eyes on.

File: src/utils/fetchWrap.ts [hunk 6]
```diff
@@ -0,0 +1,26 @@
...
+  });
+  return error;
+};
+
+const fetchHandler = (response: Response, errorID: string) => {
+  if ((!response.ok && response.status !== 300) || 400 || 500) {
+    throw TypeError(`Front End Error ID: ${errorID}`);
+  }
+  return response;
+};
```
anthony-kopczyk, line 16: Looks good, but the return type is missing. Or, maybe we should treat this like sentry. Maybe we should have this return void?

This kind of goes against the one other thing I had about middleware, but returning void is probably for the best to be explicit that it doesn't return another object.

If we do change this to modify the response object, then I think we want to rename it to `middleware`, since it's more convention that middleware will update the objects passing through it, while handlers only handle the object... if that makes sense? The term _middleware_ implies more strongly that the values passed to it will possibly be modified. And then even the middleware I think would return void.

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const fetchWrap = async (request: Request, errorID: string): Promise<unknown> => {
+  try {
+    const response = await fetch(request);
+    fetchHandler(response, errorID);
+  } catch (error) {
+    return sentryErrorHandler(error, errorID);
```
anthony-kopczyk, line 23: Instead of returning the error, we can `throw` it (same as rejecting the promise). That way places using fetchWrap can have a `try/catch` or `.then/.catch` and handle their additional processing that's specific to their request. This should also let you change the return type to `Promise<Response>`

There was a thing I proposed before about doing a `finally` block and enabling places calling the fetchWrapper to implement their own logic/callback/promise for the sentry handler. But then, that's another question.

Is this fetch wrapper to specifically catch errors like 300s, 400s, 500s, etc, or is it also to report errors that spawn from bad code. I think the error boundary would be for catching that bad code, right? So then this would be adequate without the `finally` thing. Is that right?

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+  return request;
```
anthony-kopczyk, line 25: Should this be `return response;`?

anthony-kopczyk (changes requested):
Looks really good overall. I like the latest refactor. I do have quite a few comments, mostly relating to return types and that sort of thing. I asked about the scope of this wrapper and if there is additional scope it should entail or not. I don't think it should? but, I wanted to make sure I'm on the right page.

src/utils/fetchWrap.ts
TS000: there is no response variable 🤷

src/utils/fetchWrap.ts
```diff
@@ -14,12 +15,13 @@ const fetchMiddleware = (response: Response, errorID: string) => {
...
+export const fetchWrap = async (request: Request, errorID: string): Promise<Response> => {
   try {
     const response = await fetch(request);
     fetchMiddleware(response, errorID);
   } catch (error) {
     return sentryErrorHandler(error, errorID);
   }
-  return Response;
+  // what do we return?
+  return ?;
```
TS000, line 26: What should we return here?

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+};
+
+export const fetchWrap = async (request: Request, errorID: string): Promise<Response> => {
+  try {
+    const response = await fetch(request);
+    fetchMiddleware(response, errorID);
+    return response;
+  } catch (error) {
+    sentryErrorLogger(error, errorID);
```
anthony-kopczyk, line 22: I know we just did this yesterday, but maybe `logWithSentry` instead?

anthony-kopczyk (commented):
Looks good, but I just worked on it with you, so not sure how much that weight holds!

anthony-kopczyk:
Oh dang. We'll need a test for fetchWrap. Not sure the best way to do it. Probably a lot of `expect(mockThing).toBeCalledWith`

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const sentryErrorLogger = (error: string, errorID: string): void => {
+  Sentry.withScope(scope => {
+    scope.setTag('Error ID', errorID);
+    Sentry.captureException(error);
+  });
+};
+
+const fetchMiddleware = (response: Response, errorID: string): void => {
+  if ((!response.ok && response.status !== 300) || 400 || 500) {
```
mjlumetta, line 11: can you explain what this condition is doing?

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,25 @@
...
+    throw TypeError(`Front End Error ID: ${errorID}`);
+  }
+};
+
+export const fetchWrap = async (request: Request, errorID: string): Promise<Response> => {
+  try {
+    const response = await fetch(request);
+    fetchMiddleware(response, errorID);
```
mjlumetta, line 19: what's the thought behind naming this function "middleware"? not quite sure it aligns with what i'd understand "middleware" to mean, but i'm also still learning about the overall vision and direction with this project.

src/utils/fetchWrap.ts
mjlumetta: out of curiosity, what's the `tokenClientMiddleware` in this example?

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,25 @@
...
+  });
+};
+
+const fetchMiddleware = (response: Response, errorID: string): void => {
+  if ((!response.ok && response.status !== 300) || 400 || 500) {
+    throw TypeError(`Front End Error ID: ${errorID}`);
+  }
+};
+
+export const fetchWrap = async (request: Request, errorID: string): Promise<Response> => {
```
mjlumetta, line 16: could we name the function something other than `fetchWrap`, maybe? would prefer to have a verb name for functions if possible. and also, if the idea is to abstract away the underlying HTTP API (is that part of the goal?), then maybe could do something more generic like `makeRequest`.

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,25 @@
+import * as Sentry from '@sentry/browser';
+
+const sentryErrorLogger = (error: string, errorID: string): void => {
+  Sentry.withScope(scope => {
+    scope.setTag('Error ID', errorID);
```
mjlumetta, line 5: looks good, glad this got worked out!! did you ever find any particular rationale for this one - as in, what distinguishes the tags from, say, `setContext` or `setFingerprints`? this looks right, i'm just curious about what the other things do... as we've talked about, Sentry docs not the best

mjlumetta (commented):
looks exciting! curious to see how it will look in context when it's substituted for the direct fetch calls in Apply and Contact Us. 

my main question on this one is what we're angling for with "middleware". as is, it doesn't look like the wrapper is extensible in the way i'd expect for what i'd understand as middleware, but that may be pending future expansion and all that. curious to hear more about what you're thinking on that.

src/utils/fetchWrap.ts
TS000: I took this error check from the apply.ts fetch call as I liked it's method of adding a top level error check to the initial call. I just added a Status update to the thrown error which should make the check more significant.

src/utils/fetchWrap.ts
TS000: how does "useFetch" sound?

src/utils/fetchWrap.ts
TS000: This part is like a placeholder to add functionality to our API calls. It follows the idea of the wrapper which was to add abstraction to our fetch calls since they _might_ become more and more complex as the project grows. Also this helps makes our fetch calls DRY.

src/utils/fetchWrap.ts
mjlumetta: i like "useFetch"! except... would that cause any issues with hooks linting or anything like that, given the naming convention collision? especially since hooks are inferred from the "use" prefix.

src/utils/fetchWrap.ts
mjlumetta: so how does the caller provide middleware operations to the wrapper? i guess that's the part i'm confused about. like right now, there's not a way for a caller to provide a series of things to this wrapper: `fetchWrap(req, errorID, doThing1, doThing2, doThing3...)`.

src/utils/fetchWrap.ts
mjlumetta: i guess my question was about the discrepancy between this condition and the one in Apply, which is `!response.ok && response.status !== 400`. so i'm wondering

1. what's the purpose of 300 and 500 statuses here?
2. i'm guessing the 400 and 500 parts are a typo/syntax error, but the way i'm reading this condition, those parts will always be truthy. so if `response.ok` is true, then the condition becomes `if (400 || 500)`, and both 400 and 500 are truthy individually.

src/utils/fetchWrap.ts
TS000: There are a myriad of places we can attach data in Sentry, I just chose tags since in the dashboard it makes a button you can click that then brings up the associated reports.

src/utils/fetchWrap.ts
TS000: That's a good point. I think this could be a hook with a little fussing. Might as well right?

src/utils/fetchWrap.ts
TS000: Good point! It might just be too restrictive atm. Maybe we just accept an object of items.
```
export const useFetch = async (obj): Promise => { ... }

File: src/utils/fetchWrap.ts (hunk 11 above)
Comment (TS000) [reply src/utils/fetchWrap.ts:11]: 300 and 500 have been removed

File: src/utils/fetchWrap.ts (hunk 12 above)
Comment (TS000) [reply src/utils/fetchWrap.ts:19]: Met with Johanna yesterday and the wrapper wont be accepting operations. It's purpose is to take fetch information, process it, and then return any errors as well as the desired return.

File: src/utils/fetchWrap.ts (hunk 13 above)
Comment (TS000) [reply src/utils/fetchWrap.ts:16]: decided to name it makeRequest

File: src/utils/fetchWrap.ts (hunk 12 above)
Comment (anthony-kopczyk) [reply src/utils/fetchWrap.ts:19]: For a request's specific needs, like validating that the response comes back with a parameter `foo` in the body. I'm curious about if that would get picked up by sentry. Is the goal of fetchWrap's sentry call to only find backend/frontend integration errors or also get the frontend error? I guess frontend specific errors could get picked up a boundary component, or we should be handling them gracefully anyway.

```
try {
  const response = await fetchWrap(request)
  const json = await response.json()

  if (!json.foo) {
    // Throw a fit of some kind, but this won't get picked up by sentry automatically, right?
    // Do we want this picked up by sentry?
    return
  }

  process(json.foo)
} catch (error) {
  // if error from sentry, this will be hit if we get a 400/500/etc
  // if error not from sentry, do we pass it to the error boundary component somehow?
}
```

I guess what I'm trying to ask is, if the backend doesn't return an expected result to the frontend (for instance, if it doesn't send `foo` because the backend API got messed up), are we handling that as a separate error from a bad response (ie. don't handle it in the fetch wrap) or would we want to pass that error to the fetch wrap component somehow to have it logged in sentry via fetchWrap, rather than a separate call to sentry.

> Met with Johanna yesterday and the wrapper wont be accepting operations. It's purpose is to take fetch information, process it, and then return any errors as well as the desired return.

Based on this, I think we wouldn't be passing it back up to fetchWrap to be logged, but even still, I'm uncertain if we'd consider a response having unexpected data as a separate error.

File: src/utils/fetchWrap.ts (hunk 1 above)
Comment (anthony-kopczyk) [reply src/utils/fetchWrap.ts:31]: @mjlumetta It was just a random name to represent a middleware. Nothing specific, named that way because I was looking into tokens and things..

File: src/utils/fetchWrap.ts [hunk 15]
```diff
@@ -0,0 +1,97 @@
...
+  request.headers.append('X-Request-ID', requestId);
+
+  try {
+    const response = await fetch(request);
+
+    let responseBody: any | string |null;
+
+    if(config.responseType === responseType.JSON){
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      responseBody = await response.json();
```
mjlumetta, line 69: will this fail if the response isn't 200?

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,97 @@
...
+
+   
+    const httpResponse:HttpResponse  = {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      body: responseBody,
+      ok: response.ok,
+      status:response.status,
+    };
+
+    handleNonNetworkError(url, requestId, config.responseType,  httpResponse );
```
mjlumetta, line 90: should this be above the `responseBody` handling?

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,97 @@
...
+    scope.setTag('Page Name', pageName);
+    Sentry.captureException(error);
+  });
+};
+
+/**
+ * Handles non network errors (response not ok and status not 200) and logs them to Sentry 
+ */
+const handleNonNetworkError = ( url: string, requestId: string,  type: string, response: HttpResponse): void => {
+  if(!response.ok){
```
mjlumetta, line 40: what about reversing this condition:

```
if (response.ok) {
  return;
}

// ... errors ...
```

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,97 @@
...
+    if(response.status === 400 && type === responseType.JSON && response?.body?.errors){
+      sentryErrorLogger(`Validation errors: ${response?.body?.errors?.join(', ')}`, requestId, url);
+    }
+    else if(response.status>=500 && response.status<=599){
+      // double check if it will be better to reject the errors in the 500 range
+      sentryErrorLogger(`Server Error:${response.status}`, requestId, url);
```
mjlumetta, line 46: probably do want reject? also, what about non-400, non-5xx responses?

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Interfaces & Constants
+ */
+interface CallFetchConfig {
+  responseType:string;
+}
+
+export interface HttpResponse {
+  ok: boolean;
+  status: number;
+  body: any | string | null;
```
mjlumetta, line 14: can we make this a generic type to avoid `any`? would also hope that that could encourage callers to define their response types explicitly and clearly and could abstract the type assertions away from the caller.

mjlumetta (commented):
left some suggestions. this looks really great. also accidentally reviewed #530, so there are a couple more comments on that, but i pulled the most important ones over here too.

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export const responseType = {
+  BLOB: 'BLOB',
+  JSON: 'JSON',
+  TEXT: 'TEXT',
+};
```
anthony-kopczyk, line 21: If we use generics for the response type, and use the `Content-Type` header from the response to determine if it's json, a blob, etc. then this variable could be removed.

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,97 @@
+import * as Sentry from '@sentry/browser';
+import { v4 as uuidv4 } from 'uuid';
+
+/**
+ * Interfaces & Constants
+ */
+interface CallFetchConfig {
+  responseType:string;
+}
```
anthony-kopczyk, line 9: We could add a `requiredProperties` object that takes an array of strings/keys. Then check that the response object contains each of those keys at run time (kind of like propTypes) to make sure the API sent back the data we expect. Not sure how fully on board we'd want to go with this but we could also add some semblance of type checking (string, function, number) or if there's an open source out of the box solution. I did a quick search but couldn't find one.

```
inteface CallFetchConfig {
  requiredFields: string[];
};

const validate = (response, config) => {
  Object.keys(config.requiredFields).forEach((field: string) => {
    if (response[field] === undefined) {
      throw new Error(`Response does not include required field ${field}`);
    }
  });
};
}
```

anthony-kopczyk (commented):
This looks good overall. Two main points:

- Is there runtime validation we want to check that properties exist on the response as we expect?
- I had some thoughts on responseType definition in CallFetchConfig and potentially reading the content-type header to determine what additional processing to call (response.json(), response.text(), etc)

src/utils/fetchWrap.ts
mjlumetta: ^ i really love this suggestion and think it's a huge benefit to get out of this kind of abstraction. for, like, a serious solution, JSON schema springs to mind (used quite a bit on the `vets-api` and VA.gov side, i believe, although not familiar with exactly how they use it). that might be heavy-handed, though (personally thinking yes, it is), so what you have looks pretty great to me and would be plenty for the problem we need to solve here.

src/utils/fetchWrap.ts
TS000: We tested this method with: 200, 404, 503, 400, 500 and none of them have broken the script. Is there an edge case you have in mind? I'm happy to test any 🕵️‍♀️

src/utils/fetchWrap.ts
TS000: We need the responseBody method to grab error details. See line 42.

src/utils/fetchWrap.ts
mjlumetta: i guess there are a few questions i have here that aren't quite my initial question.

- is there any particular reason for parsing the response body before checking the status? i don't think this is necessarily a bad idea, but it does require alignment in terms of the response body media type between success responses and error responses.
- should the backend align errors between success and error responses? looks like 404s return HTML. 500s return JSON. how do we handle that mismatch?

idk. what kind of testing did y'all do?

src/utils/fetchWrap.ts
TS000: In the native fetch calls non of the http code reject. 

Logging the other types of code, we need to be careful no to overflow sentry with too much data as it could make error correlation messy. 🤯

* Informational responses (100-199) <- Not sure we need to record these
* Successful responses (200-299) <- Not sure we need to record these
* Redirects (300-399) <- Not sure we need to record these
* Client errors (400-499) <- Identifying which of these codes to record will be important since some of them are expected results.
* Server errors (500-599) <- All of these need to be recorded since server errors are important to log.

Something to address is if we want to reject 500 errors so devs using the wrapper can address them in the catch block. We can do that or keep things as they are which is returning a resolved Promise allowing the error to be handled via the status code response.  🥅

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,114 @@
+import * as Sentry from '@sentry/browser';
+import { v4 as uuidv4 } from 'uuid';
+​
+// Interfaces & Constants
+
+interface CallFetchConfig {
+  responseType: string;
+}
+​
+export interface HttpResponse<T> {
```
mjlumetta, line 10: is the `T` used where this is consumed?

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,114 @@
...
+  ok: boolean;
+  status: number;
+  body: T | string | null;
```
mjlumetta, line 13: can we make this just `T`? if the caller wants a string or null, they could call `makeRequest<string>` or `makeRequest<null>`. and that way, `makeRequest` can be unambiguously typed by the caller.

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,114 @@
...
+    responseBody = (response.headers.get('Content-type')?.indexOf('application/json') !== -1) ? await response.json() : await response.text();
+  } catch(e){
+    responseBody = null;
+  }
+​
+  // Create sentry errors based on status
+  if(response.status === 400 && type === responseType.JSON && responseBody?.errors){
+    sentryErrorLogger(`Validation errors: ${responseBody?.errors?.join(', ')}`, requestId, url);
+  } else if(response.status === 404){
+    sentryErrorLogger(`API not found: ${response.status}`, requestId, url);
```
mjlumetta, line 53: i don't think we can assume this request is for an API? we only have 1/3 requests that are looking at any API-related currently (metadata yes, apply and contact us no). also, this raises the question of whether or not the request is to the developer portal backend... hmm (not a question for this PR, per se)

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,114 @@
...
+// Handles non network errors (response not ok and status not 200) and logs them to Sentry 
+
+const handleNonNetworkError = async( url: string, requestId: string,  type: string, response: Response): Promise<HttpResponse> => {
+​
+  // Tries to resolve the response to obtain error details
+  let responseBody: any | string | null;
+​
+  try{
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    responseBody = (response.headers.get('Content-type')?.indexOf('application/json') !== -1) ? await response.json() : await response.text();
```
mjlumetta, line 44: can we move this down inside the 400 block? it's not used anywhere else.

i would also say to assume it's JSON and cast it to `{ errors: string[] }`... although if you put it down in that block, i think you've already made that assumption. (would also maybe add a comment that we are making that assumption so it's explicit?)

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,114 @@
...
+  responseType: string;
+}
+​
+export interface HttpResponse<T> {
+  ok: boolean;
+  status: number;
+  body: T | string | null;
+}
+
+export enum responseType {
```
mjlumetta, line 16:
```suggestion
export enum ResponseType {
```

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,114 @@
...
+  const requestId: string = uuidv4() ;
+  
+  // Common Headers
+  request.headers.append('X-Request-ID', requestId);
+​
+​
+  try{
+    const response = await fetch(request);
+​
+    let responseBody: any | string |null;
```
mjlumetta, line 80: can we make this `T`?

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,114 @@
+import * as Sentry from '@sentry/browser';
+import { v4 as uuidv4 } from 'uuid';
+​
+// Interfaces & Constants
+
+interface CallFetchConfig {
+  responseType: string;
```
mjlumetta, line 7:
```suggestion
  responseType: ResponseType;
```

i think?

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,114 @@
...
+​
+    let responseBody: any | string |null;
+​
+    if(!response.ok){
+      return handleNonNetworkError(url, requestId, config.responseType,  response );
+    }
+ 
+    if(config.responseType === responseType.JSON){
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      responseBody = await response.json();
```
mjlumetta, line 88:
```suggestion
      responseBody = await response.json() as T; 
```

should be able to cast all of these response bodies to the expected success body type

mjlumetta (commented):
left some comments that i _think_ are in the spirit of what we talked about yesterday, but feel free to disagree. lmk if you want to pair again - i'll be around today until 5 Eastern.

mjlumetta (approved):

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,111 @@
...
+      httpResponse.body = await response.text() as T;
+    }
+
+    if (config.responseType === ResponseType.BLOB) {
+      httpResponse.body = await response.blob() as T;
+    }
+
+    return httpResponse as HttpResponse<T>;
+  // eslint-disable-next-line @typescript-eslint/no-implicit-any-catch
+  } catch (error) {
```
anthony-kopczyk, line 107: If we mark error as `unknown` can we remove the `eslint-disable-next-line`.

This makes me wonder as well if sentryErrorLogger should take an `unknown` type object to log. A generated `Error` by the system will be `Exception-like`. That's what `Sentry.captureException` says it wants. 

The only other issue is seeing what `Promise.reject` generated, if it's an (Error object) or if it's literally the JSON object we pass into it. But that would be more accurate data than a string as well.

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  // Create sentry errors based on status
+  if (response.status === 400 && typeof responseBody === 'object' && responseBody?.errors) {
+    sentryErrorLogger(`Validation errors: ${responseBody.errors.join(', ')}`, requestId, url);
+  } else if (response.status === 404) {
+    sentryErrorLogger(`Route not found: ${response.status}`, requestId, url);
+  } else if (response.status >= 500 && response.status <= 599) {
+    sentryErrorLogger(`Server Error: ${response.status}`, requestId, url);
+  }
```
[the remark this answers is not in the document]
anthony-kopczyk: Ooh. I just noticed that the exceptions are still captured, just not logged with sentry. Okay. Might be something else to add to JSDoc at the top.

"Rejects all status codes that are considered errors (400-599). For server errors (500-599), bad request errors (400), and route not found (404) errors,  they get logged with sentry as server errors."

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  // All http codes outside the 200 range will be rejected
```
anthony-kopczyk, line 65: This comment is a little misleading  I think.It's probably best to take it out completely since this function actually any requests that is passed to it, regardless of its status code. The function itself assumes the response is an error.

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,111 @@
...
+  const requestId: string = uuidv4();
+
+  // Common Headers
+  request.headers.append('X-Request-ID', requestId);
+
+  try {
+    const response = await fetch(request);
+
+    if (!response.ok) {
+      return await handleNonNetworkError(url, requestId, config.responseType,  response);
```
anthony-kopczyk, line 85: Is `await` necessary on here since we're turning a promise anyway?

anthony-kopczyk: Hmm.. or maybe I'd rather this be a thrown error to make it more obvious? I think you recently refactored some of this judging by the linter. Maybe that's something to discuss on another day as well.

anthony-kopczyk (commented):
I'm specifically questioning if we need the config ResponseType field, if we can use the content type header and remove the ResponseType interface completely. I'd like feedback on that or to know if it's already been discussed before approving.

I have various other comments about code comments, function names, and typing, but nothing that requires immediate attention before merging.

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,111 @@
...
+      httpResponse.body = await response.json() as T;
+    }
+
+    if (config.responseType === ResponseType.TEXT) {
+      httpResponse.body = await response.text() as T;
+    }
+
+    if (config.responseType === ResponseType.BLOB) {
+      httpResponse.body = await response.blob() as T;
+    }
```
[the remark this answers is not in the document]
TS000: I don't think a switch statement will be a better change as we don't have a huge if else chain going. I think 3 ifs is fine. The partial is for helping keep things as typed as possible while still having the flexability of maybe not knowing exactly what comes in.

src/utils/makeRequest.ts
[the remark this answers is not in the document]
TS000: I think a new ticket for these types of error can be made, it was decided that the current error checking is sufficient for the replacement of current fetch calls (see /apply.ts).

src/utils/makeRequest.ts
TS000: > Hmm.. or maybe I'd rather this be a thrown error to make it more obvious? I think you recently refactored some of this judging by the linter. Maybe that's something to discuss on another day as well.

the linter was being cruel and we had to make changes to appease it.

src/utils/makeRequest.ts
TS000: removed.

src/utils/makeRequest.ts
TS000: > This makes me wonder as well if sentryErrorLogger should take an unknown type object to log.

Argument of type 'unknown' is not assignable to parameter of type 'string'.

> But that would be more accurate data than a string as well.

Doesn't type string cover string-like objects like json?

anthony-kopczyk: > Argument of type 'unknown' is not assignable to parameter of type 'string'.

Yeah. The function definition of `sentryErrorLogger` would need updated to accept `unknown` instead of `string`. And then, instead of `throw new Error(error)` we can put `throw error`. I guess that gets into a dev methodology question though. If we're catching an error then it should always be an `Error` right? But since we're also catching rejected promises, that's not necessarily the case. Exceptions, as messy as they are in Java, are typed and everything and you can put multiple catch clauses. So with these exceptions I'm not sure how to handle it, since they aren't really the same at all.

```
const sentryErrorLogger = (error: unknown, errorID: string, endpointUrl: string): void => {
```

Since we're more or less just intercepting the error to log it, then we can just throw that same error instead of wrapping it in another error object. I think you can throw any kind of object, so maybe we'd be good.
```
} catch (error: unknown) {
    sentryErrorLogger(error, requestId, url);
    throw error;
  }
```

Looking at the sentry documentation, it talks about captureException taking 'exception-like' values in order to log a traceback. I can't find a definition for 'exception-like' though. Have you been able to test what this looks like in cases where we just pass a string versus where we pass objects? My main concern here, beyond switching to `unknown` is that the exception will get logged regardless of what type that parameter is/what fields `error` actually has.

anthony-kopczyk: > Doesn't type string cover string-like objects like json?

If an error is thrown as `new Error('my message')` then it's an object and not string-like. If you have a json string, then it's covered.

src/utils/makeRequest.ts
anthony-kopczyk: One semi-related question.

This values that's being returned. It's returning a reject promise. That rejected promise is not caught by the `catch` inside of this function, right? Or is it? I'm like 80% sure it's _not_ being caught by the `catch` in this function.

src/utils/makeRequest.ts
anthony-kopczyk: I think for now we can leave this as-is. Functionally it's not really going to make a change and the code is understandable.

src/utils/makeRequest.ts
TS000: > I'm like 80% sure it's not being caught by the catch in this function.

I'll run some tests to see, but before Mike and I started it was and I didn't think we changed that side of things.

src/utils/makeRequest.ts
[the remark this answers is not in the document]
anthony-kopczyk: I'm very very hesitant about the use of Partials, but like the other comments, I think we can leave this as-is as well. It's easy to understand the code, and I think if we did have a bug it wouldn't be hard to catch, since if we have a type error, then we'll probably start by looking at anywhere we cast an object `as Type`. Would find it very quickly.

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,111 @@
...
+    scope.setTag('Error ID', errorID);
+    scope.setTag('Endpoint Url', endpointUrl);
+    scope.setTag('Page Name', pageName);
+    Sentry.captureException(error);
+  });
+};
+
+// Handles non network errors (response not ok and status not 200) and logs them to Sentry
+
+const handleNonNetworkError = async <T extends unknown>(url: string, requestId: string,  type: string, response: Response): Promise<HttpResponse<T>> => {
```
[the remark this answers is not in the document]
TS000: I'm not sure the benefit as the current comment seems clear.

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,111 @@
...
+  // Tries to resolve the response to obtain error details
+  let responseBody: { errors: string[] } | null | string;
+
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    responseBody = (response.headers.get('Content-type')?.includes('application/json')) ? await response.json() : await response.text();
```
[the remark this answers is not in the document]
mjlumetta: i'm not sure about the approach of the dynamic content type reading. i think that the caller probably does need to know the expected response type of a success response at the very least, similar to setting an `Accept` header.

that said, don't think we need to work that out in this PR. what i would be interested in for a subsequent PR is getting some kind of block comment at the top of this file explaining the assumptions that the wrapper makes. for example, what servers are we making requests to? what content types do we want to parse? what are the known errors? etc.

src/utils/makeRequest.ts
mjlumetta: the rejected promise shouldn't be caught by the catch, i don't think, but i don't see why we'd want it to go to the catch in this function? i think this deserves a little more thought, too, but i'd like to refine it as we're testing with the  current endpoints.

src/utils/makeRequest.ts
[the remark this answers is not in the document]
mjlumetta: i'd be interested in talking about the Partial question synchronously, but i don't know have anything concrete to suggest about this code specifically - flagging if you want to talk about it.

src/utils/makeRequest.ts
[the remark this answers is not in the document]
anthony-kopczyk: 👍

anthony-kopczyk (dismissed):
My bunches of comments have all been addressed.

johanna-adhoc (dismissed):
💯

src/utils/makeRequest.ts
anthony-kopczyk: > the rejected promise shouldn't be caught by the catch, i don't think, but i don't see why we'd want it to go to the catch in this function? i think this deserves a little more thought, too, but i'd like to refine it as we're testing with the current endpoints.

Correct. I think we do not want it to be caught. I wanted to make sure it wouldn't get caught accidentally.

src/utils/makeRequest.ts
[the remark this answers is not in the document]
anthony-kopczyk: Sounds good. I think if we use `Partial` again and this comes up then we can have a discussion on it. Since most of our types are pretty strongly declared, I don't know if we'll be using Partial regularly.

johanna-adhoc (approved):
I tested the implementation with all the http code ranges (success and expected errors) and it is working as expected.

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,107 @@
...
+    Sentry.captureException(error);
+  });
+};
+
+// Handles non network errors (response not ok and status not 200) and logs them to Sentry
+const handleNonNetworkError =  async (url: string, requestId: string,  type: string, response: Response): Promise<HttpErrorResponse> => {
+  // Tries to resolve the response to obtain error details
+  let responseBody:  Record<string, unknown> | null | string = null;
+
+  if (response.headers.get('Content-type')?.includes('application/json')) {
```
anthony-kopczyk, line 47: The documentation for headers.get shows Content-Type with the T in type capitalized? Do you know if this get converts the string to match any cases? I know there was another place of code where we used all lower case and it functioned properly, but curious if you know.

https://developer.mozilla.org/en-US/docs/Web/API/Headers/get

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,107 @@
...
+      }
+
+      if (config.responseType === ResponseType.BLOB) {
+        httpResponse.body = await response.blob() as T;
+      }
+
+      return resolve(httpResponse as HttpResponse<T>);
+    } else {
+      const errorResponse =  await handleNonNetworkError(url, requestId, config.responseType,  response);
+      return reject(errorResponse);
```
anthony-kopczyk, line 100: Cool! I didn't know using async in a `then` gave access to `resolve` and `reject` like this without needing to create the promise and pass them in!

src/utils/makeRequest.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+  })
+    .catch(error => {
+      sentryErrorLogger(error, requestId, url);
+      return reject(new Error(error));
+    });
```
anthony-kopczyk, line 106: I think this is equivalent to the previous functionality, but this is more clean in terms of knowing when we are returning a rejected promise/what we're catching versus what we're passing back to later be caught. (that sentence is probably confusing but I can't think of another way to word it)

I like it!

anthony-kopczyk (approved):
Latest changes look good to me. Previous comments were addressed, and I really like some of the new modifications.

src/utils/makeRequest.ts
anthony-kopczyk: Nevermind to this question! I learned to actually read and it does say it's case-insensitivte in the docs, just not where I was expecting. Was expecting it to say 'is not case sensitive'

> The name of the HTTP header whose values you want to retrieve from the Headers object. If the given name is not the name of an HTTP header, this method throws a TypeError. The name is case-insensitive.

mjlumetta (approved):
lgtm to me!! excited to see this module applied to our current requests ✨

PR #498, opened 2020-10-19
Fix ESLint errors in Redux reducers

mjlumetta:
### Description

fixes ESLint errors in `src/reducers` for [API-1323](https://vajira.max.gov/browse/API-1323).

### Process

- [n/a] Docs updated


src/reducers/index.spec.ts
```diff
@@ -63,16 +63,16 @@ describe('application', () => {
       ['organization', constants.UPDATE_APPLICATION_ORGANIZATION],
     ];
 
-    inputToActionMap.forEach(([fieldName, actionName]) => {
+    inputToActionMap.forEach(([fieldName, actionName]: [string, string]) => {
```
TS000, line 66: I didn't know you could do that!

TS000 (approved):
TYLG!

src/reducers/api-versioning.ts
```diff
@@ -39,7 +39,7 @@ export const getDocURL = createSelector(
     if (!versionInfo) {
       return initialDocUrl;
     }
-    return `${process.env.REACT_APP_VETSGOV_SWAGGER_API}${versionInfo.path}`;
+    return `${process.env.REACT_APP_VETSGOV_SWAGGER_API || ''}${versionInfo.path}`;
```
nathanbwright, line 42: Curious why you are defaulting to an empty string here... If this is important behavior should it be defaulted everywhere `process.env.REACT_APP_VETSGOV_SWAGGER_API` is used?

mjlumetta: yes, good call! actually did this on another PR - it's in `src/constants/index`. will replace.

functionally, this is like a non-null assertion. the linting error for `@typescript-eslint/restrict-template-expressions` doesn't like possibly undefined values in templated strings. there is no difference at runtime - both alternatives will blow up if the env variable is undefined - but the URL will be `/my/path` instead of `undefined/my/path`, which i guess is slightly nicer but also doesn't matter very much.

nathanbwright (dismissed):
Looks good to me!

nathanbwright (approved):

PR #500, opened 2020-10-19
Fix ESLint errors in Apply components

mjlumetta:
### Description

fixes ESLint errors in Apply components for [API-1323](https://vajira.max.gov/browse/API-1323).

### Process

- [n/a] Docs updated


src/containers/apply/ApplyForm.tsx
```diff
@@ -56,7 +58,7 @@ const selectedApis = (props: ApplyProps) => {
 
 const anyOAuthApisSelected = (props: ApplyProps) => {
   const apiIdsByField = selectedApis(props).flatMap(
-    formField => APPLY_FIELDS_TO_URL_FRAGMENTS[formField],
+    formField => APPLY_FIELDS_TO_URL_FRAGMENTS[formField] as string,
```
anthony-kopczyk, line 61: Wanted to find a way to get rid of `as string` here because it's not as safe, and then you have to go looking to make sure the code can be trusted.. I'm not sure if this is a good change, but it would work:

If you change the definition of `APPLY_FIELDS_TO_URL_FRAGMENTS` to make it of type `Record<string, string>` then you won't need `as string` because it is defined to only contain string types. It makes the original definition less clean, but should indicate all of its held values are strings.

```
export const APPLY_FIELDS_TO_URL_FRAGMENTS = {
  benefits: 'benefits',
  claims: 'claims',
  communityCare: 'community_care',
  confirmation: 'veteran_confirmation',
  facilities: 'facilities',
  health: 'fhir',
  vaForms: 'vaForms',
  verification: 'veteran_verification',
} as Record<string, string>;
```

anthony-kopczyk: We might be able to apply this to other places or in the future.

anthony-kopczyk (commented):
Look good to me but have a suggestion for at least one place where we have to cast the type to string.

src/containers/apply/ApplyForm.tsx
mjlumetta: yeah, i think this is a great suggestion. pushing the type definition upstream seems like a definite improvement.

to me, the root of this problem is that there's an extremely loose coupling between the docs (with the API definitions from `src/apiDefs`) and the Apply page that we've never gotten around to fixing and causes the need for this mapping. ideally we could get rid of that entirely by aligning these little string ids, but that'd require either changing the routes here or changing the identifiers on the backend... hmm. might ask in Slack.

another solution that could go with yours would be canonizing the keys to be some enum called `APIIdentifier` or something: `type APIIdentifier = 'benefits' | 'claims' | 'communityCare' | ...`. it might not entirely solve the issue, though, and i'm not sure whether `APIIdentifier`, in general, would be the route identifier or the backend identifier. dunno... thoughts on all this?

src/containers/apply/ApplyHeader.tsx
```diff
@@ -3,24 +3,20 @@ import * as React from 'react';
...
+const ApplyHeader = (): JSX.Element => (
+  <div>
+    <PageHeader header="Apply for VA Lighthouse Developer Access" />
+    <p
+      className={classNames('usa-font-lead', 'vads-u-font-family--sans', 'vads-u-margin-bottom--2')}
+    >
+      This page is the first step towards developing with VA Lighthouse APIs. The keys and/or
+      credentials you will receive are for sandbox development only. When your app is ready to go
+      live, you may&nbsp;
+      <Link to="/go-live">request production access</Link>. Please submit the form below and
```
gileswells, line 15: You should be able to remove this `&nbsp;` and have the link move up to the previous line if you have other changes to make.

gileswells (commented):
Overall looks good to me but leaving it open for the `as string` discussion above.

src/containers/apply/ApplySuccess.tsx
```diff
@@ -55,13 +51,13 @@ const apisToEnglishApiKeyList = () => {
...
-function OAuthCredentialsNotice({
+const OAuthCredentialsNotice = ({
   clientID,
   clientSecret,
   email,
   selectedApis,
-}: IOAuthCredentialsNoticeProps) {
-  const apiNameList = selectedApis.map(k => apisToEnglishOAuthList[k]);
+}: OAuthCredentialsNoticeProps) => {
+  const apiNameList = selectedApis.map(k => apisToEnglishOAuthList[k] as string);
```
nathanbwright, line 60: Can we get away for using `as` here and line 83 too?

mjlumetta: yes i'll make that change on both!

src/containers/apply/ApplyHeader.tsx
mjlumetta: done!

nathanbwright (approved):
Looks good to me!

src/containers/apply/ApplyForm.tsx
anthony-kopczyk: I don't have any useful thoughts right now but something to take a look at later maybe. I was originally going to suggest an enum but then wasn't sure about how that would work with mapping. The original intention was to move the api definitions outside of the app some day right? However we do it, we should do it in a way compatible with that idea, or maybe that means the solution doesn't have to be that good, if moving the api definitions would result in another restructure.

PR #506, opened 2020-10-20
Fix ESLint errors in documentation and refactor SwaggerDocs to use VersionMetadata

mjlumetta:
### Description

fixes ESLint errors in `src/containers/documentation` for [API-1323](https://vajira.max.gov/browse/API-1323). also spawned some refactoring, which included

- removing the `IVersionInfo` interface from `SwaggerDocs` and replacing it with the `VersionMetadata` interface from `src/types/index`, since the two interfaces are identical and the types directory is a more appropriate place for a shared interface.
- refactoring the handling of version metadata to remove some unnecessary nesting that the developer portal doesn't care about. previously, we would take the whole `APIMetadata` object and pass it around `SwaggerDocs`, the Redux state, and Swagger UI plugins, but really, we only care about `APIMetadata.meta.versions`. so i took the response body from the metadata request, extracted the versions array at that point, and updated every subsequent place that it's used to refer to a versions array rather than the whole metadata object.
- deleting `ProviderIntegrationGuide` and inlining it in `Routes` instead. didn't really need its own file.

### Process

  - used tests for regression checks
- [n/a] Docs updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "build": "node scripts/build.js",
     "lint": "./node_modules/.bin/tslint --config tslint.json --project tsconfig.json",
     "lint:ci": "./node_modules/.bin/tslint --config tslint.ci.json --project tsconfig.json -t junit > lint-results.xml",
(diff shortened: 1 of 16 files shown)


nathanbwright (approved):
Not super familiar with SwaggerDocs, but it looks good to me.

ediiotero (approved):
looks good

PR #507, opened 2020-10-21
API-138: Block web spiders from dev urls

gileswells:
### Description
https://vajira.max.gov/browse/API-138

Through the webpack scripts for start and build we're swapping in robots.txt config files based on the environment at build. This should mean that future deploys to https://dev-developer.va.gov/ should explicitly block web spiders from crawling them.

When Jenkins runs it's builds and posts the links, the /robots.txt on the production one should match the one on prod currently and all others should have a `Disallow: /` rule on them to block all spiders.

### Process
- [ ] Docs updated


gileswells:
Direct links to robots files:
https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/45243d3/dev/robots.txt
https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/45243d3/staging/robots.txt
https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/45243d3/production/robots.txt

config/webpack.config.dev.js
```diff
@@ -344,6 +345,10 @@ module.exports = {
       tsconfig: paths.appTsConfig,
       tslint: paths.appTsLint,
     }),
+    // Place the appropriate robots.txt in public folder
+    new CopyPlugin({
+      patterns: [{ from: 'config/robots.dev.txt', to: 'robots.txt' }],
```
mjlumetta, line 350: what was the motivation for including this in the dev server? my thought would have been that the robots.txt doesn't really matter locally, but i could be not thinking of something here.

mjlumetta (commented):
looks great to me! good call on the plugin. only question is whether it's necessary for the dev server config

mjlumetta:
@gileswells not totally relevant to your changes, but i'm curious and you probably know more about `robots.txt` and SEO than me. why do we use `Disallow: ` as opposed to `Allow: *` or `Allow: /`? is there a difference, functionally or by convention?

config/webpack.config.dev.js
gileswells: Ummmmmm... To me I feel like not having a robots.txt seems more odd than having on that's specifically on dev. I mean it really doesn't help either way since it's not like we're showing the one that allows spiders to do anything. I think I like the idea of leaving it because if someone ever needs to show something on their local through something like an ngrok tunnel this will still apply and prevent it from being spidered.

gileswells:
> @gileswells not totally relevant to your changes, but i'm curious and you probably know more about `robots.txt` and SEO than me. why do we use `Disallow: ` as opposed to `Allow: *` or `Allow: /`? is there a difference, functionally or by convention?

I don't know why it was setup that way originally but in the grand scheme of things it really doesn't matter. Like there isn't a "more right way" of doing it.

mjlumetta:
> Like there isn't a "more right way" of doing it.

^ good enough for me 😄

config/webpack.config.dev.js
mjlumetta: that is 1000% not a case i had thought of. i would probably still default to leaving it out, but if you feel it's valuable, i'm cool with that. going to approve.

mjlumetta (dismissed):
🙌 ✨

mjlumetta (approved):

PR #508, opened 2020-10-21
Fix ESLint errors in root of src directory

mjlumetta:
### Description

fixes ESLint errors in the root of the `src` directory for [API-1323](https://vajira.max.gov/browse/API-1323).

miscellaneous effects:

- removes the unused parts of `registerServiceWorker`, explains that we only make sure there is no service worker, and leaves references for how to recreate this functionality in the future if desired.
- adds typing for the serialized version of the app state in session storage.
- removes unit tests for `CurlForm` from `npm run test:unit` because they are too damn slow. 🐢 

### Process

- [n/a] Docs updated


ediiotero (dismissed):
looks good 👍

src/sentenceJoin.ts
```diff
@@ -1,17 +1,17 @@
...
     return '';
   } else if (words.length === 1) {
     return words[0];
   } else if (words.length === 2) {
     return `${words[0]} and ${words[1]}`;
   } else {
-    const wordsDup = Object.assign([], words);
-    const lastWord = wordsDup.pop();
+    const wordsDup = Object.assign([], words) as string[];
+    const lastWord = wordsDup.pop() as string;
```
nathanbwright, line 11: What about something like:
```
const wordsDup = [...words];
const lastWord = wordsDup.pop() || '';
```

nathanbwright (commented):
Other than that one comment - it looks good to me!

src/sentenceJoin.ts
mjlumetta: yeah, maybe `wordsDup` part? i'm not sure i like the `|| ''` in this case, since we are so sure - based on those if statements - that this string will be defined. i don't think it's _bad_ to assert that something is a certain type in some cases as long as we're not gaming the compiler.

ediiotero (approved):
looks good 👍

PR #510, opened 2020-10-21
DRAFT: API-2540 - migrate SwaggerDocs to function component/hooks [HELP NEEDED]

alexandervalencia:
### Description
Migrating the `SwaggerDocs` container from class to functional component using hooks.

Currently not working:
- I believe one of the main issues is that the "`componentDidMount()`" replacement uses stale values even though they are updated. `setMetadataAndDocUrl()` is called, but within the scope of the useLayoutEffect, `docUrl` is still an empty string, so when `renderSwaggerUI()` is called, nothing is rendered because docUrl is still considered `""` even though it has been set in the state by `setMetadataAndDocUrl()`
- Another problem is with infinite loading when trying to switch versions. With the way that `usePrevious()` captures information, `prevVersion` and `version` will always be different as they ping pong back and forth:
```js
if (prevVersion !== version) { // "current" !== "0.0.1" so triggers a re-render
  // re-render
}
// on re-render, now prevVersion is "0.0.1" and version is "current" so trigger another re-render
// loop until we break
```

I'm not sure how to solve the issue as the container previously used `prevProps` to get the information which is not the same as a `useRef` which is what `usePrevious` is providing.

Any help fixing these issues would be appreciated!

### Process
- [ ] Docs updated

### DIFF
--- a/src/containers/documentation/SwaggerDocs.tsx
+++ b/src/containers/documentation/SwaggerDocs.tsx
@@ -1,9 +1,10 @@
 import * as Sentry from '@sentry/browser';
+import { History } from 'history';
 import * as React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useHistory, useLocation } from 'react-router-dom';
-import { usePrevious } from 'src/hooks/previous/Previous';
 import SwaggerUI from 'swagger-ui';
+import { usePrevious } from '../../hooks';
 import * as actions from '../../actions';
 import { APIDocSource } from '../../apiDefs/schema';
 import { getDocURL, getVersion, getVersionNumber } from '../../reducers/api-versioning';
@@ -25,6 +26,52 @@ export interface VersionInfo {
   internal_only: boolean;
 }
 
+const getMetadata = async (metadataUrl?: string): Promise<APIMetadata | null> => {
+  if (!metadataUrl) {
+    return null;
+  }
+  try {
+    const request = new Request(`${metadataUrl}`, {
+      method: 'GET',
+    });
+    const response = await fetch(request);
+    return response.json() as Promise<APIMetadata>;
+  } catch (error) {
+    Sentry.captureException(error);
+    return null;
+  }
+};
+
+const setSearchParam = (history: History, queryString: string, version: string) => {
+  const params = new URLSearchParams(queryString);
+  if (params.get('version') !== version) {
+    params.set('version', version);
+    history.push(`${history.location.pathname}?${params.toString()}`);
+  }
+};
+
+const renderSwaggerUI = (
+  dispatch: React.Dispatch<actions.SetRequestedAPIVersion>,
+  docUrl: string,
+  versionNumber: string,
+  metadata: APIMetadata | null,
+) => {
+  const handleVersionChange = (requestedVersion: string) => {
+    dispatch(actions.setRequstedApiVersion(requestedVersion));
+  };
+  if (document.getElementById('swagger-ui') && docUrl.length !== 0) {
+    const plugins = SwaggerPlugins(handleVersionChange);
+    const ui: System = SwaggerUI({
+      dom_id: '#swagger-ui',
+      layout: 'ExtendedLayout',
+      plugins: [plugins],
+      url: docUrl,
+    }) as System;
+    ui.versionActions.setApiVersion(versionNumber);
+    ui.versionActions.setApiMetadata(metadata as APIMetadata);
+  }
+};
+
 const SwaggerDocs = (props: SwaggerDocsProps): JSX.Element => {
   const dispatch: React.Dispatch<
     actions.SetRequestedAPIVersion | actions.SetInitialVersioning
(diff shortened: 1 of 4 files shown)


src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,110 +1,62 @@
...
-    this.setSearchParam();
-  }
-
-  private setSearchParam() {
-    const version = this.props.version;
-    const params = new URLSearchParams(this.props.location.search);
+  const history = useHistory();
+  const location = useLocation();
+  const prevApiName = usePrevious(props.apiName);
+  const prevVersion = usePrevious(version);
```
TS000, line 41: I was getting a type error here. I think you're missing the interface for that includes a version type:
`const prevVersion = usePrevious<VersionInfo>(version);`

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -118,26 +70,68 @@ class SwaggerDocs extends React.Component<ISwaggerDocsProps> {
...
+    setSearchParam();
+    renderSwaggerUI();
+  }, [renderSwaggerUI, setMetadataAndDocUrl, setSearchParam]);
+
+  /**
+   * componentDidMount()
+   */
+  React.useLayoutEffect(() => {
+    void setupAndRenderSwaggerUI();
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
anthony-kopczyk, line 105: This could have stale values in it because its dependencies aren't listed.

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -118,26 +70,68 @@ class SwaggerDocs extends React.Component<ISwaggerDocsProps> {
...
+
+  /**
+   * componentDidUpdate()
+   */
+  React.useLayoutEffect(() => {
+    if (prevApiName !== props.apiName) {
+      void setupAndRenderSwaggerUI();
+    } else if (prevVersion !== version) {
+      setSearchParam();
+      renderSwaggerUI();
```
anthony-kopczyk, line 115: Setting the search params here should update the url search query string to have the new version, which should then update prevVersion since version will update. Or I think it should? The swagger redux, indirectness, makes this tricky.

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -118,26 +70,68 @@ class SwaggerDocs extends React.Component<ISwaggerDocsProps> {
...
+  React.useLayoutEffect(() => {
+    void setupAndRenderSwaggerUI();
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  /**
+   * componentDidUpdate()
+   */
+  React.useLayoutEffect(() => {
+    if (prevApiName !== props.apiName) {
+      void setupAndRenderSwaggerUI();
```
anthony-kopczyk, line 112: This looks suspect to me on the infinite refreshes. Does prevApiName get updated?

anthony-kopczyk (commented):
Looks good. Tried asking some questions to help with the debugging.

src/containers/documentation/SwaggerDocs.tsx
anthony-kopczyk: I just looked closer. It should be getting set indirectly via the setupAndRender -> setSearchParams

anthony-kopczyk:
@alexandervalencia I think I got this working but not how I expected.

I added all used dependencies and removed the `eslint-disable-line`. Then I removed the effect that was for the original render, not expecting it to resolve the infinite redirect.
```
React.useLayoutEffect(() => {
    void setupAndRenderSwaggerUI();
  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```

I think `setupAndRenderSwaggerUI` was being updated causing both effects to get called and caused a back and forth chain effect between the two. The original render still works because prevApiName initializes to null and doesn't match whatever api name is given. Would be more testing to do but it seems promising. I can commit and push?

alexandervalencia:
@anthony-kopczyk Yeah, I think the reason why the `componentDidMount()` like aspect (the SwaggerUI properly loading up on the initial call was the reasoning I laid out in my first point in the description:
> I believe one of the main issues is that the "componentDidMount()" replacement uses stale values even though they are updated. setMetadataAndDocUrl() is called, but within the scope of the useLayoutEffect, docUrl is still an empty string, so when renderSwaggerUI() is called, nothing is rendered because docUrl is still considered "" even though it has been set in the state by setMetadataAndDocUrl()
I tried adding docUrl as a dependency but that also causes an infinite loop

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: For some reason, this doesn't cause the ping-pong issue I mentioned in my description, only the `prevVersion` and `version` `if` check are causing the loop

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -118,26 +70,68 @@ class SwaggerDocs extends React.Component<ISwaggerDocsProps> {
...
+  };
+
+  const renderSwaggerUI = React.useCallback(() => {
+    const handleVersionChange = (currentVersion: string) => {
+      dispatch(actions.setRequstedApiVersion(currentVersion));
+      setSearchParam();
+    };
+    if (document.getElementById('swagger-ui') && docUrl.length !== 0) {
+      console.log('renderSwaggerUI()');
+      const plugins = SwaggerPlugins(handleVersionChange);
```
mjlumetta, line 82: this line in particular bothers me for some reason because i feel like building the Swagger plugins shouldn't be tied up with the component lifecycle this way. as i'm reading it, there's a dependency chain from `SwaggerPlugins` -> `handleVersionChange` -> `setSearchParam` -> `version`. i would like to see if we can pull the plugins part out somehow, if not the others. i really don't have a good way of doing that thought out at all, but it seems like kind of a tell... the Swagger plugins shouldn't really change between renders except for the fact that they're now more tightly coupled with the closure they're instantiated in.

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -118,26 +70,68 @@ class SwaggerDocs extends React.Component<ISwaggerDocsProps> {
...
+  const setupAndRenderSwaggerUI = React.useCallback(async () => {
+    await setMetadataAndDocUrl();
+    setSearchParam();
+    renderSwaggerUI();
+  }, [renderSwaggerUI, setMetadataAndDocUrl, setSearchParam]);
+
+  /**
+   * componentDidMount()
+   */
+  React.useLayoutEffect(() => {
```
mjlumetta, line 103: why use `useLayoutEffect` instead of `useEffect`? is it because the `#swagger-ui` div needs to be rendered, and only `useLayoutEffect` can guarantee that (has to "read layout from the DOM and synchronously re-render", to quote the docs)? trying to follow on the details.

src/containers/documentation/SwaggerDocs.tsx
mjlumetta: a few questions here...

1. is the thought that running this with potentially stale values is fine because they can't be stale the first time?
2. if yes to (1), what does happen on subsequent renders (when presumably you could be rendering either a new version of an existing API or a new API)?
3. my _extremely_ rough understanding of hooks is that `useEffect` - or, in this case, `useLayoutEffect` - replaces both `componentDidMount` and `componentDidUpdate`. and, looking at the second `useLayoutEffect`, it looks like the first branch of the if statement would cover the `componentDidMount` case? my understanding of `usePrevious` is that it always returns null on the first render, meaning that `prevApiName` cannot equal `props.apiName` on the first render. so could we eliminate the first occurrence of this hook?

mjlumetta (commented):
leaving a bunch of comments here. haven't necessarily dug into it too deep, but i've got a lot of questions. definitely a tricky one.

one overarching question i have is why we need to memoize all the callbacks here. Johanna sent me some resources a little while ago suggesting caution around using `useCallback`. it might be that that's left an outsized impression on my thinking, but i'm wondering what benefit we are getting from memoizing. i don't think we have any computationally expensive operations we're offloading here, and i don't see any referential equality that matters other than memoizing other downstream callbacks correctly? would be curious to learn more about your approach here. here's one of the articles Johanna sent me: https://kentcdodds.com/blog/usememo-and-usecallback 

overall would say that this is a great opportunity to take a step back and try to rework how this component is working. i don't think the previous version was all that great, and i think the difficulty of converting this component to hooks is a symptom of that fact. would love to get something a bit more solid coming out of this change if we can.

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Yeah, the issue seems to be the way the redux store is updated and returning new values and useLayoutEffect not getting those new values until after re-rendering, so the values that renderSwaggerUI uses are stale even though they have been set by setSearchParam

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,110 +1,62 @@
...
     }
-  }
+  }, [history, location.search, version]);
 
-  private async setMetadataAndDocUrl() {
-    const { openApiUrl, metadataUrl } = this.props.docSource;
-    const metadata = await this.getMetadata(metadataUrl);
-    this.props.setInitialVersioning(openApiUrl, metadata);
-  }
+  const setMetadataAndDocUrl = React.useCallback(async () => {
```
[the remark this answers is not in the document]
alexandervalencia: If the functions aren't wrapped in `useCallback` and set as dependencies for the `useLayoutEffect`/`useEffect` calls, ESLint complains. So for now they are wrapped that way.

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,110 +1,62 @@
...
+    const { openApiUrl, metadataUrl } = props.docSource;
+    const currentMetadata = (await getMetadata(metadataUrl)) as string | undefined;
 
-  private async getMetadata(metadataUrl?: string): Promise<any> {
+    dispatch(actions.setInitialVersioning(openApiUrl, currentMetadata));
+  }, [dispatch, props.docSource]);
+
+  const getMetadata = async (metadataUrl?: string): Promise<any> => {
```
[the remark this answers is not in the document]
alexandervalencia: I don't see why not

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -118,26 +70,68 @@ class SwaggerDocs extends React.Component<ISwaggerDocsProps> {
...
-          plugins: [plugins],
-          url: this.props.docUrl,
-        }) as System;
-        ui.versionActions.setApiVersion(this.props.versionNumber);
-        ui.versionActions.setApiMetadata(this.props.metadata);
-      }
+  };
+
+  const renderSwaggerUI = React.useCallback(() => {
+    const handleVersionChange = (currentVersion: string) => {
```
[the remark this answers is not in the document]
alexandervalencia: Any suggestions? `version` is already in the scope of the function component.

What do you think of `requestedVersion`?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -118,26 +70,68 @@ class SwaggerDocs extends React.Component<ISwaggerDocsProps> {
...
+      const ui: System = SwaggerUI({
+        dom_id: '#swagger-ui',
+        layout: 'ExtendedLayout',
+        plugins: [plugins],
+        url: docUrl,
+      }) as System;
+      ui.versionActions.setApiVersion(versionNumber);
+      ui.versionActions.setApiMetadata(metadata as APIMetadata);
+    }
+  }, [dispatch, docUrl, metadata, setSearchParam, versionNumber]);
```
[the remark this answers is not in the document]
alexandervalencia: I don't think a lot of this is necessary, but mostly it's there to satisfy ESLint, I'm open to making changes but the function has to be passed on to SwaggerPlugins and as a rule you are not allowed to call hooks outside of function components: `React Hook "useDispatch" is called in function "handleVersionChange" that is neither a React function component nor a custom React Hook function.` is the error I get when trying to call `useDisptach` inside the function rather than set dispatch as a dependency

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: I don't have a good understanding of Swagger plugins, so I am completely open to changing things up, but still figuring out exactly how that should be done.

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: From the React docs on hooks:
> However, not all effects can be deferred. For example, a DOM mutation that is visible to the user must fire synchronously before the next paint so that the user does not perceive a visual inconsistency. (The distinction is conceptually similar to passive versus active event listeners.) For these types of effects, React provides one additional Hook called useLayoutEffect. It has the same signature as useEffect, and only differs in when it is fired.

Based on reading up on useEffect/useLayoutEffect, it seems, with the way that we're using it to render SwaggerUI, it makes sense to use `useLayoutEffect` but I could be wrong on that and am open to changing it if it makes sense to use `useEffect` instead.

alexandervalencia:
I agree that we should only use one instance of `useLayoutEffect` to replace both lifecycle methods. But it does get much more complicated to ensure everything is updating correctly since the more dependencies you add the more likely `useLayoutEffect` is to be called again.

Just still sussing out how to properly accomplish that

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,110 +1,62 @@
...
-  private handleVersionChange(version: string) {
-    this.props.setRequestedApiVersion(version);
-    this.setSearchParam();
-  }
-
-  private setSearchParam() {
-    const version = this.props.version;
-    const params = new URLSearchParams(this.props.location.search);
+  const history = useHistory();
+  const location = useLocation();
```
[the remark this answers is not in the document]
alexandervalencia: using both is simply how the component was originally working

PR #511, opened 2020-10-22
API-3129: Remove Newsletter Signup from /apply

gileswells:
### Description
Removes the newsletter signup link from the apply page in preparation for future work with newsletter signups

### Process
- [ ] Docs updated


mjlumetta (approved):
critics are calling it "a slam dunk" 🏀

PR #513, opened 2020-10-26
Enable more ESLint rules from main ESLint package

mjlumetta:
### Description

turns on a number of [ESLint rules](https://eslint.org/docs/rules/) from the main `eslint` package for [API-1323](https://vajira.max.gov/browse/API-1323). for scoping, rules in this PR are from the Possible Errors and Best Practices section in the ESLint docs.

this will be the first in a series of PRs adding or adjusting ESLint rules. most changes should reflect our current practices and shouldn't be a terribly large change.

#### Considered But Not Done

- `no-loss-of-precision`. we don't use floating point numbers ever, really. maybe will be useful in a future world where we're displaying, like, requests per minute usage or something, but not right now.
- `no-magic-numbers`. mostly didn't want to enable this one only to have to turn it off in every test suite. we can enforce this one as we see fit manually.
- `prefer-named-capture-group`. seems like a good idea but we rarely use capture groups to actually save the captured match.
- `no-void`. the `void` operator seems useful for voiding hanging Promises; otherwise would enable this
- `no-else-return`. too strict.

### Process

- [n/a] Docs updated

### Requested feedback


src/containers/Beta.tsx
```diff
@@ -396,30 +397,36 @@ export default class BetaPage extends React.Component<Record<string, unknown>, B
     );
   }
 
-  private toggleChecked = (key: string) => () => {
-    const update = {};
-    update[key] = !this.state[key];
-    this.setState(update);
-  };
+  private toggleChecked(key: string) {
```
gileswells, line 400: Wasn't one of the new ESLint rules one that enforces arrow functions? If so what's happening here that changes that?

gileswells: Also this happens throughout this file but only commenting it once.

src/containers/documentation/DocumentationRoot.e2e.ts
```diff
@@ -70,11 +70,11 @@ describe('position sticky', () => {
   });
 });
 
-describe('inavlid cagetories', () => {
-  for (const path of ['', 'docs/quickstart']) {
-    it(`should redirect "explore/invalid/${path}" to "explore"`, async () => {
+describe('invalid cagetories', () => {
+  it('should redirect invalid paths to "/explore"', async () => {
```
gileswells, line 74: Why the switch in order of the `it` and `for`?

src/containers/Beta.tsx
mjlumetta: it's from `no-invalid-this`. the distinction is the change to writing these functions as class methods rather than class properties that happen to be functions. the `prefer-arrow/prefer-arrow-functions` rule doesn't apply to class methods. 

i'm not sure whether this change is strictly necessary or not... i don't think it's a problem, though. in any case, i realllllyyy would like to kill this file entirely if we can...

src/containers/documentation/DocumentationRoot.e2e.ts
mjlumetta: the linting error was from `no-loop-func`, which i believe was a false positive. but also, it's come up before that defining tests inside a loop can be costly - think @rtravitz brought that up on a previous test-related PR. Ryan, if you have any insight/resources related to that, please feel free to share.

ediiotero: Not sure if this helps but would test.each or it.each help with this?

ediiotero (approved):
Just a question below, but otherwise everything looks good

src/containers/documentation/DocumentationRoot.e2e.ts
mjlumetta: ohhhh damnnn... probably, yeah! thanks @ediiotero, i did not know about that

ediiotero (approved):
looks good

PR #514, opened 2020-10-26
Enable stylistic ESLint rules from main ESLint package

mjlumetta:
### Description

turns on a number of [ESLint rules](https://eslint.org/docs/rules/) from the main `eslint` package for [API-1323](https://vajira.max.gov/browse/API-1323). for scoping, rules in this PR are from the Variables and Stylistic Issues section in the ESLint docs.

#### Considered But Not Done

- [`max-lines-per-function`](https://eslint.org/docs/rules/max-lines-per-function). not particularly helpful for longer JSX functions (debatable whether we should break those up) or for React components written with hooks (more clear-cut that they can be long since hook-related functions must be defined in function component's body).
- [`max-len`](https://eslint.org/docs/rules/max-len). too much work to correct right now.
- [`no-continue`](https://eslint.org/docs/rules/no-continue). probably not something we're going to want to see in our code, but maybe too strict to lint for, can leave to the author's and reviewers' discretion.
- [`indent`](https://eslint.org/docs/rules/indent). doesn't really work with Typescript, i don't think.
- [`one-var`](https://eslint.org/docs/rules/one-var) and [`operator-linebreak`](https://eslint.org/docs/rules/operator-linebreak). too strict, imo.
- [`camelcase`](https://eslint.org/docs/rules/camelcase) and [`id-match`](https://eslint.org/docs/rules/id-match). would like to lint for camel-case in some or all cases, but there are too many external identifiers we have to respect and also it's not all that important. maybe down the line.

### Process

- [n/a] Docs updated

#### Specific to This PR

this PR in particular includes stylistic rules that may or may not be useful. please let me know if you think anything is too strongly opinionated or if you think something is missing. i will default to omitting anything deemed too opinionated.

Jest test suites have some combination of `max-lines` and `max-nested-callbacks` disabled because they don't play nice with the structure of the test file. i think this is probably acceptable as long as source files and test files are linted with the same config, but would like feedback on that.

#### Generic


src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -13,18 +13,18 @@ import './AuthorizationDocs.scss';
...
-      return (
-        <div className="va-api-authorization-docs">
-          <PageHeader halo={category.name} header="Authorization" />
-          <OAuth />
-        </div>
-      );
-    } else {
-      return <ApiKeyAuth apiCategoryKey={apiCategoryKey} />;
-    }
+  if (category == null) {
```
gileswells, line 16: This reversed. Was there an upstream change that facilitated the need for this change?

gileswells: Nevermind. I see the removal below to change this to a positive match first.

gileswells (approved):
Overall I like this and agree with the ruleset you have. 

I would like to see `max-len` but that's something that can come later as I can only imaging that could be a PR in itself. When it comes to `no-return-assign` I'm on board with leaving it enabled. Best practice in my mind to know exactly what you're getting out of a function.

I agree with warnings and how it should either be an on or off thing. To me if I'm going to see warnings that don't really need to be fixed I'm still going to watch to fix them because a squiggly underline is still bothersome to me. Best to have a firm position on something so there isn't ambiguity.

src/containers/documentation/AuthorizationDocs.tsx
mjlumetta: yeah, this was from `no-negated-condition`

ediiotero (approved):
Looks great

PR #515, opened 2020-10-26
Enable several ES6-related rules from main ESLint module

mjlumetta:
### Description

turns on a number of [ESLint rules](https://eslint.org/docs/rules/) from the main `eslint` package for [API-1323](https://vajira.max.gov/browse/API-1323). for scoping, rules in this PR are from the ECMAScript 6 section in the ESLint docs.

fun thing i learned while doing the `prefer-destructuring` rule: you can destructure variables that have already been declared!! see

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration 
- https://stackoverflow.com/questions/32138513/how-to-destructure-an-object-to-an-already-defined-variable

#### Considered But Not Done

- [`sort-imports`](https://eslint.org/docs/rules/sort-imports). too much work to fix.

### Process

- [n/a] Docs updated

### Requested feedback

generic questions from #513 and #514.


ediiotero (approved):
Looks good.
Thanks for the info on destructuring, I learned something too 👍

PR #516, opened 2020-10-27
API-3122: Use redux hooks for ApplyForm and ApplySuccess

anthony-kopczyk:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-3122

- Removed the old redux connect() and mapStateToProps()
- Added some handling for if the application is undefined in the
ApplySuccess component

_This fixes that error in the routes that it can't accept and object!_

### Process
- [N/A] Docs updated

### Requested feedback

- Thoughts on throwing an error in `ApplySuccess` for the 'eventually existent' error boundary component to catch? Going to developer.va.gov/applied gives a blank page right now anyway, so it's not any worse than the current implementation.


src/Routes.tsx
```diff
@@ -40,7 +40,7 @@ export const SiteRoutes: React.FunctionComponent = () => (
       render={() => (
         <Flag
           name={['signups_enabled']}
-          component={ApplyForm}
+          render={ApplyForm}
```
anthony-kopczyk, line 43: Need to use `render` instead of `component`.

`render` takes a function that returns a `JSX.Element` like our functions using hooks do.
`component` requires a full on react component, whether a regular class or a `FunctionComponent`

src/containers/apply/ApplySuccess.tsx
```diff
@@ -130,4 +126,12 @@ const ApplySuccess = (props: ApplySuccessResult) => {
...
 };
 
-export default connect(mapStateToProps)(ApplySuccess);
+const ApplySuccess = (): JSX.Element => {
+
+  const result: ApplySuccessResult | undefined = useSelector((state: RootState) => state.application.result);
+
+  // Result can be undefined so we need this to be conditional
+  return result ? <ApplySuccessContent result={result} /> : <div>Error</div>;
+};
```
anthony-kopczyk, line 135: I made a content component for displaying that actual content, and I assume then we can make an error component for the case that the application isn't defined. But, that said, if we're adding the error boundary component and an 'Oops!' page, then maybe we don't need to worry about this case and let the error boundary component capture that error?

TS000: I'm into directing to an error boundary page.

package-lock.json
```diff
@@ -29650,7 +29650,7 @@
       "dependencies": {
         "json5": {
           "version": "1.0.1",
-          "resolved": "https://registry.npmjs.org/json5/-/json5-1.0.1.tgz",
+          "resolved": "http://registry.npmjs.org/json5/-/json5-1.0.1.tgz",
           "integrity": "sha512-aKS4WQjPenRxiQsC93MNfjx+nbF4PAdYzmd/1JIj8HYzqfbu86beTuNgXDzPknWk0n0uARlyewZo4s++ES36Ow==",
           "dev": true,
           "requires": {
```
TS000, line 29656: why did everything move to a non SSL page?

anthony-kopczyk: Umm. good question. No idea. This is the package lock. Guessing it did this by itself? I'll look into it. Good find. I would have missed this completely.

anthony-kopczyk: There were a few dependencies where this happened and were in the master branch. Something to keep an eye on as we move forward. The only way I was able to fix it was to clean my npm cache, manually change the existing package-lock to https and then npm install.

https://github.com/npm/npm/issues/16938#issuecomment-339863980

src/containers/apply/ApplySuccess.tsx
anthony-kopczyk: Updated this to directly throw an error that would then be caught by the error boundary component. When we add the boundary component, we could create some kind of scheme for custom error identifiers. Like for this one the issue is that the page has been gone to when something it needs doesn't exist. More specifically, this would happen if the person navigates straight to the page. If we had a code for that, we could send that code in the error and have the error boundary component say something like "You can't navigate directly to this page. Sorry!" Almost like websites having different pages for 400s versus 500s http codes.

src/containers/apply/ApplySuccess.tsx
```diff
@@ -130,4 +126,18 @@ const ApplySuccess = (props: ApplySuccessResult) => {
...
 };
 
-export default connect(mapStateToProps)(ApplySuccess);
+const ApplySuccess = (): JSX.Element => {
+  const result: ApplySuccessResult | undefined = useSelector((state: RootState) => state.application.result);
+
+  /*
+   * If there is no result in redux then this page can't render
+   * Throw an error to be catched by the error boundary component
+   */
```
anthony-kopczyk, line 135: I'm on board taking this comment out, but I put it in for now since this isn't a convention we have, and I'm not sure whether it's a convention we want.

src/containers/apply/ApplySuccess.tsx
```diff
@@ -130,4 +126,18 @@ const ApplySuccess = (props: ApplySuccessResult) => {
   );
 };
 
-export default connect(mapStateToProps)(ApplySuccess);
+const ApplySuccess = (): JSX.Element => {
+  const result: ApplySuccessResult | undefined = useSelector((state: RootState) => state.application.result);
+
+  /*
+   * If there is no result in redux then this page can't render
+   * Throw an error to be catched by the error boundary component
```
mjlumetta, line 134: @ediiotero and i were pair-reviewing this, and we were wondering - when will the error boundary work be done? and would you mind including a link to the ticket here?

if that work's in progress or going to be done soon, sounds good. if there's any doubt about that, though, would prefer not to intentionally cause a complete rendering error, even in an edge case like this one. regardless, though, defer to your judgment on this one, wouldn't block this from merging.

mjlumetta (approved):
co-reviewed with @ediiotero and looks good to us. have a question on the timeline of the error boundary work; wondering if it would be better to include the dummy "Error" rendering rather than breaking the render entirely and then taking out the bad fallbacks throughout the app when that work is done.

src/containers/apply/ApplySuccess.tsx
anthony-kopczyk: We're waiting on the design of the page to be done before implementing. I'll have it render a div for now like I had so that we at least capture the error for now. And I'll put on the ticket that we might want to revisit this after the error boundary component is implemented.

https://vajira.max.gov/browse/API-2751

src/containers/apply/ApplySuccess.tsx
```diff
@@ -130,4 +126,14 @@ const ApplySuccess = (props: ApplySuccessResult) => {
   );
 };
 
-export default connect(mapStateToProps)(ApplySuccess);
+const ApplySuccessError = (): JSX.Element => (
+  <div>Error! Unable to render apply success</div>
```
mjlumetta, line 130: only thing i might suggest would be adding some padding on this div, like with `vads-u-padding--4` or something similar. but frankly, not a prerequisite for merging because of how rare this case is and 

one thing i wonder coming out of this ticket is whether we should save the apply success result to session storage if it exists so that consumers who have successfully applied can navigate back to see their credentials. that seems like kind of a no-brainer now that i'm saying it, because it eliminates the most natural cause of this error. should one of us make a ticket? can do that if you'd like.

mjlumetta (approved):
nice job! ✨

src/containers/apply/ApplySuccess.tsx
anthony-kopczyk: Absolutely! You can put in a ticket.

I did some preliminary research that can probably be put in the description, or I can comment on it once it's made We'll probably want to do some more research on this before implementing. 

If you mean frontend/browser sessionStorage, then this link for owasp details it should be treated as insecure, so we'll need to find a way to secure the data. We should also take into account how this could change after adding an authenticated experience, however that might look.

> the Web Storage APIs should be treated as insecure storage. Because of this, if a business solution requires the use of the localStorage or sessionStorage to store sensitive data, such a solution should encipher data and apply replay protections.

https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html

PR #517, opened 2020-10-27
API-2540: Hooks Migration SwaggerDocs Container

alexandervalencia:
### Description

VA Jira ticket: [API-2540](https://vajira.max.gov/browse/API-2540)

- Reformats SwaggerDocs to be a functional component that uses hooks
- Updates redux actions to allow setting the version to something other than `'current'` on initial load

### Process
- [ ] Docs updated

### Requested feedback
All feedback welcome.


src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,130 +1,165 @@
...
+import { useHistory, useLocation } from 'react-router-dom';
 import SwaggerUI from 'swagger-ui';
-import * as actions from '../../actions';
+import { usePrevious } from '../../hooks';
+import {
+  setVersioning,
+  SetVersioning,
+  setRequstedApiVersion,
+  SetRequestedAPIVersion,
+} from '../../actions';
```
anthony-kopczyk, line 13: These actions look odd. Are these almost duplicates from a merge with masters?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,130 +1,165 @@
...
-    this.setSearchParam();
-    this.renderSwaggerUI();
+const getMetadata = async (metadataUrl?: string): Promise<VersionMetadata[] | null> => {
+  if (!metadataUrl) {
+    return null;
   }
+  try {
+    const request = new Request(`${metadataUrl}`, {
+      method: 'GET',
+    });
```
anthony-kopczyk, line 34: Is the default method for fetch a GET? Thinking we could have left this declaration out, but it doesn't need removed. Probably good to be explicit, never know when a browser might decide to implement fetch with different defaults.

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,130 +1,165 @@
...
+    const response = await fetch(request);
+    const metadata = await (response.json() as Promise<APIMetadata>);
+    return metadata.meta.versions;
```
anthony-kopczyk, line 37: This is nice. Didn't realize we didn't need the entire object. Might be worth renaming to `getMetadataVersions`, `getVersionMetadata`, or something like that?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,130 +1,165 @@
...
-      this.setSearchParam();
-      this.renderSwaggerUI();
-    }
+const getInitialVersion = (searchQuery: string) => {
+  const params = new URLSearchParams(searchQuery ?? undefined);
+  const versionQuery = params.get('version');
+  const version = versionQuery ? versionQuery.toLowerCase() : CURRENT_VERSION_IDENTIFIER;
+
+  return version;
+};
```
anthony-kopczyk, line 50: - Return type

I like the inclusion of the CURRENT_VERSION_IDENTIFIER for setting the default version. That said I'm not sure if there's a way to make this a little more readable/obvious. I had to think a little bit to realize that it's the default if version doesn't exist in the query params.

Maybe if we remove line 49, and make 47 return right away, without the variable declaration? That way the condition is directly on the return statement. Not sure though.

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,130 +1,165 @@
...
+
+const handleVersionChange = (dispatch: React.Dispatch<SetRequestedAPIVersion>) => (
+  (requestedVersion: string) => {
+    dispatch(setRequstedApiVersion(requestedVersion));
   }
+);
```
anthony-kopczyk, line 56: Not sure if the linter has a rule or not. Do we need to define `void` when a function returns void?

@mjlumetta Not sure if we have eslint rules being added for this? or if we already decided not to enforce having `void`?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,130 +1,165 @@
...
-      return null;
+  /**
+   * CLEAR REDUX STATE ON UNMOUNT
+   */
+  React.useEffect(
+    () => () => {
+      dispatch(setVersioning('', null));
+    },
+    [], // eslint-disable-line react-hooks/exhaustive-deps
+  );
```
anthony-kopczyk, line 132: I don't know if there's a good way to document this, but it might be good to explain that we only want this on mount and unmount, so we specifically don't want these dispatch and setVersioning added as dependencies.

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,130 +1,165 @@
...
+
+  /**
+   * UPDATES URL WITH CORRECT VERSION PARAM
+   */
+  React.useEffect(() => {
+    if (prevVersion !== version) {
+      setSearchParam(history, location.search, version);
     }
-  }
+  }, [history, location.search, prevVersion, version]);
```
anthony-kopczyk, line 141: This was so tricky/frustrating with the multiple dependencies triggering the function. Needing to still check that version was the thing that actually changed.

*shakes head sadly*

anthony-kopczyk: So glad it's together now!

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,130 +1,165 @@
...
+  /**
+   * TRIGGERS RENDER OF SWAGGER UI
+   */
+  React.useEffect(() => {
+    if (docUrl) {
+      renderSwaggerUI(dispatch, docUrl, versionNumber, versions as VersionMetadata[]);
     }
-  }
-}
+  }, [dispatch, docUrl, versions, versionNumber]);
```
anthony-kopczyk, line 150: Is there a way we can get rid of the casting on `versions`. We could add that as optional to `renderSwaggerUI` maybe? I think only a subset of the APIs have version metadata (Health FHIR), but not certain.

src/reducers/api-versioning.ts
```diff
@@ -76,7 +76,8 @@ export const apiVersioning = (
     case constants.SET_REQUESTED_API_VERSION:
       return { ...state, requestedApiVersion: action.version };
     case constants.SET_VERSIONING:
-      return { ...state, docUrl: action.docUrl, versions: action.versions };
+      const requestedApiVersion = action.version || state.requestedApiVersion;
+      return { ...state, docUrl: action.docUrl, requestedApiVersion, versions: action.versions };
```
anthony-kopczyk, line 80: Woah! I was wondering about combining the set requested api version and set versioning actions! That's so crazy you did this! Woooo!!!!

src/actions/index.ts
```diff
@@ -11,8 +11,9 @@ export interface SetRequestedAPIVersion extends Action {
 
 export interface SetVersioning extends Action {
   docUrl: string;
-  versions: VersionMetadata[] | null;
   type: constants.SET_VERSIONING;
+  version?: string;
+  versions: VersionMetadata[] | null;
 }
 
 export const setRequstedApiVersion: ActionCreator<SetRequestedAPIVersion> = (version: string) => ({
```
anthony-kopczyk, line 19: Can we fix this name with the typo? I'm not sure where all it's used. Maybe it should be a separate PR. I've been happily looking past it for weeks, as I'm sure everyone else has.

src/containers/documentation/ApiDocumentation.test.tsx
```diff
@@ -35,6 +35,20 @@ const server = setupServer(
...
+    push: jest.fn(),
+  }),
+  useLocation: jest.fn().mockReturnValue({
+    hash: '',
+    key: '5nvxpbdafa',
+    pathname: '/another-route',
+    search: '',
+    state: null,
+  }),
+}));
```
anthony-kopczyk, line 50: Nice!

anthony-kopczyk (commented):
I didn't go in and test them to make sure  they're working. That said, the changes look really good. I have a lot of semantics kinds of questions, nothing that serious.

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: the PascalCase are the interfaces, the camelCases are the actual functions

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Yeah, the default method is `GET` (https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch), we could probably take it out, I didn't really touch this particular function other than keeping it line with master.

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Renamed to `getVersionsFromMetadata` 👍

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: - Good catch! return type of `string` added
- removed extra variable to return line 47

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Yeah, I had issues before when using `versionNumber` and the previous. With `version` it seems to be working correctly, plus we're now getting `version` on initial render from the params if they exist.

src/containers/documentation/SwaggerDocs.tsx
johanna-adhoc: That does not look like a good naming convention 😅

johanna-adhoc (approved):
💯 Looks good!

I tested in local and all is working as expected!

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Because `versions` can potentially be `null` (https://github.com/department-of-veterans-affairs/developer-portal/blob/5450fd1a3fc47b1a705ba2b501b4a9d182c93678/src/types/index.ts#L28) it's either casting it here.

Casting it inside of `renderSwaggerUI` or setting the `versions` param of `renderSwaggerUI` to be `VersionMetadata[] | null` but then we have to wrap `ui.versionActions.setVersionMetadata(versions)` in an `if` check.

I'm open to other solutions! Just not sure which is best when it comes to TS

src/reducers/api-versioning.ts
alexandervalencia: Yeah, the biggest issue I kept running into was `current` **ALWAYS** being set as the initial version, so I had to change it! 🙌  🎉

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,130 +1,165 @@
...
+import { useDispatch, useSelector } from 'react-redux';
+import { useHistory, useLocation } from 'react-router-dom';
 import SwaggerUI from 'swagger-ui';
-import * as actions from '../../actions';
+import { usePrevious } from '../../hooks';
+import {
+  setVersioning,
+  SetVersioning,
+  setRequstedApiVersion,
+  SetRequestedAPIVersion,
```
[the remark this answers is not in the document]
alexandervalencia: Yes, looks like one was a typo, fixed in latest commit

src/reducers/api-versioning.ts
mjlumetta, line 80: would you mind renaming the `docUrl` state property to `defaultUrl` or something like that? it's really never used if the metadata URL is present and loads successfully... although if there's not metadata then it is always that URL/it is the only URL

src/actions/index.ts
```diff
@@ -11,20 +11,23 @@ export interface SetRequestedAPIVersion extends Action {
...
 }
 
-export const setRequstedApiVersion: ActionCreator<SetRequestedAPIVersion> = (version: string) => ({
+export const setRequestedApiVersion: ActionCreator<SetRequestedAPIVersion> = (version: string) => ({
   type: constants.SET_REQUESTED_API_VERSION,
   version,
 });
 
 export const setVersioning: ActionCreator<SetVersioning> = (
   docUrl: string,
```
mjlumetta, line 25: similarly to the state comment in the reducers, would you be open to renaming `docUrl` to `defaultUrl`? open to thoughts on this.

src/actions/index.ts
```diff
@@ -11,20 +11,23 @@ export interface SetRequestedAPIVersion extends Action {
 
 export interface SetVersioning extends Action {
   docUrl: string;
-  versions: VersionMetadata[] | null;
   type: constants.SET_VERSIONING;
+  version?: string;
```
mjlumetta, line 15: if the version param in the action creator is typed as a string, can this be marked required?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,129 +1,164 @@
...
+  const versions = useSelector((state: RootState) => state.apiVersioning.versions);
+
+  const initializing = React.useRef(true);
+
+  let version = useSelector((state: RootState) => getVersion(state.apiVersioning));
+
+  if (initializing.current) {
+    initializing.current = false;
+    // Use the version from the search param only if it's the first render
+    version = getInitialVersion(location.search);
```
mjlumetta, line 100: so question: why _doesn't_ the selector give the correct version on the first render? related question (i think, maybe): why do we read the query param in `ApiDocumentation` and write it here?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,129 +1,164 @@
...
+   */
+  const { apiName } = props;
+  const { openApiUrl, metadataUrl } = props.docSource;
+
+  const prevApiName = usePrevious(apiName);
+  const prevVersion = usePrevious(version);
+
+  const setMetadataAndDocUrl = async () => {
+    const metadataVersions = await getVersionsFromMetadata(metadataUrl);
+    const initialVersion = getInitialVersion(location.search);
```
mjlumetta, line 114: so the bug here was that if you navigated directly to `developer.va.gov/explore/benefits/docs/claims?version=0.0.1` or something like that, it would render current anyway? or was this bug specific to the hooks migration?

src/containers/documentation/SwaggerDocs.tsx
mjlumetta: yeah, wait... is the whole thing that it only returns the unmount function? also, what's the consequence of not having it?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,129 +1,164 @@
...
-        plugins: [plugins],
-        url: this.props.docUrl,
-      }) as System;
-      ui.versionActions.setApiVersion(this.props.versionNumber);
-      ui.versionActions.setVersionMetadata(this.props.versions);
+  /**
+   * TRIGGERS RENDER OF SWAGGER UI
+   */
+  React.useEffect(() => {
+    if (docUrl) {
```
mjlumetta, line 146: is this condition duplicated within `renderSwaggerUI()`?

src/containers/documentation/SwaggerDocs.tsx
mjlumetta: i think that the `versions` parameters of `renderSwaggerUI` and `setVersionMetadata` and basically anywhere it's used in the Swagger UI plugins can be nullable, if that works for you.

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,129 +1,164 @@
...
+  if (params.get('version') !== version) {
+    params.set('version', version);
+    history.push(`${history.location.pathname}?${params.toString()}`);
   }
+};
 
-  private handleVersionChange(version: string) {
-    this.props.setRequestedApiVersion(version);
-    this.setSearchParam();
+/* eslint-disable max-params */
```
mjlumetta, line 63: would probably prefer bumping the rule up if you're open to it, but i'm okay with this too if you prefer it.

mjlumetta (commented):
looks really nice!! this component was so ugly, so glad it's getting refactored. i had a handful of questions that i don't quite understand; if you've got any insight into those things, would be great to talk those through. thanks Alex!

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Because the default coming from the store is `'current'`, so even if we check and load a requested version that is not initially `'current'` (say you bookmarked the url with `?version=0.0.1`), it will flicker with current FIRST then load 0.0.1 and change the params back to `0.0.1`

The query params for ApiDocumentation are for the tabs on https://developer.va.gov/explore/health/docs/fhir?version=current

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: It wasn't a bug when the component was a class but because of the different way that functional components handle hooks and refreshing state it became a bug

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Yeah, it looks like we can take the inner docUrl.length check out

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Either way is fine for me, three is probably fine, maybe one too low, but that's just my opinion

alexandervalencia:
@mjlumetta would you mind explaining https://github.com/department-of-veterans-affairs/developer-portal/pull/517#discussion_r513047427 more, I'm not sure what you're saying

alexandervalencia:
@mjlumetta regarding https://github.com/department-of-veterans-affairs/developer-portal/pull/517#discussion_r513044509 the consequence of not having the reset on unmount is that when you navigate to a different API it flickers with an initial load based on the previous data. 
 
e.g. If you were to navigate from Benefits Claims to Appeals Status, you would see a brief flash of a version selector and the docs from Benefits Claims until the store updated with the fresh state.

src/containers/documentation/SwaggerDocs.tsx
mjlumetta: i would say add this check in there or pull the other condition out here. would like to keep them together, no preference for where they're located.

src/actions/index.ts
alexandervalencia: Updated for version to be optional in the ActionCreator as well

src/reducers/api-versioning.ts
alexandervalencia: Makes sense, I can rename

src/containers/documentation/SwaggerDocs.tsx
mjlumetta: vis a vis `ApiDocumentation`, i'm referring to the fact that `actions.setRequestedApiVersion` gets called in a `useEffect` there (in the diff, since you caught that spelling mistake). and hmm... i'm wondering whether there's a way to change that. like, can we move the query param handling farther up the tree? should we set it here instead of in `ApiDocumentation`? should we not render this component until we've determined the correct version, not assuming that it's current until we've actually read the initial URL?

i'm not sure whether there are any possible improvements to make here, but it feels like there might be something we could rework so that's not the case... eh, not sure.

mjlumetta:
> @mjlumetta would you mind explaining #517 (comment) more, I'm not sure what you're saying

the versions metadata can always be null, and therefore, i made a mistake typing it as `VersionMetadata[]` instead of `VersionMetadata[] | null`. the key check is here: https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/containers/documentation/swaggerPlugins/ExtendedLayout.tsx#L24. so it can be `VersionMetadata[] | null` anywhere in the Swagger UI plugin system rather than adding a conditional here.

src/actions/index.ts
mjlumetta: what do you think about requiring it and defaulting it to the empty string, maybe?

src/containers/documentation/SwaggerDocs.tsx
mjlumetta: yeah... appending `Action` to the interfaces would be a plus if you feel like doing that

src/actions/index.ts
alexandervalencia: In the reducer we're checking if it's null or not and if it's null, we use the original default value: https://github.com/department-of-veterans-affairs/developer-portal/pull/517/files#diff-fbe8d3a1bbe0b7fa9e1d05180289519c18389a7a9294f5a2f630318aa406fdebR79

src/containers/documentation/SwaggerDocs.tsx
anthony-kopczyk: I think this is a byproduct of the really annoying thing with hooks and redux. I'm not sure there's a nice way to resolve it.

If you're in an `Effect` and you issue a redux action that updates redux state - the `Effect` stays in its own little world and doesn't get the updated redux values until it gets called a second time, even though the redux action finished inside of that function call. It's kind of nice, but it makes redux calls like this really confusing. When researching it, people said to separate things into multiple effects but that gets really confusing trying to figure out which effects are for what and when they trigger.

alexandervalencia:
@mjlumetta I believe I've made the changes you requested as best I could, let me know if you see any other issues, thanks!

src/actions/index.ts
mjlumetta: right, yeah... i think my optional suggestion here would be to require `version` in the action type and default it to the empty string in the creator instead of undefined. i don't think there are functional changes except that we're indicating explicitly that we don't know what version to show yet. this only makes sense if we're going to do other refactors, though (ie explicitly setting the version from query params up front).

gonna throw out this suggestion - i don't have anything concrete that can fit in the scope of this PR.

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,129 +1,164 @@
...
-        plugins: [plugins],
-        url: this.props.docUrl,
-      }) as System;
-      ui.versionActions.setApiVersion(this.props.versionNumber);
-      ui.versionActions.setVersionMetadata(this.props.versions);
+  /**
+   * TRIGGERS RENDER OF SWAGGER UI
+   */
+  React.useEffect(() => {
+    if (defaultUrl) {
```
mjlumetta, line 146: can we either pull out the other condition in Swagger UI to here or move this one back in there?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,129 +1,164 @@
...
-      return metadata.meta.versions;
-    } catch (error) {
-      Sentry.captureException(error);
-      return null;
+  /**
+   * CLEAR REDUX STATE ON UNMOUNT
+   */
+  React.useEffect(
+    () => () => {
+      dispatch(setVersioning('', null));
```
mjlumetta, line 128: can we maybe make a special action creator called `resetVersioning` or something like that that will abstract away these default values? and/or make a comment explaining why this is necessary?

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -72,7 +72,7 @@ const ApiDocumentation = (props: ApiDocumentationProps): JSX.Element => {
   const apiVersion = queryParams.get('version');
 
   React.useEffect((): void => {
-    dispatch(actions.setRequstedApiVersion(apiVersion));
+    dispatch(actions.setRequestedApiVersion(apiVersion));
```
mjlumetta, line 75: can we move this into `SwaggerDocs`?

src/containers/documentation/SwaggerDocs.tsx
```diff
@@ -1,129 +1,164 @@
...
+  if (document.getElementById('swagger-ui')) {
+    const plugins = SwaggerPlugins(handleVersionChange(dispatch));
+    const ui: System = SwaggerUI({
+      dom_id: '#swagger-ui',
+      layout: 'ExtendedLayout',
+      plugins: [plugins],
+      url: defaultUrl,
+    }) as System;
+    ui.versionActions.setApiVersion(versionNumber);
+    if (versions !== null) {
```
mjlumetta, line 78: don't think we should have this condition; it's reflective of a typing error in our Swagger UI plugins. would prefer to have that correctly reflected here.

mjlumetta (commented):
couple of small cleanup things, should be able to approve once addressed. nice job!!

src/containers/documentation/ApiDocumentation.tsx
alexandervalencia: Unfortunately no, because this addresses the Tab changes requesting new api versions, and the Tabs live in `ApiDocumentation`.

I think that's part of what makes these containers messy, it's almost like ApiDocumentation and SwaggerDocs should be one container but that would make a single, very large file.

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Re-added the length check here.

src/containers/documentation/ApiDocumentation.tsx
mjlumetta: don't quite follow what you're saying here? the tab changes have to do with a different query param, which would trigger a reload of the version stuff but doesn't conflict with it directly. this action is specific to the `version` query param.

agreed that the boundary between these components isn't very clear, and we should file that away for future improvements. not sure what that looks like.

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Updated and I believe I've correctly addressed the issue by updating the System type to allow for `null`

src/containers/documentation/SwaggerDocs.tsx
alexandervalencia: Added

mjlumetta (dismissed):
💯 ✨

johanna-adhoc (approved):
💯

PR #518, opened 2020-10-27
Enable several rules from the ESLint import plugin and uninstall ESLint JSDoc plugin

mjlumetta:
### Description

does 2 things:

1. uninstalls `eslint-plugin-jsdoc` since we don't use JSDoc. this plugin was suggested by the `tslint-to-eslint-config` autogeneration tool.
2. adds a bunch of rules from [`eslint-plugin-import`](https://github.com/benmosher/eslint-plugin-import).

one thing to note is that i added explicit dependencies on `reselect`, `@braintree/sanitize-url`, and `swagger-client` for the [`import/no-extraneous-dependencies`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md) rule.

#### Considered But Not Done

[`import/exports-last`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/exports-last.md) and [`imports/grouped-exports`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/group-exports.md). could be nice in theory but it's not really in line with our styles as they stand.

### Process

- [n/a] Docs updated

### Requested feedback

_specific to this PR:_ should we prefer [`import/no-duplicates`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-duplicates.md) over ESLint's [`no-duplicate-imports`](https://eslint.org/docs/rules/no-duplicate-imports) rule?


ediiotero (dismissed):
Looks good 👍

ediiotero (approved):
👍

PR #519, opened 2020-10-28
Enable several ESLint React rules

mjlumetta:
### Description

enables several [`eslint-plugin-react`](https://github.com/yannickcr/eslint-plugin-react) rules for [API-1323](https://vajira.max.gov/browse/API-1323).

the branch is off `API-1323/eslint-refinement-4` for #518 because i was initially trying to do them in one PR but it got too big. can merge the smaller #518 first and then target master.

#### Considered But Not Done

- [`react/sort-comp`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md). seems nice in theory but not really necessary, and it doesn't apply to function components, which we expect to be the majority of our components now.
- [`react/forbid-prop-types`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md). specificity in `propTypes` would be nice, but we need to focus on much more basic usage of `propTypes` first.
- [`react/require-default-props`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md). seemed interesting, but there are legitimate use cases for props that are genuinely optional, so not enabling.
- [`react/jsx-fragments`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-fragments.md). could be nice to standardize on one of the syntaxes but def not necessary.
- [`react/jsx-one-expression-per-line`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-one-expression-per-line.md). intriguing but too strict.
- [`react/jsx-props-no-spreading`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md). i consider props spreading usually not a good idea, but there are some legit cases for its use and i don't feel changing them is worth the effort.

### Process

- [n/a] Docs updated

### Requested feedback


.eslintrc.js
```diff
@@ -103,6 +103,10 @@ const coreESLintRules = {
 /**
  * A lot of import rules are already covered by the Typescript compiler, so we don't use any
  * of the eslint-plugin-import presets. These rules are helpful additions.
+<<<<<<< HEAD
+ * https://github.com/benmosher/eslint-plugin-import/tree/master/docs/rules
+=======
+>>>>>>> origin/API-1323/eslint-refinement-4
```
ediiotero, line 109: Looks like there is a conflict here

ediiotero (commented):
Just had a comment, looks like something left over from a merge conflict

TS000 (approved):
Looks great!

PR #521, opened 2020-10-28
Move forked repos to a permanent home

gileswells:
### Description
This PR updates the packages that we've been using on a fork to be hosted on the DSVA github organization. In the grand scheme of things there should be 0 changes visible as it's all just package based.

### Process
- [ ] Docs updated


mjlumetta (commented):
this looks correct to me, but i can't get it to run locally right now. getting this error:

```
Module not found: Can't resolve 'markdown-component-loader' in '/Users/mikelumetta/developer-portal'
```

there's also some failure in the e2e tests that appear to be related to being unable to load Markdown components. will revisit tomorrow, but if you have any insights into the test failures, let me know.

alexandervalencia (dismissed):
LGTM 🚀

package.json
```diff
@@ -138,7 +138,7 @@
     "js-yaml-loader": "^1.2.2",
     "jsdom": "^16.4.0",
     "lodash.noop": "^3.0.1",
-    "markdown-component-loader": "github:adhocteam/markdown-component-loader#dv-3146-from110",
+    "markdown-component-loader": "github:department-of-veterans-affairs/markdown-component-loader#develop",
```
mjlumetta, line 141: when i was previously updating `react-router-sitemap`, @nfasulo shared this about the referencing a specific commit: "Kalil found that pointing to a particular commit greatly reduced the time it took to run npm install." so i don't know if we want to do that here or not, but submitting it for suggestion.

mjlumetta: similarly, should we standardize on the way we're referencing these repos? i think it's 6 on one hand, a half dozen on the other, but aligning on either `git+https://<url>` or `github:<org name>/<repo name>` might be nice. also v optional.

mjlumetta (approved):
looks good from my perspective, everything is working now. i left a couple of optional touch-up suggestions that could be useful, but those are up to you. will re-approve if you make either/both changes.

also, did you find any explanation/resources on the prepare vs prepublish thing that you can share? i'm a little curious why it wasn't happening with the old repos.

package.json
gileswells: > also, did you find any explanation/resources on the prepare vs prepublish thing that you can share? i'm a little curious why it wasn't happening with the old repos.

I hadn't. I just found that small branch that was being referenced and it was causing the `lib` folder to be created on install vs it not being there from the main develop branch. Once I made that change it worked.

gileswells: I've updated to link directly to the commit hash.

package.json
```diff
@@ -138,7 +138,7 @@
     "js-yaml-loader": "^1.2.2",
     "jsdom": "^16.4.0",
     "lodash.noop": "^3.0.1",
-    "markdown-component-loader": "github:department-of-veterans-affairs/markdown-component-loader#develop",
+    "markdown-component-loader": "github:department-of-veterans-affairs/markdown-component-loader#4fc9ec4ceae45c189a29befc5e9cfa3b136b4a71",
```
mjlumetta, line 141: can you merge this commit to develop on the repo? otherwise looks great

mjlumetta (approved):
one more tiny thing, noticed that the commit referenced for `markdown-component-loader` isn't on `develop`.  think it would be good to get that on the base branch, but this PR looks fine overall

gileswells:
@mjlumetta I just changed that recently. I don't have it on develop so that it's possible to merge and PR against the main package if we ever have a need to.

mjlumetta:
> @mjlumetta I just changed that recently. I don't have it on develop so that it's possible to merge and PR against the main package if we ever have a need to.

okay, yeah, that's cool too. the reason i suggested having it on develop would be so that it's easy for people to find the commit. i think either is okay, just depends which thing should be easier or not. maybe if we're going this way it could be good to add a commit documenting the branch we're using in our `markdown-component-loader` repo at the top of the README so that we can find it easily.

PR #522, opened 2020-10-28
Add ESLint accessibility plugin and uninstall TSLint

mjlumetta:
### Description

this PR removes TSLint from the project for [API-1323](https://vajira.max.gov/browse/API-1323). that includes the following steps:

- installed [`eslint-plugin-jsx-a11y`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y) to replace all of the `react-a11y-*` TSLint rules.
- uninstalled TSLint and all its plugins.
- removes TSLint disable inline comments.
- updated the [`fork-ts-checker-webpack-plugin`](https://www.npmjs.com/package/fork-ts-checker-webpack-plugin) to use ESLint instead of TSLint. also upgraded it to 5.2.1 because (a) our version was way behind and (b) the new version assumes that you're using ESLint, not TSLint.
- removes `npm run eslint` and makes `npm run lint` call ESLint.
- removes TSLint from CI.
  - in Jenkins, we renamed the TSLint stage to ESLint and removed the separate stage calling `npm run eslint`.
  - the CodeBuild pipeline never had both ESLint and TSLint stages, so switching `npm run lint` to call ESLint instead of TSLint takes care of that.
- adds a separate CI config like `tslint.ci.json` so that devs can make `console.log` statements locally (since the dev server type checker is now looking at ESLint).

### Process

- [n/a] Docs updated

### Requested feedback

did i miss anything?


nathanbwright (approved):
Looks good!

ediiotero (approved):
Looks good!

PR #523, opened 2020-10-29
API-3120:  hook migration  embedded youtube video, markdown page, hero

TS000:
### Description
Minor movement to modern Hooks.

- [ NA ] Docs updated

### Requested feedback
All feedback welcome.


src/Routes.tsx
```diff
@@ -6,7 +6,7 @@ import { getDeactivatedFlags } from './apiDefs/deprecated';
 import { getEnvFlags } from './apiDefs/env';
 import { getApiCategoryOrder, getApiDefinitions } from './apiDefs/query';
 import { APIDescription } from './apiDefs/schema';
-import MarkdownPage from './components/MarkdownPage';
+import { MarkdownPage } from './components/markdownPage/MarkdownPage';
```
anthony-kopczyk, line 9: This can be changed to `import { MarkdownPage } from './components';`

src/components/embeddedYoutubeVideo/EmbeddedYoutubeVideo.tsx
```diff
@@ -12,7 +12,7 @@ interface EmbeddedYoutubeVideoProps {
 const YOUTUBE_SERVICE = 'youtube';
 const YOUTUBE_BASE = 'https://www.youtube.com/embed/';
```
anthony-kopczyk, line 14: We should be able to add prop types for url and title.

Here's some code that I'm guessing isn't 100% correct. :)
```
EmbeddedYoutubeVideo.propTypes = {
  url: propTypes.string.isRequired,
  title: propTypes.string.isRequired,
}
```

src/utils/fetchWrap.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const fetchWrap = async (request: Request, errorID: string): Promise<unknown> => {
+  try {
+    const response = await fetch(request);
+    fetchHandler(response, errorID);
+  } catch (error) {
+    return sentryErrorHandler(error, errorID);
+  }
+  return request;
+};
```
anthony-kopczyk, line 26: Guessing this isn't supposed to be in this PR.

src/containers/Home.tsx
```diff
@@ -6,7 +6,7 @@ import apiDefinitions, { apiCategoryOrder } from '../apiDefs/data/categories';
 import padlockImg from '../assets/homepage-padlock.svg';
 import apiImg from '../assets/homepage-reliable-api.svg';
 import { CardLink } from '../components';
-import Hero from '../components/Hero';
+import { Hero } from '../components/hero/Hero';
```
anthony-kopczyk, line 9: Should be able to update this to import from `../components` too

anthony-kopczyk (commented):
Looks good for the most part. A few imports can be updated but that's about it.
Looks like some code from fetchWrap got into this PR by accident too.

src/components/hero/Hero.tsx
```diff
@@ -2,9 +2,9 @@ import classNames from 'classnames';
 import * as React from 'react';
 import { Link } from 'react-router-dom';
 
-import logo from '../assets/hero-logo.svg';
+import logo from '../../assets/hero-logo.svg';
 
-const Hero: React.FunctionComponent = (): JSX.Element => (
+const Hero = (): JSX.Element => (
```
mjlumetta, line 7: this change is fine because the `React.FunctionComponent` doesn't provide a ton of benefit, but it does have a small benefit. if you're working in an IDE, the Typescript compiler should provide intellisense about properties like `defaultProps` and `propTypes`. if you were already aware of that and decided to get rid of it, please ignore this comment.

src/components/embeddedYoutubeVideo/EmbeddedYoutubeVideo.tsx
mjlumetta: yeah, would love to see that added - could also use that `type EmbeddedYoutubeVideoProps = PropTypes.InferProps<typeof propTypes>` pattern here, in that case. 

would also ask, if we're not doing something along those lines, what the purpose of the changes in this file are. i'm cool with the style change if that's what you prefer to follow, but if we'd like to follow that overall, then we can throw in a linting rule for it and follow it across the project. lmk and i'll add it in my final PR.

mjlumetta (commented):
seconding every @anthony-kopczyk said, plus a couple style comments. looks good overall!

src/Routes.tsx
TS000: Ah thank you!

src/utils/fetchWrap.ts
TS000: no, and I'm not sure why it made it in. this happened to another pr recently. maybe it's because it's a draft?

src/components/hero/Hero.tsx
TS000: Ya know I thought it was redundant, I'll put it back thank you.

PR #525, opened 2020-10-30
Enable several Typescript ESLint rules and fix errors

mjlumetta:
### Description

final PR for [API-1323](https://vajira.max.gov/browse/API-1323). this PR adds several rules from the [`@typescript-eslint/eslint-plugin`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin) module. some notes:

1. Typescript ESLint has several rules that are extensions of core ESLint rules. this PR tidies those up so that the Typescript ESLint rules are always preferred.
2. this PR adjusts a handful of rules to be warnings in development and errors in CI. rationale: console logging, unused variables, incomplete typings, and whitespace are fairly normal in development and shouldn't necessarily cause the dev server not to compile. everything that is a warning locally should be an error in CI, though, or there's not a ton of point in linting for it. also, `fork-ts-checker-webpack-plugin` has a "compiled with warnings" state in its newer version, which is nice imo.
3. adds comments to `.eslintrc.js` to document plugins and decisions made.
4. checked against the `eslint:recommended`, `@typescript-eslint/recommended`, and `@typescript-eslint/recommended-requiring-type-checking` presets to make sure there were no redundant or unnecessary rules.

### Process

- [n/a] Docs updated

### Requested feedback

* is [`@typescript-eslint/explicit-function-return-type`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md) too strict/overkill? a lot of changes in this PR are from that.
* how do we feel about [`@typescript-eslint/no-implicit-any-catch`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-implicit-any-catch.md)? it seems to be recommended, and like it could be beneficial (?), but it also doesn't seem like that helpful of a fix... dunno.


TS000 (dismissed):
Great job, this is going to help a ton!

.eslintrc.js
```diff
@@ -1,8 +1,22 @@
...
+ * Notes
+ *  1. The Typescript ESLint plugin has several rules that extend rules in the core ESLint module. In
+ *  general, we prefer to use the Typescript ESLint version. You can see a list of the extended rules
+ *  here: https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#extension-rules
+ *  2. If any rule should apply to committed code but is painful in development, it should be included
+ *  as a warning in this file and marked as an error in .eslintrc.ci.js.
+ * 
+ * Presets included
+ *  - Typescript ESLint: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended.ts
+ *    = ESLint via Typescript ESLint: https://github.com/eslint/eslint/blob/master/conf/eslint-recommended.js
```
alexandervalencia, line 16: @mjlumetta Is this a typo?

```suggestion
 *   - ESLint via Typescript ESLint: https://github.com/eslint/eslint/blob/master/conf/eslint-recommended.js
```

src/containers/documentation/swaggerPlugins/index.tsx
```diff
@@ -14,16 +14,16 @@ const SwaggerPlugins = (versionHandler: (newVersion: string) => void): Plugins =
...
+    authorizeBtn: (): null => null,
   },
   fn: {
     curlify,
   },
   statePlugins: {
     spec: {
       wrapSelectors: {
-        allowTryItOutFor: () => () => false,
+        allowTryItOutFor: () => ((): boolean => false),
```
alexandervalencia, line 26: Should this be:
```suggestion
        allowTryItOutFor: (): () => boolean => ((): boolean => false),
```

or is that syntax overkill? it seems less clear with the first function being typed and probably unnecessary, but just wanted to ask.

src/types/constants/actions/apply.ts
```diff
@@ -1,36 +1,36 @@
...
+export const UPDATE_APPLY_ORGANIZATION_VALUE = 'UPDATE_APPLICATION_ORGANIZATION';
+export type UPDATE_APPLY_ORGANIZATION = typeof UPDATE_APPLY_ORGANIZATION_VALUE;
 
-export const UPDATE_APPLICATION_DESCRIPTION = 'UPDATE_APPLICATION_DESCRIPTION';
-export type UPDATE_APPLICATION_DESCRIPTION = typeof UPDATE_APPLICATION_DESCRIPTION;
+export const UPDATE_APPLY_DESCRIPTION_VALUE = 'UPDATE_APPLICATION_DESCRIPTION';
+export type UPDATE_APPLY_DESCRIPTION = typeof UPDATE_APPLY_DESCRIPTION_VALUE;
 
-export const UPDATE_APPLY_OAUTH_APP_TYPE =
+export const UPDATE_APPLY_OAUTH_APP_TYPE_VAL =
```
alexandervalencia, line 16: Seems long, but I think it would be better to not break the naming convention here
```suggestion
export const UPDATE_APPLY_OAUTH_APP_TYPE_VALUE =
```

src/types/swagger-client.d.ts
```diff
@@ -34,8 +34,8 @@ declare module 'swagger-client' {
   }
 
   export interface Operation {
-    parameters: Parameter[];
-    requestBody: RequestBody;
+    parameters?: Parameter[];
+    requestBody?: RequestBody ;
```
alexandervalencia, line 38:
```suggestion
    requestBody?: RequestBody;
```

alexandervalencia (dismissed):
I left a few comments, and there was an extra space. Up to you if you want to make any of the changes before merging, otherwise LGTM 🚀 

Thanks for doing an incredible amount of work related to this ticket!!! 🥇

src/containers/documentation/swaggerPlugins/index.tsx
mjlumetta: hmm... it's a good question. what about not inlining all of this? i'm a bit reluctant to add more parens and arrows to the line bc it's starting to make me dizzy, but how about something like this:

```
const allowTryItOutSelector = (): boolean: => false;

...

    allowTryItOutFor: (): => allowTryItOutSelector,
    // OR
    allowTryItOutFor: (): (() => boolean) => allowTryItOutSelector,
```

src/types/constants/actions/apply.ts
mjlumetta: yeah... i set a max identifier length of 32, so this breaks that rule if i add the "UE". so here are the options i can think of:

1. disable the max length rule with a comment here
2. rename it to something that spells out "VALUE" but removes "OAUTH", like `UPDATE_APPLY_APP_TYPE_VALUE` or `UPDATE_APPLY_CLIENT_TYPE_VALUE`. did this with the redirect URI; makes the OAuth part implied but does match the pattern.
3. leave as is

thoughts? will go with whichever you prefer.

src/types/swagger-client.d.ts
mjlumetta: good catch, thanks!

mjlumetta: also, context on this: interesting that `@typescript-eslint` doesn't have an extension for ESLint's `semi-spacing` rule 🤔  would have expected this to be caught but it's not because it's in an interface

src/containers/documentation/swaggerPlugins/index.tsx
alexandervalencia: Agreed on not adding more parens.

Would it be out of place to add a type? 
(name of type could change at your discretion of course)
```js
type predicate = () => boolean;

...

        allowTryItOutFor: (): predicate => ((): boolean => false),
```

I'm REALLY not hung up on this, I'd be okay with your suggestion as well (`(): (() => boolean) => allowTryItOutSelector,`)

src/types/constants/actions/apply.ts
alexandervalencia: I like option 2, but let's just leave it as is for now. 👍

src/types/swagger-client.d.ts
alexandervalencia: hmmm, yeah, I wonder why they don't have something that seems so simple

src/containers/documentation/swaggerPlugins/index.tsx
mjlumetta: ok! going with the `allowTryItOutSelector` route... mostly because i'm not sure i get `predicate` as the type name, and not sure what to call it? anyway, probably gonna move on hear, like you said.

alexandervalencia (dismissed):
Thanks for the updates! Approved! 🚀

mjlumetta:
cc @duganth-va for CodeBuild changes.

cicd/buildspec-ci.yml
```diff
@@ -159,10 +159,11 @@ phases:
           docsLink="https://github.com/${REPO}/blob/master/docs/testing.md#visual-regression-testing"
           comment="Visual regression testing failed. Review these diffs and then [update the snapshots](${docsLink}). <br><br> ${links}"
           ;;
-        unit|lint|security|e2e|accessibility)
-          if [[ $phase == "lint" ]]; then report=lint-results.xml; else report=test-report.xml; fi
-          aws s3 cp --no-progress ${report} s3://${S3_ERROR_BUCKET}/${S3_BRANCH}/${COMMIT_HASH}/
-          link="[report](https://console.amazonaws-us-gov.com/s3/buckets/${S3_ERROR_BUCKET}/${S3_BRANCH}/${COMMIT_HASH}/)"
+        lint)
+          comment='Linting failed. Run "npm run lint:ci" to see linting errors or look at the CodeBuild output.'
+        unit|security|e2e|accessibility)
```
duganth-va, line 164: 🚫  One minor syntax edit:
```
        lint)
          comment='Linting failed. Run "npm run lint:ci" to see linting errors or look at the CodeBuild output.'
          ;;
```

duganth-va (commented):
🎸  Thank you for removing all of the extra spaces at the end of my lines 😅  Need to invest into some yaml tooling. 
One small syntax change for the case statement and looks good. You really did have to go through a ton of files 💯

src/apiDefs/query.ts
```diff
@@ -26,10 +26,11 @@ const lookupApiByFragment = (apiKey: string): APIDescription | null => {
   const hasMatchingIdentifier = (apiDesc: APIDescription): boolean =>
     apiDesc.urlFragment === apiKey;
   const apiResult = getAllApis().find(hasMatchingIdentifier);
-  return apiResult || null;
+  return apiResult ?? null;
 };
 
-const lookupApiCategory = (categoryKey: string): APICategory | null => apiDefs[categoryKey] || null;
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- should check category existence
```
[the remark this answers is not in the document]
mjlumetta: yes. unfortunately, the compiler says that `apiDefs[categoryKey]` is always an `APICategory` object and never undefined... i'm not sure why that is. i could try to tweak the types so that's not the case if you'd like. if you can see anything different in terms of what's going on here or in terms of approaching it, i'd love to hear that. otherwise, though, seems like a bug in Typescript from what i'm seeing?

src/apiDefs/data/health.ts
```diff
@@ -24,7 +24,7 @@ import {
 } from '../../content/apiDocs/health';
 import { APIDescription } from '../schema';
 
-const swaggerHost: string = process.env.REACT_APP_VETSGOV_SECONDARY_SWAGGER_API || '';
+const swaggerHost: string = process.env.REACT_APP_VETSGOV_SECONDARY_SWAGGER_API ?? '';
```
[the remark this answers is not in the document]
gileswells: For the other guy in the room that can't get Google to give me an answer, what does `??` do?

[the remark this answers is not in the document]
mjlumetta: it's called the nullish coalescing operator (didn't know the name of until i googled it just now 😂 ). it resolves to the second operand _only_ if the first operand is null or undefined, as opposed to `||`, which resolves to the second operand if the first operand is falsy in any way. so, as an example, `0 ?? 100` evaluates to 0. 

- release notes for 3.7: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing 
- lint rule: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md

gileswells (approved):
LGTM

PR #526, opened 2020-10-30
API-2447 - Add testing for VersionSelect

alexandervalencia:
### Description
Tests added for the VersionSelect component using `react-testing-library` to match our new desired testing pattern.

### Process

- [ ] Docs updated

### Requested feedback

All feedback welcome


src/containers/documentation/swaggerPlugins/VersionSelect.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+describe('VersionSelect', () => {
+  beforeEach(() => {
+    updateVersionMock.mockReset();
+  });
+
+  it('renders successfully', () => {
+    const getSystem = mockGetSystem();
+    render(<VersionSelect getSystem={getSystem} />);
+
+    expect(screen.getByRole('button')).toHaveTextContent('Select');
```
mjlumetta, line 63: would also add that i think we can check that the select itself gets rendered with `screen.getByRole('combobox')`. would be nice to check for that as well. similarly, if we want to check for the aria-label/accessible name - which i think is frequently a good practice - we could also extend that to be `screen.getByRole('combobox', { name: 'Version Selection' })`.

src/containers/documentation/swaggerPlugins/VersionSelect.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  describe('when the button is clicked', (): void => {
+    it('fires Swagger UI updateVersion action', (): void => {
+      const getSystem = mockGetSystem(versionMetadata);
+      render(<VersionSelect getSystem={getSystem} />);
+
+      const select = screen.getByDisplayValue('1.0.0 - Current Version');
+      expect(select).toBeInTheDocument();
+
+      fireEvent.change(select, { target: { value: '0.0.1' } });
```
mjlumetta, line 86: should we use `userEvent.selectOptions()` here and in the other select change locations? https://github.com/testing-library/user-event#selectoptionselement-values

src/containers/documentation/swaggerPlugins/VersionSelect.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  it('renders successfully', () => {
+    const getSystem = mockGetSystem();
+    render(<VersionSelect getSystem={getSystem} />);
+
+    expect(screen.getByRole('button')).toHaveTextContent('Select');
+  });
+
+  it('has select options when metadata is not null', () => {
```
mjlumetta, line 66: this test name seems a bit different from the test content, or i had different expectations, at least. the test seems to be checking that the value of the select control responds appropriately to change events. i'd like to update the name for this test to reflect that, but i also think it'd be beneficial to add a new test that checks that the select options get rendered correctly. so could be something like `getByRole(select, 'option', { name: '1.0.0 - Current Version' })`, and then also checking that the value is '1.0.0', and the same thing for the other version.

src/containers/documentation/swaggerPlugins/VersionSelect.test.tsx
nwalker-va, line 63: 🎸  Overall, your PR looks great! Well done. My only question is about this test. I'm just wondering if there is anything more specific that can be looked for to make sure the VersionSelect rendered successfully? I haven't actually looked at the component, so I'm not sure. I just didn't know if looking for a select button might be too generic, but then again, this is a select, so...it's probably fine haha.

nwalker-va (approved):

src/containers/documentation/swaggerPlugins/VersionSelect.test.tsx
mjlumetta: should we also add a check that the current version and/or the Redux version is selected by default? could be in this test or a separate test.

src/containers/documentation/swaggerPlugins/VersionSelect.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const getSystem = mockGetSystem(versionMetadata);
+    const { getByDisplayValue } = render(<VersionSelect getSystem={getSystem} />);
+
+    const select = getByDisplayValue('1.0.0 - Current Version');
+    expect(select).toBeInTheDocument();
+
+    fireEvent.change(select, { target: { value: '0.0.1' } });
+
+    expect(getByDisplayValue('0.0.1 - Previous Version (Internal Only)')).toBeInTheDocument();
```
mjlumetta, line 75: could also add an assertion here that the version change handler does _not_ get called by changing the select without clicking the button

src/containers/documentation/swaggerPlugins/VersionSelect.test.tsx
alexandervalencia: Thanks Nicole! The entirety of the component is a select and a button.

VersionSelect

Although VersionSelect's parent won't render it if there's no options, technically the select can accept null instead of an Array of values for the options. So checking for the select to have options won't work, but a render without options will still contain the Select button.

mjlumetta (commented):
looks really good!! i don't have any issues with what's here, but i have some suggestions for additional assertions/tests that we could add.

1. check for the existence of the select element in the "renders successfully" test
2. check for the existence of the expected options explicitly
3. check that the initial version is the expected version
4. check that the version change handler does not get called if the button isn't clicked
5. check the value of the select initially and on change

these are all additive, so whatever call you want to make on whether or not they're valuable is alright with me.

the other thing i'd add, which is similarly not necessary but is something i've been leaning towards, is that i have a slight preference for `getByRole` where the target element in question has a distinctive role, like the select box. the reason for this preference is that it can have a somewhat stronger correlation with how users interact with the site; for example, in this case, the user browsing visually or with a screenreader will probably first identify that the select box is a control that they can interact with and then process the specific values available to them. this is a tiny, tiny detail, and it's also a bit of hypothesis on my part, but i figured i'd raise it for discussion.

src/containers/documentation/swaggerPlugins/VersionSelect.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+    updateVersionMock.mockReset();
+  });
+
+  it('renders successfully', () => {
+    expect(screen.getByRole('combobox', { name: 'Version Selection' })).toBeInTheDocument();
+    expect(screen.getByRole('button')).toBeInTheDocument();
+  });
+
+  it('has the options passed by the version metadata', () => {
+    const options = screen.getAllByRole('option') as HTMLOptionElement[];
```
mjlumetta, line 68: should we add a check that the length of the options array is 2, to be explicit? could be more explicit, but diminishing returns, etc. liking this test, would consider that extra

mjlumetta (approved):
this is nice! great job Alex, like how you've covered this component ✨

mjlumetta (approved):

PR #527, opened 2020-10-30
API-3125: Hook Migration MainNavItem, SelectedApis 

TS000:
### Description
Hook migration. Jira ticket [here](https://vajira.max.gov/browse/API-3125).

- [ NA ] Docs updated

### Requested feedback
All feedback welcome


src/components/mainNavItem/MainNavItem.tsx
```diff
@@ -21,39 +22,54 @@ interface MainNavItemProps {
...
+        </div>
+      )}
+    </>
+  );
+};
+
+MainNavItem.propTypes = {
+  excludeLargeScreen: PropTypes.bool,
+  excludeSmallScreen: PropTypes.bool,
+  onClick: PropTypes.func,
```
mjlumetta, line 66: should we add the other prop types here?

mjlumetta (approved):
good to go if it passes CI (validly, contact me if you run into timeout issues) 💯 ✨

src/components/mainNavItem/MainNavItem.tsx
TS000: I ran npm run lint:ci, and npm run lint locally and didnt find any errors. I'm barefoot so I can't check the Jenkin errors (no SOCKS). Any ideas on what could be wrong is greatly appreciated.

TS000: > should we add the other prop types here?

We could. I only added the default ones as without those it was not passing. Let me know if you think it's a good idea and I'll add the rest.

anthony-kopczyk: Hopefully helpful info below: (Not sure how much you've used prop types before, so 
 I apologize if you've already seen this)

When adding the properties, make sure to put `isRequired` on the non-optional fields. It's the opposite of typescript in that we have to add extra for required fields instead of adding the `?` for optional fields
> `excludeLargeScreen: PropTypes.bool.isRequired,`

The `apiDefs/schema.ts` file includes some definitions for proptypes using `shape` if you're interested. I think that will be useful when handling `largeScreenProps` since it's a more complicated object.

There's also `React.InferProps` that might be useful. I'm not sure how well it handles complex objects but worth a shot.

src/components/mainNavItem/MainNavItem.tsx
```diff
@@ -21,39 +22,54 @@ interface MainNavItemProps {
...
+          </NavLink>
+        </div>
+      )}
+      {!excludeSmallScreen && (
+        <div className={mobileOnly()}>
+          <NavLink onClick={onClick} {...sharedProps}>
+            {children}
+          </NavLink>
+        </div>
+      )}
```
anthony-kopczyk, line 58: I know this is the correct way to put these conditions, but something about it always makes me question if there's a better way. Maybe if we made more components (that included the div and nav link). That doesn't actually turn out much nicer because you still have to pass in all of the props. Some day react will have a new version that will make us refactor all of these. I know it! haha!

anthony-kopczyk (commented):
Similar comment on the propTypes, but other than this it looks good!

src/components/mainNavItem/MainNavItem.tsx
TS000: `excludeLargeScreen: PropTypes.bool.isRequired,` kicks an error when set to "false" in the default props.

src/components/mainNavItem/MainNavItem.tsx
mjlumetta: yeah, i think more expressive prop types would be awesome if we can have them.

fwiw, the big thing i learned here is that `propTypes` defines the props the component's consumer has to pass, versus the Typescript type of props, which is what the component has available to it internally. so props in `defaultProps` are not required in `propTypes` but are require in `MainNavItemProps`.

src/components/mainNavItem/MainNavItem.tsx
```diff
@@ -21,39 +22,61 @@ interface MainNavItemProps {
...
+    </>
+  );
+};
+
+MainNavItem.propTypes = {
+  activeClassName: PropTypes.string,
+  children: PropTypes.oneOfType([
```
mjlumetta, line 65: can this be `PropTypes.node`?

src/components/mainNavItem/MainNavItem.tsx
```diff
@@ -21,39 +22,61 @@ interface MainNavItemProps {
...
+    PropTypes.object,
+    PropTypes.string]).isRequired,
+  className: PropTypes.string.isRequired,
+  excludeLargeScreen: PropTypes.bool,
+  excludeSmallScreen: PropTypes.bool,
+  largeScreenProps: PropTypes.object.isRequired,
```
mjlumetta, line 71: should we try `PropTypes.shape` here?

src/components/mainNavItem/MainNavItem.tsx
```diff
@@ -21,39 +22,61 @@ interface MainNavItemProps {
...
+    </>
+  );
+};
+
+MainNavItem.propTypes = {
+  activeClassName: PropTypes.string,
+  children: PropTypes.oneOfType([
+    PropTypes.object,
+    PropTypes.string]).isRequired,
+  className: PropTypes.string.isRequired,
```
mjlumetta, line 68: shouldn't be required, i don't think

mjlumetta (commented):
miscellaneous minor comments on `propTypes`. nothing serious, ping me if you need an approval later

src/components/mainNavItem/MainNavItem.tsx
TS000: I tried shape and it kicked out errors.

src/components/mainNavItem/MainNavItem.tsx
TS000: @mjlumetta updated, let me know.

mjlumetta (approved):
✨

src/components/mainNavItem/MainNavItem.tsx
TS000: > ✨

Mind helping a barefooted dev and checkout why Jenkins is failing?

mjlumetta: Jenkins is in good shape, but CodeBuild is timing out on the e2e tests... this is a pretty big problem right now. we might want to circle up on the platform eng side and determine how we want to approach this right now (talked to Tom Dugan a little bit about it on Friday)

src/components/mainNavItem/MainNavItem.tsx
anthony-kopczyk: Hm. That's interesting. I wonder if it does a truthy/falsey check instead of checking for undefined when you use isRequired.

On a separate-ish note, looking here: https://reactjs.org/warnings/dont-call-proptypes.html:, they don't want us using PropTypes.shape like I have in that `schema.ts` file. So that's fun.

PR #528, opened 2020-10-30
Add Docker Login for CI pipelines

duganth-va:
### Description
Adds docker login steps for Jenkins and CodeBuild pipelines to alleviate concerns pending the anyonmous pull rate limit enforcement from Docker. 

If you're making non-content updates make sure you have read the development guide: https://github.com/department-of-veterans-affairs/developer-portal/blob/master/docs/development.md
-->

### Process

- [ ] Docs updated


mjlumetta (dismissed):
looks good 💯 only thing i might add would be that that this could benefit from a comment to explain why there is a login, since we're not pushing any images to the registry or pulling any private images. that's non-blocking, and other than that, good to go 🚀 ✨

cicd/buildspec-ci.yml
```diff
@@ -89,6 +91,8 @@ phases:
         EOF
           done
         fi
+        # Login to Docker Hub prior to pulling Dockerfile's base image node:12. This prevents rate limiting from Docker Hub. 
+      - echo "${DOCKER_PASSWORD}" | docker login --username ${DOCKER_USERNAME} --password-stdin
```
rtravitz, line 95: Are you echoing the password in because of some sort of `set +x` in Codebuild that would otherwise leak the interpolated value?

rtravitz (approved):
Just updating with another approval after the comment dismissed @mjlumetta's.

rtravitz:
It failed the e2e tests in Codebuild, which seems more like a timing issue than an indication of an actual failure, considering it passed in Jenkins. I haven't been paying as much attention to the frontend lately—is this something we're seeing consistently?

cicd/buildspec-ci.yml
duganth-va: 🗣️  Yeah good question. I think there are two things to note: 

1. CodeBuild will mask any value on output that was passed defined in the `parameter-store` section. If we were to use `docker login --username ${DOCKER_USERNAME} --password ${DOCKER_PASSWORD}` it would work and we wouldn't be concerned with displaying sensitive values.   

2.  We use `password-stdin` from Docker's recommendation. 
>To run the docker login command non-interactively, you can set the --password-stdin flag to provide a password through STDIN. Using STDIN prevents the password from ending up in the shell’s history, or log-files.

We have to echo the variable to pass it to to `password-stdin` for it to work properly.
https://docs.docker.com/engine/reference/commandline/login/

Hopefully that explanation checks out. 😅

duganth-va:
> It failed the e2e tests in Codebuild, which seems more like a timing issue than an indication of an actual failure, considering it passed in Jenkins. I haven't been paying as much attention to the frontend lately—is this something we're seeing consistently?

Seems to be happening ever since this commit https://github.com/department-of-veterans-affairs/developer-portal/commit/a8ff2bf15dfdae57aaaadb4cae124397c417e621

PR #529, opened 2020-10-30
API-3120: Hook Migration EmbeddedYoutubeVideo, MarkdownPage, Hero

TS000:
### Description
Moved components to folders and updated hooks. Jira ticket [here](https://vajira.max.gov/browse/API-3120).

- [ NA ] Docs updated

### Requested feedback
All feedback is welcome.


src/Routes.tsx
```diff
@@ -6,7 +6,7 @@ import { getDeactivatedFlags } from './apiDefs/deprecated';
 import { getEnvFlags } from './apiDefs/env';
 import { getApiCategoryOrder, getApiDefinitions } from './apiDefs/query';
 import { APIDescription } from './apiDefs/schema';
-import MarkdownPage from './components/MarkdownPage';
+import { MarkdownPage } from './components/markdownPage/MarkdownPage';
```
mjlumetta, line 9:
```suggestion
import { MarkdownPage } from './components';
```

mjlumetta (commented):
one carry-over comment but this is in good shape!

src/Routes.tsx
mjlumetta: also: i think another option would be to inline `MarkdownPage` in this file if you'd prefer. it's only used here, and it's like 3 tags deep. could be nice to reduce the size of `src/components` by one directory.

mjlumetta (dismissed):

anthony-kopczyk (approved):
Looks good to me!

mjlumetta:
@TS000 can this one close? no diff with the base branch, looks like all those commits were in #531

TS000:
> @TS000 can this one close? no diff with the base branch, looks like all those commits were in #531

I need some reviews. Other than that I think it should be gtg.

PR #530, opened 2020-11-02
[DRAFT] Fetch wrapper proposal.

johanna-adhoc:
**Proposal of implementation for Story API-2900**
A wrapper for the native fetch that easily allows to include common headers and handle common generic errors.


TS000 (approved):

src/utils/callFetch.ts
```diff
@@ -0,0 +1,97 @@
...
+  responseType:string;
+}
+
+export interface HttpResponse {
+  ok: boolean;
+  status: number;
+  body: any | string | null;
+}
+
+export const responseType = {
```
mjlumetta, line 17: should we try exploring Typescript enums for this kind of stuff? i've never used them before and have no idea how easy or not they are

src/utils/callFetch.ts
```diff
@@ -0,0 +1,97 @@
+import * as Sentry from '@sentry/browser';
+import { v4 as uuidv4 } from 'uuid';
+
+/**
+ * Interfaces & Constants
```
mjlumetta, line 5: should we define these types somewhere in `src/types`?

src/utils/callFetch.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+interface CallFetchConfig {
+  responseType:string;
+}
+
+export interface HttpResponse {
+  ok: boolean;
+  status: number;
+  body: any | string | null;
```
mjlumetta, line 14: can this be made a generic type on `callFetch`? besides getting rid of the `any`, it could help encourage callers to explicitly and clearly define their types so that this module does all the type handling based on an unknown response.

mjlumetta (commented):
this looks really amazing, Johanna! have some miscellaneous questions about tweaks to make, but this looks so good. like that we're incorporating a UUID to associate requests.

PR #531, opened 2020-11-02
API-3126:  Hook Migration:  QuickStartWrapper and TestingNotice

TS000:
### Description
Hook migration as well as reusable component folder transfer.
Jira ticket: [here](https://vajira.max.gov/browse/API-3126)

### Process
- [ NA ] Docs updated

### Requested feedback
All feedback welcome 👋


src/components/TestingNotice.tsx
```diff
@@ -5,7 +5,7 @@ import * as React from 'react';
 
 import './TestingNotice.scss';
 
-const TestingNotice: React.FunctionComponent = () => (
+const TestingNotice: React.FunctionComponent = ():JSX.Element => (
```
mjlumetta, line 8: probably want a space here, i think this will be a linting error as of #525. if it doesn't fail the current linting, though, your call

mjlumetta (approved):
looks good to me. appears to have a few commits from #529, which it would be good to get rid of and/or get merged into master first, but otherwise everything is good to go!

mjlumetta (approved):
looks great! ✨

PR #533, opened 2020-11-04
Update Apply and Contact Us URLs to use new /internal/developer-portal/public/* routes

mjlumetta:
### Description

changes Apply and Contact Us URLs to the new routes for [API-2787](https://vajira.max.gov/browse/API-2787). see the preceding PRs for context:

- backend app routes: https://github.com/department-of-veterans-affairs/developer-portal-backend/pull/105 
- gateway updates: https://github.com/department-of-veterans-affairs/devops/pull/7913

to test, go through the Apply and Contact Us flows in the portal. if both flows succeed, the changes are working. to test completely, try on all 3 review instances for dev/prod/staging, or point your local to those hosts and test each.

### Process

  - see Description for manual testing instructions
- [n/a] Docs updated


TS000 (approved):
LGTM 🤖

gileswells:
@mjlumetta My only question is whether or not the back end is setup to accept form submissions on both urls during the changeover. Do we have a 2 step process on this so that if one build goes out before the other they will continue to work? Even if both deploy at the same time, users on the portal would still post to the old url while they are navigating around the app since the new url would only show up after a full page reload.

mjlumetta:
@gileswells i think the answer to your question is yes, we do have a 2 step process. both versions are available concurrently as of now (although we still need to deploy to prod in the backend, which has to happen before this PR gets merged). but yeah, the idea is that we get these changes out with both versions available on the gateway and in the backend server, and then we'll remove them from the gateway, and then from the backend application itself. 

it's a good point that we should probably not take those routes down immediately after we deploy this update; i hadn't thought about people having an older version already loaded and cached in their browser window. so we should probably leave a window of a few days between deploying these changes and applying the changes to take down the old routes from the gateway. i don't know, what do you think is a good minimum for that window?

gileswells:
@mjlumetta a day? Shouldn't need to be too much. A couple hours would be enough honestly.

gileswells (approved):
LGTM

PR #536, opened 2020-11-10
Create feature flag for new oauth docs

gileswells:
### Description
https://vajira.max.gov/browse/API-3405
This adds a new environment config of `REACT_APP_AUTH_DOCS_V2` to the app and with it allows devs to opt in by setting that to true in `.env.local` which will add a new Authorization link to the left nav for the rework of the auth flow.

### Process
- [ ] Docs updated

### Requested feedback
Any and all comments welcome. This is just meant to be a jumping off point and some things are expected to not be fully functional as they're covered by other tickets.


src/containers/documentation/DocumentationRoot.tsx
```diff
@@ -92,6 +95,7 @@ const ExploreSideNav = (): JSX.Element => {
               )}
               {categoryKey !== 'benefits' &&
                 apiCategory.apis.some(api => !!api.oAuth) &&
+                !getFlags()[FLAG_AUTH_DOCS_V2] &&
```
mjlumetta, line 98: fun fact: the most recent version of `flag` has hooks that would be good for this! specifically, in this case, we can use `useFlag([FLAG_AUTH_DOCS_V2])`. we'd have to export `useFlag` from `flags.ts`, because it's returned by `createFlags`, but seems like a plus to use. https://www.npmjs.com/package/flag#useflag 

it'd probably be beneficial to slowly replace our uses of `getFlags` directly in our React tree with `useFlag` or `useFlags` in the long run, although it's fine for now. but definitely i would think `useFlag` is probably better for a single flag.

src/types/constants/index.ts
```diff
@@ -23,3 +23,4 @@ export const APPLY_FIELDS_TO_URL_FRAGMENTS = {
 export const APPLY_STANDARD_APIS = ['benefits', 'facilities', 'vaForms', 'confirmation'];
 export const APPLY_OAUTH_APIS = ['claims', 'communityCare', 'health', 'verification'];
 export const PAGE_HEADER_ID = 'page-header';
+export const FLAG_AUTH_DOCS_V2 = 'auth_docs_v2';
```
mjlumetta, line 26: slight preference for putting "flag" at the end of this name, like `AUTH_DOCS_V2_FLAG`

mjlumetta (commented):
looks good so far. i'd suggest going one step further in this PR, which would be adding the route to the site and creating a dummy component (maybe that's just an MDX component with an h1 in it or whatever). would like to see that much covered in this PR/ticket so that we can test out the navigating to a page part. everything else looks good, and it should be good to go once that's done. (had a couple small suggestions but they're not super significant and shouldn't necessarily be blocking.)

src/types/constants/index.ts
gileswells: I had gone with Flag at the front so that if we have future environment flag names we can have them all being alphabetically sorted via that prefix.

ediiotero (commented):
Looks good, just had a question. 
Should we update the README for the .env.local vars to include REACT_APP_AUTH_DOCS_V2?

src/containers/documentation/DocumentationRoot.tsx
anthony-kopczyk: I can't believe I missed that. I looked through the flag documentation while refactoring components to try and figure out how it worked. 🤦

gileswells:
> Should we update the README for the .env.local vars to include REACT_APP_AUTH_DOCS_V2?

I don't think we need to in this case. Without adding it you're still going to have a functional app. The ones mentioned in the README are critical to it's function on local vs this one that's optional and will default to the status quo of what's on prod currently.

mjlumetta:
oh, one other thing i thought of here was that if we have a stubbed out Auth page, we may want to stub out the tests for that as well. may or may not fit on this PR, will leave that to your judgment as you implement the stubbed page. they wouldn't have very much content at this stage, certainly.

src/containers/documentation/DocumentationRoot.tsx
```diff
@@ -131,7 +137,10 @@ const DocumentationRoot = (): JSX.Element => {
               {oldRouteToNew.map(routes => (
                 <Redirect key={routes.from} exact from={routes.from} to={routes.to} />
               ))}
-              {!shouldRouteCategory && <Redirect from="/explore/:apiCategoryKey" to="/explore" />}
+              <Route path="/explore/authorization" component={AuthorizationDocs} exact />
+              {!shouldRouteCategory && !authDocsV2 && (
```
gileswells, line 141: Fix this for when v2 flag is active
Also fix multiple matching routes when /explore/authorization is triggering the redirect on v2

src/containers/documentation/AuthorizationDocsLegacy.scss
```diff
@@ -0,0 +1,47 @@
+@import '../../styles/variables';
```
mjlumetta, line 1: is this a direct copy of `AuthorizationDocs.scss`? i was initially thinking that there wouldn't need to be separate SCSS files for the two components, but upon further reflection, i can see why as we develop the new page the styles will diverge and we'll need to separate them at some point. so i understand, can you share a little about your thinking behind pulling this file out? was that what you were trying to go for, and either way, are there other reasons?

src/flags.ts
```diff
@@ -13,7 +14,8 @@ export interface AppFlags {
   signups_enabled: boolean;
 }
 
-const { Flag, FlagsProvider } = createFlags<AppFlags>();
+// eslint-disable-next-line @typescript-eslint/unbound-method
```
mjlumetta, line 17: if you happen to know why the override is necessary here, might be nice to document in a comment. i can't see off the bat why that would be, though, and it matches the `flag` documentation, so it should be fine as far as i can see.

mjlumetta (commented):
the one outstanding thing i can see is that we may want to have the card links on the documentation overview landing page and on the category landing page. that is, if the feature flag is on, there should be an Authorization card on the top level, and there should not be an Authorization card on each of the category pages. otherwise, i'm ready to approve. also, don't think that's an absolute requirement since the production version still looks right, but would be good to get it in now.

src/containers/documentation/AuthorizationDocsLegacy.scss
gileswells: `.va-api-authorization-docs(-legacy)`

The two different files have a different wrapping class which was intentional. This way any changes to the new one `.va-api-authorization-docs` won't affect the old docs.

src/flags.ts
gileswells: I wasn't able to figure it out myself.

src/containers/documentation/AuthorizationDocsLegacy.scss
mjlumetta: okay, cool. that's generally what i was thinking, sounds good to me

ediiotero (approved):
Looks good! 👍

mjlumetta (approved):
great job, let's roll 💯 ✨

PR #537, opened 2020-11-10
API-3124: Move PageHeader and CodeWrapper to unique folders

anthony-kopczyk:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-3124

With these files moved to unique folders the components folder is
cleaned up some more, and we can import the two components from
'./components' since they no longer use default exports.

As a side effect of this change, because of how the mdx imports work, we
have to encapsulate the component names in both curly braces and quotes
so that when the imports from the mdx files are injected into the
generated js files, they have the correct value instead of saying:
import [object Object] from '../components';

### Process
- [N/A] Docs updated

### Requested feedback
I had to update the imports in the `mdx` files so that the keys would get placed properly in the generated js files. Without encapsulating in curly braces, they fail because with our new component structure we no longer use default exports. In addition, I had to wrap them in quotes. Otherwise the generated file get `import [object Object] from 'file';` instead of the actual import key (`import { Component } from 'file';`). That's no so much a feedback request as a heads up.


src/content/apiDocs/oauthTechnical.mdx
```diff
@@ -1,8 +1,9 @@
 ---
 imports:
-  CodeWrapper: ../../components/CodeWrapper.tsx
+  '{ CodeWrapper }': ../../components
```
TS000, line 3: This is interesting. I've never had to worked with a mdx file like this. Good to know!

TS000 (dismissed):
Looks great 👍

src/content/apiDocs/oauthTechnical.mdx
mjlumetta, line 3: the thing that i don't love about this is that it's requiring you to keep Javascript and YAML string mechanics in your mind at the same time... but on the other hand, cool that you can do this 😄

mjlumetta (approved):

src/content/apiDocs/oauthTechnical.mdx
anthony-kopczyk: 100%. It looks like we could change this to a regular import statement, but I didn't want to make extra changes. If that's something we want to do/look into, I can make a jira ticket. That way the import looks more familiar/easier to understand.

mjlumetta: what do you mean by "regular import statement", in this case? i'm not too opinionated about it either way, more curious what you had in mind bc i'm not familiar. if you feel like making the ticket would be beneficial, go for it, but i'm good with this as is.

anthony-kopczyk: I think we can put normal typescript code in.

`import { CodeWrapper } from '../../components';`

mjlumetta: directly in there? i wouldn't think we'd be able to do that because it would treat it as JSX... but i'm not sure. anyway

mjlumetta (approved):

PR #538, opened 2020-11-10
Fix bug in CurlForm from treating shallow-merged React state as if it were deep-merged

mjlumetta:
### Description

fixing a bug in the `CurlForm`. https://lighthouseva.slack.com/archives/CDNA4T352/p1605024572060900

### Process

  - ran `CurlForm.test.tsx`
  - part of the problem is that the `CurlForm` unit tests are disabled temporarily because they are very cost-intensive right now. need to refactor to be much more efficient, will create/flag this ticket after this PR gets merged.
- [n/a] Docs updated
  - bug would definitely qualify as an accessibility issue, fixed


gileswells (approved):
LGTM

PR #539, opened 2020-11-10
Api 3123: React redux hooks for OAuthAppInfo and DeveloperInfo

anthony-kopczyk:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-3123

This ticket was to convert the two components to use react redux hooks.
- OAuthAppInfo
- DeveloperInfo

### Process
- [N/A] Docs updated

### Requested feedback
I don't think there's anything specific that I have to request for feedback. Let me know how it looks!


src/containers/apply/DeveloperInfo.tsx
```diff
@@ -1,85 +1,68 @@
...
-
 type DeveloperInfoDispatch = ThunkDispatch<RootState, undefined, actions.UpdateApplicationAction>;
 
-const mapDispatchToProps = (dispatch: DeveloperInfoDispatch): DeveloperInfoDispatchProps => ({
-  updateEmail: (oldValidation?: string) => (value: ErrorableInput): void => {
+const DeveloperInfo = (): JSX.Element => {
+  const dispatch = useDispatch<DeveloperInfoDispatch>();
+
+  const email = useSelector((state: RootState) => state.application.inputs.email);
+  const updateEmail = (oldValidation?: string) => (value: ErrorableInput): void => {
```
mjlumetta, line 15: i _think_ we can take this closure out of here because the value of `oldValidation` is already in scope here. functionally, it's the current validation on the email from line 14. so we can do 

```
const email = useSelector((state: RootState) => state.application.inputs.email);
const updateEmail = (value: ErrorableInput): void => {
  dispatch(actions.updateApplicationEmail(value, email.validation));
};
```

what do you think? this sort of indirection is not great in the first place, so if you have any thoughts on the bigger picture, they're definitely welcome. you may have already seen it, but there's a length comment in the Apply actions file explaining the context for this.

src/containers/apply/OAuthAppInfo.tsx
```diff
@@ -1,44 +1,29 @@
...
-    dispatch(actions.updateApplyOAuthRedirectURI(value, oldValidation));
-  },
-});
+  };
 
-const OAuthAppInfo: React.FunctionComponent<OAuthAppInfoProps> = (props: OAuthAppInfoProps) => {
-  const { oAuthApplicationType, oAuthRedirectURI } = props;
+  const oAuthRedirectURI = useSelector((state: RootState) => state.application.inputs.oAuthRedirectURI);
+  const updateOAuthRedirectURI = (oldValidation?: string) => (value: ErrorableInput): void => {
+    dispatch(actions.updateApplyOAuthRedirectURI(value, oldValidation));
```
mjlumetta, line 22: same thing here as with email in `DeveloperInfo`

mjlumetta (approved):
looks good! had 1 small suggestion in 2 places

src/containers/apply/DeveloperInfo.tsx
anthony-kopczyk: Should be addressed in the latest commit. Tested it out. Appears to work fine.

src/containers/apply/OAuthAppInfo.tsx
anthony-kopczyk: Should be addressed in the latest commit. Tested it out. Appears to work fine.

mjlumetta (approved):

TS000 (approved):
LGTM 🏁

PR #541, opened 2020-11-11
API - 2988:  Update Accessibility Props for Hover Image

TS000:
### Description
[Jira ticket](https://vajira.max.gov/browse/API-2988)
Added alt tag props to dynamic image. Added new prop portal to News. 

### Process
Test for HoverImage now includes the proper props check with addition of the new alt tag prop.
- [✔️ ] Tests added or updated for change
- [  n/a ] Docs updated
- [ ✔️ ] Accessibility concerns have been considered and addressed

### Requested feedback
There are some greater accessibility things that AXE caught but I wanted to keep within the bounds of the ticket.


src/components/hoverImage/HoverImage.tsx
```diff
@@ -1,20 +1,27 @@
...
 
+  const setProps = (): void => {
+    setAlt(props.imageAlt);
+    setSrc(props.hoverImagePath);
+  };
+
   return (
     <img
-      alt=""
+      alt={alt}
```
mjlumetta, line 20: can you explain why we're making this change? i would've thought that this alt attribute was empty on purpose to mark the image as presentational. the reason i could see it being necessary in context would be that it's nested within a link. however, the alt doesn't actually have any effect unless you also remove the `role="presentation"` attribute. so those are my questions. if we don't want to change this image being presentational, i wonder whether this ticket is necessary. haven't tested yet, will see if that gives me any insight.

i wonder whether we might also be open to getting a quick look from a designer on this one. i'm not sure how much good duplicating the link as an icon helps the user experience? anyway.

src/components/hoverImage/HoverImage.tsx
```diff
@@ -1,20 +1,27 @@
 import * as React from 'react';
 
 interface HoverImageProps {
+  imageAlt: string;
```
mjlumetta, line 4: maybe make this optional and default to the empty string if not provided, probably? having an empty alt should be a valid choice, but i also see why we might want to make the consumer pass it explicitly.

src/components/hoverImage/HoverImage.tsx
```diff
@@ -1,20 +1,27 @@
...
   imagePath: string;
   hoverImagePath: string;
 }
 
 const HoverImage = (props: HoverImageProps): JSX.Element => {
+  const [alt, setAlt] = React.useState(props.imageAlt);
   const [src, setSrc] = React.useState(props.imagePath);
 
+  const setProps = (): void => {
+    setAlt(props.imageAlt);
```
mjlumetta, line 14: why are we changing the alt here?

mjlumetta (commented):
i have a question about the purpose of this PR. i was under the impression that the image was supposed to be presentational on purpose, in which case an empty alt would be the right call. but i could also see having an empty alt inside the link being problematic.

src/components/hoverImage/HoverImage.tsx
TS000: The reason for this change was to add an alt tag to the dynamic image. The ticket was made by @alexandervalencia he might have more to say.

src/components/hoverImage/HoverImage.tsx
TS000: [A11y](https://www.a11yproject.com/posts/2013-01-14-alt-text/). Essentially since this area changes we wanted a way to dynamically tell screen readers what the image is.

mjlumetta: i guess what i was trying to ask here was about whether the alt is actually dynamic. does the alt change on hover? there are two props for the image source, one for no hover and one for hover. there's one prop for the alt, so i'm wondering whether it would actually change the DOM on hover events or reset the same alt text.

to take a step back, though, does the image conceptually change on hover?

src/containers/News.tsx
```diff
@@ -64,6 +64,7 @@ const MediaItem = ({ item }: { item: NewsItemData }): JSX.Element => {
       <div aria-hidden>
         <a href={item.url} tabIndex={-1}>
           <HoverImage
+            imageAlt={item.title}
```
gileswells, line 67: As this image is purely for decoration and not something that includes any specific information about news item it decorates I would argue that the proper alt text here is an empty string because it doesn't add anything of consequence and doubling up on the title would just add frustration for screen reader users.

src/components/hoverImage/HoverImage.tsx
gileswells: I concur in allowing this to be optional and having it default to an empty string.

gileswells: Also personally I would prefer to see this property just named `alt` as opposed to the current `imageAlt` so that it uses the more expected prop name.

src/components/hoverImage/HoverImage.test.tsx
```diff
@@ -9,20 +9,21 @@ afterEach(cleanup);
 
 const props = {
   hoverImagePath: 'hover',
+  imageAlt: 'United States',
```
gileswells, line 12: Making `imageAlt` or `alt` optional would need an additional unit test to cover when the prop is omitted and defaults to an empty string.

gileswells: Also currently setting the alt text, while it's being done in this test, isn't actually being confirmed with an `expect()`

gileswells (changes requested):

src/containers/News.tsx
TS000: > doubling up on the title would just add frustration for screen reader users.

Is there a better option for the alt text since no alt text is provided?

src/components/hoverImage/HoverImage.test.tsx
TS000: > Also personally I would prefer to see this property just named alt as opposed to the current imageAlt so that it uses the more expected prop name.

With this direction would it also make more sense to change imagePath to url and hoverImagePath to hoverUrl?

src/components/hoverImage/HoverImage.tsx
TS000: > Also personally I would prefer to see this property just named alt as opposed to the current imageAlt so that it uses the more expected prop name.

With this direction would it also make more sense to change imagePath to url and hoverImagePath to hoverUrl?

src/components/hoverImage/HoverImage.test.tsx
TS000: made some updates to test

src/containers/News.tsx
TS000: I ended up setting the tag to props.alt and it seems to be picking up a tag 🚀

src/components/hoverImage/HoverImage.tsx
TS000: ended up making the alt tag directly updated my props rather than updating on hover

src/components/hoverImage/HoverImage.tsx
TS000: Naming has been changed and we are now checking for an empty string. I would like some eyes on my test however. I was confused on checking between a string and an empty string since they are both strings.

src/components/hoverImage/HoverImage.tsx
```diff
@@ -10,11 +11,11 @@ const HoverImage = (props: HoverImageProps): JSX.Element => {
 
   return (
     <img
-      alt=""
+      alt={props.alt ? props.alt : ''}
```
mjlumetta, line 14: what do you think about `props.alt ?? ''` for the syntax of this?

src/containers/News.tsx
```diff
@@ -28,6 +28,7 @@ interface NewsSection extends DataSection {
 }
 
 export interface NewsItemData {
+  alt: string;
```
mjlumetta, line 31: can we remove this, actually? this property doesn't exist in the `news.yml` schema

mjlumetta (commented):
so, think this looks good from the code perspective, but my question would be whether or not we want to make this change. because the news media video icons don't actually have alt text, it feels like we're solving for a future situation where the `HoverImage` component is used for something else that does need alt text. so i'm cool with making the change to cover that scenario, but i wanted to make sure we're all on board with making the change for that reason.

src/containers/News.tsx
TS000: Yes.

src/components/hoverImage/HoverImage.tsx
TS000: I agree it is a bit redundant.

TS000: It actually throws an error if I use your script. I think it's better to leave it as it was.

src/containers/News.tsx
TS000: If I remove it from the interface it's no longer available in the component and throws a type error.

TS000: Should I update the yml file associated with this?

alexandervalencia:
Hey @TS000 I was wondering what you thought about writing the component this way to use a default value as opposed to using a ternary to set the alt text. We would also set the prop to optional in the interface:
```js
interface HoverImageProps {
  alt?: string;
  imagePath: string;
  hoverImagePath: string;
}

const HoverImage = ({ alt = '', imagePath, hoverImagePath }: HoverImageProps): JSX.Element => {
  const [src, setSrc] = React.useState(imagePath);

  return (
    <img
      alt={alt}
      role="presentation"
      src={src}
      onMouseEnter={(): void => setSrc(hoverImagePath)}
      onMouseLeave={(): void => setSrc(imagePath)}
    />
  );
};
```

alexandervalencia:
This would solve the issue @mjlumetta was talking about with how the `News` component is currently *not* passing an `alt` prop to the `HoverImage` component

mjlumetta (commented):
still have the question about whether there should be functional changes to the site as a result of this PR. for me, the core question is whether the video icons on the News page should change from presentational to functional by way of having a non-empty alt. would like to understand the thinking behind that more clearly - maybe it's worth syncing up on this one?

src/components/hoverImage/HoverImage.tsx
```diff
@@ -1,20 +1,21 @@
...
 
-const HoverImage = (props: HoverImageProps): JSX.Element => {
-  const [src, setSrc] = React.useState(props.imagePath);
+const HoverImage = ({ alt = '', imagePath, hoverImagePath }: HoverImageProps): JSX.Element => {
+  const [src, setSrc] = React.useState(imagePath);
 
   return (
     <img
-      alt=""
-      role="presentation"
```
mjlumetta, line 14: i think it is appropriate to conditionally include the presentation role if the alt is empty. i don't think it matters at all in context because it's nested under that `aria-hidden` ancestor, but we might want it elsewhere. i'm okay with this as is given that it's not necessarily applicable at the moment, though.

mjlumetta (approved):
approved ✅ left a small non-blocking comment about the role question. will re-approve if you choose to make changes, but no objections if you want to merge as is

src/components/hoverImage/HoverImage.tsx
TS000: I'm into addressing it the need arises.

PR #543, opened 2020-11-11
API: 3074 - Uninstall connected-react-router as a Dependency from Dev Portal

TS000:
### Description
[Jira Ticket](https://vajira.max.gov/browse/API-3074)
Transitioned the Dev Portal away from the NPM package `connected-react-router
`
### Process
Updated both components using CRR as well as redux connections and type casts relating to this transitioning.
- [  n/a ] Docs updated

### Requested feedback
All feedback welcome.


mjlumetta (dismissed):
this all looks great to me 🥇 💅  only thing i'd ask here is about that last AC on the ticket. it'd be great to upgrade `history` to latest in this ticket as well bc `connected-react-router` was the reason it had to be v4. doesn't have to be this PR, but wanted to ask if you were planning on doing that now. can be a follow-up ticket too if you'd prefer.

TS000:
> this all looks great to me 🥇 💅 only thing i'd ask here is about that last AC on the ticket. it'd be great to upgrade `history` to latest in this ticket as well bc `connected-react-router` was the reason it had to be v4. doesn't have to be this PR, but wanted to ask if you were planning on doing that now. can be a follow-up ticket too if you'd prefer.

On it!

TS000:
Looks like upgrading History requires more than just a quick change in the package.json. There's a function that History 4 no longer exports in 5 so I'm going to request another ticket for a History upgrade.

mjlumetta (approved):
✨

PR #545, opened 2020-11-12
Remove "are you a vet" banner from the developer portal

ediiotero:
### Description
[API-209](https://vajira.max.gov/browse/API-209)
Removed the "Are you a vet" banner from the home page. 

Jest image snapshots have been updated


gileswells (approved):
#thatwaseasy LGTM

mjlumetta (approved):

PR #549, opened 2020-11-16
Use Router instead of BrowserRouter so that React Router uses our custom history

mjlumetta:
### Description

when we uninstalled `connected-react-router`, we accidentally introduced a bug in the Apply flow. for some reason, using separate history objects [within `BrowserRouter`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/BrowserRouter.js#L11) and in our app means that the site routing doesn't respond to programmatic changes from our history object. to fix this issue, we need to use `Router` and pass our history object as a prop: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/BrowserRouter.js#L30

going to deploy as soon as this gets merged.

### Process

  - we need integration tests for this, may try to add some but also need to get this change deployed
- [n/a] Docs updated

#### Manual Testing

to verify the changes, go through the Apply flow and verify that you land on the success page with your credentials. the old behavior is that it will never render the success page with the credentials.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
 
 import classNames from 'classnames';
-import { Route, BrowserRouter } from 'react-router-dom';
+import { Route, Router } from 'react-router-dom';
 import { Footer, Header, PageContent } from './components';
 import { FlagsProvider, getFlags } from './flags';
+import { history } from './store';
 
 import 'highlight.js/styles/atom-one-dark-reasonable.css';
 import './styles/base.scss';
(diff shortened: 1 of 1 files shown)


johanna-adhoc (approved):

anthony-kopczyk (approved):
Looks good to me.

As a follow up for maybe another ticket, Is there a way we can get rid of the custom history (no reason other than me thinking it would be nice to get out of the redux store so we'd only interface with it via hooks)?

gileswells (approved):
Regarding the comments on the slack thread the console warnings when running the unit test don't break the build and don't show up while using the site. Something to debug later but not pressing.

mjlumetta:
> As a follow up for maybe another ticket, Is there a way we can get rid of the custom history (no reason other than me thinking it would be nice to get out of the redux store so we'd only interface with it via hooks)?

@anthony-kopczyk i was wondering that too. can write that ticket, or if you'd like to, feel free. i think it's a good idea.

thanks to everyone for your speedy responses on this one! going to merge and deploy.

PR #550, opened 2020-11-18
API-3542 - Test Header component

alexandervalencia:
## Story [API-3542](https://vajira.max.gov/browse/API-3542)
Adds tests in accordance with our new pattern of using react-testing-library.

### Process
- [ ] Docs updated

### Requested feedback

If there are more tests cases you think would be helpful, please let me know. This is a first pass at new tests for this component.


nwalker-va (dismissed):
Looks good to me 👍 I don't really have anything to comment on.

src/components/header/Header.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+    );
+  });
+
+  it('should render the header', () => {
+    expect(screen.getByText(/Lighthouse/)).toBeInTheDocument();
+  });
+
+  describe('when the menu button is clicked', () => {
+    it('displays the mobile menu', () => {
+      const mobileNav = screen.getByRole('navigation');
```
mjlumetta, line 26: since it's the same nav element displayed responsively, maybe remove the "mobile" part from the name?

mjlumetta (approved):
lgtm!! going to throw out a few suggestions that we might add, but it's up to you whether or not you think they're significant enough to include. i don't have a lot of strong opinions.

1. could be good to spot-check the skip nav link because it's an important accessibility feature. to be fair, this might be more meaningfully covered in an integration test, so i think on that basis we could choose to leave it out here (that said, the integration test doesn't exist yet, so).
2. could be good to spot-check the "Request an API Key" button because it's such a crucial part of the site (the main way that consumers get to sandbox self-service).
3. veterans crisis line... i would lean towards no on testing this, normally, but i'm wondering if there are any legal or policy requirements that require us to display this info. if yes, is that worth a spot-check?

in any case, approving and good with merging as is. will re-approve if you decide to make any changes.

alexandervalencia:
@mjlumetta I made some changes you suggested, would appreciate another look at this. Thanks!

src/components/header/Header.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+        </Router>
+      </FlagsProvider>,
+    );
+  });
+
+  it('should render the header', () => {
+    expect(screen.getByText(/Lighthouse/)).toBeInTheDocument();
+    expect(screen.getByText(/Veterans Crisis Line/)).toBeInTheDocument();
+    expect(screen.getByText('Skip to main content')).toBeInTheDocument();
+    expect(screen.getAllByText('Request an API Key')[0]).toBeInTheDocument();
```
mjlumetta, line 24: would prefer (a) making these separate tests and (b) doing a getByRole + accessible name on the last two to specify that they're links, and checking the link destinations. still fine with approving as is so will leave it up to you what changes you'd like to make.

mjlumetta (approved):
suggested a couple tweaks to the new changes, but good with approving ✅  can re-approve if you make any other changes

src/components/header/Header.test.tsx
alexandervalencia: I guess I'm not sure why we would make them separate tests? None of these elements are being dynamically rendered. I could understand checking the links being separate but could you give me a little bit more explanation on each being separate? Thanks @mjlumetta !

mjlumetta: i guess my reasoning would be that i view these things as logically distinct enough that it'd be slightly more comprehensible to see separate pass/fails in the Jest output for each part of the header. so, for example, if the skip nav link breaks, i would prefer to still see that the apply link is working, or, conversely, that it's broken. 

given that, like you said, none of this content is dynamic and there are still a small number of things we're checking, this is a very small tradeoff, so i want to acknowledge that it is my preference and def not necessary if this way's easier for others to reason about. if that scenario happens, you fix the one thing, run it again, no problem. also, if breaking them out comes with any noticeable performance cost, then the current version seems preferable. so it's totally up to you.

alexandervalencia: Also, changes made, let me know what you think. Thanks!

mjlumetta (approved):
looks beautiful ✨

PR #551, opened 2020-11-18
API-1848: Fix Swagger Description Links for Font, Size, and Color

TS000:
### Description
Jira Ticket [here](https://vajira.max.gov/browse/API-1848).
Added a11y coloring to links in the swagger docs main scss file.

- [ na ] Tests added or updated for change
- [ na ] Docs updated
- [ na ] Accessibility concerns have been considered and addressed

### Requested feedback
Feedback welcomed.


src/styles/swagger.scss
```diff
@@ -18,8 +18,13 @@
     }
 
     a {
-      color: #4c565c;
+      color: #004795;
       transition: none;
+      font-size: 1.7rem;
```
anthony-kopczyk, line 23: I've never seen `rem` but looked them up. Neat!

src/styles/swagger.scss
```diff
@@ -18,8 +18,13 @@
...
+
+      &:visited {
+        color: #4c2c92
+      }
```
anthony-kopczyk, line 27: This is more general question for the Jira ticket. - I don't see these colors defined anywhere else, and so I'm having trouble understanding the basis of the ticket of making it consistent across the platform. Are these colors defined outside our project somewhere (in the vads themes?) that my 'Find All' isn't working to find any other uses of these colors? If not, I'm wondering if we're actually trying to 'un-override' default browser colors here.

These changes look good to accomplish the stated goal of the ticket.

@mjlumetta Saw your comment on the jira ticket, not sure if you might have answers ready to go?

anthony-kopczyk (approved):
This looks good to me to accomplish the goal of the jira ticket, regardless of the outcome of my semantics comment.

src/styles/swagger.scss
mjlumetta: yeah, so, i think the default styles do come from USWDS via VADS. here's [the style block](https://github.com/uswds/uswds/blob/v1.6.12/src/stylesheets/elements/_typography.scss#L17) for links in the USWDS package. i'm not sure what VADS actually exports, though, and i thought that it didn't export either of these colors. looking back at [the VADS color utility docs](https://design.va.gov/utilities/color), though, it looks like the regular color of #004795 is exported as `.vads-u-color--link-default`. that might be slightly preferable to have explicit, but it doesn't really matter too much imo.

the styles we are overriding are Swagger UI's. if you toggle off Tyler's changes in the browser dev tools, you can see that where those styles are coming from.

PR #552, opened 2020-11-18
API-3029: Set colors so contrast is at least 4.5:1

anthony-kopczyk:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-3029

This overrides the portion of our color theme in our markdown that's causing color contrast ratios not to be high enough. I looked through the apps some and did not find anywhere unexpected that this color might show up.

### Process

- [N/A] Docs updated

### Requested feedback
If anybody knows the app more thoroughly to know potential problem side-effect areas, I'd appreciate if you can check those. I verified the app as best I could, but I'm not sure if there are any pages I didn't think to verify.


TS000 (dismissed):
Looks great, not only is the new pink a help for a11y but it looks better with the green!

mjlumetta (changes requested):
what you've got here looks good to me, but i'm still seeing some color contrast errors on the auth page - fewer, but they're still there. also, one thing we should include in this PR is stripping out the override in the axe tests that disables color contrast checks; enabling those checks again should give us confidence the problem is fixed. https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/base.accessibility.ts#L17

---

**EDIT:** upon further investigation, it looks like they're false positives. turning the color contrast checks back on in the `test:accessibility` suite does give two of these false positives (as opposed to 5 in my browser extension). it appears like it's contrasting the `hljs-variable` class with the color `#d19a66` with something with a white background. (i think the `pf` element it's contained in? that doesn't make a ton of sense, though, because even though the `pf` element is the `hljs-variable` element's parent, it doesn't have a background color and has white text color. it's also not actually the background.) functionally, these are the two "<yourRedirectURL>" tokens. 

so i don't know... we don't want to disable the color contrast axe checks like this. at the same time, we need a way to indicate that this is a confirmed false positive. so... how do we do that

src/styles/atom-one-dark-reasonable-overrides.scss
```diff
@@ -0,0 +1,10 @@
+/*
+ * Overrdie $F92672 pink colors to meet accessibility requirements with the
+ * background color
+ */
+.hljs-keyword, .hljs-operator {
+  color: #F96080;
+}
+.hljs-pattern-match {
```
mjlumetta, line 8: any particular reason this one has a separate block?

anthony-kopczyk: In the theme it's defined as a separate block, so I left it the way they had it there.

anthony-kopczyk:
> so i don't know... we don't want to disable the color contrast axe checks like this. at the same time, we need a way to indicate that this is a confirmed false positive. so... how do we do that

I'll look into this some today to see if there's an approach and then if not, might make a separate jira ticket for it. On top of that I don't think my axe browser plugin is working properly, so I need to figure that out too.

anthony-kopczyk:
> what you've got here looks good to me, but i'm still seeing some color contrast errors on the auth page - fewer, but they're still there. also, one thing we should include in this PR is stripping out the override in the axe tests that disables color contrast checks; enabling those checks again should give us confidence the problem is fixed. https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/base.accessibility.ts#L17
> 
> **EDIT:** upon further investigation, it looks like they're false positives. turning the color contrast checks back on in the `test:accessibility` suite does give two of these false positives (as opposed to 5 in my browser extension). it appears like it's contrasting the `hljs-variable` class with the color `#d19a66` with something with a white background. (i think the `pf` element it's contained in? that doesn't make a ton of sense, though, because even though the `pf` element is the `hljs-variable` element's parent, it doesn't have a background color and has white text color. it's also not actually the background.) functionally, these are the two "" tokens.
> 
> so i don't know... we don't want to disable the color contrast axe checks like this. at the same time, we need a way to indicate that this is a confirmed false positive. so... how do we do that

I think the problem is with how the plugin detects foreground and background colors when overflow is set. If I set overflow to hidden in the CSS the errors go away when I run the accessibility tests. The plugin isn't scrolling or something and so I think it's checking with the background color outside of the actual code block. That explains why we can't find an issue with the theme css itself too.

Going to take a little bit more time to double check and to see if there's a solution. If there isn't a solution or if I can't find one by end of the day I'll put in a jira ticket explaining the issue so we can tackle it in the future. It's also possible we could add a css change for the accessibility test to hide overflow or something. I don't think that's a good solution because of potential side effects of missing other errors, but it would work temporarily.

anthony-kopczyk:
The issue mentioned in the previous comment (about it not scrolling and thus detecting the wrong background color) should be fixed in version 4.1.0 of axe-core. The latest version of axe-core that jest-axe is using is axe-core 4.0.1 right now. Once jest-axe is updated to use axe-core 4.1.0, we should be able to update jest-axe and that issue should be resolved.

https://github.com/dequelabs/axe-core/releases/tag/v4.1.0

Made a Jira Ticket: https://vajira.max.gov/browse/API-3737

mjlumetta:
@anthony-kopczyk asked about this problem in the Ad Hoc accessibility huddle and it appears like there's not a good programmatic solution to this problem, so we might have to leave the exception. think there are a couple of things we can/should do to tie off this issue and document those false positives, but don't put any effort into solving it if you were doing that. gonna circle back after i get out of this meeting.

src/base.accessibility.ts
```diff
@@ -14,9 +14,22 @@ describe('Accessibility tests', () => {
     await page.addScriptTag({ path: require.resolve('axe-core') });
 
     /**
-     * THIS IS BAD. code highlighting on the Authorization page has several color contrast
-     * issues. upgrading axe-core from 3.3.1 to 4.0.2 and jest-axe from 3.2.0 to 4.0.0 introduced
-     * stricter rules, which is good, but we can't resolve this issue without UX input.
+     * Code highlighting on the Authorization page has issue with colo contrast
```
mjlumetta, line 17: would suggest making explicit that the issue is a false positive in `axe-core` and not a real color contrast issue, if you're okay with making another commit. that's implied/explained by the rest of the comment, though, so non-blocking

mjlumetta (approved):

src/base.accessibility.ts
anthony-kopczyk: Added a note in a new commit

mjlumetta (approved):
thanks for bearing with me through the whole `axe-core` discussion!! and nice job on finding the bug fix in the release. glad the real issues are fixed ✨

PR #555, opened 2020-11-20
Use React Helmet to define unique, descriptive page titles for WCAG SC 2.4.2

mjlumetta:
### Description

for [API-2139](https://vajira.max.gov/browse/API-2139).

Addresses Success Criterion 2.4.2: Page Titled from WCAG 2.1. The specific failure we're solving is F25,
"Failure of SC 2.4.2 due to the title of a Web page not identifying the contents".

References:
- WCAG 2.1 Success Criterion 2.4.2: Page Titled: https://www.w3.org/WAI/WCAG21/Understanding/page-titled
- WCAG Failure 25: https://www.w3.org/WAI/WCAG21/Techniques/failures/F25
- React Helmet: https://github.com/nfl/react-helmet

page titles are based roughly on the page headers for each page, and I've tried to place the `Helmet` uses near the `PageHeader` component (or h1 in the rare cases we use it directly). would like feedback on the page titles themselves.

### Process

  - when there are integration tests, i think we can maybe test this there, but for now we should be fine to leave this as manually tested.
- [n/a] Docs updated
  - fixes an accessibility issue and i don't see any others getting introduced here

### Requested feedback

- are the page titles themselves good titles to use?
- I considered trying to make something generic that we could loop in, such as trying to include the `Helmet` inside the `PageHeader`, but ultimately decided against it because it makes it more complicated without providing enough benefit. sometimes the page title and header aren't going to be identical, so it's easier not to genericize at this point.


TS000 (approved):
LGTM ⛑️

ediiotero (approved):
This looks good, I'm ok with the titles you used.

PR #558, opened 2020-11-20
API-3543 - adds testing for PageContent component

alexandervalencia:
## Story [API-3543](https://vajira.max.gov/browse/API-3543)
Adds tests in accordance with our new pattern of using react-testing-library.

### Process
- [ ] Docs updated

### Requested feedback

If there are more tests cases you think would be helpful, please let me know. This is a first pass at new tests for this component.


src/components/pageContent/PageContent.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    expect(screen.getByRole('main')).toBeInTheDocument();
+  });
+
+  it('loads new page content after navigating to a different route', async () => {
+    expect(screen.getByText(/A Veteran-centered API platform/)).toBeInTheDocument();
+
+    const documentationLink = screen.getByText('Read our documentation');
+    userEvent.click(documentationLink);
+
+    const documentationPageDescription = await screen.findByText('Documentation');
```
mjlumetta, line 35: what do you think about maybe replacing `findByText('Documentation')` with `findByRole('heading', { name: 'Documentation' })`, in this test and the following one? doesn't look like there are any "Documentation" strings on the home page now, but this seems like a test condition that could be sensitive to changes in the page content. might also be beneficial to do a similar thing on the "Read our documentation" link, although i think that's far less likely to be an issue.

mjlumetta (commented):
looks good! like what you've got here. i can think of a few more cases that might be valuable to test here if you think they're worthwhile: 

1. on navigation to an anchor link in the same page, there's no scrolling and the target element has focus (scrolling being the more important part but we don't want to accidentally focus main)
2. on navigation to an anchor link in a different page, there's no scrolling and the target element has focus (same considerations as 1)
3. on navigation, the main element gets focused

lmk! would be fine with merging as is but want to raise those for consideration.

src/components/pageContent/PageContent.test.tsx
alexandervalencia: That makes sense to me!

rtravitz (commented):
Ready to go other than addressing @mjlumetta's suggestion!

alexandervalencia:
@mjlumetta @rtravitz I updated the tests to use `getByRole` and also adjusted to use the Apply page instead of Documentation page. I felt like it was a bit clearer and that the targets that I used ("Request an API Key") would be less likely to change then the documentation link (though that was a minor part of the decision).

mjlumetta (approved):
lgtm ✨

PR #561, opened 2020-11-30
Add ARIA label to tabs on FHIR documentation page

mjlumetta:
### Description

The [ARIA guidance](https://www.w3.org/TR/wai-aria-practices/#tabpanel) for tab components says the following: "If the tab list has a visible label, the element with role tablist has aria-labelledby set to a value that refers to the labelling element. Otherwise, the tablist element has a label provided by aria-label." We did not previously implement this, so the screen reader would read out the first tab without any context. Now VoiceOver reads "Veterans Health API (FHIR) implementations, tab group" before entering the tab list.

Loosely tied to [API-1845](https://vajira.max.gov/browse/API-1845). Not really what the ticket was originally about, but since that issue got fixed incidentally, I played around to verify some things about the accessibility of said tabs.

I also upgraded `react-tabs` while I was playing around with this, and I added some detail to the PR template for docs + accessibility.

### Process

- [X] Docs updated
  - added some context here: https://community.max.gov/display/VAExternal/Component+Design


TS000 (approved):
Nice additions!

gileswells (approved):
LGTM

PR #562, opened 2020-12-01
Api 3786: Use semantic tags on main page

anthony-kopczyk:
### Description
This switches some div tags that should be p tags.

Jira Ticket: https://vajira.max.gov/browse/API-3786

### Process
- [N/A] Docs updated

### Requested feedback
I'd like general feedback on the changes I made. - whether I should change anything visually or if I should change css, etc.

Visually the "Here's how you know" section looks identical, but I had to add css classes to those paragraphs to remove margins. I could take those css classes out and let the paragraphs be styled semi-naturally, but I didn't think that looked very good.

The main page paragraphs do look different (that's why visual regression tests are currently failing). I'd like feedback on how those looks. With the old version, there is a manual line break entered on the page. Changing to paragraphs that line break doesn't make much sense. The easiest way to fix that is letting the paragraphs have their default margins. Let me know if this looks good. If I can get buy-in I'll update the visual regression tests so they match the new main page.


src/components/banner/Banner.tsx
```diff
@@ -37,8 +37,9 @@ const GuidanceBox = (props: GuidanceBoxProps): JSX.Element => (
       alt={props.icon_alt ?? undefined}
     />
     <div className="vads-u-margin-bottom--1p5">
-      <div><strong>{props.title}</strong>
-      </div>
+      <p className="vads-u-font-weight--bold vads-u-margin--0">
```
mjlumetta, line 40: any particular reason this is a paragraph? my knowledge of where it is semantically correct to use `<p>` versus you don't necessarily have to when it gets down to these smaller snippets is not very clear at all. if you have any insight into the standards there, would love to learn more.

mjlumetta (dismissed):
lgtm!!

i do have one question about the purpose of the ticket: "As a user navigating with a screen reader, I want paragraphs to use the html semantic p tag, so that I know the purpose of the content." is this supposed to make a functional change that the user will notice? i tested with VoiceOver on Mac and didn't see any difference; it regards both as text elements. it's possible that's not the case for JAWS or NVDA and what have you, of course. wondering if someone had been testing this and noticed it could be improved and, if so, what the environment was. with that said, this is more semantically correct regardless, so i think it's unequivocal that it is an improvement.

src/components/banner/Banner.tsx
anthony-kopczyk: I worked with JAWS a few years ago. The guy I learned it from used it primarily by navigating jumping through headings, paragraph tags, list tags, etc - whereas voiceover seems to want you to go through each element using the arrow keys. Since the text is in a div, I think JAWS users would miss it at first and need to go back. I think the most similar way to how you can use JAWS in the voiceover rotor. That said, JAWS also includes a next element like VoiceOver, so maybe not?

In regard to the small snippets, I didn't think about that for the title props. I think for users navigating by html landmarks/p/heading tags, having it as a p tag will be better than having it as a div. I'm interested in thoughts from other devs as well because I'm really not certain.

Hmm.. I just found some apple shortcuts that allow you to navigate via 'next heading' and such, which looks really similar to JAWS. Going to play around with that some. Looks like navigating to plain text doesn't care if it's a div or anything.

**Convenience Links**
- Voiceover Rotor: https://support.apple.com/en-us/HT204783
- JAWS Shortcuts/Commands: https://www.freedomscientific.com/SurfsUp/Quick_Keys.htm
- Apple Voiceover Shortcuts: https://www.apple.com/voiceover/info/guide/_1131.html

anthony-kopczyk: Also, thank you for your comment so much! That got me looking for shortcuts for voiceover. Voiceover felt really awkward to use but with those other shortcuts, it's so much better for me!

mjlumetta: thanks for all this detail Anthony!! this is so helpful. going to chalk up my initial impulse to being a novice with VoiceOver, or any screen reader 😂 this is a good reminder to use more of that functionality, ty

TS000 (dismissed):
Looks great 🥇

johanna-adhoc (changes requested):
I noticed the prop name being "title" if the string displayed here is a title wouldn't the div tag be replaced by a heading? (not sure about the heading level though)

src/components/banner/Banner.tsx
mjlumetta: what do y'all think about using a different noun for this text besides "title"? i think that making it a true heading is awkward for the reasons Anthony raised, that it doesn't really that way in the document flow. maybe "lead"?

johanna-adhoc: I don't think the issue is the name of the prop, the text is actually performing as a title for the description written below. 
The best approach I see is wrapping the title in an h1, BUT also replacing the outer div tag for an article tag. By doing this the h1 heading of the article is ignore for the main content hierarchy content so it will not introduce any issues. 
_the <h1>s inside the <article> elements “become” logical <h2>s, because <article>, like <section>, is a sectioning element._

This article has info about it!
https://www.smashingmagazine.com/2020/01/html5-article-section/

mjlumetta: oh wow that's really cool!! yeah, if that tests well, then that sounds great to me. one question that i still have is does wrapping something in an article at the top level with an h1 - so still a logical h1 to assistive technologies - violate the "one h1 per page" guideline? in any case, thanks for sharing that, that's super informative 🙏

johanna-adhoc (approved):
Approving this PR. Issues with the GuidandeBox component will be addressed with another story.

johanna-adhoc (approved):
LGTM

PR #569, opened 2020-12-04
API-3736/support contact us form fetch migration

johanna-adhoc:
### [API-3736](https://vajira.max.gov/browse/API-3736)
Fetch logic migrated to use the wrapper. This component no longer needs to handle integration with Sentry.

### Process
- [N/A] Docs updated
  
### DIFF
--- a/src/containers/support/SupportContactUsForm.test.tsx
+++ b/src/containers/support/SupportContactUsForm.test.tsx
@@ -1,4 +1,3 @@
-// import * as Sentry from '@sentry/browser';
 import '@testing-library/jest-dom/extend-expect';
 import { fireEvent, render, waitFor } from '@testing-library/react';
 import { mount } from 'enzyme';
@@ -10,6 +9,11 @@ import * as React from 'react';
 
 import { CONTACT_US_URL } from '../../types/constants';
 import SupportContactUsForm from './SupportContactUsForm';
+
+jest.mock('uuid', () => ({
+  __esModule: true,
+  v4: jest.fn(() => '123'),
+}));
 
 const server = setupServer(
   rest.post(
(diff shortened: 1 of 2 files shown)


TS000 (approved):
Looks great, nice to see the wrapper in action!

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -124,37 +124,20 @@ const SupportContactUsForm = (props: SupportContactUsFormProps): JSX.Element =>
...
-        }
-        return response;
-      })
-      .then(response => response.json())
-      .then((json: { errors?: string[] }) => {
-        if (json.errors) {
-          throw Error(`Contact Us Form validation errors: ${json.errors.join(', ')}`);
-        }
-      })
+    }, { responseType: ResponseType.TEXT }).then(() => {
```
anthony-kopczyk, line 135: Should we be using ResponseType.JSON here since the accept header indicates json should be returned?

src/containers/support/SupportContactUsForm.tsx
```diff
@@ -124,37 +124,20 @@ const SupportContactUsForm = (props: SupportContactUsFormProps): JSX.Element =>
...
+      // do nothing and return void
+    })
       .catch(error => {
-        Sentry.withScope(scope => {
-          scope.setLevel(Sentry.Severity.fromString('warning'));
-          Sentry.captureException(error);
-        });
+        throw error;
```
anthony-kopczyk, line 139: Just to confirm, this error can be either the returned error response from sentry or a generic error. Is there any kind of special handling we want to do here, or is throwing this enough? I'm guessing there isn't anything to do, since we've already logged to sentry for network errors?

src/containers/support/SupportContactUsForm.test.tsx
```diff
@@ -78,55 +82,6 @@ describe('SupportContactUsForm', () => {
...
-    expect(Sentry.Severity.fromString).toHaveBeenCalledWith('warning');
-
-    expect(Sentry.captureException).toHaveBeenCalledWith(
-      Error('Contact Us Form validation errors: unknown error'),
-    );
-
-    captureException.mockRestore();
-    fromString.mockRestore();
-    scope.mockRestore();
-  });
```
anthony-kopczyk, line 128: Woo! Being able to remove now unneeded tests for the win!

src/containers/support/SupportContactUsForm.test.tsx
```diff
@@ -11,6 +10,11 @@ import * as React from 'react';
 import { CONTACT_US_URL } from '../../types/constants';
 import SupportContactUsForm from './SupportContactUsForm';
 
+jest.mock('uuid', () => ({
+  __esModule: true,
+  v4: jest.fn(() => '123'),
+}));
```
anthony-kopczyk, line 16: If '123' is pulled out into a separate variable that might make it more clear around line 72 that this is here in order to get a predictable request-id in the test.

anthony-kopczyk: Had another thought: Is this something we'll need commonly in tests as we transition components over to use `makeRequest`? This might be a good target to move to a test util in the future. Nothing to change now, just a thought.

anthony-kopczyk (commented):
I think this is good to merge, especially with Tyler's approval. I have a couple comments here and there but nothing major, most of it asking questions since I'm not that familiar with this component.

src/containers/support/SupportContactUsForm.tsx
johanna-adhoc: no, the actual server doesn't return a JSON but text... maybe the headers should be the ones changed but this is how it was coded originally so I left it like it was.

src/containers/support/SupportContactUsForm.tsx
johanna-adhoc: The handling of errors are now happening in the wrapper and that is why sentry migration is not needed in individual components anymore. However this response is used in another component that makes changes in the UI so the throw here is happening so that other "Form" component gets to know the right text to present in the UI.

src/containers/support/SupportContactUsForm.test.tsx
johanna-adhoc: so moving forward the wrapper test should be the only one checking for sentry error log integration. I would say for individual components we can either check that we are calling the wrapper with the correct information or we can test if we are calling the native fetch with the right information.

If the component is adding extra specific logs that are not already handled by the wrapper, only then we would need to check sentry integration in the component's test file.

anthony-kopczyk (approved):
Looks good. The comments I had have all been addressed.

PR #573, opened 2020-12-08
API-3757: Test for Hero Component

TS000:
There isn't any logic in the hero component so other than testing that the component mounts I tested the img src.

- [ ☑️ ] Tests added or updated for change
- [ ] Docs updated
  
### Requested feedback

If anyone has any other ideas for a test on this component I'm all ears👂👂.


src/components/hero/Hero.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+import 'jest';
+import '@testing-library/jest-dom/extend-expect';
+import {  render } from '@testing-library/react';
+import { BrowserRouter as Router } from 'react-router-dom';
+import * as logoImage from '../../__mocks__/fakeImage';
+import { Hero } from './Hero';
+
+describe('Hero', () => {
+  it('checks image source', () => {
+    const { queryByRole } = render(<Router><Hero  /></Router>);
```
anthony-kopczyk, line 11: It looks like there's an extra space on the Hero tag

anthony-kopczyk (dismissed):
Can we add a test to check for the 'Request an API key' link? Maybe one for the heading text too, but I'm not sure on that one.

src/components/hero/Hero.test.tsx
anthony-kopczyk: Oh. When I was looking at react testing library things, I saw that you can use 'screen' instead of retrieving { getByRole } from render. I think screen is essentially an alias for the returned object. I'm not hardset on doing it one way or the other, but I think the testing-library way is to use screen. Feels odd using a global like that though.

TS000: Thanks! Added the more precise "screen" query provided by Testing-Library and fixed the extra space.

TS000:
> Can we add a test to check for the 'Request an API key' link? Maybe one for the heading text too, but I'm not sure on that one.

I've added a test for the api url. I didnt think the Heading text needed a test since it's something that could be changed regularly.

src/components/hero/Hero.tsx
```diff
@@ -37,6 +37,7 @@ const Hero: React.FunctionComponent = (): JSX.Element => (
             id="hero-request-key"
             to="/apply"
             className="usa-button vads-u-width--full medium-screen:vads-u-width--auto"
+            role="link"
```
mjlumetta, line 40: out of curiosity, what was the motivation for adding the role explicitly as opposed to using the implicit role?

johanna-adhoc (changes requested):
> > Can we add a test to check for the 'Request an API key' link? Maybe one for the heading text too, but I'm not sure on that one.
> 
> I've added a test for the api url. I didnt think the Heading text needed a test since it's something that could be changed regularly.

And that is exactly why the text content needs regression testing, so any change made by a developer should be conscious enough to also involve update of the unit test of the component.

src/components/hero/Hero.tsx
TS000: I'm new to Testing-Library and saw that you could target by role. The role wasnt set so I added one after looking up a11y best practice for links.

> out of curiosity, what was the motivation for adding the role explicitly as opposed to using the implicit role?

Are you implying that the role will be implicitly set and I cant still search by role "link" and it will still work?

src/components/hero/Hero.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+    const heroImage = queryByRole('presentation');
+
+    expect(heroImage).toBeInTheDocument();
+    expect(heroImage).toHaveAttribute('src', (logoImage.default));
+  });
+});
```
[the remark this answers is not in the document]
TS000: Is my check for '/apply' done incorrectly?

src/components/hero/Hero.tsx
mjlumetta: yeah, absolutely; this is probably a good thing to call out then, for Testing Library and ARIA roles in general. lots of elements have implicit roles based on the tag + sometimes other characteristics. for example, `a` tags have the link role implicitly _if_ they have an href, and otherwise no implicit role, and Testing Library will respect those implicit roles. they have a link from the `*ByRole` docs in the text of that body to [this table of roles in the ARIA spec](https://www.w3.org/TR/html-aria/#docconformance). it's been a super helpful reference for me while writing unit tests and learning about how these roles work in general.

TS000: You're a legend, thank you!

johanna-adhoc (approved):

anthony-kopczyk (approved):
Looks good to me

PR #575, opened 2020-12-10
Redirect legacy auth page to new auth page

ediiotero:
### Redirect legacy page to new auth page

[API-3471](https://vajira.max.gov/browse/API-3471)

Added a Redirect from explore/:apiCategoryKey/docs/authorization to the new auth page when the auth docs v2 flag is true

### Requested feedback

Looking forward to feedback 🙂


src/containers/documentation/DocumentationRoot.tsx
```diff
@@ -141,11 +142,13 @@ const DocumentationRoot = (): JSX.Element => {
               {!shouldRouteCategory && <Redirect from="/explore/:apiCategoryKey" to="/explore" />}
```
mjlumetta, line 142: not quite within the scope of your ticket, but can you fix the opposite of this case while you're in here? the under-construction v2 docs are currently exposed in prod. if you wouldn't mind wrapping the route on line 141 in a check for the flag, that'd be awesome!

mjlumetta (commented):
changes look perfect to me 💅 ✨ 

i do have a couple small notes that could be added to this PR that are sort of tangential. one is that the v2 docs are currently available in prod and not flagged at the route level; you can't get to them by site navigation, but they're accessible. not part of your ticket, but would be great to fix while we're here. two, one thing that could be a helpful but not required addition to this PR (somewhat in the spirit of the previous suggestion, actually) is adding an integration test for this behavior. (although actually, not sure how that would work with environment variables. but that's something we can discuss if you're interested in adding one!)

mjlumetta (dismissed):
oh I meant to approve, actually. but will re-approve if you choose to make any of those changes, of course. thanks Eddie!!

mjlumetta (approved):
yeah, looks good to go! tests look perfect, great job on those 💅 💯

PR #576, opened 2020-12-11
Api 3759/side nav test coverage

johanna-adhoc:
### [API-3759](https://vajira.max.gov/browse/API-3759)

Added test coverage for SideNav Component

### Process
- [N/A] Docs updated

Screen Shot 2020-12-11 at 10 14 11 AM


TS000 (approved):
Looks great!

src/components/sideNav/SideNav.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <li data-testid="item1">item one</li>
+      <li data-testid="item2">item two</li>
+      <li data-testid="item3">item three</li>
+    </>
+  );
+
+  it('checks stickyfilljs library is not setup setup when ref to nav is null or undefined.', () => {
+    applyStickiness(null);
+    expect(Stickyfill.addOne).not.toHaveBeenCalled();
+  });
```
anthony-kopczyk, line 28: Does it make sense to move the test for applyStickiness into it's own `describe`? Then, in the other two tests, the render could go within a `beforeEach`

anthony-kopczyk (approved):
This looks good.

One potential suggestion might be that we could move `applyStickiness` into a hook. I don't think we use this anywhere else in the code at the moment though? It would simplify the SideNav component, since it would only `useStickiness` instead of a `useRef` and `useEffect`.

src/components/sideNav/SideNav.test.tsx
johanna-adhoc: yeah, we could move the applyStickiness to its own describe. The "before each" part you are mentioning I am not understanding well.

johanna-adhoc:
> This looks good.
> 
> One potential suggestion might be that we could move `applyStickiness` into a hook. I don't think we use this anywhere else in the code at the moment though? It would simplify the SideNav component, since it would only `useStickiness` instead of a `useRef` and `useEffect`.

I don't think moving it to a custom hook is necessary as this logic is not likely to be reused.

src/components/sideNav/SideNav.test.tsx
anthony-kopczyk: Since the render is exactly the same in both tests, the render could go in a `beforeEach` function. I don't think this is necessary, and maybe it's better to be explicit in each individual test what we want to render anyway.

```
describe('thing', () => {
  beforeEach(() => {
    render(<div></div>);
  });

  it('renders div', () => {
    expect(thing);
  });

  it('does something', () => {
    expect(other thing);
  });
});
```

anthony-kopczyk (approved):
Looks good!

TS000 (approved):
LGTM 🚀

PR #577, opened 2020-12-11
API: 2935 - Remove HoverImage Component

TS000:
### Description
Jira ticket [here](https://vajira.max.gov/browse/API-2935)
Removed HoverImage and replaced it's functionality with CSS.

- [ ] Docs updated
  
### Requested feedback

Feedback welcome.


anthony-kopczyk (dismissed):
This looks good to me, but I did have some thoughts that maybe we could further improve this in other tickets. I'm unsure if these suggestion are worth pursuing though.

When changing the size of the page the new div snaps to different sizes where the original slowly gets smaller as you change the screen size. I don't think this is necessarily something to block the PR though.

The other thought, and this is how it works in prod already, so again nothing to block this PR, is that the other Digital media each have a play button after the link. This link, and 'play button', has no function and does not get read out loud by a screen reader (using NVDA on Chrome Windows). I'm not sure if this could be confusing, since it doesn't quite match the pattern of the other Digital media links.

TS000:
I went through and checked the most used viewports to make sure there weren't any bad jumps.

johanna-adhoc:
I think this class name should be something like "hover-image-videoplayer" given that the image to be displayed is actually really specific and not generic and totally reusable

anthony-kopczyk (approved):
Looks good

johanna-adhoc (approved):
👍

PR #582, opened 2020-12-15
API-3968/govwebsite banner screen reader enhancements

johanna-adhoc:
### Description

As a user using a screen reader I want to be able to easily and accurately navigate through the header/banner content area that explains it is an official government website on the dev portal

Currently Screen reader gets caught in images.

This ticket addresses only issues with the dropdown for government website information

### Process
- [ ] Docs updated
  - [X] If you've made any UI changes, you've run an axe check using the axe browser extension that reported no new issues


src/components/banner/Banner.tsx
```diff
@@ -101,11 +103,13 @@ const Banner = (): JSX.Element => {
                     'small-desktop-screen:vads-u-margin-left--1',
                     'small-desktop-screen:vads-u-margin-top--0',
                   )}
+                  aria-label="Here&apos;s how you know. Read more information on how to know this is an official website"
```
1Copenut, line 106: What do you think about labeling this "Here's how you know this is an official website" ? I think it captures what you're looking to say and doesn't repeat the word "know".

src/components/banner/Banner.tsx
```diff
@@ -122,8 +126,10 @@ const Banner = (): JSX.Element => {
                 'vads-l-grid-container',
                 'vads-u-background-color--gray-dark')}
               aria-hidden={accordionVisible ? 'false' : 'true'}
+              aria-labelledby="toggle-how-you-know-dropdown"
```
1Copenut, line 129: Let's change this to an `aria-label` and have the text match the accessible label of the button: "Here's how you know this is an official website". That way if you want to change the accessible button label later, this region name won't change with it.

src/components/banner/Banner.tsx
```diff
@@ -35,13 +35,14 @@ const GuidanceBox = (props: GuidanceBoxProps): JSX.Element => (
       className={classNames('va-api-banner-icon', 'vads-u-margin-right--1', 'vads-u-margin-top--0p5')}
       src={props.icon}
       alt={props.icon_alt ?? undefined}
+      aria-hidden="true"
     />
-    <div className="vads-u-margin-bottom--1p5">
```
1Copenut, line 39: In testing the screen readers, the `div > p` markup pattern felt a little more natural. Adding the `article` and `heading` created two additional "stops" for screen readers and the text didn't flow as nicely. This is completely a subjective call; feel free to keep this valid markup if it makes sense for your team.

1Copenut (commented):
@johanna-adhoc I added a few comments after some local testing with VoiceOver and JAWS. Thank you for improving this element!

anthony-kopczyk (approved):
This looks good to me. Running through with a screen reader this seems much smoother than before, with the aria-label attributes. The region being named so similar to the button was a little offsetting for me, but that may be standard on sites when using screenreaders. I'm still getting used to them.

src/components/banner/Banner.tsx
```diff
@@ -101,11 +103,13 @@ const Banner = (): JSX.Element => {
                     'small-desktop-screen:vads-u-margin-left--1',
                     'small-desktop-screen:vads-u-margin-top--0',
                   )}
+                  aria-label="Here&apos;s how you know this is an official website"
```
mjlumetta, line 106: so the idea here is that we want to override the text for the button for the screen reader only? would like to understand this part of the change a bit deeper.

src/components/banner/Banner.tsx
```diff
@@ -122,8 +126,10 @@ const Banner = (): JSX.Element => {
                 'vads-l-grid-container',
                 'vads-u-background-color--gray-dark')}
               aria-hidden={accordionVisible ? 'false' : 'true'}
+              aria-label="Here&apos;s how you know this is an official website"
+              role="region"
```
mjlumetta, line 130: if we're adding the `aria-label`, would it be preferable to make this tag a section and get the ARIA role implicitly?

also, React JSX question - do characters need to be HTML-encoded in attributes? not familiar with the rules around that, don't know if i've ever encountered it.

mjlumetta (commented):
thanks @johanna-adhoc, this is such a huge improvement! I left a couple of smaller detail questions that I'm curious about, but think this would be fine to go as is.

src/components/banner/Banner.tsx
johanna-adhoc: So basically when you are reading a page with the screen reader each element kinda reads independently from each other. There is no context of elements that are together belong together and elements apart are not related. 

So if you use the screen reader to read from left to right (forward) it would read something like this:
Move to element to the right -> "An official website of the United States government" 
Move to next element to the right -> "Here's how you know"

So at this point the user could assume these two elements are related... but there is nothing for sure. User cannot see the position of these elements to understand that they are in fact correlated under the same idea/context.

If you try to read the page backwards it starts to read even more odd and the context is totally lost.

Move to element to the left -> "Talk to veterans crisis line now" 
Move to next element to the left -> "Here's how you know"

Context is totally lost. So basically with screen readers the elements need to have sufficient information on themselves to let the user know what the elements is really about and what is doing.

With this change the element is descriptive by itself it would read "Here's how you know this is an official website". So know we clearly see what is the topic this collapsible button will describe in more detail.

Following this idea is the reason why the region has been assigned a label that ties together the collapsible button with the area that in fact expands and collapse @anthony-kopczyk. I hope this also answer your question. For more detail on the handling of collapsible/accordion elements please see this link https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html

src/components/banner/Banner.tsx
johanna-adhoc: I normally always try to use encoded characters when dealing with HTML and because the value of the attribute will be HTML at the end that is why I wrote it that way. I am not sure there is a preference and the screen reader seems to be working fine either way encoded or decoded.

TS000 (approved):
Looks great! 👯‍♂️

src/components/banner/Banner.tsx
mjlumetta: okay, gotcha, sounds good to me!

src/components/banner/Banner.tsx
mjlumetta: ok, gotcha. I've got a couple questions off of this discussion; one is a more philosophical design question about the context + screen reader discussion, and another is about the region. and thanks for the callout of the accordion design pattern, that's helpful to know in context of this design.

1. the design question is basically: when does that context get lost for adjacent elements for screen reader users? this is something that I feel like I have a really poor grasp of as a sighted user who primarily interprets web pages visually. so like, when I use the screen reader, I feel like my visual mental model doesn't really translate to the mental model presented by the screen reader. when I use the screen reader, I do experience that loss of context, but I'm also a total novice with the screen reader, and I'm often considering elements in isolation and navigating very slowly. so I'm wondering, how do I and others of us who primarily interpret sites visually (which I assume is most of us, but I also don't want to make generalizations about things like that without being certain) know when navigating between adjacent elements is genuinely a loss of context and when it's a lack of familiarity on my part? because I could totally understand this being a loss of context like you're saying, and I could also totally believe that I'm no good at interpreting pages with the screen reader, and that users with visual disabilities who regularly use a screen reader could easily interpret this without special handling and might even find it repetitive. like, I could see the adjacency of these elements indicating their relation to screen reader users much like I read left to right with my eyes when reading forward, and I could see them knowing they'd need to keep going back for more context when reading backward, as I would reading right to left. tl;dr I genuinely do not know how to make that distinction in cases like these, and if you have any insight you can share about how to make that determination, would love to get a deeper understanding.
2. for the reading backwards case, I'm wondering if it would be beneficial to make an outer region containing the "An official website..." text, the button, and the accordion that would be labeled by the "An official website" bit? that could potentially alleviate the reading backwards case, although it might also add an unnecessary extra stop. (although also making it a section would potentially make it easier to skip over? again, total novice.)

thanks for your detailed explanation. hoping to get a better understanding of how to navigate these considerations and would love to hear your thoughts.

src/components/banner/Banner.tsx
```diff
@@ -122,8 +126,10 @@ const Banner = (): JSX.Element => {
                 'vads-l-grid-container',
                 'vads-u-background-color--gray-dark')}
               aria-hidden={accordionVisible ? 'false' : 'true'}
+              aria-label="Here&apos;s how you know this is an official website"
```
mjlumetta, line 129: can this be an `aria-labelledby` pointing back to the button? or would that lose the extra context from the `aria-label`?

src/components/banner/Banner.tsx
1Copenut: 👍 I agree with the current implementation and appreciate the discussion thread!

src/components/banner/Banner.tsx
1Copenut: This "internal" aria-label feels like most compact way to give the button a semantically meaningful label. If users are moving by pressing `TAB` or by the form elements menu on screen readers, they'll hear the `aria-label` text and have clear context for the button's purpose.

1Copenut (approved):

src/components/banner/Banner.tsx
johanna-adhoc: 1. I cannot say for sure how a person with a disability and more experience uses the SR. The little that I know is that the reading speed at which they use it is setup is super fast and they have a mastery in the use of landmarks and shortcuts beyond my skills. I personally test screen reader changes in the following way:
a) I literally close my eyes and see if for what I hear I am getting the same information and understanding on what I see.
b) I make a first reading of the whole page/area in an automated way with the commands: Control + Option +A
c) Do a read by moving manually into each element from left to right (forward).
d) Do a read by moving manually into each element from right to left (backward).

I always try to check for official examples from the w3c accessibility websites. I start with those implementations and then tweak and test until the content sounds more human like with the testing steps I explained above. If possible I reach for someone with more experience to double check the changes, but pretty much that is how I do it.

2. I did tried the section/role region element at the higher level but it didn't added a more human like reading in my opinion so I didn't follow that path.

mjlumetta (approved):
works great! I left one small comment about making an empty alt for the icons since we're hiding them, but not necessary to address here. thanks also for the knowledge sharing on this PR, this was a useful learning experience for me.

PR #584, opened 2020-12-17
Auth Docs: Code/Scopes Switcher dropdown

gileswells:
### Description

https://vajira.max.gov/browse/API-3407
This adds in the dropdown for code blocks and scopes tables. All oauth code blocks that had a dropdown in the [figma](https://www.figma.com/proto/1TwAA9NRhVeNjp8wfQYm1D/Authorization?node-id=2%3A79&viewport=-1451%2C-2942%2C0.7296842336654663&scaling=min-zoom) are able to be set within the appropriate APIDescription in the apiDefs folder.

Assumptions I made:

- Only currently available APIs are listed in the dropdown for code/scope swapping
- Only externally visible APIs are listed (Address Validation is hidden by this)

### Process

1. Set the `.env.local` value for `REACT_APP_AUTH_DOCS_V2=true`
2. Navigate to /explore/authorization
3. Observe that code blocks and the scopes table now all swap together when one is changed

- [ ] Docs updated
  

src/apiDefs/query.ts
```diff
@@ -22,6 +22,12 @@ const getApiCategoryOrder = (): string[] => apiCategoryOrder;
 const getAllApis = (): APIDescription[] =>
   Object.values(getApiDefinitions()).flatMap((category: APICategory) => category.apis);
 
+const getAllOauthApis = (): APIDescription[] =>
+  getAllApis()
+    .filter((item: APIDescription) => !!item.oAuth && !item.vaInternalOnly)
```
mjlumetta, line 27: i think the `vaInternalOnly` check probably comes from the Address Validation error, so i think it can probably be eliminated regardless, but i don't know that we'd want to exclude internal APIs in this function overall.

src/apiDefs/schema.ts
```diff
@@ -62,11 +62,30 @@ export interface APIDescription {
   readonly vaInternalOnly: boolean;
   readonly trustedPartnerOnly: boolean;
   readonly oAuth?: boolean;
+  readonly openidDocs?: OpenIDCodeExamples;
+  readonly pkceDocs?: PKCECodeExamples;
+  readonly scopes?: string[];
```
mjlumetta, line 67: for these + the `oAuth` boolean - what do you think about the idea of nesting this config inside some kind of "authorization info" object? might be useful to group it together. not necessary at this point, per se, but think it's worth considering.

src/components/codeWrapper/CodeWrapper.tsx
```diff
@@ -1,14 +1,32 @@
...
   children: React.ReactNode;
+  options?: APIDescription[];
+  selectedOption?: string;
+  onSelectionChange?: (event: React.ChangeEvent<HTMLSelectElement>) => void;
 }
 
 /* eslint-disable jsx-a11y/no-noninteractive-tabindex -- required for keyboard scrolling */
 const CodeWrapper = (props: CodeWrapperProps): JSX.Element => (
   <div className="code-wrapper" tabIndex={0}>
+    {props.options && props.selectedOption && (
```
mjlumetta, line 16: what's the reason for including the select within the code wrapper, as opposed in another component that uses the code wrapper? it seems to me like this could be an edge case for the code wrapper overall (you know, TBD depending on platform growth), and i wonder if it would be better to avoid creating that conditional.

regardless of the answer to that, i think it is somewhat problematic for the select to be after the code wrapper block in the tab order. for scrollable code blocks, the order is "code block -> tab to select an API -> reverse tab back to code block to scroll", and it seems like the experience would be better if we eliminated that extra navigation.

src/components/oauthDocs/BuildingOIDCContent.tsx
```diff
@@ -1,38 +1,47 @@
+/* eslint-disable react/no-unused-prop-types */
```
mjlumetta, line 1: unused props being the ones that are splatted down?

src/components/oauthDocs/AuthCodeFlowContent.tsx
```diff
@@ -1,154 +1,208 @@
...
-    </p>
-
-    <AuthCodeDeleteGrantRequest />
-  </section>
-);
-
-AuthCodeFlowContent.propTypes = {};
+interface AuthCodeFlowContentProps {
+  apiDef: APIDescription | null;
+  options: APIDescription[];
```
mjlumetta, line 11: do `options` need to be a prop, or can we use the array of OAuth APIs directly?

src/components/oauthDocs/AuthCodeFlowContent.tsx
```diff
@@ -1,154 +1,208 @@
...
+  selectedOption: string;
+  onSelectionChange: (event: React.ChangeEvent<HTMLSelectElement>) => void;
+}
+
+const AuthCodeFlowContent = (props: AuthCodeFlowContentProps): JSX.Element => {
+  const wrapperProps = {
+    onSelectionChange: props.onSelectionChange,
+    options: props.options,
+    selectedOption: props.selectedOption,
```
mjlumetta, line 20: it seems like we're directly passing down the selected API + change handler for several levels in the tree. should we consider representing that in the Redux store to simplify the direct handling at those levels?

src/components/oauthDocs/ScopesContent.tsx
```diff
@@ -0,0 +1,172 @@
+/* eslint-disable complexity */
```
mjlumetta, line 1: why are we disabling this rule here?

src/components/oauthDocs/ScopesContent.tsx
```diff
@@ -0,0 +1,172 @@
...
+import DefaultScopes from '../../content/apiDocs/oauth/Scopes.mdx';
+
+interface ScopesContentProps {
+  apiDef: APIDescription | null;
+  options: APIDescription[];
+  selectedOption: string;
+  onSelectionChange: (event: React.ChangeEvent<HTMLSelectElement>) => void;
+}
+
+// eslint-disable-next-line arrow-body-style
```
mjlumetta, line 16: same question about disabling the rule

src/components/oauthDocs/ScopesContent.tsx
```diff
@@ -0,0 +1,172 @@
...
+                <td>
+                  <code>launch/patient</code>
+                </td>
+                <td>
+                  A permission setting to obtain the patient&apos;s identifier in the token response
+                  when the app is launched from an EHR.
+                </td>
+              </tr>
+            )}
+            {hasClaimScope && (
```
mjlumetta, line 57: did we ever get actual content for this? cc @charleystran in particular

src/components/oauthDocs/ScopesContent.tsx
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// eslint-disable-next-line arrow-body-style
+const ScopesContent = (props: ScopesContentProps): JSX.Element => {
+  const scopes = props.apiDef?.scopes ?? [];
+  const hasClaimScope = scopes.some(element => element.startsWith('claim.'));
+  const hasPatientScope = scopes.some(element => element.startsWith('patient/'));
+
+  return (
+    <>
```
mjlumetta, line 23: i'm pretty confused about the design of this component. can you share more about the rationale behind this structure? most of the time, non-OIDC scopes are not shared across authorization servers. the only exception is `launch/patient`. i would've anticipated that we would have a conditional at the level of the authorization server and render the scopes for that authorization server as one unit, rather than going scope by scope.

src/containers/documentation/DocumentationRoot.tsx
```diff
@@ -75,7 +75,7 @@ const ExploreSideNav = (): JSX.Element => {
     <SideNav ariaLabel="API Docs Side Nav">
       <SideNavEntry key="all" exact to="/explore" name="Overview" />
       <Flag name={[FLAG_AUTH_DOCS_V2]}>
-        <SideNavEntry key="authorization" exact to="/explore/authorization" name="Authorization" />
+        <SideNavEntry key="authorization" to="/explore/authorization" name="Authorization" />
```
mjlumetta, line 78: is this removal in anticipation of the query param? curious as to why this change happened

mjlumetta (commented):
leaving a bunch of questions and comments. have some questions about the config and whether we can slim that down and about data flow through the component tree. let me know if you want to talk about any of this in more detail! like seeing this come together.

src/apiDefs/data/verification.ts
```diff
@@ -17,7 +17,39 @@ const verificationApis: APIDescription[] = [
     enabledByDefault: true,
     name: 'Address Validation API',
     oAuth: true,
+    openidDocs: {
```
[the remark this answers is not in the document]
gileswells: I'll kill these and update that line

src/apiDefs/data/benefits.ts
```diff
@@ -19,7 +19,31 @@ const benefitsApis: APIDescription[] = [
...
+        'HTTP/1.1 400 Bad Request\nContent-Type: application/json\nCache-Control: no-store\nPragma: no-cache\n\n{\n  "error": "invalid_request",\n  "error_description": "Invalid email address."\n}',
+      authRevokeTokenAccess:
+        "POST /oauth2/revoke HTTP/1.1\nHost: sandbox-api.va.gov\nContent-Type: application/x-www-form-urlencoded\nAuthorization: Basic {base64 encoded *client id* + ':' + *client secret*}\ntoken={your access token}&token_type_hint=access_token",
+      authRevokeTokenRefresh:
+        "POST /oauth2/revoke HTTP/1.1\nHost: sandbox-api.va.gov\nContent-Type: application/x-www-form-urlencoded\nAuthorization: Basic {base64 encoded *client id* + ':' + *client secret*}\n\ntoken={your refresh token}&token_type_hint=refresh_token",
+      authUrl:
+        'https://sandbox-api.va.gov/oauth2/authorization?\n  client_id=0oa1c01m77heEXUZt2p7\n  &redirect_uri=<yourRedirectURL>\n  &response_type=code\n  &scope=profile openid offline_access launch/patient claim.read claim.write\n  &state=1AOQK33KIfH2g0ADHvU1oWAb7xQY7p6qWnUFiG1ffcUdrbCY1DBAZ3NffrjaoBGQ\n  &nonce=o5jYpLSe29RBHBsn5iAnMKYpYw2Iw9XRBweacc001hRo5xxJEbHuniEbhuxHfVZy',
+    },
+    pkceDocs: {
```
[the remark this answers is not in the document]
gileswells: In the structure that Eddie created the OpenID docs and the PKCE docs are split into different components. The split was to keep that segmentation throughout.

src/apiDefs/data/health.ts
```diff
@@ -37,7 +37,37 @@ const healthApis: APIDescription[] = [
     enabledByDefault: true,
     name: 'Community Care Eligibility API',
     oAuth: true,
+    openidDocs: {
+      authManageAccount: 'POST /oauth2/token HTTP/1.1\nHost: sandbox-api.va.gov',
```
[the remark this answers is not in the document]
gileswells: That may be well and good. My reasoning was that nobody is perfect and by keeping the code samples as whole and yes repetitive is better prepared for some api that we might be asked to included that doesn't go through this same endpoint. It _shouldn't_ happen but I would not at all be surprised if it did.

If you're confident in there not being an issue with potential down the road things I can rework this to be much smaller and just swapping out base path and scopes.

src/apiDefs/data/health.ts
```diff
@@ -91,7 +121,47 @@ const healthApis: APIDescription[] = [
...
+      authPostToken:
+      'POST /oauth2/token HTTP/1.1\nHost: sandbox-api.va.gov\nContent-Type: application/x-www-form-urlencoded\n\ngrant_type=authorization_code\n&code=z92dapo5\n&state=af0ifjsldkj\n&redirect_uri=<yourRedirectURL>\n&code_verifier=ccec_bace_d453_e31c_eb86_2ad1_9a1b_0a89_a584_c068_2c96',
+      authUrl:
+        'https://sandbox-api.va.gov/oauth2/authorization?\n  client_id=0oa1c01m77heEXUZt2p7\n  &redirect_uri=<yourRedirectURL>\n  &response_type=code\n  &scope=profile openid offline_access launch/patient patient/AllergyIntolerance.read patient/Condition.read patient/DiagnosticReport.read patient/Immunization.read patient/Medication.read patient/MedicationOrder.read patient/MedicationRequest.read patient/MedicationStatement.read patient/Observation.read patient/Patient.read patient/Procedure.read\n  &state=1AOQK33KIfH2g0ADHvU1oWAb7xQY7p6qWnUFiG1ffcUdrbCY1DBAZ3NffrjaoBGQ\n  &code_challenge_method=S256\n  &code_challenge=gNL3Mve3EVRsiFq0H6gfCz8z8IUANboT-eQZgEkXzKw',
+    },
     releaseNotes: FhirApiReleaseNotes,
+    scopes: [
+      'profile',
+      'openid',
+      'offline_access',
```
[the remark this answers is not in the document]
gileswells: Being that it's not a really complex repetition I prefer to just list them here rather than include them from elsewhere. Not something I'm looking to die on a hill for so I'm ok moving it to a constant.

src/apiDefs/query.ts
```diff
@@ -22,6 +22,12 @@ const getApiCategoryOrder = (): string[] => apiCategoryOrder;
...
+    .sort((a, b) => (a.name > b.name ? 1 : -1))
+    .map((item: APIDescription) => item);
```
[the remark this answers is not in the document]
gileswells: There is absolute is no purpose. Just an absent minded line of code.

src/apiDefs/query.ts
gileswells: I think it is a fair assumption to limit access to these so even with the Address Validation fix I think leaving it here is a good idea. If in the future we want to change that it's not a hard fix.

src/apiDefs/schema.ts
gileswells: I can get behind that.

src/components/codeWrapper/CodeWrapper.tsx
gileswells: eh I missed that. I'll rework that. Also that's a good time to unify the drop downs across the code wrapper and table.

src/apiDefs/data/health.ts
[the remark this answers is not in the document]
mjlumetta: I'm pretty confident about this not being significant enough of an issue that we shouldn't do it now. I think that a change of that scale would require rethinking either how we document OAuth or some larger changes in the OAuth implementation, so I think we're safe to refactor this to be smaller for this round of changes.

src/apiDefs/data/health.ts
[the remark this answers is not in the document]
mjlumetta: not a big issue for me either way for the same reason. I'm okay with leaving them here explicitly too.

src/apiDefs/query.ts
mjlumetta: why do you think it's a fair assumption? that could be correct, but I wouldn't necessarily assume so myself. I agree that it would be an easy fix in either direction, but I have a preference for not filtering internal only APIs because it is more opinionated to add that extra condition. 

it's also a naming thing; like, by default, I would not assume that internal-only classification has anything to do with being an OAuth API (although it _kind of_ does in reality in the sense that VA-internal apps often already have authorization to the veteran's data, but that's tangential to the point). so if the name were to reflect something like "get public OAuth APIs", then I'd understand why this condition goes in this function.

src/apiDefs/data/benefits.ts
```diff
@@ -19,7 +19,31 @@ const benefitsApis: APIDescription[] = [
     enabledByDefault: true,
     name: 'Benefits Claims',
     oAuth: true,
+    openidDocs: {
+      authManageAccount: 'POST /oauth2/token HTTP/1.1\nHost: sandbox-api.va.gov',
```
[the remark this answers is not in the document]
gileswells: I grabbed it from the existing code samples and missed that there was a difference. Will update.

src/apiDefs/query.ts
gileswells: That makes sense and if someone wants just public or just internal they could filter that themselves very easily after pulling all.

src/components/oauthDocs/AuthCodeFlowContent.tsx
gileswells: I had done this so that there would be less querying of the list of APIs. Grabbing that array within multiple components seemed to be more like repetitive code than just passing them as props was. I'm not really stubborn on this issue. I'll update it.

src/components/oauthDocs/AuthCodeFlowContent.tsx
gileswells: I thought we had removed redux but I guess that was just the redux connected router. Yeah redux would be the right place for that.

src/components/oauthDocs/ScopesContent.tsx
gileswells: The scopes table has a lot of conditions in it and to me it didn't seem like breaking them out into lots of smaller files made the most sense. Loading up a single file to setup the scopes table vs having to load up potentially lots of files for each scope parent (`patient/*` or `claim.*`) seemed like the better option.

src/components/oauthDocs/ScopesContent.tsx
gileswells: My thought process here was that because the scopes do have some overlap in what's under different parent items was just to make it a singular table. For example this allows us to keep the `patient/*` parent to both the Community Care API and the Health API and we just list off which scopes apply in the right column. I'm not sure if I followed your question properly but that's how I read it.

src/components/oauthDocs/ScopesContent.tsx
gileswells: I asked Darcy via DM and this is what she gave me.

src/containers/documentation/DocumentationRoot.tsx
gileswells: Query param wasn't the initial reason. When you click the hash links down the page the side nav would show no links highlighted before this change.

src/apiDefs/data/benefits.ts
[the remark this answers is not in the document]
mjlumetta: ah, i see. i want to clarify the terminology a little bit here, because that's mainly what was throwing me off. PKCE is a part of OpenID Connect; both the traditional authorization code flow docs and the PKCE docs are nested under the overall OIDC heading. (technically PKCE is a variant of the authorization code flow, also, but we've decided to make this distinction for practical purposes.)

src/apiDefs/query.ts
mjlumetta: exactly, yup!!

also, tiny thing i just noticed - do we need the `!!` operator on the `oAuth` condition? iirc, that's a boolean already, right?

src/components/codeWrapper/CodeWrapper.tsx
mjlumetta: yeah, that totally makes sense, would be good to have a single source for the OAuth API picker

src/components/oauthDocs/AuthCodeFlowContent.tsx
mjlumetta: yeah, I think that's probably unlikely to be a huge concern right now since we're querying something in local memory. when we move to a real external database for OAuth APIs, I'm sure we will want to eliminate the multi-querying, but I imagine in that case we'd cache the list in the Redux store and then use `useSelector`.

I'm also fine with leaving it in here if that's your preference, but maybe rename to `oauthAPIs` in that case? would make it a bit clearer what those options are in context.

src/components/oauthDocs/AuthCodeFlowContent.tsx
mjlumetta: sounds good. curious what your thoughts are on extending the structure of the app state overall - I think it's an open question of how to structure the state as we expand past, like, 2 items.

src/components/oauthDocs/ScopesContent.tsx
mjlumetta: gotcha, gotcha. good to know, thanks

src/containers/documentation/DocumentationRoot.tsx
mjlumetta: ahhh... huh, interesting. good catch there!

src/components/oauthDocs/ScopesContent.tsx
mjlumetta: ahh, okay. I think we do have different mental models of the API scopes in that way. part of it, I think, is that I didn't really see the `patient/*` rows in Community Care and Health as being the same since they are mutually exclusive sets, which goes more back to my understanding of the overall presentation in [this comment](https://github.com/department-of-veterans-affairs/developer-portal/pull/584#discussion_r546055707). 

understanding that, I see the rationale for combining Health and Community Care now. I still think that Verification and Claims should get separate tables, though, because there shouldn't be any overlap between those two or between either of those and either of the Health APIs. what are your thoughts on that approach?

src/components/oauthDocs/ScopesContent.tsx
mjlumetta: gotcha. in that case, my thoughts are in [this comment](https://github.com/department-of-veterans-affairs/developer-portal/pull/584#discussion_r546063307)

src/components/oauthDocs/ScopesContent.tsx
gileswells: Was causing an issue. Now isn't. 🤷‍♂️

src/apiDefs/data/benefits.ts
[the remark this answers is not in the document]
gileswells: Based on how this is just the auth docs I added the `basePath` to the slimmed down `oAuthInfo` property BUT it's never being used anywhere. Each of these code samples just hits the auth endpoints which, from what I can tell, don't hit any of the services' base paths. I feel like having it there is good because we can likely use it BUT for right now it's not being used.

src/apiDefs/query.ts
gileswells: It's an optional boolean property so while undefined does evaluate falsy it's a personal preference of mine to add the `!!` when there's ambiguity to what's being used for the condition. It works both with and without it.

src/apiDefs/data/benefits.ts
```diff
@@ -19,6 +19,10 @@ const benefitsApis: APIDescription[] = [
     enabledByDefault: true,
     name: 'Benefits Claims',
     oAuth: true,
+    oAuthInfo: {
+      basePath: '/services/claims/v1',
```
mjlumetta, line 23: may have had a small miscommunication here around "base path" here; I wasn't very clear with that terminology. I was meaning to refer to the _authorization server base path_, but totally understand how that wasn't clear. will flag where this is used in the components.

src/apiDefs/query.ts
mjlumetta: ahh yes okay, I'm totally on board with coercing it to a boolean if it's ambiguous. mostly I wish it weren't optional in the first place. carry on

src/apiDefs/schema.ts
```diff
@@ -62,11 +62,17 @@ export interface APIDescription {
   readonly vaInternalOnly: boolean;
   readonly trustedPartnerOnly: boolean;
   readonly oAuth?: boolean;
+  readonly oAuthInfo?: OAuthInfo;
```
mjlumetta, line 65: liking how this looks so far, would be good with committing this. I have a suggestion for a potential optional enhancement around the Typescript handling, which would involve nesting the `oAuth` boolean inside `oAuthInfo` and repurposing `oAuthInfo` as a general "authorization info" object. so we could have something like 

```
interface APIDescription {
  ...
  readonly authInfo: AuthorizationInfo;
}

interface AuthorizationInfo {
  readonly oAuth: boolean;
  readonly oAuthBasePath?: string;
  readonly scopes: string[];
}
```

we could refactor the `AuthorizationInfo` interface even further by dividing it into separate types for OAuth and key auth and then doing a type union.

```
interface OAuthInfo {
  readonly oAuth: true;
  readonly basePath: string;
  readonly scopes: string[];
}

interface KeyAuthInfo {
  readonly oAuth: false;
}

type AuthorizationInfo = OAuthInfo | KeyAuthInfo;
```

we could go even further in order to (a) anticipate the possibility of other auth types and (b) add slightly more descriptiveness by changing the `oAuth` boolean to an enum, too.

```
enum AuthorizationType {
  OAuth = 'OAuth',
  KeyAuth = 'KeyAuth',
}

interface OAuthInfo {
  readonly authType: AuthorizationType.OAuth;
  readonly basePath: string;
  readonly scopes?: string[];
}

interface KeyAuthInfo {
  readonly authType: AuthorizationType.KeyAuth;
}
```

now, obviously, I've gone a little bit off the rails here, and I cannot stress strongly enough that all of this is optional (and def focus on the other stuff first). but if you're interested in trying this TS refactor, I think it could be neat.

src/components/apiSelector/APISelector.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import './APISelector.scss';
+
+interface APISelectorProps {
+  options: APIDescription[];
+  selectedOption: string;
+  onSelectionChange: (event: React.ChangeEvent<HTMLSelectElement>) => void;
+}
+
+const APISelector = (props: APISelectorProps): JSX.Element => (
```
mjlumetta, line 12: great job on this component! perfect 👨‍🍳 👌  🙌

src/components/oauthDocs/AuthCodeFlowContent.tsx
```diff
@@ -1,154 +1,216 @@
...
+  onSelectionChange: (event: React.ChangeEvent<HTMLSelectElement>) => void;
+}
+
+const AuthCodeFlowContent = (props: AuthCodeFlowContentProps): JSX.Element => {
+  const selectorProps = {
+    onSelectionChange: props.onSelectionChange,
+    options: props.options,
+    selectedOption: props.selectedOption,
+  };
+  const authUrl = `\`\`\`plaintext\nhttps://sandbox-api.va.gov/oauth2/authorization?\n  client_id=0oa1c01m77heEXUZt2p7\n  &redirect_uri=<yourRedirectURL>\n  &response_type=code\n  &scope=${props.apiDef?.oAuthInfo?.scopes.join(' ') ?? ''}\n  &state=1AOQK33KIfH2g0ADHvU1oWAb7xQY7p6qWnUFiG1ffcUdrbCY1DBAZ3NffrjaoBGQ\n  &nonce=o5jYpLSe29RBHBsn5iAnMKYpYw2Iw9XRBweacc001hRo5xxJEbHuniEbhuxHfVZy\n\`\`\``;
```
mjlumetta, line 22: this is where we need to consume the authorization server base path, in place of `/oauth2/authorization`. you can find those base paths in [API-1643](https://vajira.max.gov/browse/API-1643) or wherever you got the OpenID config like Beau suggested [here](https://lighthouseva.slack.com/archives/CB0DV4HC5/p1608047238115600?thread_ts=1608044563.112900&cid=CB0DV4HC5).

another question about these strings that @ediiotero and I discussed briefly when we were pairing on the code styles: now that these strings aren't stored in the API data, what do you think about using the markdown components Eddie created with props support? I think we may have discussed the need to type those components being an additional requirement, which would be something we'd have to tackle, but I think it could provide a little more readability for devs to look at the markdown if that's doable.

mjlumetta (commented):
left a couple comments on the most recent commits. the two things I saw were clarifying a small issue with the base path and a completely, totally optional type refactor on the authorization info schema. also, really nice job on the API selector, looks v polished.

src/components/oauthDocs/AuthCodeFlowContent.tsx
gileswells: The issue that I found with adding props to the markdown files is that through the webpack markdown loader we're using you cannot add props to the imported files. I haven't looked into if you can do something similar directly with importing a file through `ReactMarkdown` but that could be an option to consider. 

I'm updating the auth urls.

src/apiDefs/schema.ts
gileswells: I don't really like the idea of nesting the primary `oAuth` boolean within an object. If we want to resolve the `!!` item raised within another comment and make this a requirement for all APIs so that he boolean is required everywhere now this nesting is going to be required everywhere. To me the option of having neighboring fields only when needed vs having a nested object with optional params I would prefer the first one. Two different paths to the same end in the grand scheme of things though.

mjlumetta: I guess I was viewing requiring this config explicitly and in one place as a positive. yes, it is more verbose, but it means that each API is explicit about its authorization info; it would allow us to eliminate some optional chaining in these components and express key auth conditions as "it is a key auth API" rather than "it is not an OAuth API" in the apply form and such. in any case, I'm good with the current format.

src/apiDefs/query.test.ts
```diff
@@ -73,7 +73,7 @@ describe('query module', () => {
       const verificationApi = lookupApiCategory('verification');
       expect(verificationApi).not.toBeNull();
       expect(verificationApi?.apis.length).toBeGreaterThanOrEqual(3);
-      expect(verificationApi?.apis.map(api => !!api.oAuth).filter(m => m).length).toEqual(2);
+      expect(verificationApi?.apis.map(api => !!api.oAuth).filter(m => m).length).toEqual(1);
```
gileswells, line 76: Fixing address validation's miscategorized oAuth boolean forced this change and not some other underlying issue in the logic.

src/components/oauthDocs/ScopesContent.tsx
gileswells: I'll circle back with you on this tomorrow so we can discuss because I'm not sure I'm following your thought process here.

src/types/index.ts
```diff
@@ -28,7 +28,12 @@ export interface APIVersioning {
   requestedApiVersion: string;
 }
 
+export interface APISelection {
+  selectedApi: string;
+}
+
 export interface RootState {
+  apiSelection: APISelection;
```
mjlumetta, line 36: same thing about names, and also wondering whether we want to nest this within another object, like a top-level "documentation" key. not necessary at this point, but I think we're going to want to start grouping related pieces together, and OAuth API could be one (potentially nested together with the version, but that would be additional refactoring on top of the nesting).

mjlumetta (commented):
overall this is looking really good. here's what I'm thinking in terms of remaining things to look at:

1. comments on here: have some feedback about clarifying the "OAuth" in the Redux state.
2. design: need the dummy API fallbacks.
3. want to revisit the markdown problem. can pair on this.

don't think this necessarily needs to be said, but to make it explicit, don't worry about the optional refactors at this point. I've pointed a couple out but really don't worry about those. the 3 things above are what feels important.

I'll hang out for a while tonight while you're still working on this in case you want another review before I'm out. can also pair on the markdown thing just to see how doable that is.

src/actions/index.ts
```diff
@@ -20,6 +20,15 @@ export interface SetVersioning extends Action {
   versions: VersionMetadata[] | null;
 }
 
+export interface ResetAPISelection extends Action {
```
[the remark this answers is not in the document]
gileswells: My thought on this being an API Selection vs being an OAuth Api Selection is that there's no reason that I can currently see to limit this to just OAuth APIs. We're passing in the options to the selector component so that it's agnostic, I figure the store should be agnostic as well.

src/components/oauthDocs/PKCEAuthContent.tsx
```diff
@@ -1,107 +1,144 @@
...
-    <p>
-      The response will return a new <code>access_token</code> and <code>refresh_token</code>, if
-      you requested the <code>offline_access</code> scope.
-    </p>
-  </section>
-);
 
+const PKCEAuthContent = (): JSX.Element => {
+  const selectedApi = useSelector((state: RootState) => state.apiSelection.selectedApi);
+  const apiDef = lookupApiByFragment(selectedApi);
```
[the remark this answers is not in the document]
gileswells: I'm going to pass on that for now mainly because I haven't used a reselect selector before but I'll add that to future research/learning sticky note.

src/types/index.ts
gileswells: I could see nesting as we get larger. Now wouldn't be a bad time if it wasn't such a bad time on the calendar.

src/components/oauthDocs/PKCEAuthContent.tsx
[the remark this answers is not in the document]
mjlumetta: absolutely, yeah, sounds good

ediiotero (approved):
LGTM 👍

PR #586, opened 2020-12-21
Api 3497: Import extend-expect for all unit tests

anthony-kopczyk:
### Description

This PR imports extend-expect for all of our unit tests, so that we do not need to import it in each one we want to use it. This PR also removes the import statements from any tests that included it. Note that any new tests written at the same time of this PR are not covered, so we'll want to make a mental note to check for this on our PRs for a few weeks. Note that if a stray unit test did include the import it won't cause any negative effects besides having an extra line in the file.

Jira Ticket: https://vajira.max.gov/browse/API-3947

### Process

- [N/A Docs updated
  
### Requested feedback
I had to create a new js file for setting the up 'post jest initialization' for the import. Any suggestions regarding those are greatly appreciated, as well as any ideas for clarification in the configuration files.


mjlumetta (dismissed):
this is awesome!!

johanna-adhoc (approved):
💯

PR #587, opened 2020-12-21
Resolve missing Link component

anthony-kopczyk:
### Description

There isn't a Jira ticket for this, but I ran into a similar issue while working https://vajira.max.gov/browse/API-2991, and the fix took about 5 minutes. I'm not sure if there's a better way to approach this however. I can retroactively put in a ticket for this if we deem we want to look into this more before merging.

The ApiDocumentation test was having issues that the React 'Link' tag was not defined. This was caused by a mock of the react router dom. It includes 'useHistory' and 'useLocation' but not 'Link'. To rememdy this, we can use jest.requireActual and a spread operator to effectively only mock those two methods of the node library and use the native package versions of Link, useParams, and any other values of react-router-dom.

### Process

- [N/A] Docs updated
  
### Requested feedback

If anybody knows of better jest mock functions to accomplish this let me know! I see a few, including auto-generated mocks. I'm unsure if any work internally similar to this. A `jest.mockActual` method as syntactic sugar for what I have would be very nice, but I didn't see one available.


src/containers/documentation/ApiDocumentation.test.tsx
```diff
@@ -35,6 +35,8 @@ const server = setupServer(
 );
 
 jest.mock('react-router-dom', () => ({
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+  ...(jest.requireActual('react-router-dom') as Record<string, unknown>),
```
anthony-kopczyk, line 39: I needed the eslint disable here because it requires the type be an object to use the spread operator, but `requireActual` returns type `any`, which our linter doesn't like. So, I casted it to Record, but then the linter says the cast is unnecessary. It's a fun cyclic issue.

mjlumetta (approved):
looks great!! ✨

johanna-adhoc (approved):
LGTM

PR #588, opened 2020-12-21
API-3541 test: add tests for ApplySuccess component

alexandervalencia:
### Description

This PR is for [API-3529](https://vajira.max.gov/browse/API-3529); it adds testing for the ApplySuccess component, displayed after a user successfully signs up for an API Key.

image

- [ ] Docs updated
  
### Requested feedback

Any feedback welcome


src/containers/apply/ApplySuccess.test.tsx
```diff
@@ -0,0 +1,202 @@
...
+      });
+
+      it('displays the API Key generated by the backend', () => {
+        expect(screen.getByText('Your VA API key is:')).toBeInTheDocument();
+        expect(screen.getByText('elf-friend')).toBeInTheDocument();
+      });
+
+      it('displays the provided email address', () => {
+        expect(
+          screen.getByText(/You should receive an email at gimli@eredluin\.com/gm),
```
rtravitz, line 104: What does the `/gm` give you here? I'm not super familiar with React Testing Library. The docs I see just list `/i` as an ignore case option for the substring match. Assuming these are others?

alexandervalencia: Unfortunately, when the testing-library parses out the DOM it writes it as:

```html
<!-- (In this example "・" represents a whitespace.) -->

You should receive an email at・
gimli@eredluin.com・
````
So we need the `m` to make the regex multi-line, but I accidentally left the `g` in, so I'll take it out 👍

nwalker-va (approved):
Looks good to me 👍

PR #590, opened 2020-12-23
Api 3837/health api encourage r4

ramosan-lit:
### Description

Update the Health API portal to encourage users to use the latest recommended/required version of FHIR.
Updating release notes
### Process

- [x] Docs updated
  

src/apiDefs/data/health.ts
```diff
@@ -93,7 +93,7 @@ const healthApis: APIDescription[] = [
     oAuth: true,
     releaseNotes: FhirApiReleaseNotes,
     trustedPartnerOnly: false,
-    urlFragment: 'fhir',
+    urlFragment: 'fhir?tab=r4',
```
gileswells, line 96: Rather than changing this via the `urlFragment` which is used in many different places I would suggest you move the R4 part of the `docSources` a few lines above in this file to be first in the list so that it loads first by way of being the first tab. If you want you can add the recommendation to that as an `apiIntro` as the other two have.

gileswells (changes requested):
I'd much prefer to see the tabs re-ordered putting R4 first instead of the urlFragment trying to force the desired tab to open.

gileswells (approved):
This looks much better to me. 👍

PR #592, opened 2020-12-23
API - 3512: Create Test For API - 2900: Fetch Wrapper

TS000:
### Description
[Jira ticket here.](https://vajira.max.gov/browse/API-3512)

File             | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-----------------|---------|----------|---------|---------|-------------------                       
  makeRequest.ts |     100 |    93.75 |     100 |     100 | 47,59  (partial coverage)           

### Process

Lot's of parts covered however, some things to consider for a new ticket in the future:
The fetch wrapper is able to read JSON and Text but the logs get created in very specific situations. This may not be ideal for the type of error reporting we want.

Error codes 404 and 5XX create a log but don’t care about the response content.

Error code 400 only logs if the response is a JSON with a field error. That decision was made because supposedly that is how the response is always sent from the Backend Dev Portal.

One solution is that we could create a log when the code is 404 AND the response content is Text. How do we know the response text doesn’t contain sensitive information?

When the JSON with the error array (ln 55 of makeRequest.tsx) is returned there should be logic that controls what goes in the response so no sensitive information is exposed .

- [ ✖️ ] Tests added or updated for change

### Requested feedback

All feedback welcome, this is the biggest unit text I've ever written. 😮


johanna-adhoc (approved):
LGTM

src/utils/makeRequest.test.ts
```diff
@@ -0,0 +1,319 @@
...
+      },
+      method: 'POST',
+    };
+
+    await makeRequest<ExpectedResponse>(errorUrl, init).catch(() => {
+      expect(spyFetch).toHaveBeenCalledWith(headerDataError);
+      expect(withScope).toHaveBeenCalled();
+      expect(Sentry.captureException).toHaveBeenCalledWith(expect.objectContaining({
+        message: 'Unexpected end of JSON input',
+      }));
```
anthony-kopczyk, line 204: If makeRequest does not throw an error for some reason these expect statements won't get called and the test will pass incorrectly, or will they be tested anyway?

anthony-kopczyk: This occurs in a few places. I didn't comment on each one

src/utils/makeRequest.test.ts
```diff
@@ -0,0 +1,319 @@
+/* eslint-disable max-lines */
+import * as Sentry from '@sentry/browser';
+import 'jest';
+import {  MockedRequest, rest, restContext  } from 'msw';
```
anthony-kopczyk, line 4: Looks like there's an extra space in these curly braces

anthony-kopczyk (commented):
Looks good overall. I have a concern over some expect statements occurring within `.catch` statements. I'm not sure if jest somehow knows to execute those statements regardless of if an error is thrown or not.

src/utils/makeRequest.test.ts
TS000: _[If the promise is rejected, the test will automatically fail.](https://jestjs.io/docs/en/asynchronous)_. Does this help?

johanna-adhoc: FYI Adding expects inside then and catch blocks is a normal thing when implementing test cases. 

Regarding the scenario of encountering that the method makeRequest "does not throw an error for some reason" then we would need to pinpoint what specific part of the logic allows for that specific scenario to happen, if the behavior is correct then we would need to add a new test case, and if the behavior is not correct then we would need to revisit the logic and create a new test case regardless.

anthony-kopczyk: @TS000 That helps a lot. I didn't know about some of those techniques. With our scenario it looks like there should be calls to `expect.assertions` to make sure the statements in the catch are called. A couple paragraphs lower it says:

> If you expect a promise to be rejected, use the .catch method. Make sure to add expect.assertions to verify that a certain number of assertions are called. Otherwise, a fulfilled promise would not fail the test.

```
test('the fetch fails with an error', () => {
  expect.assertions(1);
  return fetchData().catch(e => expect(e).toMatch('error'));
});
```

src/utils/makeRequest.test.ts
```diff
@@ -0,0 +1,325 @@
...
+      },
+      method: 'POST',
+    };
+
+    const testResponse = await makeRequest<Blob>(
+      requestUri,
+      init,
+      { responseType: ResponseType.BLOB },
+    );
+    expect.assertions(1);
```
anthony-kopczyk, line 111: Since the following assertion isn't within a `.catch` or conditional spot, I don't think we need it here.

src/utils/makeRequest.test.ts
```diff
@@ -0,0 +1,325 @@
...
+      body: 'json for you',
+      headers: {
+        accept: 'application/json',
+        'content-type': 'application/json',
+      },
+      method: 'POST',
+    };
+
+    await makeRequest<ExpectedResponse>(errorUrl, init).catch(() => {
+      expect.assertions(3);
```
anthony-kopczyk, line 168: I think `expect.assertions` needs to be called outside the catch block. I'll leave it to you to test though :)

```
expect.assertions(3);
await makeRequest<... catch(() => {
  expect(thing);
}
```

anthony-kopczyk (commented):
I'm unsure of the implementation of expect.assertions. The code looks good otherwise.

src/utils/makeRequest.test.ts
johanna-adhoc: I think the assertion would only make sense if it is outside the catch block. The only part I am wondering is that because the catch block is async code that would mean that the assertion expect would need to be at the end and needs to be implemented in conjunction with an await waitFor so it actually waits for the catch block to be executed.

TS000: You are partly correct. Anthony and I did some invasive changes to makeRequest so we could check to make sure tests were running correctly. We discovered that assertions inside the catch block would unexpectedly succeed while assertions outside the block would fail as expected. Placing the assertions before or after the block made no difference and is shown in an [official Jest example](https://jestjs.io/docs/en/expect#expectassertionsnumber) as before.

johanna-adhoc (approved):

anthony-kopczyk (approved):
Looks good to go!

PR #593, opened 2020-12-24
Update styling of code blocks

ediiotero:
### Update styling of code blocks

Add styling to API Selector in the new Auth docs. Added React Syntax Highlighter for code blocks. Removed React Markdown.


src/containers/documentation/AuthorizationDocs.scss
```diff
@@ -55,4 +56,25 @@
       list-style-type: square;
     }
   }
+
+  .api-selector {
+    select {
+        background-color:$color-gray;
```
gileswells, line 62: I would prefer to see these styles on the `.api-selector` selector rather than on the select element. Reason being, while we're not implementing the copy button currently it feels more correct to me that the background and rounded corners and whatnot be on the container rather than an element within the container.

gileswells (changes requested):
Just some structural changes to what the CSS is applied to.

src/components/codeWrapper/CodeWrapper.scss
```diff
@@ -1,7 +1,7 @@
 .code-wrapper {
-  margin: 16px 0;
+  margin: 0;
```
gileswells, line 2: This change here is what I believe is causing the 2 back to back code blocks to now have no gap between them at all. While the visual regressions pass with the new images I believe that the more correct solution is to update this change to account for when 2 code wrappers are siblings on the page. Perhaps something similar to this?

```css
.code-wrapper + .code-wrapper {
  margin-top: 16px;
}
```

gileswells (changes requested):
The 2 back to back code wrappers now have no gap between them.

src/components/codeWrapper/CodeWrapper.scss
ediiotero: Thanks, I guess I missed that

mjlumetta:
something about the API selector looks broken? not sure why that is here.

src/components/oauthDocs/AuthCodeFlowContent.tsx
```diff
@@ -51,7 +54,9 @@ const AuthCodeFlowContent = (): JSX.Element => {
 
       <APISelector {...selectorProps} />
       <CodeWrapper>
-        <ReactMarkdown>{authUrl}</ReactMarkdown>
+        <SyntaxHighlighter language="plaintext" style={syntaxColor}>
```
mjlumetta, line 57: want to do a bit of a roundup with Okapi eng + any other frontend folks on the conversion from markdown to Javascript strings to discuss what direction we think is most beneficial here. I still have a preference for the markdown setup because it makes the code blocks more legible/editable/etc for me at least; don't know if others feel similarly or not. there are some wonky spacing issues with a couple of the code blocks (newlines, indentation) that I'm not sure whether it's due to being lost in those strings or due to something with `react-syntax-highlighter` that we should fix regardless of how we choose to approach this question. anyway, would like hear people's thoughts and get any other relevant info. like, for example, does React Syntax Highlighter give us better control of highlighting in some way? I'm good with whatever the team thinks is best, but I'm not entirely clear on what the rationale for React Syntax Highlighter over the markdown approach is at the moment.

src/containers/documentation/AuthorizationDocs.scss
```diff
@@ -5,15 +5,10 @@
   td {
     line-height: 1.6;
   }
-
-  code,
-  .code-wrapper {
-    // Needs both due to covering of both code blocks and inline formatted code/parameter names
```
mjlumetta, line 11: why was this block removed? I think this comment still applies. see here:

NoBackgroundOnCodeBlockScroll

mjlumetta (commented):
overall this looks really great! couple of CSS tweaks suggested, etc. the one bigger question I'd like to address in this ticket is the syntax highlighter approach with the new module vs the old markdown approach. would like to have a discussion on the benefits of each; either could work, but I'd like for us to get a little clearer together on why we're making the decision we're making.

src/containers/documentation/AuthorizationDocs.scss
ediiotero: sorry I missed that, thanks for catching it

src/containers/documentation/AuthorizationDocs.scss
```diff
@@ -28,6 +27,8 @@
     padding: 2rem;
     margin: 0;
     white-space: pre;
+    border-top-left-radius: 0;
+    border-top-right-radius: 0;
```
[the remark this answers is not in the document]
ediiotero: thanks fixed that

src/components/apiSelector/APISelector.scss
```diff
@@ -0,0 +1,18 @@
+@import '../../styles/variables';
+
+.api-selector {
+    select {
+        border: none;
+        min-width: 100%;
+        padding: 1rem 1.5rem;
+        margin-bottom: 0;
+        margin-top: 2rem;
+        color: $color-white;
```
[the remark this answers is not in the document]
ediiotero: taking the color out of this file and using the AuthorizationDocs file to handle color

src/containers/documentation/AuthorizationDocs.scss
```diff
@@ -51,8 +48,34 @@
...
+    }
+  }
+
+  .scopes-content {
+    .api-selector {
+      select {
+        border: 1px solid $color-black;
+        background-color: $color-white;
+        border-radius: 0;
+        color: $color-black;
```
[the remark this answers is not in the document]
ediiotero: changed it to $color-base. 
Is that the color you're talking about?

src/containers/documentation/AuthorizationDocs.scss
```diff
@@ -51,8 +48,34 @@
...
+
+  .api-selector {
+    background-color: $color-gray;
+    border-top-left-radius: 6px;
+    border-top-right-radius: 6px;
+
+    select {
+      option {
+          background-color: $color-white;
+          color: $color-black;
```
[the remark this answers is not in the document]
ediiotero: was able to remove the background color but need to keep the font color for the options as we made the select fonts white

src/components/oauthDocs/AuthCodeFlowContent.tsx
ediiotero: open to talking about it

src/containers/documentation/AuthorizationDocs.scss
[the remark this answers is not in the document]
mjlumetta: yup, that looks like it!

src/containers/documentation/AuthorizationDocs.scss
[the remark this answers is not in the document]
mjlumetta: cool cool, sounds good!

src/containers/documentation/AuthorizationDocs.scss
```diff
@@ -51,8 +51,41 @@
...
+          color: $color-base;
+      }
+    }
+  }
+
+  .scopes-content {
+    .api-selector {
+      background-color: $color-white;
+      select {
+        border: 1px solid $color-base;
```
mjlumetta, line 77: can we make the border color `$color-gray`? trying to match it up with the default styles as seen in the version selector.

src/containers/documentation/AuthorizationDocs.scss
```diff
@@ -51,8 +51,41 @@
...
     border-left: 2px solid $color-gray-light;
     list-style-type: none;
+
     ul {
       list-style-type: square;
     }
   }
+
+  .api-selector {
+    background-color: $color-gray;
```
mjlumetta, line 61: wondering if this style and the corresponding override on line 75 might be better suited in the `select` block next to the text color... don't think it makes much of a difference functionally, but I think of those two things going together usually. if other people think about them similarly, might be worth grouping together for ease of reading.

src/components/codeWrapper/CodeWrapper.scss
```diff
@@ -1,9 +1,13 @@
...
```
mjlumetta, line 2: if this margin is 0, can we cut this line with no change in behavior?

src/components/oauthDocs/AuthCodeFlowContent.tsx
```diff
@@ -16,17 +17,19 @@ const AuthCodeFlowContent = (): JSX.Element => {
...
-  const authRevokeGrant = `\`\`\`http\nDELETE ${apiDef?.oAuthInfo?.baseAuthPath ?? '/oauth2'}/grants HTTP/1.1\nHost: sandbox-api.va.gov\nContent-Type: application/x-www-form-urlencoded\n\n{\n  "client_id": {client_id},\n  "email": {test account email}\n}\n\`\`\``;
-  const authRevokeGrantError = '\`\`\`http\nHTTP/1.1 400 Bad Request\nContent-Type: application/json\nCache-Control: no-store\nPragma: no-cache\n\n{\n  "error": "invalid_request",\n  "error_description": "Invalid email address."\n}\n\`\`\`';
+  const authUrl = `https://sandbox-api.va.gov${apiDef?.oAuthInfo?.baseAuthPath ?? '/oauth2'}/authorization?\n  client_id=0oa1c01m77heEXUZt2p7\n  &redirect_uri=<yourRedirectURL>\n  &response_type=code\n  &scope=${apiDef?.oAuthInfo?.scopes.join(' ') ?? 'profile openid offline_access'}\n  &state=1AOQK33KIfH2g0ADHvU1oWAb7xQY7p6qWnUFiG1ffcUdrbCY1DBAZ3NffrjaoBGQ\n  &nonce=o5jYpLSe29RBHBsn5iAnMKYpYw2Iw9XRBweacc001hRo5xxJEbHuniEbhuxHfVZy`;
```
mjlumetta, line 20: not necessarily related to code blocks styles, but can we make the dummy val for the base path something like "/oauth2/{api}/v1"? I think we probably want to avoid giving the consumer the impression that the /oauth2 path on its own is a meaningful value.

src/components/codeWrapper/CodeWrapper.scss
```diff
@@ -1,9 +1,13 @@
...
   overflow-x: auto;
 
   & > pre {
     margin: 0;
     overflow: visible;
   }
 }
+
+.code-wrapper + .code-wrapper {
```
mjlumetta, line 11: I don't think that this block is doing anything right now. the only back to back blocks we have right now have an API selector, so the selector doesn't apply. also think that this is something that can be handled by the auth docs stylesheet; the code wrapper should probably be agnostic to how it's positioned with respect to other code wrappers, or other elements in general, as long as it provides reasonable defaults.

mjlumetta (commented):
looking really good! all of these comments are pretty detail-oriented. think that the code blocks implementation is the main thing to look at in order to get this committed.

src/components/codeWrapper/CodeWrapper.scss
ediiotero: this was added for the legacy auth docs to fix a visual regression failure. I moved it to the AuthorizationDocsLegacy.scss file

src/components/oauthDocs/AuthCodeFlowContent.tsx
ediiotero: will be addressed in another ticket, per our convo in the working session

src/containers/documentation/AuthorizationDocs.scss
ediiotero: Thas as done because of @gileswells comment [above](https://github.com/department-of-veterans-affairs/developer-portal/pull/593#discussion_r549370662) I thought it made sense but can change if needed.

src/containers/documentation/AuthorizationDocs.scss
ediiotero: done

src/components/codeWrapper/CodeWrapper.scss
ediiotero: done

src/components/codeWrapper/CodeWrapper.scss
mjlumetta: ahhh I missed that, thanks for pointing that out. moving to AuthorizationDocsLegacy.scss sounds good!

src/components/oauthDocs/AuthCodeFlowContent.tsx
mjlumetta: going to be resolved in a follow-up ticket. Eddie, would you mind creating a follow-up ticket in this epic before closing out API-3414?

src/containers/documentation/AuthorizationDocs.scss
mjlumetta: ahh I see. it's up to you where you want it to go, in that case. generally speaking, I feel like it can be beneficial to look at text and background color alongside each other, but that's more of a personal preference.

src/components/oauthDocs/ScopesContent.tsx
```diff
@@ -24,7 +24,7 @@ const ScopesContent = (): JSX.Element => {
   const hasPatientScope = scopes.some(element => element.startsWith('patient/'));
 
   return (
-    <>
+    <section className="scopes-content">
```
mjlumetta, line 27: maybe add an `aria-labelledby="scopes"` on this section? although that's not strictly necessary because this is sort of inconsistent on this page due to the different implementations of MDX vs pure React. should probably straighten that out at some point to make it consistent for screen readers - not sure which way is preferable (ARIA role of "region" vs no role).

mjlumetta (approved):
looks good to me! left one suggestion that you can take or leave; will re-approve if you make any more changes. nice work on this one 🚀 ✨

src/components/oauthDocs/ScopesContent.tsx
ediiotero: Same here maybe we can talk about it during another working session

PR #594, opened 2020-12-30
API-3411 / Add query param for selected API on Auth docs

gileswells:
### Description

https://vajira.max.gov/browse/API-3411

This borrows the structure of how api version does their query string manipulation on changes to that dropdown.

### Process

- [ ] Docs updated
  

src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -1,17 +1,66 @@
...
 
 import './AuthorizationDocs.scss';
 
-export const AuthorizationDocs = (): JSX.Element => {
+interface AuthorizationDocsProps {
+  location: Location;
+}
+
+const AuthorizationDocsPropTypes = {
+  // Leave as any for now until we can use the location react hooks
```
TS000, line 24: I wasn't aware we couldn't use the location react hooks 😭

ediiotero (commented):
Looks good! I had a problem with seeing the ui reflect the correct api when I copy and paste the url into another tab. Not sure if its something on my end, but might be worth another look.

src/containers/documentation/AuthorizationDocs.tsx
gileswells: Honestly I'm not sure about that. I copied this from the ApiDocumentation component and figured someone had a good reason for this but not many people were here to ask about it.

src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -1,17 +1,66 @@
...
+  const { location } = props;
   const dispatch: React.Dispatch<ResetOAuthAPISelection> = useDispatch();
+  const initializing = React.useRef(true);
+  let api = useSelector((state: RootState) => state.oAuthApiSelection.selectedOAuthApi);
+  const prevApi = usePrevious(api);
+
+  if (initializing.current) {
+    initializing.current = false;
+
+    api = getInitialApi(location.search);
```
mjlumetta, line 53: how does this initial API get set in the Redux state if they're not the same?

src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -1,17 +1,66 @@
...
+  }
+
+  /**
+   * UPDATES URL WITH CORRECT VERSION PARAM
+   */
+  React.useEffect(() => {
+    if (prevApi !== api) {
+      setSearchParam(history, location.search, api);
```
mjlumetta, line 61: runs the first time, which sets `?api=` as the query string. think we need `api` to be a non-empty string to actually run this function.

src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -1,17 +1,66 @@
...
+
+const getInitialApi = (searchQuery: string): string => {
+  const params = new URLSearchParams(searchQuery || undefined);
+  const apiQuery = params.get('api');
+  return apiQuery ? apiQuery.toLowerCase() : DEFAULT_OAUTH_API_SELECTION;
+};
+
+const setSearchParam = (history: History, queryString: string, api: string): void => {
+  const params = new URLSearchParams(queryString);
+  if (params.get('api') !== api) {
```
mjlumetta, line 36: this condition is very similar to the one on line 60, with the difference that one is for the Redux val and one is for the URL val. can we get rid of one, or do we need both? either way, why?

mjlumetta (commented):
looks pretty good on the first pass, left a couple of comments. the big thing i can see is that the API isn't getting selected based on the query param on the initial load right now. another thing that i don't know where to put but is probably important is that the hash links in the page get rid of the query string, which is bad, presumably. that one might be very straightforward or very difficult to work with, depending on the cause.

src/containers/documentation/AuthorizationDocs.tsx
gileswells: changing to the hook works well.

src/containers/documentation/AuthorizationDocs.tsx
gileswells: The function works with blank string or `none` depending on how we want it to look in the url. Personally I'd prefer it to be a blank string if we go with an empty selection option.

src/containers/documentation/AuthorizationDocs.tsx
gileswells: I removed the one on line 60. No need for both looking at it.

mjlumetta: sounds good. the only thing I'd say after thinking about it a little further is that I have a slight preference for the other location, because it's not entirely clear where it's called that it's conditional, but that's also a v small consideration.

src/containers/documentation/AuthorizationDocs.tsx
mjlumetta: oh wait another question - what happens if the query param doesn't refer to a real API?

mjlumetta (commented):
looks pretty good to me at this point. everything works, just about - the only other thing I'd call out would be looking into those hash links, plus a `useEffect` hook I think we're going to need. trying to think if there are any edge cases I'm missing but don't think so.

also, maybe could add some integration tests for this functionality?

src/containers/documentation/AuthorizationDocs.tsx
gileswells: So if there isn't a matching API it will default to no selection.

mjlumetta: yeah, I saw that it's functionally working when I tried it out. I guess my question there is, should we explicitly prevent the value of the `selectedOAuthApi` in the Redux state from being a value that doesn't correspond to a real API? I doubt that it matters that much right now, but it could be good to safeguard against for the future (for example, it gives us more confidence when pulling the API definition that the state value refers to a real API). I'll leave that up to your judgment whether or not that's important to guard against rn.

gileswells: I've updated this to set the redux store within this method and renamed it accordingly.

src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -1,17 +1,65 @@
...
+  const apiQuery = params.get('api');
+  const api = apiQuery ? apiQuery.toLowerCase() : DEFAULT_OAUTH_API_SELECTION;
+  dispatch(setOAuthApiSelection(api));
+};
+
+const setSearchParam = (history: History, queryString: string, api: string): void => {
+  const params = new URLSearchParams(queryString);
+  if (params.get('api') !== api) {
+    params.set('api', api);
+    history.push(`${history.location.pathname}?${params.toString()}`);
```
mjlumetta, line 37: should include the hash as well here, I think?

we need to handle this case both ways:

1. clicking an in-page anchor link should not remove the query string.
2. selecting a new API should not remove the hash anchor.

mjlumetta (commented):
one remaining thing is keeping the hash anchors in sync with the query strings and vice versa. should be good to approve one there's handling for that, I think. (again, this is part of the huge pain of that zenscroll bug, because if we didn't need to use React Router Hash Link for in-page links, we wouldn't have to worry about that side of things.... 😭 😞 )

src/containers/documentation/AuthorizationDocs.tsx
gileswells: So for some reason when you add the hash into this it will force the page all the way to the top regardless of what the hash is. On a different hand I think that we should be removing the hash on this makes sense because when you're sharing the url it's not guaranteed that you were scrolled to that point anymore. If we wanted to add a feature where we keep the hash updated as you're scrolling through the page that would make sense to me but currently I'm of the opinion that removing is an acceptable solution.

mjlumetta: hmm... do you know why it will force the page to the top? not sure why that is, but might need to tweak something in `PageContent` or elsewhere. I don't have clear thoughts on what the best thing to do here is, but I would say removing it like you've described is an acceptable solution as well and we can fine-tune later if need be.

however, the exception that I can see as it currently stands is that it removes the hash if you navigate to the page in the URL bar with a specific hash, and I'm pretty sure that's not what we want there. so at least on the first load we need to preserve that hash, I would think. not sure if solving it in that specific case means that it makes more sense to solve it for the general case as well or if we want to preserve it only on first load. (focus does get set correctly still though, which is good.)

semi-related, bit tangential, but not sure where else to put this: we need to apply the `... location` to both skip nav links, because they have a similar thing where setting the search param is overriding the hash. 🤔 

thoughts?

mjlumetta (commented):
all the links look good, but I found a couple edge cases that we might need to address 🤕 the rabbit hole keeps on going

ediiotero (dismissed):
looks good to me. The only question I have is are we going to default this to Claims or is that going to be done on another ticket?

mjlumetta:
> The only question I have is are we going to default this to Claims or is that going to be done on another ticket?

@ediiotero this is a good question. I would say different - this PR is almost done and that would add significant (enough) scope. I think we might need a general cleanup ticket to do design + copy + a11y tweaks that we find as we do our final reviews on this work.

gileswells:
> this PR is almost done and that would add significant (enough) scope.

Not too much scope as it's a 2 line change to default and remove the no selection option from the dropdown. I'm also adding in 3 lines to default back to claims if someone tries to link to something like `?api=banana`

ediiotero (approved):
LGTM 👍

src/components/oauthDocs/AuthCodeFlowQueryParamsTable.mdx
```diff
@@ -3,12 +3,12 @@ imports:
...
+| Query Parameter | Required     | Values |
+| --------------- | ------------ | ------ |
+| `client_id`     | **Required** | The `client_id` issued by the VA API Platform team |
+| `redirect_uri`  | **Required** | The URL you supplied. The user will be redirected to this URL after authorizing your application. |
+| `response_type` | **Required** | Supported response types: `code` |
+| `state`         | **Required** | Ensures authorization flow integrity. See <HashLink to={{ ...location, hash: '#security-considerations' }}>security DEAD LINK</HashLink>. |
```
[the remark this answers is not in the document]
gileswells: I missed this before the commit / merge

PR #599, opened 2021-01-14
Update CodeBuild deploy to deploy from the main release without the "cb-" prefix

mjlumetta:
### Description

this PR updates the CodeBuild deploy to deploy from the main release as opposed to the secondary one with a "cb-" prefix. currently, CodeBuild creates a release in Github with a tag of the form `cb-developer-portal/v0.0.0`, and Jenkins creates the main release with a tag of the form `developer-portal/v0.0.0`. so, then, this PR is updating the CodeBuild deploy to use the release that is currently created by Jenkins but will be created by CodeBuild in the near future.

note that this deploy job does not run automatically anywhere right now. this PR will allow us to manually test the deploy job in dev and staging from a commit hash and from a release. once we have confidence in the deploy job, we will then cut over the dev and staging deploys that run on merges to master to be run from CodeBuild instead of Jenkins.

this PR also turns on PR comments in CodeBuild in preparation for the cutover. this change will temporarily cause redundant comments (roughly) from CodeBuild and Jenkins.

### Process

  - manually testing deploys  
- [n/a] Docs updated
  - [n/a] If you have made structural changes to the site, the [Developer Portal Content Types](https://community.max.gov/x/qUwOg) docs have been updated
  - [n/a] If you have made/confirmed any decisions about user interactions and accessibility, the [Accessible Component Design](https://community.max.gov/x/xS8mgg) docs have been updated
  
  - [n/a] If you've made any UI changes, you've run an axe check using the axe browser extension that reported no new issues


cicd/buildspec-deploy.yml
```diff
@@ -45,31 +45,28 @@ phases:
...
+          VERSION=${RELEASE#"developer-portal/v"}
           for f in *.bz2; do
             mv ${f} "${f%-${VERSION}.tar.bz2}.tar.bz2"
           done
-        fi
-      # Download and prepare S3 assets if deploying from commit hash.
-      - |
-        if [[ ${COMMIT_HASH} ]]; then 
+        elif [[ ${COMMIT_HASH} ]]; then 
           aws s3 sync --no-progress s3://${S3_ARCHIVE_BUCKET}/cb-${COMMIT_HASH:0:7}/ .
```
duganth-va, line 65: Did we decide to change this functionality for this step, I cant remember. I know we said we would not modify the release part for CI but I can't remember if we decided against referencing the Jenkins Archive for the manual deploys.

duganth-va: This line in particular: 
` aws s3 sync --no-progress s3://${S3_ARCHIVE_BUCKET}/cb-${COMMIT_HASH:0:7}/ .`

cicd/buildspec-deploy.yml
```diff
@@ -45,31 +45,28 @@ phases:
...
-          VERSION=${RELEASE#"cb-developer-portal/v"}
+          VERSION=${RELEASE#"developer-portal/v"}
           for f in *.bz2; do
             mv ${f} "${f%-${VERSION}.tar.bz2}.tar.bz2"
           done
-        fi
-      # Download and prepare S3 assets if deploying from commit hash.
-      - |
-        if [[ ${COMMIT_HASH} ]]; then 
+        elif [[ ${COMMIT_HASH} ]]; then
```
duganth-va, line 64: 💯

duganth-va (approved):
Looks good to me, pretty sure we can test this prior to merge to master if you'd like.

cicd/buildspec-deploy.yml
mjlumetta: my thought here - and I think this is what we said when we met a few weeks ago, although not 100% sure - was that we could leave this as is? I think either archive is fine, but if we do reference the Jenkins archive, we're going to have to change it later anyway. if we change this from `cb-${COMMIT_HASH:0:7}` to `${COMMIT_HASH}` now, we're going to need to flip it back from `${COMMIT_HASH}` to `${COMMIT_HASH:0:7}` later, assuming that we stick with the short ref (which I'm open to changing, but I think it's more readable because I only look at the first few characters anyway when I'm looking at a ref).

duganth-va: Sweet! I don't have a strong preference either just couldnt remember if we came to any conclusion, but this makes sense.

mjlumetta:
@duganth-va I'm going to try out the manual deploy in the lower envs before merge like you suggested

duganth-va (approved):
💯

rtravitz (approved):
One question, but this seems ready to move on.

cicd/buildspec-deploy.yml
```diff
@@ -45,31 +45,32 @@ phases:
...
             mv ${f} "${f%-${VERSION}.tar.bz2}.tar.bz2"
           done
-        fi
-      # Download and prepare S3 assets if deploying from commit hash.
-      - |
-        if [[ ${COMMIT_HASH} ]]; then 
-          aws s3 sync --no-progress s3://${S3_ARCHIVE_BUCKET}/cb-${COMMIT_HASH:0:7}/ . 
+        elif [[ ${COMMIT_HASH} ]]; then 
+          # confirm that commit has an archived build
+          aws s3 ls s3://${S3_ARCHIVE_BUCKET}/cb-${COMMIT_HASH:0:7}/ || exit 1
```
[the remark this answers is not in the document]
mjlumetta: yes, they will. currently, there are archived builds with a key of the long ref created by Jenkins and builds with a key of `cb-` + the short ref created by CodeBuild. we could've changed it so that CodeBuild archives it with the short ref only as a key now, or we could've changed this to reference the Jenkins archived build with the long ref now. we didn't want to remove the prefix while there were two kinds of builds so that there would be short ref keys and long ref keys, because that could be kind of confusing in the short term (wouldn't be very clear on the difference if you're looking at the S3 bucket). Tom and I discussed using the Jenkins archived builds [here](https://github.com/department-of-veterans-affairs/developer-portal/pull/599#discussion_r557700751) and why we think it doesn't really matter at the moment.

PR #600, opened 2021-01-14
Api 4200/oauth docs link

ediiotero:
### Description
[Story: API4200](https://vajira.max.gov/browse/API-4200)

Adding an Alert type link under the header of all the Auth APIs. Will link to the page with the API the user came from preselected in the dropdown. Also, remove links to the old auth page on the Health quick start page.

### Process
The link should only appear for auth APIs.


src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -78,6 +87,14 @@ const ApiDocumentation = (props: ApiDocumentationProps): JSX.Element => {
    */
   return (
     <Flag name={['hosted_apis', apiDefinition.urlFragment]}>
+      {(currentOauthUrl.length > 0 && authDocsV2) && (
+        <div className="api-docs-oauth-link">
+          <h3 className="usa-alert-heading">Authentication and Authorization</h3>
```
ediiotero, line 92: Wasn't sure if an h3 is the right way to go here. This came from the VA Alert box.

mjlumetta: should be an h2 to maintain the header order since it's right below the h1. we can update the styling if it looks too big or anything like that.

although... my question about this is what the experience is like in terms of interpreting what comes below the h2 but outside of the box. so it's clear to sighted users that it's not describing the intro, but potentially not to screen reader users. gonna have to think about this one.

src/containers/documentation/ApiDocumentation.tsx
```diff
@@ -68,6 +74,9 @@ const ApiDocumentation = (props: ApiDocumentationProps): JSX.Element => {
   const dispatch = useDispatch();
   const queryParams = new URLSearchParams(location.search || undefined);
   const apiVersion = queryParams.get('version');
+  const oauthApis = getAllOauthApis().filter((item: APIDescription) => !isApiDeactivated(item));
+  const currentOauthUrl = oauthApis.filter(api => api.urlFragment === apiDefinition.urlFragment);
```
mjlumetta, line 78: this logic feels a bit circular to me. i'm reading it as "given 1 API, we fetch the complete list of OAuth APIs and see if the current API is in that list". but, in order to fetch the list of all APIs, `getAllOauthApis` is checking each of them - specifically `api.oAuth` - see if the API is an OAuth API. so i think we can use `!!api.oAuth` directly here.

mjlumetta (commented):
couple of small comments but overall looks great!

src/containers/documentation/ApiDocumentation.tsx
ediiotero: ok will clean this up a bit

src/containers/documentation/ApiDocumentation.tsx
mjlumetta: the best thing i could think of here for the latter concern would be to give that containing div a role of `region` and probably in that case attach an `aria-labelledby`. unclear how well that would work.

mjlumetta (approved):
looks good to me! I would _like_ to address that screen reader issue if possible before merging. I left a comment for something to try, but other than that, not sure what the best solution is. so yeah - let's try to give that a crack, but everything else is perfect.

mjlumetta (approved):
nice work @ediiotero ✨

PR #603, opened 2021-01-19
Update facilitiesReleaseNotes.mdx

deugster-va:
### Description

Release notes for additional VAST field


as-va (approved):

rtravitz (approved):
Looks fine from a functional perspective. @darduda is away for a while, so I'd say let's roll it in for now and we can address it in the future if there's a problem for some reason.

PR #604, opened 2021-01-20
API-4441 - Update Developer Portal landing page

alexandervalencia:
### Description
This PR supports [API-4441](https://vajira.max.gov/browse/API-4441) by adding new button variants, updating the look of the Developer Portal and adding a feature flag for new API Publishing related content and navigation.

### Process

- [ ] Docs updated
  

rtravitz (approved):
Haven't been in here in a while, but I don't find anything objectionable. Keeping the API Publishing button out of production works as advertised. 

Might double-check with others who are living with this every day if they're comfortable with the way the button overrides are added.

src/components/header/Header.tsx
```diff
@@ -10,12 +10,7 @@ import Search from '../search/Search';
 import TestingNotice from '../TestingNotice';
 import './Header.scss';
 
-const buttonClassnames = classNames(
-  'usa-button',
-  'vads-u-background-color--white',
-  'vads-u-color--primary-darkest',
-  'vads-u-margin-right--2',
-);
+const buttonClassnames = classNames('usa-button', 'usa-button-default', 'vads-u-margin-right--2');
```
mjlumetta, line 13: is `usa-button-default` a class from USWDS?

src/components/pageContent/PageContent.test.tsx
```diff
@@ -32,20 +32,22 @@ describe('PageContent', () => {
...
+    expect(documentationPageHeader).toBeInTheDocument();
   });
 
   it('scrolls the window to the top position after navigation', async () => {
-    userEvent.click(screen.getByRole('link', { name: 'Request an API Key' }));
+    userEvent.click(screen.getByRole('link', { name: 'Read the Docs' }));
 
-    await screen.findByRole('heading', { name: 'Apply for VA Lighthouse Developer Access' });
+    await screen.findByRole('heading', {
+      name: "Explore usage policies and technical details about VA's API offerings.",
```
mjlumetta, line 49: any reason for testing this as opposed to the h1? same for the test above. (don't really have an qualms about this functionally, but I'm also not really sure that this should be a heading. which, that's fine if we have to tweak this test again if that changes down the line.)

src/flags.ts
```diff
@@ -7,6 +7,7 @@ import { APIDescription } from './apiDefs/schema';
 export interface AppFlags {
   api_publishing: boolean;
   auth_docs_v2: boolean;
+  api_publishing: boolean;
```
mjlumetta, line 10: duplicate of line 8 on the merge

src/styles/buttons.scss
```diff
@@ -0,0 +1,51 @@
...
+      color: $color-primary-darker;
+    }
+
+    &:active,
+    &.usa-button-active {
+      background-color: $color-primary-alt-lightest;
+    }
+  }
+
+  &.usa-button-secondary {
```
mjlumetta, line 32: if we're completely overriding the color scheme of this button style, can we use a custom class with a prefix like `.va-api` rather than repurposing the USWDS class and overriding its styles? would make a similar suggestion for `usa-button-default`, except that I couldn't find it in the USWDS styles. thoughts on that?

mjlumetta (commented):
looks good overall! have some questions about how we're arranging the button styles in terms of USWDS styles vs custom styles and how those converge/diverge. also one functional TS issue from the merge commit that needs to be addressed prior to merging.

src/components/pageContent/PageContent.test.tsx
alexandervalencia: I think I was worried about "Documentation" being a little too general, but I checked again and it doesn't show up more than once in the PageContent area... I'll update it as it's more clear what we're targeting and what page we expect to be on.

src/styles/buttons.scss
alexandervalencia: Thanks @mjlumetta. 

There wasn't a way to achieve the desired stylings using utilities alone. 
My original idea was to follow the existing naming pattern for the button classes as the only existing precedence for custom classes is the `va-api-hero` classes which were replaced with utilities.

I like the idea of distinguishing that these are specifically VA API classes, I'll make the update to the names

src/components/hero/Hero.test.tsx
```diff
@@ -1,47 +1,50 @@
 import * as React from 'react';
 import 'jest';
-import {  render, screen } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { BrowserRouter as Router } from 'react-router-dom';
+import { FlagsProvider, getFlags } from '../../flags';
 import * as logoImage from '../../__mocks__/fakeImage';
 import { Hero } from './Hero';
 
+const flags = { ...getFlags(), api_publishing: true };
```
anthony-kopczyk, line 9: I can't get over how much I like destructuring

anthony-kopczyk (dismissed):
Had some time to take a look at this. Everything seems good to me, but I'm not 100% on our CSS styling principles.

src/components/header/Header.tsx
alexandervalencia: It's a custom class to describe the requested `default` variant

src/styles/buttons.scss
mjlumetta: yeah thanks for making this update! I think there are a few more examples of custom classes if you do a grep/other search for `va-api-` in the `src` directory. there's also a small file in `src/styles/lib.scss` that has some custom utilities that follow the VADS utility pattern with values/properties that aren't included in the design system.

src/styles/buttons.scss
```diff
@@ -0,0 +1,49 @@
...
+button,
+[type='button'],
+[type='submit'],
+[type='reset'] {
+  &.va-api-button-default {
+    background-color: $color-white;
+    color: $color-primary;
+
+    &:hover,
+    &.va-api-button-hover {
```
mjlumetta, line 19: only suggestion I'd make on this version would be removing the `.va-api-button-hover` and `.va-api-button-active` class selectors and using only the corresponding pseudo-classes, since we're not explicitly using those CSS classes at the moment. thanks for making these changes! approving now

mjlumetta (dismissed):
looks good! one suggestion but it's non-blocking ✨

mjlumetta (approved):
re-approving ✅

PR #609, opened 2021-01-21
Upgrade react-dev-utils to 11.0.1

anthony-kopczyk:
### Description
Jira Ticket: https://vajira.max.gov/browse/API-4498

Upgrades react-dev-utils to its latest version - 11.0.1

### Process

- [N/A] Docs updated
  

scripts/start.js
```diff
@@ -73,10 +73,25 @@ choosePort(HOST, DEFAULT_PORT)
...
+      config,
+      appName,
+      urls,
+      useYarn: false,
+    });
     // Load proxy config
     const proxySetting = require(paths.appPackageJson).proxy;
-    const proxyConfig = prepareProxy(proxySetting, paths.appPublic);
+
+    // prepareProxy takes a servedPathname but whether it exists or not it does
```
TS000, line 86: Interesting!

TS000 (dismissed):

config/webpackDevServer.config.js
```diff
@@ -101,7 +109,8 @@ module.exports = function(proxy, allowedHost) {
       // We do this in development to avoid hitting the production cache if
       // it used the same host and port.
       // https://github.com/facebookincubator/create-react-app/issues/2272#issuecomment-302832432
-      app.use(noopServiceWorkerMiddleware());
+      const servedPath = '.';
```
mjlumetta, line 112: how does this work? guessing the change is that the path is now required?

scripts/start.js
mjlumetta: couple of questions about this comment:

1. "whether it exists or not" - is that the pathname or the proxy config object in `package.json`?
2. when you say "the docs say it should be there", do you mean the `prepareProxy` call and its results or the `servedPathname` arg?

wondering how this should work given that we don't have any `proxy` config in `package.json`. with that said, if this is useful in terms of keeping us close to how Create React App is designed, that's good by me.

mjlumetta (dismissed):
nice job! couple of questions for my own understanding but looks good to merge. ✨ 🚀

scripts/start.js
anthony-kopczyk: I probably should have taken some more time on that comment. `prepareProxy` takes three argument. The third is servedPathname. If I don't pass a variable in there (or pass `undefined` I suppose) everything still seems to run fine. Since it's a variable in the docs though I figured I should at least set it to something and pass it in, to match the docs.

NPM docs link:
https://www.npmjs.com/package/react-dev-utils#prepareproxyproxysetting-string-apppublicfolder-string-servedpathname-string-object

config/webpackDevServer.config.js
anthony-kopczyk: Yes. This is similar to the change with `preparedProxy` except this actually causes a crash if it's absent.

This brings up a question. The NoopServiceWorkerMiddleware looks for `${servedPath}/service-worker.js` I thought we had a service-worker.js but couldn't find one anywhere in our files. Should this be set to something specific?

NPM docs link:
https://www.npmjs.com/package/react-dev-utils#noopserviceworkermiddlewareservedpath-string-expressmiddleware

scripts/start.js
anthony-kopczyk: With us not having the proxy it may just be getting ignored - would explain why an undefined isn't causing problems? It's probably carry over from the Create React App configurations.

I'd say we could look at taking that configuration out, but then if we ever need it again - not sure if that could end up making it more complicated.

mjlumetta: thanks for the clarifications, Anthony!

> With us not having the proxy it may just be getting ignored - would explain why an undefined isn't causing problems? It's probably carry over from the Create React App configurations.

I would _guess_ that this might be the case? part of what I was wondering is what is the effect of passing the empty string, and I'm wondering whether the empty string would cause some kind of error if we actually had proxy config. if it ignores this arg no matter what because `proxySetting` is undefined, then it's kind of hard to tell.

> I'd say we could look at taking that configuration out, but then if we ever need it again - not sure if that could end up making it more complicated.

I'm also not sure about this either. it may not matter that much either way. I think if we need it in the future, we'd probably have to figure out how it works and how to configure it properly at that time either way. I would say that there might be some value to leaving in default CRA config that we don't use so we do understand some kinds of features/tools that are available - but also, this is CRA config circa mid 2018, so a lot of these tools have probably changed, and CRA has probably changed a lot too. I'm fine either way.

config/webpackDevServer.config.js
mjlumetta: yeah, that's a good question. we have a file at `src/registerServiceWorker.js` that specifically unregisters any service workers that may have been previously registered. iirc, at one time we used them by default as enabled in Create React App, and we left that `unregister()` in case there were any remaining in client browsers. (think there's a more detailed comment in that file.)

similar to the `prepareProxy` config, I think we can remove this config if we'd like. I tried to get `http://localhost:3001/service-worker.js` to see if it auto-created this file, but it doesn't look like it? I'm also fine with leaving it if we feel like it's beneficial to leave some of this default CRA config that we don't use, but I don't think we need it either. pretty ambivalent - whatever you decide is fine by me.

mjlumetta (approved):
re-approving ✅

mjlumetta:
@anthony-kopczyk found the `service-worker.js` file...

Screen Shot 2021-01-25 at 10 16 32 AM

TS000 (approved):
Re-Approved 👍

PR #610, opened 2021-01-22
Add API Publishing landing page

christineyost:
### Add API Publishing landing page

This PR adds the API Publishing landing page described in this [ticket](https://vajira.max.gov/browse/API-4439)

Also added the TwoColumnLayout component for api-publishing page and reused it on several pages that have the same layout.

### Process

- [ ] Docs updated
  

.eslintrc.js
```diff
@@ -370,5 +370,11 @@ module.exports = {
         'max-nested-callbacks': 'off',
       },
     },
+    {
+      files: ['**/*.tsx'],
+      rules: {
+        'react/prop-types': 'off',
+      }
+    }
```
christineyost, line 378: Turned this rule off because it wasn't allowing parameter de-structuring

anthony-kopczyk: @mjlumetta Did this rule come up before when you were working on switching over to eslint? Not sure if we hit on this before. I'm all for destructuring in the method signature if we can.

anthony-kopczyk (commented):
This looks really good! Massive PR.

I don't have any changes specifically to request. I think if other reviews have the same suggestions then maybe we can change those to requests.

One other thing I forgot to comment as I was looking through the files is that we started adding some react prop-types to components to get runtime checking. I think the argument can be made that we only need that on components that are pulling from external interfaces. It may be worth considering adding those.

`GuidanceBox` in the `Banner.tsx` file has some PropTypes and uses `PropTypes.InferProps` to build the typescript typing.

src/containers/publishing/components/sectionWithIcon/SectionWithIcon.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { FC } from 'react';
+
+interface SectionWithIconProps {
+  imageFile: string;
+  header: string;
+}
+
+const SectionWithIcon: FC<SectionWithIconProps> = ({ children, imageFile, header }) => (
```
mjlumetta, line 8: is this component generic enough that it might be able to go in `src/components`?

src/containers/publishing/components/sectionWithIcon/SectionWithIcon.tsx
```diff
@@ -0,0 +1,20 @@
...
+  <section className="vads-u-display--flex vads-u-align-items--flex-start">
```
mjlumetta, line 9: should this section have an `aria-labelledby` pointing to the h3?  adding an `aria-labelledby` will make this section a region, meaning that screen readers will announce it as such when entering and leaving it. one benefit in particular that I can see to this behavior would be that the user won't think the card links are under the last h3, since it will announce the end of the region.

src/paths.tsx
```diff
@@ -0,0 +1,4 @@
+export const PUBLISHING_PATH = '/api-publishing';
+export const PUBLISHING_ONBOARDING_PATH = `${PUBLISHING_PATH}/process`;
+export const PUBLISHING_EXPECTATIONS_PATH = `${PUBLISHING_PATH}/expectations`;
+export const SUPPORT_CONTACT_PATH = '/support/contact-us';
```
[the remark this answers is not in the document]
mjlumetta: +1 to this - think it would be great to have a `src/types/constants/paths.ts` file or something like that. related question: any particular reason for making this a TSX file instead of plain old TS?

src/containers/News.test.tsx
```diff
@@ -40,23 +40,14 @@ describe('News', () => {
...
-      expect(cardLinks[index].children[0]).toHaveTextContent(section.title);
-      expect(cardLinks[index].children[1]).toHaveTextContent(section.description);
-    });
+    const newsReleasesCard = screen.getByRole('link', { name: 'News releases' });
+    const articlesCard = screen.getByRole('link', { name: 'Articles' });
+    const digitalMediaCard = screen.getByRole('link', { name: 'Digital media' });
+
+    expect(newsReleasesCard).toBeInTheDocument();
+    expect(articlesCard).toBeInTheDocument();
+    expect(digitalMediaCard).toBeInTheDocument();
```
alexandervalencia, line 50: Looks like we lost some expectations for the test... even though the test says that is `has a card link`, we should probably still check for the accuracy of the links as we were before. I could be totally wrong on this, I'd look to @mjlumetta for guidance.

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -76,39 +76,35 @@ const ReleaseNotes = (): JSX.Element => {
...
+        </SideNav>
+      }
+      right={
+        <Switch>
+          <Route exact path="/release-notes/" component={ReleaseNotesOverview} />
+          <Route
+            exact
+            path="/release-notes/deactivated"
+            component={DeactivatedReleaseNotes}
+          />
```
alexandervalencia, line 103:
```suggestion
          <Route exact path="/release-notes/deactivated" component={DeactivatedReleaseNotes} />
```

src/containers/support/Support.tsx
```diff
@@ -32,29 +32,29 @@ const sections: SupportSection[] = [
...
+      <SideNav ariaLabel="Support page side nav">
+        <SideNavEntry key="all" exact to="/support" name="Overview" />
+        {sections.map(section => (
+          <SideNavEntry key={section.id} to={`/support/${section.id}`} name={section.name} />
+        ))}
+      </SideNav>
+    }
+    right={
+      <Switch>
+        <Route exact path="/support/" render={(): JSX.Element => <SupportOverview sections={sections} />} />
```
alexandervalencia, line 46: My `prettier` is saying to do this instead:
```suggestion
        <Route
          exact
          path="/support/"
          render={(): JSX.Element => <SupportOverview sections={sections} />}
        />
```

src/containers/support/Support.tsx
```diff
@@ -32,29 +32,29 @@ const sections: SupportSection[] = [
...
+        {sections.map(section => (
```
mjlumetta, line 47: need to make changes to `SupportOverview`, `SupportFAQ`, and `SupportContactUs` to remove the section tags with `aria-label`/`aria-labelledby` as suggested above in the `TwoColumnLayout` file

mjlumetta (commented):
nice job on a huge PR! most of my comments here are either seconding @anthony-kopczyk's questions/suggestions or raising accessibility concerns with the new layout. I am thrilled that you extracted that common layout into a shared component - that's been a long time coming and it's super exciting to see that refactor!

src/components/twoColumnLayout/TwoColumnLayout.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { FC } from 'react';
+import classNames from 'classnames';
+
+interface TwoColumnLayoutProps {
+  left: React.ReactNode;
+  right: React.ReactNode;
+  className?: string;
+}
+
+const TwoColumnLayout: FC<TwoColumnLayoutProps> = ({ left, right, className }) => (
```
alexandervalencia, line 10: You can satisfy the prop-types lint rule by supplying the type for the object you're destructuring:
```ts
{ left, right, className }: TwoColumnLayoutProps
```

(This makes the line too long so it has to look like this:
```suggestion
const TwoColumnLayout: FC<TwoColumnLayoutProps> = ({
  left,
  right,
  className,
}: TwoColumnLayoutProps) => (
```

.eslintrc.js
alexandervalencia: I don't think we need to add this rule, see my comment on `TwoColumnLayout`

src/components/twoColumnLayout/TwoColumnLayout.tsx
```diff
@@ -0,0 +1,23 @@
...
+  <div className={classNames('vads-u-padding-y--5', className)}>
+    <div className="vads-l-grid-container">
+      <div className="vads-l-row">
+        {left}
+        <section aria-label="content" className={classNames('vads-l-col--12', 'medium-screen:vads-l-col--8')}>
```
alexandervalencia, line 15: This line is too long:
```suggestion
        <section
          aria-label="content"
          className={classNames('vads-l-col--12', 'medium-screen:vads-l-col--8')}
        >
```

src/components/twoColumnLayout/TwoColumnLayout.tsx
```diff
@@ -0,0 +1,23 @@
+import React, { FC } from 'react';
```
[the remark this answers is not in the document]
christineyost: Learned about this from @RockslideJr !

src/components/twoColumnLayout/TwoColumnLayout.tsx
christineyost: Agree, but was hoping not to do this since it should be inferred from `FC<TwoColumnLayoutProps>`

src/components/twoColumnLayout/TwoColumnLayout.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+const TwoColumnLayout: FC<TwoColumnLayoutProps> = ({ left, right, className }) => (
+  <div className={classNames('vads-u-padding-y--5', className)}>
+    <div className="vads-l-grid-container">
+      <div className="vads-l-row">
+        {left}
+        <section aria-label="content" className={classNames('vads-l-col--12', 'medium-screen:vads-l-col--8')}>
+          {right}
+        </section>
```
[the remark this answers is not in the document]
christineyost: Yes, this is only used in places where the left is nav. This sounds good to me!

src/containers/publishing/components/sectionWithIcon/SectionWithIcon.test.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import checkImage from '../../../../assets/check.svg';
+import SectionWithIcon from './SectionWithIcon';
+
+describe('SectionWithIcon', () => {
+  beforeEach(() => {
+    render(
+      <SectionWithIcon imageFile={checkImage} header="Test">
+        This is a child.
```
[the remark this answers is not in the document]
christineyost: Happy to do so! 😄

src/containers/publishing/components/sectionWithIcon/SectionWithIcon.tsx
christineyost: Yes, it is very generic. I wasn't sure if it should go there yet because no other pages are currently making use of anything like it, but if generic components typically go in `src/components` even if they're not yet shared I can move this there

src/containers/publishing/components/sectionWithIcon/SectionWithIcon.tsx
christineyost: Yes, this sounds like a good idea! Will add it in!

src/paths.tsx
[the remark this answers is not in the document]
christineyost: Sounds good! I noticed the paths in `src/types/constants/index.ts` look like they're all paths to the backend. Should all these paths live in the same `src/types/constants/paths.ts` file?

Making this a TSX file was a mistake 😄

src/components/twoColumnLayout/TwoColumnLayout.tsx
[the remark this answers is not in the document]
christineyost: Is it ok to reference the `PAGE_HEADER_ID` constant directly in this file? Or would it be better to have that passed in?

src/containers/News.test.tsx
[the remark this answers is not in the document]
christineyost: This was changed because with the original TwoColumnLayout code we lost the `'News'` aria label, but now that that's back in this test doesn't need to change.

src/containers/publishing/Publishing.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { FC } from 'react';
+import { Route, Switch } from 'react-router';
+import TwoColumnLayout from '../../components/twoColumnLayout';
+import { SUPPORT_CONTACT_PATH, PUBLISHING_EXPECTATIONS_PATH, PUBLISHING_ONBOARDING_PATH, PUBLISHING_PATH } from '../../paths';
+import { SideNav, SideNavEntry } from '../../components';
+import PublishingIntroduction from './components/publishingIntroduction';
```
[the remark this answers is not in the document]
christineyost: The idea of this pattern was to separate the main Publishing page from the other components. This directory is meant to hold Publishing-specific components like the SectionWithIcon which is not used anywhere else on the site. I also included PublishingIntroduction here because it's a kind of secondary component that is rendered by the main Publishing page. 

Happy to hear feedback on this!

src/containers/publishing/components/sectionWithIcon/SectionWithIcon.tsx
```diff
@@ -0,0 +1,20 @@
...
+    <div className="vads-u-display--flex vads-u-justify-content--center vads-u-flex--1">
+      <img src={imageFile} alt="" role="presentation" />
```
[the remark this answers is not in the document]
christineyost: Got it! I thought `presentation` was necessary to do what `alt=""` does, but I see it's necessary. Removed this role and updated the tests to find the image by the `img` role instead

.eslintrc.js
christineyost: Shouldn't the second typing be unnecessary since the prop type is provided in `FC<ContentWithNavProps>`

src/components/sideNav/SideNav.test.tsx
```diff
@@ -31,21 +30,14 @@ describe('SideNav', () => {
...
 
     const skipContentLink = screen.queryByText('Skip Page Navigation');
     expect(skipContentLink).toBeInTheDocument();
     expect(skipContentLink).toHaveAttribute('href', '/#page-header');
 
-    const navigation = screen.queryByRole('navigation');
-    expect(navigation?.className).toContain('custom-class');
-    expect(navigation).toBeInTheDocument();
-    expect(navigation).toHaveAttribute('aria-label', ariaLabel);
-
```
christineyost, line 48: These checks are now unnecessary since this is in the `ContentWithNav` component now

src/components/sideNav/SideNav.tsx
```diff
@@ -51,14 +44,13 @@ const SideNav = (props: SideNavProps): JSX.Element => {
       >
         Skip Page Navigation
       </HashLink>
-      <nav
```
christineyost, line 54: Moving the `nav` tag and aria label to `ContentWithNav` will now include the `Skip Page Navigation` element in the `nav` section. Is this ok?

src/containers/News.tsx
```diff
@@ -75,63 +76,59 @@ const MediaItem = ({ item }: { item: NewsItemData }): JSX.Element => {
   );
 };
 
+// eslint-disable-next-line no-confusing-arrow
 const NewsItem = ({ item, media }: { item: NewsItemData; media: boolean }): JSX.Element =>
-  (media ? <MediaItem item={item} /> : <ItemDescription item={item} />);
+  media ? <MediaItem item={item} /> : <ItemDescription item={item} />;
```
christineyost, line 81: Prettier is getting rid of these parenthesis for me, but then eslint complains about a `no-confusing-arrow` rule violation. Should we turn this rule off?

src/containers/documentation/ApiPage.test.tsx
```diff
@@ -80,10 +80,6 @@ describe('ApiPage', () => {
       expect(lookupApiCategoryMock).toHaveBeenCalledWith('lotr');
     });
 
-    it('renders region', () => {
-      expect(screen.getByRole('region')).not.toBeNull();
-    });
-
```
christineyost, line 86: Removed this since the region was moved to the `ContentWithNav` component

src/paths.tsx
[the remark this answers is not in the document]
mjlumetta: that's a good callout on frontend paths vs backend paths. I'd vote for keeping them separate, I think? might be worth adding a smaller block comment at the top to clarify the purpose of the file, although idt that's necessary

src/components/twoColumnLayout/TwoColumnLayout.tsx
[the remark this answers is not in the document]
mjlumetta: you can reference it directly in this file, it shouldn't need to be a prop or anything like that. (if that's what you mean there - want to make sure I'm responding to what you're actually asking)

src/containers/publishing/Publishing.tsx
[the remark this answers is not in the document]
mjlumetta: imo, it's fine to put any of those components directly in the `src/containers/publishing` directory. for other pages on the site that map to directories in `src/containers` (`apply`, `documentation`, `releaseNotes`, and `support`), you can see a top-level root component and then several other components for sub-pages that show up in the side nav, and sometimes also smaller components too. that said, if you want to nest some components, I think that's fine too; I don't feel like I get the `publishing/components` directory structure right off the bat, necessarily, but I'm also not opposed either. so I'm open to discussion on what you and other people feel is most useful here. could also consider other nested directory structures as well.

more general side note on the containers/components pattern: I don't necessarily find it super useful, personally. whether that's true in general or because we don't follow it very strictly, I'm not sure, but we definitely use it very loosely. so it might be worth discussing as a larger group how useful or not that is at some point down the line.

src/containers/documentation/DocumentationRoot.tsx
```diff
@@ -129,39 +128,29 @@ const DocumentationRoot = (): JSX.Element => {
...
+          <Route exact path="/explore/" component={DocumentationOverview} />
+          <Route exact path="/explore/:apiCategoryKey" component={CategoryPage} />
+          <Route exact path="/explore/:apiCategoryKey/docs/authorization">
+            {authDocsV2 ? <Redirect to="/explore/authorization" /> : <AuthorizationDocsLegacy />}
+          </Route>
+          <Route exact path="/explore/:apiCategoryKey/docs/quickstart" component={QuickstartPage} />
+          <Route exact path="/explore/:apiCategoryKey/docs/:apiName" component={ApiPage} />
+        </Switch>
+      }
+      navAriaLabel="API Docs Side Nav"
```
mjlumetta, line 151: a side note unrelated to your changes but perhaps worth discussing here: is it worth considering alternative wordings for this label? the "side" in "side nav" is really a visual description that is unlikely to be very meaningful to many screen reader users (users with complete/near-complete blindness, mainly). what do we think about something like "API Docs Page Navigation" (and the corresponding values for other pages)?

src/components/contentWithNav/ContentWithNav.tsx
```diff
@@ -0,0 +1,37 @@
...
+          className={classNames(
+            'vads-l-col--12',
+            'vads-u-padding-right--5',
+            'medium-screen:vads-l-col--4',
+          )}
+        >
+          {nav}
+        </nav>
+        <section
+          aria-labelledby={PAGE_HEADER_ID}
```
mjlumetta, line 27: sorry to drop one more thing on this `aria-labelledby` consideration, but might need to add an optional prop to override `PAGE_HEADER_ID`... in `CategoryReleaseNotes`, there's a different id for that section that includes the h1 and the halo above it. looks like we might need to remove that `section` tag because there are 2 in that case and then use that prop in this case. lmk if you have thoughts about this - I _think_ that's the last stray wrinkle re: this refactor.

mjlumetta (commented):
looks really close to me. got a couple more questions and suggested tweaks but I think it's in great shape

src/containers/News.tsx
anthony-kopczyk: No idea if this will work, but maybe this structure will satisfy both? I'd hope the left one right after the arrow should keep it from being 'confusing' and prettier shouldn't try and clean up the parentheses if they're on separate lines.
```
const NewsItem = ({ item, media }: { item: NewsItemData; media: boolean }): JSX.Element => (
    media ? <MediaItem item={item} /> : <ItemDescription item={item} />
);
```

src/containers/releaseNotes/ReleaseNotes.tsx
```diff
@@ -31,7 +30,7 @@ const SideNavAPIEntry = (props: SideNavAPIEntryProps): JSX.Element => {
               <small>Internal VA use only.</small>
             </span>
           )}
-          {(api.vaInternalOnly && api.trustedPartnerOnly) ? <br /> : null}
+          {api.vaInternalOnly && api.trustedPartnerOnly ? <br /> : null}
```
anthony-kopczyk, line 33: Why were these parentheses removed? Prettier? I like them in this case because it groups the condition - makes it a little easier for me to read.

src/containers/documentation/DocumentationRoot.tsx
anthony-kopczyk: I like this. Maybe 'Page Navigation for API Docs', so screen readers that use the site regularly can hear 'Page Navigation' and know if they want to skip, without waiting for the page title. Or maybe we don't need the page title? Not sure. I'm on board this being another ticket/PR if this ends up slowing this one down.

anthony-kopczyk (commented):
Changes look good. Haven't found anything major to note, but I see others have some more questions so I'm just commenting for now.

src/components/sideNav/SideNav.tsx
christineyost: I like this change, definitely cleaner than what I was doing. Adding that "Skip Page Navigation" to the main nav made some tests weird too - they're cleaner now

src/containers/documentation/ApiPage.test.tsx
```diff
@@ -80,10 +80,6 @@ describe('ApiPage', () => {
       expect(lookupApiCategoryMock).toHaveBeenCalledWith('lotr');
     });
 
-    it('renders region', () => {
```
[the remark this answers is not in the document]
christineyost: This region is now rendered by the ContentWithNav component instead of the ApiPage

src/containers/documentation/DocumentationRoot.tsx
christineyost: I like this too! Agree, let's make another ticket unless we come to a decision on the wording quickly.

src/containers/releaseNotes/ReleaseNotes.tsx
christineyost: Yes, this was Prettier

src/containers/News.tsx
christineyost: Unfortunately this doesn't work either. Prettier really doesn't want the parenthesis 😞

src/components/sideNav/SideNav.tsx
alexandervalencia: Agreed with @mjlumetta on this one, having the `<nav>` tag outside of the `SideNav` component feels off to me. I like the suggestion he gave as well

alexandervalencia (commented):
Great work taking in feedback and continuing to work it through, I think @mjlumetta and @anthony-kopczyk covered some great points already!

alexandervalencia (approved):
LGTM! Great work Christine! 🚀

alexandervalencia (approved):
LGTM ✨ 🚀

src/containers/News.tsx
anthony-kopczyk: Wow. That's intense. Yeah, like Mike I don't have a preference one way or the other on handling this. That would have been nice if it had worked.

src/components/contentWithNav/ContentWithNav.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface ContentWithNavProps {
+  nav: React.ReactNode;
+  content: React.ReactNode;
+  navAriaLabel: string;
+  contentAriaLabelledBy?: string;
+  className?: string;
+}
+
+const ContentWithNav: FC<ContentWithNavProps> = ({ nav, content, navAriaLabel, contentAriaLabelledBy, className }) => (
```
mjlumetta, line 14: this final product for this component looks amazing 💅 🙌

src/containers/publishing/components/sectionWithIcon/SectionWithIcon.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+    );
+  });
+
+  it('renders successfully', () => {
+    const region = screen.getByRole('region', { name: 'Hobbits' });
+    expect(region).toBeInTheDocument();
+  });
+
+  it('contains the expected content', () => {
+    const image = screen.getByRole('img');
```
mjlumetta, line 21: i'm a bit surprised this test works given that the [spec](https://www.w3.org/TR/html-aria/#docconformance) says images with `alt=""` shouldn't have any role, but if that's how the Testing Library interprets it, works for me 🤷

mjlumetta (approved):
great job Christine!! ✨

src/containers/publishing/components/sectionWithIcon/SectionWithIcon.test.tsx
christineyost: Yeah, that is a bit odd, but this works...

PR #611, opened 2021-01-25
Api 4416/mdx props

ediiotero:
### Using Props with MDX
Story: [API-4416](https://vajira.max.gov/browse/API-4416)

Using props with MDX files to be used with the new API selector in the new Auth docs page and remove the mdx files we are not using.

- [ ] Docs updated

### Requested feedback

Had trouble getting syntax highlighting to work without the Syntax Highlighter component. The goal is to get it working as in the old auth docs and use props. Lost highlighting and some styling without Syntax Highlighter component in the MDX.


src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -0,0 +1,151 @@
+---
+imports:
+  '{ HashLink }' : 'react-router-hash-link'
+  '{ Link }' : 'react-router-dom'
+  '{APISelector, CodeWrapper }': ../index
```
gileswells, line 5: nitpicky space needed here
```suggestion
  '{ APISelector, CodeWrapper }': ../index
```

src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -0,0 +1,151 @@
...
+<SyntaxHighlighter language="plaintext" style="Prism">
+{{ props.propObj.codeGrant }}
+</SyntaxHighlighter>
+
+</CodeWrapper>
+
+Use the following format, in HTTP basic authentication, for your request using the returned code and state parameters.
+
+- Use your client ID and client secret as the HTTP basic authentication username and password.
+- Be sure to replace `< yourRedirectURL >` with the redirect URL that you provided during registration.
```
gileswells, line 45: In other instances we've used `&lt;` and `&gt;` instead of adding spaces like this to prevent the markdown loader from trying to render a component. Does that not work here?

src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -0,0 +1,151 @@
...
+<APISelector options={ props.propObj.options } selectedOption={ props.propObj.selectedOption } />
+<CodeWrapper>
+
+<SyntaxHighlighter language="http" style="Prism">
+{{ props.propObj.postToken }}
+</SyntaxHighlighter>
+
+</CodeWrapper>
+
+The authorization server will respond with an <HashLink to='#id-token'>access token</HashLink>. If you requested the `offline_access` scope, you will also receive a`refresh_token`. The response will look like this:
```
gileswells, line 56: Do these `<HashLink />`s still hold the query string parameter for the selected API? I had an issue where I'd need to include the location object in order for these to work properly.

src/types/md.d.ts
```diff
@@ -9,6 +9,8 @@ declare module '*.mdx' {
   interface MarkdownComponentProps {
     className?: string;
     style?: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    propObj?: any;
   }
```
gileswells, line 14: I'd probably add a comment about why you're disabling the any check here just for clarity sake.

gileswells (changes requested):
Looks solid overall. I wonder if now's the right time to consider bringing the code samples fully into the mdx files and relying less on the `propObj` to bring in the whole code sample but instead the base url and scopes (if markdown doesn't throw a fit of course).

src/types/md.d.ts
ediiotero: changed this to unknow, but yeah you're right. I'll add a comment

src/components/oauthDocs/AuthCodeFlowContent.mdx
ediiotero: seems to be working

src/components/oauthDocs/AuthCodeFlowContent.mdx
ediiotero: I removed the spaces and it seems to work fine. Maybe because of the backticks?

src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -0,0 +1,218 @@
...
+
+**NOTE:** the <HashLink to='#id-token'>access token</HashLink> will only work for the API and scopes for which you have previously initiated authorization. If you need additional scopes in the future, you will need to build a new authorization URL with the additional scopes and have the Veteran grant consent again.
+
+Use the `refresh_token` to obtain a new `access_token` after its expiry by sending the following request.
+
+<APISelector options={ props.propObj.options } selectedOption={ props.propObj.selectedOption } />
+<CodeWrapper>
+
+```http
+POST {{props.propObj.apiDef?.oAuthInfo?.baseAuthPath ?? '/oauth2'}}/revoke HTTP/1.1
```
mjlumetta, line 123: not part of this PR, but think this is actually the token endpoint

```suggestion
POST {{props.propObj.apiDef?.oAuthInfo?.baseAuthPath ?? '/oauth2'}}/token HTTP/1.1
```

src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -0,0 +1,218 @@
...
+The response will return a new `access_token` and `refresh_token`, if you requested the `offline_access` scope.
+
+#### Manage Account {#manage-account}
+
+The manage endpoint directs end users to a URL where they can view which applications currently have access to their data and can make adjustments to these access rights (grants).
+
+<CodeWrapper>
+
+```http
+POST {{props.propObj.apiDef?.oAuthInfo?.baseAuthPath ?? '/oauth2'}}/token HTTP/1.1
```
mjlumetta, line 142:
```suggestion
POST {{props.propObj.apiDef?.oAuthInfo?.baseAuthPath ?? '/oauth2'}}/manage HTTP/1.1
```

src/components/oauthDocs/PKCEAuthContent.mdx
```diff
@@ -0,0 +1,130 @@
+---
+imports:
+  '{ HashLink }' : 'react-router-hash-link'
+  '{ Link }' : 'react-router-dom'
+  '{ APISelector, CodeWrapper }': ../index
+  'SyntaxHighlighter' : 'react-syntax-highlighter'
+  '{ Prism }' : 'react-syntax-highlighter/dist/cjs/prism'
```
mjlumetta, line 7:
```suggestion
```

left over stuff

src/containers/documentation/AuthorizationDocs.scss
```diff
@@ -82,10 +83,8 @@
   }
 
   .api-selector + .code-wrapper {
-    pre > code {
-      border-top-right-radius: 0;
-      border-top-left-radius: 0;
-    }
+    border-top-right-radius: 0;
+    border-top-left-radius: 0;
```
mjlumetta, line 87: what's the cause of the changes in this file, here and above?

src/types/md.d.ts
```diff
@@ -6,9 +6,13 @@ declare module '*.mdx' {
    * configuration data. This constrained type signature is used here as a statement of our
    * preference to limit our use the component loader to its basic functionality.
    */
+  /**
+   * adding 'propObj' so the mdx files are open to any props that are passed
```
mjlumetta, line 10: would you mind folding this sentence into the previous comment and updating the previous comment to reflect the current changes? they're addressing the same thing, and the previous comment is no longer accurate as of this change.

src/components/oauthDocs/AuthCodeFlowContent.tsx
```diff
@@ -17,257 +13,16 @@ const AuthCodeFlowContent = (): JSX.Element => {
...
-  const syntaxColor = Prism;
+  const propObj = {
+    apiDef,
+    options: selectorProps.options,
+    selectedOption: selectorProps.selectedOption,
+  };
 
   return (
     <section aria-labelledby="authorization-code-flow">
-      <h3 id="authorization-code-flow" tabIndex={-1}>
```
mjlumetta, line 36: if all of the markup for this component is in the MDX file now, do we still need this TSX file?

src/components/oauthDocs/AuthCodeFlowContent.mdx
mjlumetta: yeah, this isn't working for me, although the issue isn't removing the query string per se. here's what i'm seeing:

1. link has an href to `/explore/authorization#id-token`.
2. clicking the link sets the URL to that path + hash.
3. the top-level `AuthorizationDocs` component re-renders, and the hook sets the query param to the currently selected API from the Redux store. in the process, it removes the hash, so the URL is `<host>/explore/authorization?api=claims` (or whichever API). so I think there is an issue to be addressed here, which would require using the `location` object.

mjlumetta (commented):
bunch of miscellaneous comments and questions, plus a few proofreading things not specific to this PR. looking good in general.

.babelrc
```diff
@@ -1,6 +1,8 @@
 {
   "plugins": [
-    "@babel/plugin-proposal-class-properties"
+    "@babel/plugin-proposal-class-properties",
+    "@babel/plugin-proposal-optional-chaining",
+    "@babel/plugin-proposal-nullish-coalescing-operator"
```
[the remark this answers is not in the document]
ediiotero: They were needed for the optional chaining in props for the mdx files.

src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -0,0 +1,218 @@
+---
+imports:
+  '{ HashLink }' : 'react-router-hash-link'
+  '{ Link }' : 'react-router-dom'
+  '{ APISelector, CodeWrapper }': ../index
+  'SyntaxHighlighter' : 'react-syntax-highlighter'
+  '{ Prism }' : 'react-syntax-highlighter/dist/cjs/prism'
```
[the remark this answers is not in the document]
ediiotero: thanks missed those

src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -0,0 +1,218 @@
...
+  'AuthCodeFlowQueryParamsTable': ./AuthCodeFlowQueryParamsTable.mdx
```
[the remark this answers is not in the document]
ediiotero: yeah you're right I'll fix that

src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -0,0 +1,218 @@
...
+
+**NOTE:** We provide a sample [Node.JS](https://nodejs.org/en/) application for demonstrating how to get up and running with our OAuth system. You can find the complete source code for it on our [GitHub](https://github.com/department-of-veterans-affairs/vets-api-clients/tree/master/samples/oauth_node).
+
+Begin the OpenID Connect authorization by using the authorization endpoint, query parameters, and scopes listed below.
+
+<APISelector options={ props.propObj.options } selectedOption={ props.propObj.selectedOption } />
+<CodeWrapper>
+
+```plaintext
+https://sandbox-api.va.gov{{props.propObj.apiDef?.oAuthInfo?.baseAuthPath ?? '/oauth2'}}/authorization?
```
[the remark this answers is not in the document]
ediiotero: ok I'll make that change

src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -0,0 +1,218 @@
...
+
+After a Veteran gives authorization for you to access their data, their browser will be redirected to your application with the response shown below, which returns the `code` and `state` parameters you must use to make a request to our authorization service. We require the state parameter for all authorization code grant flows.
+
+<CodeWrapper>
+
+```plaintext
+GET <yourRedirectURL>?
+  code=z92dapo5
+  &state=af0ifjsldkj
+  Host: <yourRedirectHost>
```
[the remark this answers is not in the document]
ediiotero: I guess I missed this comment, the doc has been updated since I worked on the static content. I'll remove it

src/components/oauthDocs/PKCEAuthContent.mdx
ediiotero: done

src/containers/documentation/AuthorizationDocs.scss
ediiotero: removing the SyntaxHighlighter component removed the border-radius syles and had to adjust

src/types/md.d.ts
ediiotero: done

src/components/oauthDocs/AuthCodeFlowContent.tsx
ediiotero: I'll work on merging this file and the PKCE content file into the Building OIDC Content file

src/types/md.d.ts
gileswells: There's a different implementation method for this change in another PR and despite suggesting using a `propObj` param initially I think that I prefer the other method now having seen it.

https://github.com/department-of-veterans-affairs/developer-portal/pull/615/files?diff=unified&w=1#diff-38e9706b08e494d5b2d4e43026117b5e5091e090ddd07716508ba634334bdf95R10

gileswells (changes requested):
Check the implementation for mdx props on #615 and consider changing over to that style with wildcard prop names.

.babelrc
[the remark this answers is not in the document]
mjlumetta: yeah, makes sense. feels strange to have to have it from Babel and TS, but that's fine

mjlumetta (approved):
I've got a couple general proofreading things for the auth docs, but your changes look good 💯

mjlumetta:
oh wait - should `react-syntax-highlighter` be uninstalled?

gileswells (approved):
LGTM Nice work on making this 1000% easier to manage going forward.

mjlumetta:
@gileswells ready for review whenever you're ready

src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -195,10 +196,7 @@ DELETE {{props.apiDef?.oAuthInfo?.baseAuthPath ?? '/oauth2/{api}/v1'}}/grants HT
 Host: sandbox-api.va.gov
 Content-Type: application/x-www-form-urlencoded
 
-{
-    "client_id": {client_id},
-    "email": {test account email}
-}
+client_id={client_id}&email={test account email}
```
gileswells, line 199: My only thing that I don't particularly like is the use of `test account email` here. I'd prefer that it stick with the snake case we have of pseudo fields like `client_id`. Also if you're calling it `test_account_email` then I would argue that `test_account_client_id` would be a more harmonious variable name for `client_id` but my vote would just be to use `email_address` because even though these are sandbox examples I would think that `client_id` pairs with `email_address` better.

Not worth failing the review on this alone but just something I noted.

gileswells (approved):
LGTM

src/components/oauthDocs/AuthCodeFlowContent.mdx
mjlumetta: it's a good point on the snake case. I don't think `test_account_client_id` makes much sense in that way because the client id isn't actually associated with a test user, although the email is.

PR #613, opened 2021-01-27
Api 4583/update header footer

christineyost:
### Description
This PR is for [API-4583](https://vajira.max.gov/browse/API-4583). It updates the header so the main title is 'Lighthouse APIs' instead of only 'Lighthouse'. It also updates the links in the footer (removes Micro-consulting and adds Contact Us) and shifts them to be aligned with the right side of the page.

### Process

- [ ] Docs updated
  

src/components/footer/Footer.test.tsx
```diff
@@ -1,13 +1,56 @@
 import * as React from 'react';
+import { render, screen } from '@testing-library/react';
 
 import 'jest';
```
rtravitz, line 4: I know you didn't change this, but do we need explicitly import jest?

rtravitz (approved):
This looks good to me from a functional standpoint. I don't feel qualified to champion any accessibility standards here, however. Would be good to get somebody else.

src/components/footer/Footer.test.tsx
christineyost: Nope! This isn't necessary

src/components/footer/Footer.test.tsx
```diff
@@ -1,13 +1,55 @@
...
-import 'jest';
+import { render, screen } from '@testing-library/react';
 
 import { shallow } from 'enzyme';
+import { BrowserRouter as Router } from 'react-router-dom';
+import { FlagsProvider, getFlags } from '../../flags';
 import { Footer } from './Footer';
 
 describe('Footer', () => {
   it('There should be no more beta footer banner', () => {
```
mjlumetta, line 10: this test is an old one for something that's been removed for a while. maybe we can get rid of it here, get rid of `enzyme`, and kill the inner describe block?

mjlumetta (dismissed):
yeah looks good to me! only thing I'd point out, other than removing that test that you didn't write if you feel up to that, would be that adding "APIs" to the header makes the breakpoints a little messier... so if we want to massage the size of the button and search field or the link text or something, that's a thought.

Screen Shot 2021-01-28 at 11 32 25 AM

src/components/footer/Footer.tsx
```diff
@@ -21,70 +22,71 @@ const Footer: React.FunctionComponent = (): JSX.Element => (
...
-        <div
+    <div
+      className={classNames(
+        'vads-u-display--flex',
+        'vads-u-margin-x--2',
+        'vads-u-flex-direction--column',
+        'medium-screen:vads-u-margin-x--4',
+        'medium-screen:vads-u-flex-direction--row',
+      )}
+    >
```
RockslideJr, line 33: I think this div is extraneous. If we move these classes onto the footer, change the margin to padding, and remove this div i think we would get the same result?

src/components/footer/Footer.tsx
```diff
@@ -21,70 +22,71 @@ const Footer: React.FunctionComponent = (): JSX.Element => (
...
+        className={classNames(
+          'vads-u-margin-y--1p5',
+          'vads-u-flex--auto',
+          'medium-screen:vads-u-margin-y--2p5',
+        )}
+      >
+        <a href="https://www.va.gov">
+          <img src={logo} className="va-api-footer-logo" alt="Department of Veterans Affairs" />
+        </a>
+      </div>
```
RockslideJr, line 44: i didn't do the checking on this one, but we can probably eliminate this div too.

RockslideJr (approved):
A few thoughts on removing extra divs, but looks great!

src/components/header/Header.scss
```diff
@@ -16,6 +16,11 @@
   position: relative;
 }
 
+a.va-api-apply-button {
+  font-size: 1.4rem;
```
mjlumetta, line 20: bit worried about this font size... think we're going to want it to be at least 16px/1.6rem for accessibility/easy reading reasons. as is, there's still a small ~15 pixel window where the banner wraps, and so I think it's okay if we leave a small amount of wrapping, even if it's a wider window. the thing that I'd actually suggest if we wanted the wrapping to never happen would be to mess with the "VA | Lighthouse APIs" text size... maybe should've led with that on the other comment. sorry for being misleading there.

src/components/header/Header.tsx
```diff
@@ -72,7 +74,7 @@ const Header = (): JSX.Element => {
                 'medium-screen:vads-u-font-size--2xl',
```
mjlumetta, line 74:
```suggestion
                'medium-screen:vads-u-font-size--xl',
                'small-desktop-screen:vads-u-font-size--2xl',
```

my bad, my recommendation earlier was unnecessarily vague and not very helpful, sorry about that. here's what I would suggest might be more helpful. I'm also fine if there is some wrapping; it's more important that everything be large enough to read than that the wrapping be pixel-perfect. sorry if the first comment was misleading.

mjlumetta (commented):
think my comments earlier about the wrapping/responsive styles were probably not the most helpful - should've been more specific. don't think there's enough size to cut off those components (search box is probably okay being smaller, but the text on the Apply link probably needs to be at least 16px). I left a suggestion for how to responsively add a 30px level for the banner link text, but I am also okay with some wrapping happening too (not ideal but also not the end of the world! that's what flex is for, after all).

src/components/footer/Footer.tsx
```diff
@@ -4,88 +4,79 @@ import { Link } from 'react-router-dom';
...
-            'vads-u-margin-y--1p5',
-            'medium-screen:vads-u-margin-y--2p5',
-            'vads-l-col--12',
-            'medium-screen:vads-l-col--4',
-          )}
-        >
-          <a href="https://www.va.gov">
-            <img src={logo} className="va-api-footer-logo" alt="Department of Veterans Affairs" />
+  <footer role="contentinfo">
+    <div className={yellowDivStyles} />
```
mjlumetta, line 48: any particular reason for making this its own div instead of a border-top on the footer?

mjlumetta (approved):
looks great!! one question but it shouldn't block merging 😄 🚀

mjlumetta (approved):

PR #615, opened 2021-01-29
API-4440 - API Publishing Expectations page

alexandervalencia:
### Description

This PR supports [API-4440](https://vajira.max.gov/browse/API-4440) by adding an API Expectations page.

Because there was so much content, I opted to use markdown files (`.mdx`). In regards to testing that the markdown is being correctly loaded: our Jest configuration does not currently handle loading markdown. I explored trying to make that a reality, but unfortunately, with packages that are currently available (`jest-mdx-loader` and `jest-transformer-mdx` were the two I tested), it's not possible to load the markdown files for testing—at least not possible in a "non-hacky" way.

At some point in the future if we want to test markdown content properly across the entire developer-portal, I would recommend looking into adapting the custom fork we have of the `markdown-component-loader` to be adapted to Jest.

### Process

- [ ] Docs updated
  
### DIFF
--- a/src/containers/publishing/components/publishingExpecations/PublishingExpectations.test.tsx
+++ b/src/containers/publishing/components/publishingExpecations/PublishingExpectations.test.tsx
@@ -2,13 +2,16 @@ import React from 'react';
 import { render, screen } from '@testing-library/react';
 import { MemoryRouter } from 'react-router';
 import { PublishingExpectations } from './PublishingExpectations';
+import { FlagsProvider, getFlags } from '../../../../flags';
 
 describe('PublishingExpectations', () => {
   beforeEach(() => {
     render(
-      <MemoryRouter initialEntries={['/api-publishing/expectations']}>
-        <PublishingExpectations />
-      </MemoryRouter>,
+      <FlagsProvider flags={getFlags()}>
+        <MemoryRouter initialEntries={['/api-publishing/expectations']}>
+          <PublishingExpectations />
+        </MemoryRouter>
+      </FlagsProvider>,
     );
   });
 
@@ -25,47 +28,17 @@ describe('PublishingExpectations', () => {
     });
   });
 
-  describe('security', () => {
+  describe.each([
+    ['Security'],
+    ['Documentation'],
+    ['Versioning'],
+    ['Monitoring'],
+    ['Support'],
+    ['Modernization'],
+    ['Authentication and Authorization'],
+  ])('%s', name => {
     it('renders succesfully', () => {
-      expect(screen.getByRole('heading', { name: 'Security' })).toBeInTheDocument();
-    });
-  });
-
-  describe('documentation', () => {
-    it('renders succesfully', () => {
-      expect(screen.getByRole('heading', { name: 'Documentation' })).toBeInTheDocument();
-    });
-  });
-
-  describe('versioning', () => {
-    it('renders succesfully', () => {
-      expect(screen.getByRole('heading', { name: 'Versioning' })).toBeInTheDocument();
-    });
-  });
-
-  describe('monitoring', () => {
-    it('renders succesfully', () => {
-      expect(screen.getByRole('heading', { name: 'Monitoring' })).toBeInTheDocument();
-    });
-  });
-
-  describe('support', () => {
-    it('renders succesfully', () => {
-      expect(screen.getByRole('heading', { name: 'Support' })).toBeInTheDocument();
-    });
-  });
-
-  describe('modernization', () => {
-    it('renders succesfully', () => {
-      expect(screen.getByRole('heading', { name: 'Modernization' })).toBeInTheDocument();
-    });
-  });
-
-  describe('authentication and authorization', () => {
-    it('renders succesfully', () => {
-      expect(
-        screen.getByRole('heading', { name: 'Authentication and Authorization' }),
-      ).toBeInTheDocument();
+      expect(screen.getByRole('heading', { name })).toBeInTheDocument();
     });
   });
 });
--- a/src/containers/publishing/components/publishingExpecations/PublishingExpectations.test.tsx
+++ b/src/containers/publishing/components/publishingExpecations/PublishingExpectations.test.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
 import { MemoryRouter } from 'react-router';
-import { PublishingExpectations } from './PublishingExpectations';
 import { FlagsProvider, getFlags } from '../../../../flags';
+import { PublishingExpectations } from './PublishingExpectations';
 
 describe('PublishingExpectations', () => {
   beforeEach(() => {
--- a/src/containers/publishing/components/publishingExpecations/PublishingExpectations.test.tsx
+++ b/src/containers/publishing/components/publishingExpecations/PublishingExpectations.test.tsx
@@ -1,0 +1,71 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { MemoryRouter } from 'react-router';
+import { PublishingExpectations } from './PublishingExpectations';
+
+describe('PublishingExpectations', () => {
+  beforeEach(() => {
+    render(
+      <MemoryRouter initialEntries={['/api-publishing/expectations']}>
+        <PublishingExpectations />
+      </MemoryRouter>,
+    );
+  });
+
+  it('renders successfully', () => {
+    expect(
+      screen.getByRole('heading', { name: 'Expectations for Lighthouse APIs' }),
+    ).toBeInTheDocument();
+  });
+
+  it('renders all accordions in a closed state', () => {
+    const buttons = screen.getAllByRole('button');
+    buttons.forEach(button => {
+      expect(button.getAttribute('aria-expanded')).toBe('false');
+    });
+  });
+
+  describe('security', () => {
+    it('renders succesfully', () => {
+      expect(screen.getByRole('heading', { name: 'Security' })).toBeInTheDocument();
+    });
+  });
+
+  describe('documentation', () => {
+    it('renders succesfully', () => {
+      expect(screen.getByRole('heading', { name: 'Documentation' })).toBeInTheDocument();
+    });
+  });
+
+  describe('versioning', () => {
+    it('renders succesfully', () => {
+      expect(screen.getByRole('heading', { name: 'Versioning' })).toBeInTheDocument();
+    });
+  });
+
+  describe('monitoring', () => {
+    it('renders succesfully', () => {
+      expect(screen.getByRole('heading', { name: 'Monitoring' })).toBeInTheDocument();
+    });
+  });
+
+  describe('support', () => {
+    it('renders succesfully', () => {
+      expect(screen.getByRole('heading', { name: 'Support' })).toBeInTheDocument();
+    });
+  });
+
+  describe('modernization', () => {
+    it('renders succesfully', () => {
+      expect(screen.getByRole('heading', { name: 'Modernization' })).toBeInTheDocument();
+    });
+  });
+
+  describe('authentication and authorization', () => {
+    it('renders succesfully', () => {
+      expect(
+        screen.getByRole('heading', { name: 'Authentication and Authorization' }),
+      ).toBeInTheDocument();
+    });
+  });
+});
--- a/src/containers/publishing/components/publishingExpecations/PublishingExpectations.tsx
+++ b/src/containers/publishing/components/publishingExpecations/PublishingExpectations.tsx
@@ -16,7 +16,6 @@ import { FLAG_AUTH_DOCS_V2 } from '../../../../types/constants';
 
 const PublishingExpectations = (): JSX.Element => {
   const authDocsV2 = useFlag([FLAG_AUTH_DOCS_V2]);
-  console.log(authDocsV2);
   const authenticationUrl = authDocsV2
     ? '/explore/authorization'
     : '/explore/verification/docs/authorization';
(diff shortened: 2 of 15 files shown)


src/containers/publishing/components/publishingExpecations/PublishingExpectations.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+  it('renders successfully', () => {
+    expect(
+      screen.getByRole('heading', { name: 'Expectations for Lighthouse APIs' }),
+    ).toBeInTheDocument();
+  });
+
+  it('renders all accordions in a closed state', () => {
+    const buttons = screen.getAllByRole('button');
+    buttons.forEach(button => {
+      expect(button.getAttribute('aria-expanded')).toBe('false');
```
mjlumetta, line 24: nice test!!

src/containers/publishing/components/publishingExpecations/PublishingExpectations.tsx
```diff
@@ -0,0 +1,77 @@
...
+  SecurityMarkdown,
+  SupportMarkdown,
+  VersioningMarkdown,
+} from '../../../../content/publishing/expectations';
+
+declare const window: { VetsGov?: Record<string, unknown> };
+
+const PublishingExpectations = (): JSX.Element => {
+  React.useEffect(() => {
+    // CollapsiblePanel expects a VetsGov object on the global window
```
mjlumetta, line 19: hate this so much!!! 🤮 

but actually, if I'm being practical about it... should we add this effect in `App` or somewhere else at the top level and then forego the individual declarations everywhere we need `CollapsiblePanel`?

src/content/publishing/expectations/security.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+- We rely on mature software such as: Kong (nginx), PostgreSQL, and Okta, which are used in other security-sensitive environments.
+- Our software runs in VA's [FISMA-compatible](https://www.cisa.gov/federal-information-security-modernization-act) Amazon Web Services (AWS) data center, which is subject to [VA Handbook](https://www.va.gov/vapubs/viewPublication.asp?Pub_ID=733&FType=2) 6500.3 continuous monitoring requirements.
+- We use a variety of methods to provide authentication and security.
+
+</div>
+
+Lighthouse has a full authorization to operate (ATO) at the FISMA Moderate level. We also complete a privacy impact assessment and privacy threshold analysis whenever the API gateway sends new kinds of data to new systems.
+
+The production environment is [accessible here](https://api.va.gov) and the sandbox environment is [accessible here](https://dev-api.va.gov). Our production and sandbox environments are deployed in the VA Enterprise Cloud's AWS GovCloud region, sitting behind the VA Trusted Internet Connection (TIC), and in compliance with VA’s external system network requirements.
```
mjlumetta, line 27:
```suggestion
The production environment is [accessible here](https://api.va.gov) and the sandbox environment is [accessible here](https://sandbox-api.va.gov). Our production and sandbox environments are deployed in the VA Enterprise Cloud's AWS GovCloud region, sitting behind the VA Trusted Internet Connection (TIC), and in compliance with VA’s external system network requirements.
```

src/content/publishing/expectations/security.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+We rely on mature software that is used in other security-sensitive environments: Kong (nginx), PostgreSQL, and Okta. These are all running in VA's [FISMA-compatible](https://www.cisa.gov/federal-information-security-modernization-act) AWS data center and are subject to [VA Handbook](https://www.va.gov/vapubs/viewPublication.asp?Pub_ID=733&FType=2) 6500.3 continuous monitoring requirements.
+
+### Addressing Security Risks
+
+If you have a simple API that publishes shared information about VA, there are very few security risks to address, and we use a variety of methods to provide authentication and security. You can read more about this on our [Authentication and Authorization](https://developer.va.gov/explore/verification/docs/authorization) page, but to start, here are some of the ways our authentication offers additional security:
```
mjlumetta, line 33: the Authorization link should be an internal link using React Router. technically this works, but there are two considerations here:

1. with an absolute href, it doesn't respect the host in lower environments.
2. if it's not a `Link` or a `HashLink`, it will reload the whole site when clicked. even though this is probably not the end of the world, especially if the site is cached in the user's browser, it's not consistent with the client-side routing experience, and we should probably keep the single-page experience consistent.

also, should we use the v2 auth docs flag to conditionally show the old/new URLs? this link should be to `/explore/authorization` in the near future.

src/content/publishing/expectations/authentication.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+We are in the process of an integration with the VA single sign-on platform, which will offer:
+
+- **SSOi**: single-sign on for internal users, which are VA employees or contractors using VA systems
+- **SSOe**: single sign on for external users, such as Veterans and associates signing into public-facing websites like VA.gov, or using a third-party external app integrated with VA data
+
+API keys are maintained within Kong. There is no default expiration date. If a key needs to be rotated on a recurring basis, Lighthouse can either update the key immediately, or update it on a rolling basis, where both keys work for a short period of time and the consumer facilitates the update on their end with no downtime. The information passed to the API provider does not change when the key is updated for the consumer.
+Sandbox and production access
+API consumers are granted access to the sandbox environment first, so they can test your API without data risk. When they’re ready, they become approved for production access using our [Path to Production](https://github.com/department-of-veterans-affairs/vets-api-clients/blob/master/Path-to-Production.md).
```
mjlumetta, line 52: this link points the user back to the `/go-live` page. should this link directly to that page instead?

src/content/publishing/expectations/authentication.mdx
```diff
@@ -0,0 +1,52 @@
+<div className="feature">
+
+### What we provide
```
mjlumetta, line 3: issue flagged by axe: `markdown-component-loader` generates ids for heading level (think this is due to a `markdown-it` plugin, forget the exact mechanics but can check on this) based on the text content. anyway, the lovely consistent structure of this feature div means that these headings all have duplicate ids. would you mind giving them custom ids using the `{#my-custom-id-name}` annotation? you can see examples of this on the Release Notes page, where we've run into this issue before with dates (particularly with OAuth APIs), and on the new Auth docs components.

src/content/publishing/expectations/authentication.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+APIs that don't include PII/PHI are secured using an API key, an authentication mechanism relying on the generation of a secret key, which is shared with each unique API consumer. This way, each consumer is approved before being granted access.
+
+We are in the process of an integration with the VA single sign-on platform, which will offer:
+
+- **SSOi**: single-sign on for internal users, which are VA employees or contractors using VA systems
+- **SSOe**: single sign on for external users, such as Veterans and associates signing into public-facing websites like VA.gov, or using a third-party external app integrated with VA data
+
+API keys are maintained within Kong. There is no default expiration date. If a key needs to be rotated on a recurring basis, Lighthouse can either update the key immediately, or update it on a rolling basis, where both keys work for a short period of time and the consumer facilitates the update on their end with no downtime. The information passed to the API provider does not change when the key is updated for the consumer.
+Sandbox and production access
```
mjlumetta, line 51: what's going on here - is this supposed to be a heading? looks a bit wonky when rendered.

src/content/publishing/expectations/security.mdx
mjlumetta: also gonna drop this issue Brent flagged here https://lighthouseva.slack.com/archives/C0190E2PVTQ/p1611958892171700?thread_ts=1611953150.159900&cid=C0190E2PVTQ (axe will raise it as well)

src/content/publishing/expectations/authentication.mdx
```diff
@@ -0,0 +1,52 @@
+<div className="feature">
```
mjlumetta, line 1: for these feature divs, there's no equivalent of the visual highlighting for screen readers. should we give them some kind of markup that will distinguish them as such? my initial thought would be a `role="region"` + an `aria-label`, but I'm open to suggestion.

mjlumetta (commented):
nice job on this! comments are mostly accessibility concerns + copy edits.

mjlumetta:
re: the Jest/Markdown question, that's a good concern to point out. I've got a couple quick thoughts before signing off for the week:

1. do we want to store this much content in MDX files in the first place? thinking that if we're ever storing and editing content externally in a consistent way, that may cut out the need for this conversation entirely.
2. if there's anything that urgently needs testing, integration tests are always an option. 

but yeah, would love to solve this at some point.

src/containers/publishing/components/publishingExpecations/PublishingExpectations.tsx
alexandervalencia: me too 😑  

Yeah, I think that's a good idea since we know it's going to be used in at least two spots.

src/containers/publishing/components/publishingExpecations/PublishingExpectations.test.tsx
alexandervalencia: Thanks!

src/content/publishing/expectations/authentication.mdx
brentschneider: That would make the most sense to me as well Mike.  Let's link directly to the portal `/go-live` page.

alexandervalencia: confirmed with Brent 👍

src/content/publishing/expectations/authentication.mdx
alexandervalencia: Yeah, missed a heading here. Good catch! 🧤

src/content/publishing/expectations/authentication.mdx
alexandervalencia: `region` seems like the correct role here, I think you're spot on!
ref: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Region_role

src/content/publishing/expectations/security.mdx
alexandervalencia: Great observations, I added the flag and made sure to use `Link`

alexandervalencia:
Thanks for reviewing @mjlumetta! I've hopefully addressed all the concerns you mentioned. In regards to your final thoughts on MDX. I'm with you, it would be much nicer if we could store this much content elsewhere in a more easily editable way for non-devs to handle. Until then, I think having mostly pure-content is better than large `tsx` files.

src/containers/publishing/components/publishingExpecations/PublishingExpectations.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    });
+  });
+
+  describe('security', () => {
```
mjlumetta, line 28: absolutely not necessary, but if you end up pushing another commit and are interested, might be a good use case for `describe.each`. that's purely aesthetic though

src/types/md.d.ts
```diff
@@ -7,6 +7,7 @@ declare module '*.mdx' {
    * preference to limit our use the component loader to its basic functionality.
    */
   interface MarkdownComponentProps {
+    [x: string]: unknown;
```
mjlumetta, line 10: couple small notes here: might want to check with @ediiotero on this one, since he's got similar but differently structured changes on #611. also, regardless of how we commit this functionality, the comment above is no longer accurate, so that will need to change, I think.

src/content/publishing/expectations/versioning.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+- We’ll work with you on a versioning policy.
+- Any breaking changes to your API must be published as a new version.
+- We’ll work with you to manage communication of new versions and to deprecate and deactivate previous versions.
+
+</div>
+
+Breaking changes to published API versions can be disruptive to consumers, so our policy is to ensure breaking changes aren’t published without a major version change.
+
+We define a breaking change as any change that might adversely impact a consumer’s use of an API. For example, removing an endpoint or field would be classified as a breaking change. A non-breaking change is one that does not adversely impact current users of an API, such as adding an endpoint or field. Non-breaking changes do not require a new version, but changes that could impact consumers’ experience or use of an API must have an accompanying [release note](https://developer.va.gov/release-notes).
```
mjlumetta, line 28: one more that should be a `Link`

src/content/publishing/expectations/authentication.mdx
```diff
@@ -0,0 +1,59 @@
+---
+imports:
+  '{ Link }': 'react-router-dom'
+---
+
+<div className="feature" role="region" aria-label="Authentication and Authorization Expectations">
```
mjlumetta, line 6: the region + ARIA label thing works great! only thing I'd ask here is should we tweak the labels to be "Authentication and Authorization Expectations Summary" (or "overview")? since the whole page is about expectations, might be important to use language that clarifies that these are the key points and alludes to additional info below. (especially in case someone is navigating by landmarks, don't want them to think the region is the whole body of information we're providing.)

mjlumetta (commented):
yeah, this is basically there. I've suggested a small language tweak on the ARIA labels for the feature regions, there's a missing `Link` element, and would suggest syncing up with @ediiotero on the Markdown props type def since he's working on something similar. but overall looks great, this is nearly ready to go.

src/types/md.d.ts
gileswells: I personally like this method better than what's in #611. I'm adding a review comment over on that ticket about this.

src/content/publishing/expectations/versioning.mdx
alexandervalencia: Thanks for the catch!

src/content/publishing/expectations/authentication.mdx
alexandervalencia: love this, will fix!

mjlumetta (dismissed):
lgtm! good to go pending CI checks ✨

rtravitz:
@mjlumetta Any thoughts on why that lint comment triple posted?

alexandervalencia:
Brent found two more issues; the changes are fixed, it should be okay from him, just need final approval here, Thanks!

mjlumetta (approved):
re-approving ✅

PR #616, opened 2021-01-29
Cut over main release and production deploy to CodeBuild

mjlumetta:
### Description

does a few things in support of [API-3139](https://vajira.max.gov/browse/API-3139) to make CodeBuild control the main release and production deploy:

- removes the `cb-` prefix from the S3 key used in the CodeBuild archive command.
  - also adds a `jenkins-` prefix to the corresponding key in the Jenkins archive command. (worth keeping temporarily, or should we get rid of the Jenkins archive stage entirely?)
- removes the `cb-` prefix from the CodeBuild release tag.
- changes the CodeBuild deploy that's kicked off from the release job to use the release and not the archived commit as the deployment artifacts.

see https://github.com/department-of-veterans-affairs/devops/pull/8446 for changes to Jenkins release + deploy. at the moment, we're only commenting them out.

note that production deploys will happen manually, as currently happens with the `developer-portal-backend` and other apps. previously, the production deploy happened daily Mon-Fri at 11:30 Eastern (scheduled at 10:30).

also adding the `--ci` flag to the `npm run test:visual` command. thanks to @RockslideJr, @duganth-va, and @rtravitz for that catch. open to feedback on how to configure this one - it could go in the Makefile, or we could use the `CI="true"` environment variable in `buildspec-ci.yml`.

### Process

- [ ] Docs updated
  - [n/a] If you have made structural changes to the site, the [Developer Portal Content Types](https://community.max.gov/x/qUwOg) docs have been updated
  - [n/a] If you have made/confirmed any decisions about user interactions and accessibility, the [Accessible Component Design](https://community.max.gov/x/xS8mgg) docs have been updated
  
  - [n/a] If you've made any UI changes, you've run an axe check using the axe browser extension that reported no new issues

### Requested feedback

- is it worthwhile to continue archiving the Jenkins builds with the `jenkins-` prefix, or should we remove the Jenkins archive step entirely?
- preferences on where to add the CI config for `jest-image-snapshot`? what are people's preferences for updating snapshots at the moment?
- do we need to keep the `grep ^developer-portal` check on line 51 of the deploy job as a safeguard in case there are releases in some other format for whatever unseen reason, or should we take that out?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -74,9 +74,8 @@ def reviewBucketPath() {
   return "${review_s3_bucket_name}/${shortRef}"
 }
 
-
 node('vetsgov-general-purpose') {
-  properties([[$class: 'BuildDiscarderProperty', strategy: [$class: 'LogRotator', daysToKeepStr: '60']]]);
+  properties([[$class: 'BuildDiscarderProperty', strategy: [$class: 'LogRotator', daysToKeepStr: '60']]])
   def dockerImage, args, imageTag
 
   // Checkout source, create output directories, build container
@@ -106,7 +105,7 @@ node('vetsgov-general-purpose') {
       shortRef = sh(returnStdout: true, script: 'git rev-parse --short HEAD').trim()
 
       if (prNum) {
-        envNames.each{ envName ->
+        envNames.each { envName ->
           // Set public url so review deploys handle not being at the site root
           // Set sentry DSN to send review deploy errors to the review sentry project
           writeFile(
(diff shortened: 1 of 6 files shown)


duganth-va (dismissed):
LGTM 💯 
Only thing I would add in regards to visual regression CI tests is maybe a separate `visual:ci` that has the `--ci` flag like the other tests?

mjlumetta:
> Only thing I would add in regards to visual regression CI tests is maybe a separate visual:ci that has the --ci flag like the other tests?

that's a good question. personally I'm okay with the default behavior being failure if anything's changed and I haven't passed `-u`, and I don't think there's any particular reason to add multiple commands if they're mostly redundant. but there's not really any harm in it, either. so I'm open to whatever the other frontend devs want, really.

gileswells (dismissed):
LGTM

In regards to the `visual:ci` vs `--ci` I feel like its fine how you have it since there is no large appreciable difference in how it runs.

gileswells (approved):
LGTM

rtravitz (approved):
Long time coming

duganth-va (approved):
Outstanding! next you just automate the MR creation 🤣

PR #618, opened 2021-02-01
Redirect for old vet verification auth docs to the new auth docs

ediiotero:
Story: [API-4636](https://vajira.max.gov/browse/API-4636)

### Redirect old vet verification auth docs

Adding a redirect from old auth docs to the new docs with vet verification preselected.

### DIFF
--- a/src/containers/documentation/DocumentationRoot.tsx
+++ b/src/containers/documentation/DocumentationRoot.tsx
@@ -135,12 +135,8 @@ const DocumentationRoot = (): JSX.Element => {
           {oldRouteToNew.map(routes => (
             <Redirect key={routes.from} exact from={routes.from} to={routes.to} />
           ))}
-          {authDocsV2 && (
-            <>
-              <Route path="/explore/authorization" component={AuthorizationDocs} exact />
-              <Redirect from="/explore/verification/docs/authorization" to="/explore/authorization?api=veteran_verification" />
-            </>
-          )}
+          {authDocsV2 && <Route path="/explore/authorization" component={AuthorizationDocs} exact />}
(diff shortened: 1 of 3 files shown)


mjlumetta (dismissed):
looks great!! only suggestion I'd make would be to consider adding an integration test (which would be skipped at present like the others are, presumably, but could be useful to include).

gileswells (dismissed):
Wasn't where I expected the redirect to go but I like it better here.

src/containers/documentation/DocumentationRoot.tsx
```diff
@@ -135,9 +135,8 @@ const DocumentationRoot = (): JSX.Element => {
           {oldRouteToNew.map(routes => (
             <Redirect key={routes.from} exact from={routes.from} to={routes.to} />
           ))}
-          {authDocsV2 && (
-            <Route path="/explore/authorization" component={AuthorizationDocs} exact />
-          )}
+          {authDocsV2 && <Route path="/explore/authorization" component={AuthorizationDocs} exact />}
+          {authDocsV2 && <Redirect exact from="/explore/verification/docs/authorization" to="/explore/authorization?api=veteran_verification" />}
```
mjlumetta, line 139: any particular reason for making these blocks separate this time?

mjlumetta (approved):

src/containers/documentation/DocumentationRoot.tsx
ediiotero: The previous way I had it set up caused the other authDocsV2 redirect to break, not sure why. Separated them and all the routes worked as expected and the tests passed.

mjlumetta: hmm... I'm pretty confused why that would've happened - they should be functionally equivalent.

ediiotero: yeah same here

PR #619, opened 2021-02-01
[API-4403] Focus Management

gileswells:
### Description

https://vajira.max.gov/browse/API-4403
This updates hash links and how they're handled on the auth pages specifically but also touches on the release notes pages as well. With this update the pages should properly respect hash urls on load and on in page navigation.

### Process

- [ ] Docs updated
  
### Requested feedback

This should be tested across Chrome/FireFox/Safari as some issues were not universal.


src/components/pageContent/PageContent.tsx
```diff
@@ -7,7 +7,9 @@ const focusAndScroll = (elementToFocus: HTMLElement | null): void => {
     elementToFocus.focus();
   }
 
-  window.scrollTo(0, 0);
+  if (elementToFocus?.id === 'main') {
```
mjlumetta, line 10: wondering if we should inline this `focusAndScroll` function in the effect. I don't think this conditional is necessarily bad in a separate function as a safeguard, but if we inline it, it becomes clearer that the only element we call it on is the main element.

mjlumetta (dismissed):
yeah, it's working for me in all browsers. should we add some new integration tests? that's an AC on the ticket. anyway, I think it's all looking good from what I can tell doing manual testing.

gileswells:
> should we add some new integration tests? that's an AC on the ticket.

I'm working on those now. I'm thinking that I might end up having to hold on the maintaining of focus parts of those tests for the auth docs pages specifically until they're enabled in prod.

gileswells:
To get the test running you'll need to change [this line](https://github.com/department-of-veterans-affairs/developer-portal/pull/619/files#diff-a98238eb805c61b666ae4064bc2cad399d5819f34a2d49565dda5a825b9185f0R56) to remove `.skip` and also add `REACT_APP_AUTH_DOCS_V2=true` to `jest-puppeteer.config.json` under the `envVars` array.

ediiotero (approved):
Ran e2e test locally. Looks good

ediiotero (approved):
LGTM

PR #620, opened 2021-02-01
Api 4442/update main navigation

christineyost:
### Description
This PR adds API Publishing to the main navigation as described in [API-4442](https://vajira.max.gov/browse/API-4442). It also refactors the mobile subnav currently used for Documentation into its own component so it could be easily reused for API Publishing.

### Process

- [ ] Docs updated
  
### DIFF
--- a/src/components/navBar/NavBar.test.tsx
+++ b/src/components/navBar/NavBar.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@testing-library/react';
+import { cleanup, render, screen } from '@testing-library/react';
 import 'jest';
 import * as React from 'react';
 import { BrowserRouter as Router } from 'react-router-dom';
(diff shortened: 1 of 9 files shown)


src/components/subNav/SubNav.tsx
```diff
@@ -0,0 +1,54 @@
...
+const SubNav: FC<SubNavProps> = ({ children, name, alt }) => {
+  const [subNavVisible, setSubNavVisible] = React.useState(false);
+
+  const toggleSubNav = (): void => {
+    setSubNavVisible(!subNavVisible);
+  };
+
+  return (
+    <div className={mobileOnly()}>
+      <button className={expandNavButtonStyles} onClick={(): void => toggleSubNav()} type="button">
```
mjlumetta, line 40: couple of things that weren't on the old version but should go here: `aria-expanded={subNavVisible ? 'true' : 'false'}` and `aria-controls="<id of the ul>"`. this isn't new with your refactor, they should've been on the old one, but would be great to address now.

src/components/subNav/SubNavEntry.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const linkStyles = classNames(
+  'vads-u-color--gray-dark',
+  'vads-u-display--block',
+  'vads-u-padding-left--2p5',
+  'vads-u-text-decoration--none',
+  'vads-u-width--full',
+);
+
+interface SubNavEntryProps {
```
mjlumetta, line 20: should `children` be in this interface somehow?

src/components/subNav/SubNav.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { FC } from 'react';
+import classNames from 'classnames';
+import { mobileOnly } from '../../styles/vadsUtils';
+import minusIcon from '../../../node_modules/uswds/src/img/minus.png';
+import plusIcon from '../../../node_modules/uswds/src/img/plus.png';
+
+interface SubNavProps {
```
mjlumetta, line 7: same question about `children` as for `SubNavEntry`

src/components/navBar/NavBar.tsx
```diff
@@ -189,34 +140,22 @@ const NavBar = (props: NavBarProps): JSX.Element => {
...
-                  alt="Expand Documentation"
-                  aria-label="Expand Documentation"
-                  className="va-api-expand-nav-icon"
-                />
-              </button>
-              {visibleSubNavs.documentation && (
-                <DocumentationSubNav onMobileNavClose={props.onMobileNavClose} />
-              )}
-            </div>
+            <SubNav name="Documentation" alt="Expand Documentation">
```
mjlumetta, line 143: it would be awesome if we could fold this component into the `MainNavItem` component so the `MainNavItem` abstraction takes care of everything. with that said, that's an additional refactor on top of what you've already done, so I don't view it as necessary for this PR by any means. and I know y'all are working under time constraints, so feel free to pass on this suggestion.

src/components/subNav/SubNavEntry.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+const SubNavEntry: FC<SubNavEntryProps> = ({ children, onClick, to, key }) => (
+  <li className={itemStyles} key={key}>
+    <NavLink
+      onClick={onClick}
+      exact
+      to={to}
+      className={linkStyles}
+      activeClassName="va-api-active-sub-nav"
```
mjlumetta, line 33: does this class actually do anything? looks like no from what I can see. if it doesn't, should it? I'm okay with including it as is to have an active class name that isn't just `active`, but wondering if we need to add some kind of style to that effect

mjlumetta (commented):
great job on this PR, like the `SubNav` refactor. most of my suggestions are related to further improvements on the old design rather than things specific to this PR. some are low-hanging and would be pretty easy to do on this one, but there are one or two you may or may not want to pass on for now.

src/components/subNav/SubNav.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const elrondLink = screen.getByRole('link', { name: 'Elrond' });
+    expect(elrondLink).toBeInTheDocument();
+    const galadrielLink = screen.getByRole('link', { name: 'Galadriel' });
+    expect(galadrielLink).toBeInTheDocument();
+
+    userEvent.click(button);
+
+    expect(screen.queryByRole('link', { name: 'Legolas' })).not.toBeInTheDocument();
+    expect(screen.queryByRole('link', { name: 'Elrond' })).not.toBeInTheDocument();
+    expect(screen.queryByRole('link', { name: 'Galadriel' })).not.toBeInTheDocument();
```
alexandervalencia, line 56: nice coverage!

src/components/subNav/SubNav.tsx
```diff
@@ -0,0 +1,54 @@
...
+  };
+
+  return (
+    <div className={mobileOnly()}>
+      <button className={expandNavButtonStyles} onClick={(): void => toggleSubNav()} type="button">
+        <span>{name}</span>
+        <img
+          src={subNavVisible ? minusIcon : plusIcon}
+          alt={alt}
+          aria-label={alt}
```
alexandervalencia, line 45: Do we need an `aria-label` if we have an `alt` tag? Based on my reading it seems like we don't: https://fae.disability.illinois.edu/rulesets/IMAGE_1/#:~:text=The%20alt%20attribute%20is%20the,for%20img%20and%20area%20elements.&text=The%20aria%2Dlabel%20attribute%20should,for%20img%20and%20area%20elements.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Role_Img

It looks like `aria-label` should be used when non-`` tags have a `role="image"` attribute

alexandervalencia (commented):
Looks great Christine! I had some comments and a test refactor I'd like you to take a look at and tell me what you think. Thanks!

src/components/subNav/SubNavEntry.tsx
christineyost: It doesn't have to be because the `FC` type defines children, but we can explicitly define it here as well. Do you think that would be better?

src/components/navBar/NavBar.tsx
christineyost: Agree this would be really nice, but might skip this additional refactor for now

src/components/subNav/SubNavEntry.tsx
christineyost: Doesn't look like it to me. I actually don't see any styles for this class defined anywhere - just grabbed it from the existing subnav

src/components/navBar/NavBar.tsx
mjlumetta: yes, that is more than fair!

src/components/subNav/SubNavEntry.tsx
mjlumetta: nope, didn't realize that! if the type definition includes it, should be good as is. thanks for clarifying 😄

src/components/navBar/NavBar.test.tsx
```diff
@@ -1,38 +1,74 @@
...
 import 'jest';
 import * as React from 'react';
 import { BrowserRouter as Router } from 'react-router-dom';
+import { FlagsProvider, getFlags } from '../../flags';
 
 import { NavBar } from './NavBar';
 
 const noop = (): void => undefined;
 
+const renderNavBar = (isMobileMenuVisible: boolean): RenderResult => (
```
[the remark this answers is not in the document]
christineyost: Looks good to me, agree just reading `renderNavBar(false)` sounds like it won't be rendered

src/components/subNav/SubNav.tsx
christineyost: No... Mike also suggested removing this. `alt` was also removed since this is wrapped by a button

src/components/navBar/NavBar.test.tsx
```diff
@@ -1,38 +1,94 @@
...
+        screen.getByRole('navigation').classList.contains('va-api-mobile-nav-visible'),
+      ).toBeTruthy();
+    });
+
+    it('contains the expected links', () => {
+      /*
+       * The Documentation and API Publishing navigation items are links in the desktop view
+       * and buttons in the mobile view. Both are rendered even though only one will only be
+       * visible at a time based on the view.
+       */
```
alexandervalencia, line 39: 🙌 ❤️

src/components/subNav/SubNav.tsx
```diff
@@ -6,29 +6,27 @@ import plusIcon from '../../../node_modules/uswds/src/img/plus.png';
...
-  'vads-u-margin--0',
+  'vads-u-justify-content--space-between',
   'vads-u-width--full',
+  'vads-u-margin--0',
+  'vads-u-padding--1',
+  'medium-screen:vads-u-padding--1p5',
+  'vads-u-color--gray-dark',
+  'vads-u-font-weight--normal',
+  'vads-u-line-height--4',
+  'vads-u-text-decoration--none',
```
alexandervalencia, line 24: thanks! this is wonderful!

alexandervalencia (dismissed):
Nice work @christineyost 🚀

src/components/navBar/NavBar.test.tsx
```diff
@@ -1,38 +1,94 @@
...
+        </FlagsProvider>,
+      );
+    });
+
+    it('should render the navbar', () => {
+      const navbar = screen.getByRole('navigation');
+      expect(navbar).toBeInTheDocument();
+    });
+
+    it('should use "va-api-mobile-nav-visible"', () => {
```
mjlumetta, line 28: i think you could reasonably make this the only test in this describe block and move all the others into the top-level describe block. that said, it's not blocking, definitely wouldn't make a commit just for that

src/components/subNav/SubNavEntry.tsx
mjlumetta: yeah. should be fine to leave in here though - I think it adds an a class that's just "active" if you don't specify this prop, so might as well make it specific.

src/components/subNav/SubNav.tsx
```diff
@@ -0,0 +1,57 @@
...
+    setSubNavVisible(!subNavVisible);
+  };
+
+  return (
+    <div className={mobileOnly()}>
+      <button
+        className={expandNavButtonStyles}
+        onClick={(): void => toggleSubNav()}
+        type="button"
+        aria-expanded={!!subNavVisible}
```
mjlumetta, line 42: since `subNavVisible` is already a boolean, was there any particular reason for adding the `!!` operator? not a blocker by any means, of course

src/components/subNav/SubNav.tsx
```diff
@@ -0,0 +1,57 @@
...
+        aria-controls={`${name}-sub-nav`}
```
mjlumetta, line 43: one small issue with this `aria-controls`/`id` value: with "API Publishing", it has whitespace, which is [invalid for ids](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id):

> id's value must not contain whitespace (spaces, tabs etc.). Browsers treat non-conforming IDs that contain whitespace as if the whitespace is part of the ID.

there's a utility called `toHtmlId` in `src/toHtmlId.ts`, and you can use that to convert this current string to an id-compliant string.

mjlumetta (changes requested):
great job! one small change on the `id` of the sub nav list.

src/components/subNav/SubNav.tsx
christineyost: Oops, missed that

mjlumetta (approved):
✨ 🚀

PR #622, opened 2021-02-04
Api 4444/add publishing support to contact form

RockslideJr:
### Description

[API-4444](https://vajira.max.gov/browse/API-4444)

This PR updates the contact form to include a flow for API publishing support and to make changes based off of new designs.

In the process it also implements the first step towards utilizing [formik](https://formik.org/docs/overview) for form state management. 

### Process

- [x] Docs updated
  
### Requested feedback

General feedback on the implementation for Formik is requested. This is only step one, and follow-up tickets will be introduced to the Quokka backlog to convert the other form in this application to formik.


christineyost (commented):
Added a couple questions specific to API Publishing

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.tsx
```diff
@@ -0,0 +1,100 @@
+import classNames from 'classnames';
+import React, { ReactNode } from 'react';
+import { Formik, Form, Field } from 'formik';
+import { CONTACT_US_URL } from '../../../types/constants';
+import { makeRequest, ResponseType } from '../../../utils/makeRequest';
+import '../SupportContactUsForm.scss';
+import { SupportContactUsFormState, SupportContactUsFormProps, SupportContactUsFormPropTypes, FormType, FormData } from '../../../types/contactForm';
```
alexandervalencia, line 7: I can't understand why prettier & eslint aren't complaining about this line. 🙃

src/containers/support/SupportContactUs.tsx
```diff
@@ -45,7 +61,16 @@ const SupportContactUs = (): JSX.Element => {
           </Helmet>
           <PageHeader {...headerProps} />
           <GitHubSnippet />
-          <SupportContactUsForm onSuccess={onSuccess} />
+          <Flag
+            name={[FLAG_API_PUBLISHING]}
```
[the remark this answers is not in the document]
RockslideJr: This is a great question. I see the benefit to being able to turn this off without turning off the rest of the publishing documentation, but I also see the need to turn this off with the publishing documentation if we needed to (which i doubt we will). Maybe I'll add in a new flag and check both?

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+    });
+
+    describe('default is selected', () => {
+      it('renders the description field', () => {
+        expect(screen.getByLabelText(/Describe your question or issue in as much detail as you can./));
+      });
+
+      describe('all fields are valid', () => {
+        beforeEach(async () => {
+          await userEvent.type(screen.getByLabelText(/First name/), 'Frodo', { delay: 0.00001 });
```
RockslideJr, line 48: this `{ delay: 0.00001 }` was a work around to formik producing `act` warnings. By default the type method returns a promise, but does all the typing synchronously, which doesn't allow time for all the side effects happening in Formik from the change events. Wrapping this in an `act` _does_ eliminate the act warnings, but for some reason this first field only ends up with the last letter typed as it's value (o instead of Frodo). setting the delay forces type to treat this as an actual async event, which elminates _most_ of the act warnings these test were throwing (there are still 2 but i'm not sure where they come from)

If someone has a better solution or more info on how to fix any of these problems i am all 👂 👂

alexandervalencia (commented):
Nice work @RockslideJr! I'm excited to see Formik getting used. I had a few comments to check out.

Really appreciate the thorough testing you wrote up.

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+          expect(screen.getByRole('button', { name: 'Submit' })).toBeEnabled();
+        });
+
+        describe('submitting the form', () => {
+          beforeEach(async () => {
+            userEvent.click(screen.getByRole('button', { name: 'Submit' }));
+            expect(await screen.findByRole('button', { name: 'Sending...' })).toBeInTheDocument();
+          });
+          it('sends the values', async () => {
+            expect(jsonSpy).toHaveBeenCalledWith({
```
mjlumetta, line 66: is there any particular reason to spy on `JSON.stringify` instead of testing `body: JSON.stringify({ ... fieldValues })` in the expect block? asserting that `JSON.stringify` is the particular mechanism by which it gets transformed from an object to a string seems like an implementation detail. but, then again, using the return value of `JSON.stringify` would be dependent on the order. so not sure whether there's a good way out of that 🤔

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+        });
+        it('displays the validation errors', async () => {
+          expect(await screen.findByText('First name must not be blank.')).toBeInTheDocument();
+          expect(await screen.findByText('Must be a valid email address.')).toBeInTheDocument();
+        });
+      });
+    });
+
+    describe('publishing is selected', () => {
+      beforeEach(async () => {
```
mjlumetta, line 152: instead of clicking the radio option each time, should we render the form with `defaultType={FormType.PUBLISHING}`? I don't think this test covers that case specifically, but the previous describe block has covered switching the type in an already-rendered form.

agree with Alex though, the tests are impressive 🙌

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.tsx
```diff
@@ -0,0 +1,100 @@
+import classNames from 'classnames';
+import React, { ReactNode } from 'react';
+import { Formik, Form, Field } from 'formik';
+import { CONTACT_US_URL } from '../../../types/constants';
```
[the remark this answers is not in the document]
mjlumetta: this [came up](https://github.com/department-of-veterans-affairs/developer-portal/pull/610#discussion_r563768660) on @christineyost's PR for publishing docs, and she pointed out that the existing paths here are paths to the backend, whereas all the paths in `paths.ts` are frontend paths. so there's a question of whether those belong in the same place or not. I don't have a strong preference, but if they go in the same file, would be good to add some clear block comments to distinguish the two.

src/containers/support/supportContactUsFormPublishing/components/ContactDetailsFormFields.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <div className={textFieldClasses('left')}>
+        <FormField label="Last name" name="lastName" required />
+      </div>
+    </div>
+    <div className={classNames('vads-l-row', 'vads-u-margin-top--4')}>
+      <div className={textFieldClasses('right')}>
+        <FormField label="Email address" name="email" type="email" required />
+      </div>
+      <div className={textFieldClasses('left')}>
+        <FormField label="Organization" name="organization" />
```
mjlumetta, line 28: not really a code question, but... is this field not required for publishing an API?

src/containers/support/supportContactUsFormPublishing/components/DefaultFormFields.tsx
```diff
@@ -0,0 +1,18 @@
+import classNames from 'classnames';
+import React, { FC } from 'react';
+import { FormField } from '../../../../components';
+
+const DefaultFormFields: FC = () => (
```
mjlumetta, line 5: might've said a version of this in a different comment, but I would have a strong preference for naming this `ConsumerFormFields`.

also, the API checkboxes are going away? more of a @brentschneider question, but was surprised by that

src/types/contactForm.ts
```diff
@@ -0,0 +1,50 @@
+import PropTypes from 'prop-types';
+
+export const enum FormType {
+  DEFAULT = 'DEFAULT',
+  PUBLISHING = 'PUBLISHING',
+}
+
+export const SupportContactUsFormPropTypes = {
```
mjlumetta, line 8: generally speaking, the prop types (both React and Typescript) are defined in the same file as the component. was there any particular reason you wanted to define them in a separate location? if we're interested in altering that pattern, would like to get some more discussion from other frontend portal devs going. for me personally, I do find colocating them useful.

src/utils/validators.ts
```diff
@@ -24,7 +25,22 @@ export const validateOAuthRedirectURI = (newValue: ErrorableInput): ErrorableInp
 };
 
 export const validatePresence = (newValue: ErrorableInput, fieldName: string): ErrorableInput => {
-  const presencePattern = /^(?!\s*$).+/;
-  validateByPattern(newValue, presencePattern, `${fieldName} must not be blank.`);
+  validateByPattern(newValue, PRESENCE_PATTERN, `${fieldName} must not be blank.`);
   return newValue;
 };
+
+export const validatePresenceFormik = (fieldName: string, value: string): string | undefined => {
```
mjlumetta, line 32: don't really love the `*Formik` suffix on these two validator functions, for two reasons:

1. we could conceivably use these for non-Formik purposes. _but_
2. if Formik is the new way of the world, then I think the other two functions will be going away soon? in which case, if we need a suffix that distinguishes one set of these functions as non-normative, I would say put the suffix on the others and take the `validatePresence` and `validateEmail` names for these functions.

so could go a bunch of different ways on this. I would say some options are (a) name the other functions `validatePresenceLegacy` and `validateEmailLegacy` if we want to emphasize that those functions are going away soon, (b) name the other functions `validateRequiredInput` and `validateEmailInput` to emphasize the different in function signatures if we want both to co-exist alongside each other for a while, or (c) ignore this comment and leave these names since this is only temporary until the Apply form gets refactored (assuming the old validators go away and these new ones take the old names at that point). so yeah, relatively small, but wanted to ask about it.

mjlumetta (commented):
leaving a bunch of comments on various things, most of which are naming-related/domain-related. haven't been able to do a lot of manual testing but want to get in a review before the end of the week. overall think the code looks good, and I'm on board with Formik based on seeing it in action. would also agree with Alex that the testing is pretty great.

main questions and concerns I want to raise:

1. can we name things that are currently named "default" to "consumer"? "default" doesn't mean a lot to me in the context of the platform, and I suspect it might not to devs who aren't already familiar with the app, but "consumer" feels pretty clear whether or not you know a ton about the portal.
2. on a somewhat related note, it feels pretty strange to me that `SupportContactUsFormPublishing` wraps both consumer and provider support requests.
3. not new with your PR by any means, but the prefixing of everything in the `support` directory doesn't feel very useful to me. can we take this opportunity to start mvoing away from that a bit?

src/containers/support/supportContactUsFormPublishing/components/ContactDetailsFormFields.tsx
johanna-adhoc: Are these the FormField components that will be replacing the traditional Errorable components? If that is the case, why are they defined within supportContactUsFormPublishing? I thought the idea was to migrate those Errorable so they cane used in the future forms and the migration of the older forms

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { makeRequest, ResponseType } from '../../../utils/makeRequest';
+import '../SupportContactUsForm.scss';
```
johanna-adhoc, line 6: I see in comments that the old form contactUs will be defunct once this hit prod. Is there a ticket in Quokka backlog to deal with code clean up?
Once the migration happens I don't see why we need the folder "supportContactUsFormPublishing". All the files within this folder should be at the root of support if we are following the same pattern.

src/containers/support/supportContactUsFormPublishing/components/ContactDetailsFormFields.tsx
johanna-adhoc: Are these the FormField components that will be replacing the traditional Errorable components? If that is the case, why are they defined within supportContactUsFormPublishing? I thought the idea was to migrate those Errorable so they cane used in the future forms and the migration of the older forms

johanna-adhoc (commented):
Left some comments on the code. I would also like to know if there will be a follow up to migrate the other form in the app.

TS000 (commented):
I agree with the other teams issues as I found a number that align with what they found. I don't see a reason to duplicate issues so I'll be keeping an eye on this PR. Thank you for pulling Formik into this, looks like a nice feature set!

src/components/formField/FormField.tsx
```diff
@@ -0,0 +1,46 @@
...
+      {
+        description &&
+          <div id={descriptionId}>
+            {description}
+          </div>
+      }
+      <span id={errorId} className={validationClass}>
+        <ErrorMessage name={name} />
+      </span>
+      <Field id={`formField${name}`} className={fieldClass} name={name} required={required} {...props} aria-describedby={`${errorId} ${descriptionId}`} />
```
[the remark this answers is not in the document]
RockslideJr: I'll add in `aria-invalid` and `toHtmlId` here. I agree with your point 3

src/components/subNav/SubNavEntry.tsx
```diff
@@ -20,7 +20,7 @@ const linkStyles = classNames(
 interface SubNavEntryProps {
   // the onClick prop is used to close the mobile nav, not for anything related to the native link behavior
   onClick: () => void;
-  to: string;
+  to: ComponentPropsWithoutRef<NavLink>['to'];
```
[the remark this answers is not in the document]
RockslideJr: My thought here was that since this prop is getting passed directly to the `NavLink`'s `to` prop, this prevents us from having to duplicate that type, or make updates to it when/if it changes. And most importantly, I'm using the `LocationDescription` version of the underlying `to` prop when calling this with query params.  I could switch to just using a string containing the query params though.

I know VSCode has good support for resolving this to it's underlying type when doing utilizing it's type checking functionality (hovering over types and such), which helps reduce the confusion for people who don't know the types, but I'm not sure about other editors.

TL;DR: I would push for keeping in this update, but if the general consensus is to keep it as-is I'll make the change back to `string`.

src/containers/support/SupportContactUs.tsx
```diff
@@ -45,7 +61,16 @@ const SupportContactUs = (): JSX.Element => {
...
+            render={(): ReactNode => (
+              <>
+                <p>From this page, you can ask us questions, get help or support, or get started with publishing your API. You can also find answers to many common questions on our FAQ page.</p>
+                {defaultType && <SupportContactUsFormPublishing onSuccess={onSuccess} defaultType={defaultType} />}
```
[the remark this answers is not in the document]
RockslideJr: I certainly don't mind switching around the names. I suppose generally i was just being lazy and trying to avoid renaming old files that i would remove in  a quick follow-up, but you're right that the naming as is doesn't make a lot of sense. I'm not a huge fan of calling things `Legacy` since there's always a chance that it _doesn't_ get removed, but in this case I'm pretty confidant that it will, so that seems like a good update to me!

src/containers/support/SupportContactUs.tsx
```diff
@@ -1,10 +1,15 @@
...
+import React, { ReactNode, useEffect, useState } from 'react';
+import { useLocation } from 'react-router';
 import Helmet from 'react-helmet';
+import { FormType } from '../../types/contactForm';
+import { FLAG_API_PUBLISHING } from '../../types/constants';
+import { Flag } from '../../flags';
 import { PageHeader } from '../../components';
 import SupportConfirmation from '../../content/supportConfirmation.mdx';
 import SupportContactUsForm from './SupportContactUsForm';
+import SupportContactUsFormPublishing from './supportContactUsFormPublishing/SupportContactUsFormPublishing';
```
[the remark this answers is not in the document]
RockslideJr: I'm totally on board with removing the redundant component naming. Since i need to do a follow-up PR to this at some point in the near future to remove the FF and the old version of the form, what do you think about doing the renaming in that PR?

src/containers/support/SupportContactUs.tsx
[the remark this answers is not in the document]
RockslideJr: And yes, the `defaultType` check if for it being undefined before the `useHistory` hook in the containing component resolves the query params, which doesn't happen on initial render

src/containers/support/SupportContactUs.tsx
```diff
@@ -27,6 +32,17 @@ const headerProps = {
 
 const SupportContactUs = (): JSX.Element => {
   const [sent, setSent] = React.useState(false);
+  const [defaultType, setDefaultType] = useState<FormType>();
```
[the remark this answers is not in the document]
RockslideJr: my thought with default was that this query param and variable only represents the type that the form is when first loaded, since it could be changed afterwords without updating this query param. Changing this variable once the form is loaded won't change the form state. I'm concerned that changing the naming here may confuse people into thinking that changing this value _will_ change the form state, but i'm not sure that's a huge concern, as long as the parameter on the form component is still named `defaultType` or maybe something like `initialType`. Thoughts?

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.test.tsx
RockslideJr: Yeah that's exactly what i was doing at first, but I ran into issues with the ordering. Spying on `JSON.stringify` seemed the best way to check that all the correct values were being passed without tests breaking on ordering.

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.tsx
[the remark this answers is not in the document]
RockslideJr: I think i like the idea of keeping the external and internal urls in different files, so the `paths.ts` can focus on internal paths.

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.tsx
RockslideJr: My original plan was to do the cleanup work as part of the same ticket, but that seems dumb, so I made a clean up ticket in our backlog (https://vajira.max.gov/browse/API-5000).

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.tsx
RockslideJr: The current linting on the project isn't actually running prettier. If that's something we want I think we should have a ticket somewhere to hook it up and clean up any existing prettier issues (assuming we don't already have a ticket for that somewhere)

src/containers/support/supportContactUsFormPublishing/components/ContactDetailsFormFields.tsx
RockslideJr: @mjlumetta Not as far as the designs were concerned 🤷 

The `FormField` component is indeed the component that would be replacing the errorable components, but it's currently being defined in the top level component directory

src/containers/support/supportContactUsFormPublishing/components/DefaultFormFields.tsx
RockslideJr: Yup, i double checked on the removal of the API checkboxes with Brent, but I'll let him explain more

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.test.tsx
mjlumetta: ahh yeah, okay. sounds good enough to me, no need to do anything more complicated here 🙃

src/containers/support/supportContactUsFormPublishing/components/PublishingFormFields.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <FormField
+        label="Is there anything else we should know about your API, how it’s used, or what you need from us?"
+        name="apiOtherInfo"
+        as="textarea"
+      />
+    </>
+  );
+};
+
+export default PublishingFormFields;
```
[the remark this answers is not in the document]
RockslideJr: Does this standard apply everywhere, or just in the top level components directory? I followed the pattern from the `FAQ` and `Overview` support sub-components, which use default exports

src/types/contactForm.ts
RockslideJr: I agree! I must have gotten these mixed in with the rest of the types when i was moving them around and never noticed. I'll update

src/utils/validators.ts
RockslideJr: My plan was (c), so that i wouldn't have to touch the other form is this PR.

src/containers/support/SupportContactUs.tsx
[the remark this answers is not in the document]
mjlumetta: I'm a little confused with the logic you're envisioning here. as far as I can see, changing this value _could_ change the form state - it actually depends on Formik's implementation and how the `initialValues` prop works.

1. call `setDefaultType` with a new value.
2. the form component gets a new `defaultType` prop and re-renders.
3. `Formik` re-renders with a new `initialValues` prop. what does Formik do here? presumably it would only use that prop on the very first render, but I don't know for sure that it's smart enough to do that.

I guess, though, I'm a little confused as to why this value needs to be state at all, or why it needs to be set from within a `useEffect` hook (well, I guess the second statement follows directly from the first... anyway). we already have the `useLocation` hook, so why not just read it and set a plain old variable based on the search part of the URL, and then set the `defaultType` based on that? does that make any sense, or am I missing something?

src/components/subNav/SubNavEntry.tsx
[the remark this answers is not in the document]
mjlumetta: that's cool with me! if any other reviewers have preferences, feel free to comment, but I understand that tradeoff.

src/containers/support/SupportContactUs.tsx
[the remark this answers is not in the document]
mjlumetta: and I guess as far as the name of the query param itself goes, I was thinking that "type" would be comprehensible enough to a user, more so than "default", although feel free to disagree there. and I was thinking that the user's understanding would take precedence for something that's visible to them, if that makes sense.

src/utils/validators.ts
mjlumetta: sounds good to me! as long as they're going to have the updated names eventually, that works for me

src/containers/support/SupportContactUs.tsx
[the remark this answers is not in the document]
RockslideJr: You know, honestly, I have no idea why i thought this whole setup was necessary. You are absolutely correct. Also I'm ok with changing the param name as well.

src/containers/support/SupportContactUs.tsx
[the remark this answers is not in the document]
mjlumetta: I'd agree with the not calling things `Legacy` point in general, but at the same time, one of them has to go, so 🤷 I'm also pretty confident that it's safe enough to do in this case. thanks for making the change!!

on the `defaultType` check, I'd say a couple of things:

1. said this in a separate comment, but I'm not sure it needs to be a state value? and if so, it doesn't have to be falsy on the first render.
2. if it does have to be state and it's going to cause stuff not to render on the first time around, can we handle it at a lower level? seems like the `ContactDetailsFormFields` should be unaffected... although I guess `Formik` is above them, and you need it for `initialValues`, so I think I answered my own question. so (1) is the only real consideration, unless that's not the reason you're handling it at this level.

src/containers/support/ContactUsForm/ContactUsForm.tsx
```diff
@@ -0,0 +1,108 @@
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import React, { ReactNode } from 'react';
+import { Formik, Form, Field } from 'formik';
+import { CONTACT_US_URL } from '../../../types/constants';
+import { makeRequest, ResponseType } from '../../../utils/makeRequest';
+import './ContactUsForm.scss';
+import { SupportContactUsFormState, FormType, FormData } from '../../../types/contactUsForm';
```
mjlumetta, line 8: I would suggest these could potentially be colocated as well, though that's not absolutely necessary. something that is React component state I would generally expect to be colocated, although at this point it's not directly used as state anymore, since it's handled by Formik internally

src/containers/support/ContactUsForm/ContactUsForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+      apiInternalOnlyDetails: values.apiInternalOnlyDetails,
+      apiOtherInfo: values.apiOtherInfo ?? '',
+      type: values.type,
+    };
+  }
+};
+
+const SupportContactUsFormPublishing = ({ onSuccess, defaultType }: SupportContactUsFormProps): JSX.Element => {
+  const initialValues: SupportContactUsFormState = {
+    apiDetails: '',
```
mjlumetta, line 50: what about `apiDescription` and `apiOtherInfo` and `organization` in this object? I see in the Formik docs it says

> Even if your form is empty by default, you must initialize all fields with initial values otherwise React will throw an error saying that you have changed an input from uncontrolled to controlled.

I don't see the uncontrolled => controlled error in the console, but I'm not sure how this is or should be working, either.

src/containers/support/ContactUsForm/ContactUsForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+import validateForm from './validateForm';
+
+const SupportContactUsFormPropTypes = {
+  defaultType: PropTypes.oneOf([FormType.CONSUMER, FormType.PUBLISHING]).isRequired,
+  onSuccess: PropTypes.func.isRequired,
+};
+
+type SupportContactUsFormProps = PropTypes.InferProps<typeof SupportContactUsFormPropTypes>;
+
+const processedData = (values: SupportContactUsFormState): FormData => {
```
mjlumetta, line 21: what's the relationship between `SupportContactUsFormState` and `FormData`? seems like there's a high degree of overlap... could they be the same type?

src/containers/support/ContactUsForm/components/ConsumerFormFields.tsx
```diff
@@ -0,0 +1,18 @@
+import classNames from 'classnames';
+import React, { FC } from 'react';
+import { FormField } from '../../../../components';
+
+const DefaultFormFields: FC = () => (
```
mjlumetta, line 5: `ConsumerFormFields`?

src/containers/support/ContactUsForm/components/ContactDetailsFormFields.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <legend className={classNames('vads-u-font-size--lg')}>Tell us about you</legend>
+    <div className={classNames('vads-l-row', 'vads-u-margin-top--2p5')}>
+      <div className={textFieldClasses('right')}>
+        <FormField label="First name" name="firstName" required />
+      </div>
+      <div className={textFieldClasses('left')}>
+        <FormField label="Last name" name="lastName" required />
+      </div>
+    </div>
+    <div className={classNames('vads-l-row', 'vads-u-margin-top--4')}>
```
mjlumetta, line 23: this top margin is leaving some pretty wonky spacing in certain errors cases

Screen Shot 2021-02-08 at 5 41 05 PM

mjlumetta (commented):
I would say this is in pretty good shape. have a fair number of fine tuning comments or miscellaneous cleanup of names for consistency, things like that, and I've got a couple of questions about Formik. but as far as functionality goes, I think everything's covered. may try to do a little more robust screen reader testing (did a very cursory pass this afternoon) but I would say this is getting pretty close.

src/containers/support/ContactUsForm/ContactUsForm.tsx
RockslideJr: I can see the confusion here. I think I need to rename `FormData` to something like `SubmissionData` maybe? `FormData` is a stricter type that enforces that only the data needed for whatever version of the form being submitted is present (so that you're not submitting api details with the consumer version and such)

src/containers/support/ContactUsForm/ContactUsForm.tsx
RockslideJr: Yeah I moved these into the types folder since they're just the type of the state and not the state itself

src/containers/support/ContactUsForm/ContactUsForm.tsx
RockslideJr: Ah, good catch, thanks. Not sure why the error's not showing up.

src/containers/support/ContactUsForm/components/ContactDetailsFormFields.tsx
RockslideJr: I'm interested in your thoughts on this (and @brentschneider 's thoughts as well). The top margins of these fields have been specifically set based on Brent's designs, so we're not using the default label margin. I can add in some custom css to remove the top margin that is being added by the `usa-input-error`, which will get things to look as expected with the custom margins we're setting, or we can stop using the custom margins and revert to the default stylings, which will get everything working as expected, but won't match the designs perfectly

src/containers/support/supportContactUsFormPublishing/components/DefaultFormFields.tsx
brentschneider: Correct— they are being removed. 
_At one point, there was an idea to hook the fields up to the salesforce flow, but apparently, we don't have that feature in our license._

src/containers/support/ContactUsForm/components/ContactDetailsFormFields.tsx
brentschneider: If it helps with troubleshooting, the error spacing should only push down the spacing by one token unit (1).

Error-spacing

RockslideJr: I added custom styles for this component to get this issue resolved and keep the custom spacing here. Not sure i'm a fan of it though.

brentschneider:
@RockslideJr the fixes you applied in commit [99ad59e](https://github.com/department-of-veterans-affairs/developer-portal/pull/622/commits/99ad59ef0fd8bce7bfc1a7bc915d8942d01295b1) look fantastic, thank you so much for pushing this through! 🎉

src/containers/support/supportContactUsFormPublishing/SupportContactUsFormPublishing.tsx
johanna-adhoc: That works for me! just please don't forget the ticket in the backlog :)

johanna-adhoc:
> Left some comments on the code. I would also like to know if there will be a follow up to migrate the other form in the app.
@RockslideJr  Can you also create a ticket to follow up with the other form migration?

src/containers/support/ContactUsForm/ContactUsForm.scss
```diff
@@ -0,0 +1,12 @@
+.va-api-contact-us-form {
+  label,
+  textarea {
+    max-width: 100%;
```
mjlumetta, line 4: what are these taking care of?

src/containers/support/ContactUsForm/components/ContactDetailsFormFields.tsx
mjlumetta: yeah, not sure what the best thing to do is here. not sure why `usa-input-error` adds such a huge margin in the first place. changes look more like what I'd expect, though - not such a big jump in spacing.

src/containers/support/ContactUsForm/components/ConsumerFormFields.tsx
```diff
@@ -0,0 +1,18 @@
+import classNames from 'classnames';
+import React, { FC } from 'react';
+import { FormField } from '../../../../components';
+
+const ConsumerFormFields: FC = () => (
+  <>
+    <h3 className={classNames('vads-u-margin-top--4', 'vads-u-margin-bottom--0')}>Description</h3>
```
mjlumetta, line 7: should this be an h2? it doesn't logically fall under the h2 over it. also, I don't think that h2 should be a heading in the first place.

src/containers/support/ContactUsForm/ContactUsForm.tsx
mjlumetta: yeah, I think the `SubmissionData` name is clearer, for sure. could also potentially see a name like `RequestData` along those lines. I wonder if it'd be useful to add the `ContactUs` prefix as well - I would lean yes - but I wouldn't say this requires any further refinement on the names unless you find it meaningful. I would suggest adding a comment to the types file with the distinction from your previous comment, though.

RockslideJr:
@johanna-adhoc Sorry, i missed your comment the first time around! I've created a ticket in quokka's backlog for refinement ([API-5081](https://vajira.max.gov/browse/API-5081))

johanna-adhoc (approved):
LGTM and all my questions have been addressed

mjlumetta (commented):
okay, I would say this is ready to approve, except for one big thing, which is that I can't enter text into the textareas using VoiceOver (in Safari). and I can do it for the description field on the old form. any thoughts as to why that might be? I would say user error is definitely possible, but it's not obvious to me what that is since I can do that with the prod form.

mjlumetta (dismissed):
did some more testing and didn't encounter the same issues with textareas with VoiceOver in Chrome. also, putting focus directly in the textarea with the mouse or keyboard with VoiceOver in Safari didn't have any problems, so I think it's safe to approve. might want to keep an eye on that and do more digging at some point, but this PR should be good to go.

src/containers/support/ContactUs.tsx
```diff
@@ -45,7 +56,23 @@ const SupportContactUs = (): JSX.Element => {
           </Helmet>
           <PageHeader {...headerProps} />
           <GitHubSnippet />
-          <SupportContactUsForm onSuccess={onSuccess} />
+          <Flag
+            name={[FLAG_API_PUBLISHING]}
+            render={(): ReactNode => (
+              <Flag
+                name={[FLAG_API_PUBLISHING_CONTACT_FORM]}
+                render={(): ReactNode => (
```
christineyost, line 64: If you're not happy with how this looks, we could use a similar pattern to category flags, which looks like a map of flag name to boolean... but since this is temporary anyway this seems fine if you want to keep it like this!

mjlumetta (approved):
LGTM ✨ the changes to the form aren't going to be deployed in prod yet, though?

alexandervalencia (commented):
Overall this looks great! I had a few thoughts that are hopefully helpful. Thankful for the other reviewers already covering a11y and other good points.

johanna-adhoc (approved):

src/flags.ts
```diff
@@ -31,6 +32,7 @@ const getFlags = (): AppFlags => {
 
   return {
     api_publishing: process.env.REACT_APP_API_PUBLISHING === 'true',
+    api_publishing_contact_form: process.env.REACT_APP_API_PUBLISHING_CONTACT_FORM === 'true',
```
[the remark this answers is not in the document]
RockslideJr: oh this is a great suggestion. Big Brain Time.

src/containers/documentation/ApiDocumentation.test.tsx
```diff
@@ -53,6 +53,7 @@ jest.mock('react-router-dom', () => ({
 describe('ApiDocumentation', () => {
   const defaultFlags: AppFlags = {
     api_publishing: false,
+    api_publishing_contact_form: false,
```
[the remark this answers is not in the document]
RockslideJr: I like this idea! I'm not sure i'd want to put a test specific method directly into the flags file though. What do you think about creating a `test` sub-directory in the `utils` directory, or a `utils` sub-directory in the `test` directory (which is currently only being used for storing visual regression snapshots) and putting the method in there?

alexandervalencia (approved):
LGTM 🚀

PR #624, opened 2021-02-05
Change route for Claims Attributes API to Docserver

rtravitz:
### Description
This PR supports [API-4350](https://vajira.max.gov/browse/API-4350). It changes the documentation route for the Claims Attributes API from pulling straight from the provider to using the Docserver.

### Process

All checks N/A

- [ ] Docs updated
  
### DIFF
--- a/src/apiDefs/data/benefits.ts
+++ b/src/apiDefs/data/benefits.ts
@@ -61,7 +61,7 @@ const benefitsApis: APIDescription[] = [
     description: 'Improve claim routing',
     docSources: [
       {
-        openApiUrl: `${OPEN_API_SPEC_HOST}/services/claims-attributes/v1/docs/openapi.json`,
+        openApiUrl: `${OPEN_API_SPEC_HOST}/internal/docserver/api/claims-attributes/1/oas/latest`,
       },
     ],
     enabledByDefault: false,


src/apiDefs/data/benefits.ts
```diff
@@ -61,7 +61,7 @@ const benefitsApis: APIDescription[] = [
     description: 'Improve claim routing',
     docSources: [
       {
-        openApiUrl: `${OPEN_API_SPEC_HOST}/services/claims-attributes/v1/docs/openapi.json`,
+        openApiUrl: `${OPEN_API_SPEC_HOST}/internal/docserver/api/claims-attributes/1/oas/latest`,
```
RockslideJr, line 64: are we worried about this change ending up in higher environments before the docserver gets there? This might be worth feature flagging

rtravitz: The app itself isn't even deployed to higher environments. The release of Claims Attributes itself on the Developer Portal is already feature flagged here: https://github.com/department-of-veterans-affairs/developer-portal/blob/fd0584239e4b0913c757a192f492328aedb9e5b8/.env.dev#L13

Not worth adding an extra one IMO.

RockslideJr (approved):

PR #626, opened 2021-02-09
API-4607 Remove security block requirement for curlify plugin

hinzed1127:
### Description

**Goal**: Let curl examples show for endpoints that currently doing have a security block in the generated `openapi.json` 

**Result**

Note the absence of the `Api Key` or `Bearer Token` input

Screen Shot 2021-02-09 at 4 43 23 PM

This covers changes for
https://vajira.max.gov/browse/API-4607 and
https://vajira.max.gov/browse/API-4959

### Process

- [ ] Docs updated
  

SlakrHakr (approved):

rtravitz (commented):
Everyone hold off of reviewing this one for a minute—it may be moot.

There is a top-level security element in OASv3, but that only defines the scheme and it may need to be applied on a per-operation level to make the curl form show up properly.

In the case of Claims Attributes, we'd make this fix in the Docserver.

In any case, we never want to remove the `hasSecurity` check entirely, because every API on Lighthouse is secured with either an API key or a Bearer token that comes from OAuth. Without including which one that is in the OAS, the generated curls will be useless.

hinzed1127:
closing this in favor of the approach @rtravitz outlined above

PR #630, opened 2021-02-15
[API-5128] Code Snippets Single Line Bug

gileswells:
### Description

https://vajira.max.gov/browse/API-5128

This bug would show up when viewing the auth docs on the default Benefits Claims url where all the code snippets would show up in a single line vs respecting the newline characters in the markdown files.

### Process

- [ ] Docs updated
  
### Requested feedback

I am not currently happy with the code that patches this issue. There is an underlying issue, that I was not able to determine, but the patch I applied was that changing the `key` on the code snippets would force re-render the snippet and fix the single line issue. The second issue was that even with this key change, on initial page load where the key is set from the very beginning, there wouldn't be a change to trigger the patch to re-render it.


src/components/oauthDocs/AuthCodeFlowContent.mdx
```diff
@@ -42,7 +42,7 @@ The Veteran will need to grant your application access permission. To do this, d
 
 After a Veteran gives authorization for you to access their data, their browser will be redirected to your application with the response shown below, which returns the `code` and `state` parameters you must use to make a request to our authorization service. We require the state parameter for all authorization code grant flows.
 
-<CodeWrapper>
+<div></div><CodeWrapper key={`snippet-1-${props.selectedOption}`}>
```
gileswells, line 45: This `<div></div>` is needed to prevent the markdown parser from rendering `<CodeWrapper ... >` as text and then complaining that `</CodeWrapper>` below the code sample is trying to close something that never opened.

src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -50,6 +50,9 @@ const setInitialApi = (
   const availableApis = getAllOauthApis().filter((item: APIDescription) => !isApiDeactivated(item));
   const isAnApi = availableApis.some((item: APIDescription) => item.urlFragment === apiQuery);
   const api = apiQuery && isAnApi ? apiQuery.toLowerCase() : DEFAULT_OAUTH_API_SELECTION;
+  if (api === DEFAULT_OAUTH_API_SELECTION) {
```
gileswells, line 53: This is part of the patch to force an initial key change on the code snippets to forcibly trigger a re-render when the initially selected API is the default (Benefits Claims).

src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -66,8 +69,9 @@ const AuthorizationDocs = (): JSX.Element => {
     if (initializing.current) {
       // Do this on first load
       initializing.current = false;
-
-      setInitialApi(history, location.search, dispatch);
+      setTimeout(() => {
```
gileswells, line 72: This `setTimeout()` is here because in order for the initial render to be triggered with the single line error and then re-render when the store is changed just afterwards the first render needed to be delayed slightly so that it could bug and then self correct.

gileswells (commented):
@mjlumetta Here's the jumping off point for the things that I do not like about this patch and the reasoning for them. 

I had originally considered putting these keys on the `BuildingOIDContent` component itself (or on the two included markdown components) but that would trigger a re-render of basically the whole page and cause the browser to lose focus on API change.

src/containers/documentation/AuthorizationDocs.tsx
mjlumetta: oooof, okay. so, accepting that it initially renders wrong for ~ unknown reasons ~, we know that

1. it's going to render incorrectly.
2. after the timeout clears, it sets the initial API from this hook.
3. it re-renders because of the state change, forcing the correction.

am i following that correctly?

src/containers/documentation/AuthorizationDocs.tsx
mjlumetta: gotcha. and this includes when `?api=claims` is already set explicitly, or just without the query param set at all? also not following what happens on the render after the OAuth selection is set to "none". 

why is the value "none", by the way? is it just a placeholder, or is it intended to communicate that they went straight to `/explore/authorization`, no query param?

thinking that if this is the workaround that we need to go with, then we're gonna need pretty good comments explaining what's going on here and why, to the best of our knowledge.

mjlumetta (commented):
okay, whew. first off, thanks for digging into this one; I think we can agree it's pretty messy and that it comes from some underlying issue, as you said. I'm also not thrilled with the patch and find it really hard to wrap my head around, but at the same time, gotta give you kudos for having a creative solution that appears to prevent the issue. with that said, here are a few of my lines of thought on the ways forward, with the caveat that I don't have any unified thoughts around it so they're pretty loose.

1. the key fix makes perfect sense to me. no questions there.
2. less so on the div fix; that one's really not clear to me, especially since it's only on some of them. but not the hardest part of the PR - the logic in `AuthorizationDocs` is the real tricky piece. I've left some questions about how that works, and I think we're going tot have to make some pretty extensive comments there.
3. re: the underlying issue, I'm wondering if maybe we can brainstorm that as a group. you, Eddie, and me at least, Charley if he's interested, and maybe grab a FE eng or two from the other platform teams if they're interested. if we don't come up with anything, that's fine, but wonder if something will come up if we talk it out together.
4. the `markdown-component-loader` in the VA Github we're using is 125 commits behind the `develop` branch of the main repo. I would say that we should merge the upstream changes back into here and see if that fixes anything, _but_ also there hasn't been a new release of `markdown-component-loader` in [NPM](https://www.npmjs.com/package/markdown-component-loader) in 3 years, since before our program even existed. it has 7 weekly downloads. so I have little to no idea what's up with that. my best hypothesis would be that the author/maintainers/users are the same people/org and they're pointing directly to the repo, not using a package. anyway, my question is, should we try merging in the upstream changes or at least browsing through them in case there's a fix? maybe you've already done that. or maybe we should consider looking for more actively maintained alternatives.
5. finally, would rewriting the auth docs to use more TSX and less MDX possibly be an alternate path to solving the issue? I know we were considering leaning toward that approach at one point, sorta considered it a tossup in the end, and then went to the more MDX side. I'm wondering if we pull out most stuff into TSX and then have individual MDX files for each code block where we can say `<MyCodeBlock selectedApi={api} />`, would that take some of the burden off of the `markdown-component-loader`? realize that would be moving _a lot_ of stuff around, so I understand that's not necessarily appealing.

anyway, that's a lot of miscellaneous thoughts. I would say I'm okay with merging this fix if it's what it takes to fix the issue, because that's what's most important, with main feedback for that path being that we should add code comments along the lines of what you've included in the PR. but, given that this workaround isn't very pretty and neither of us is particularly happy with it, would be good if we could put our heads together and figure out an alternative. anyway, great work digging into this and getting something that works.

src/components/oauthDocs/AuthCodeFlowContent.mdx
gileswells: Yes. Just adding the prop makes it fail and it does only fail on the specific lines where I added the `<div></div>`. Why it's only on some lines that it fails? I do not know. All I know is it makes no sense.

src/containers/documentation/AuthorizationDocs.tsx
gileswells: This is needed whenever the initial set it `claims`. It doesn't matter if the URL had the query string in it on arrival to the auth docs or not, it needs it both ways.

"none" was just a random placeholder I put in there. Knowing this PR would spark larger conversations I didn't overthink it and just tossed none in there and it did the trick for the time being. In no way do I expect this PR to get merged as is.

src/containers/documentation/AuthorizationDocs.tsx
gileswells: Yes that is the render flow that it follows.

src/components/oauthDocs/AuthCodeFlowContent.mdx
anthony-kopczyk: Not sure how the markdown renderer works. Would this be an issue with the markdown renderer that we could maybe fix, or is it likely something outside of our control? Wondering if we could wrap it in a react fragment maybe rather than have the div.

anthony-kopczyk: I don't think this is related to the div thing but thought this would be a good place to put this. I went to compare the benefits claims link to oauth versus the Health API auth docs.

The code the CodeWrapper generated in the health api docs had a `pre` without any css classes but the benefits claims one does have a css class (language-plaintext). The benefits api css class on the 'pre' tag and 'code' tags start with spaces as well which seems odd.

Removing the css class from pre doesn't fix anything, but it seems odd that the css classes seem to get applied differently for what's basically the same page.

mjlumetta: I don't quite follow your question @anthony-kopczyk. are you talking about a specific code block?

anthony-kopczyk: Just realized I think prod is showing legacy docs, but I imagine it may be the same root issue affecting both with the css class thing

mjlumetta: yeah, prod still has the legacy docs. but I don't know what the issue you're referring to is? sorry for not following here

anthony-kopczyk: My thoughts weren't very well gathered. haha!

Two separate things.
1. Can we use a react fragment to glitch this into working, like using the div
2. (read below)

In production, I see inconsistency in what's rendered from the markdown between the claims auth page and benefits auth page.

The claims auth page was rendering with:
```
<pre class=" language-plaintext">
  <code class=" language-plaintext">Stuff</code>
</pre>
```

The benefits auth page was rendering with:
```
<pre>
  <code class="language-plaintext">Stuff</code>
</pre>
```

Manually changing the css classes in the inspector didn't change anything with the layout. But it seems really odd that they're getting rendered differently when it should be the same components.

anthony-kopczyk: After talking with Giles, it looks like the error I put above only happens on my Chrome browser but does not happen in my firefox browser. I think it might be some kind of caching issue, not an issue with the code.

mjlumetta: ah huh... that is weird about the inconsistency between pre tags having/not having the language CSS class, and with the extra space. still not sure where I should look to see the examples though - which code block is it? also getting lost here because there aren't authorization docs for Claims or Benefits in prod. I am genuinely curious in case it provides any clues to the bizarre behavior that we're seeing.

src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -50,6 +50,14 @@ const setInitialApi = (
   const availableApis = getAllOauthApis().filter((item: APIDescription) => !isApiDeactivated(item));
   const isAnApi = availableApis.some((item: APIDescription) => item.urlFragment === apiQuery);
   const api = apiQuery && isAnApi ? apiQuery.toLowerCase() : DEFAULT_OAUTH_API_SELECTION;
+  if (api === DEFAULT_OAUTH_API_SELECTION) {
+    /*
+     * We need to trigger a change to the API selection so the key will change and trigger a
+     * re-render of the <CodeWrapper /> component to fix the bug.
+     * Because other APIs besides the default include that change we only need to do it on claims.
```
mjlumetta, line 57: when you say "other APIs besides the default include that change", you mean the change on line 61, right? since the keys _have_ to change anyway, we don't need to artificially force a change. might be worth clarifying that part explicitly.

also, can you remind me when/how the "none" value gets set back to "claims"? might be worth noting that tas well.

src/containers/documentation/AuthorizationDocs.tsx
```diff
@@ -66,8 +74,13 @@ const AuthorizationDocs = (): JSX.Element => {
     if (initializing.current) {
       // Do this on first load
       initializing.current = false;
-
-      setInitialApi(history, location.search, dispatch);
+      setTimeout(() => {
+        /*
+         * This is needed because the MDX loader will needs to be stepped back from the initial
+         * render to allow that first render to complete (badly) before we do a good render
```
mjlumetta, line 80: if you need to make another commit, might be worth saying "we do not know why the first render is bad, and we are fixing it by changing the loader". think that's obvious to those of us who've been looking at said issue, but others looking later might need a little more context.

mjlumetta (approved):
looks good to me. I've left a couple thoughts on comments in case you make another commit, but should be good to go. I'll re-approve if you make any other changes.

src/components/oauthDocs/AuthCodeFlowContent.mdx
anthony-kopczyk: If I was going straight to the URL or through the sidenav then it was correct, but if I got to the authorization docs through the link inside of the Benefits Claims page (under Design -> Authentication), it had the inconsistency, even though it was going to the same url.

mjlumetta (approved):
re-approving

PR #631, opened 2021-02-15
API-3904: Dependabot Setup

anthony-kopczyk:
### Description

Jira Ticket: https://vajira.max.gov/browse/API-3904

### Process

- [N/A] Docs updated
  
### Requested feedback

If this looks good to others I think this is good to merge. I'm not sure how to really test it beyond waiting for the first of the next month. We could set this to daily and do another PR to update at a later date if desired. Let me know in the review comments.


johanna-adhoc (dismissed):
LGTM

johanna-adhoc:
@anthony-kopczyk Actually I do have a suggestions. Can you add labels to the automatic PRs? Add the label dependency and dependabot (dependabot label does not exist yet so you will need to create it first)

anthony-kopczyk:
> @anthony-kopczyk Actually I do have a suggestions. Can you add labels to the automatic PRs? Add the label dependency and dependabot (dependabot label does not exist yet so you will need to create it first)

Setup Guide (quoted below): https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates

> By default, Dependabot raises all pull requests with the 'dependencies' label. If more than one package manager is defined, Dependabot includes an additional label on each pull request.

This paragraph continues saying that if we have more than one package manager defined (we don't right now), it would also add an 'npm' tag for npm dependency updates. I can modify this to be 'dependency' and 'dependabot', but I'll wait to get feedback. Not sure if you knew that it should have some default labels.

## Other config options

This actually brings up a few other things I was thinking, but thought they might come up in the reviews.

###  Commit Messages

> By default, Dependabot attempts to detect your commit message preferences and use similar patterns. Use the commit-message option to specify your preferences explicitly.

This section continues detailing we can add a commit prefix, a commit prefix for dev dependencies, and 'scope' to include a list of the updated dependencies. This aspect I figured we could see how dependabot behaves and do a PR at a later date to update most of these options as we like.

### Reviewers

We can set reviewers to be added to the PRs if we want. Not sure if we want to do this or if we want our teams to assign them to themselves when they take a look.

### Time

We can set specific time to check for these updates. The current settings should result in the first of the month at a specific hour and minute (can't remember at the moment. Think it's like 5 AM UTC)

### Versioning Strategy

By default dependabot will try to upgrade the npm versions. We can change this to only update the lock file or to only update when it's required by a newer version. I think we'd probably want to keep this at the default, but added just in case.

johanna-adhoc:
@anthony-kopczyk 🤔 so the dependencies label will be automatically added, but what about a dependabot label. I am just trying to see how a PR can be easily recognized that it was created by dependabot. The dependabot label could help with that similar how we have the koala label unless there is a better way.

anthony-kopczyk:
Had a conversation with Gabriel. The security update PRs should be enabled via a setting in the repository. After looking further, this dependabot yml checks for version upgrades - separate from if there are any actual security issues. I didn't realize that before. Doesn't change much for the PR itself, but maybe changes the way we look at this PR.

I'm assuming this config will be ready by the security update PRs and create the same labels. We'll find out for sure when this kicks off the first time.

mjlumetta (dismissed):
yeah, this looks good to me. I had been under the impression that we could configure this to be minor + patch updates only, but it looks like that's not the case. that might change the strategy in the [dependency management doc](https://community.max.gov/x/YfzCgQ). but anyway, this looks like what we'd want to me.

mjlumetta:
also, I'm guessing the plan would be to merge after #632 goes in so that Dependabot doesn't start opening up PRs once merged?

anthony-kopczyk:
@mjlumetta

> yeah, this looks good to me. I had been under the impression that we could configure this to be minor + patch updates only, but it looks like that's not the case. that might change the strategy in the dependency management doc. but anyway, this looks like what we'd want to me.

Yeah. Looks like there's no way to achieve that right now.

> also, I'm guessing the plan would be to merge after #632 goes in so that Dependabot doesn't start opening up PRs once merged?

With the current settings it should only check the first of each month. But I'm not sure if it needs to be on the master branch to start working.

I expect us to be update this file semi-regularly. Right now there's nothing under 'Ignores' to ignore specific packages. It's likely we'll get some packages that we'll want ignored specifically in cases where we're unable to make the upgrades.

mjlumetta:
re: timing on the first of the month vs when merged, here's what I was reading:

> When you first enable version updates, you may have many dependencies that are outdated and some may be many versions behind the latest version. Dependabot checks for outdated dependencies as soon as it's enabled. You may see new pull requests for version updates within minutes of adding the configuration file, depending on the number of manifest files for which you configure updates.

https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates#about-version-updates-for-dependencies 

the `schedule.interval` option documentation doesn't account for this, so it sounds like they conflict? unsure what takes precedence there. but if you merge and it opens PR that are redundant with Johanna's updates, we can always close those PRs.

anthony-kopczyk:
> re: timing on the first of the month vs when merged, here's what I was reading:
> 
> > When you first enable version updates, you may have many dependencies that are outdated and some may be many versions behind the latest version. Dependabot checks for outdated dependencies as soon as it's enabled. You may see new pull requests for version updates within minutes of adding the configuration file, depending on the number of manifest files for which you configure updates.
> 
> https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates#about-version-updates-for-dependencies
> 
> the `schedule.interval` option documentation doesn't account for this, so it sounds like they conflict? unsure what takes precedence there. but if you merge and it opens PR that are redundant with Johanna's updates, we can always close those PRs.

Thanks! Yeah. We can wait to merge this to try and prevent duplicates

.github/dependabot.yml
```diff
@@ -0,0 +1,56 @@
...
+        versions: [11.x]
+      # Version 10+ has a dependency with postcss-loader v5
+      - dependency-name: autoprefixer
+        versions: [10.x]
+      # Version 7+ needs webpack 5
+      - dependency-name: copy-webpack-plugin
+        versions: [7.x]
+      # Version 5+ needs webpack 5
+      - dependency-name: html-webpack-plugin
+        versions: [5.x]
```
anthony-kopczyk, line 53: Preset some of these to the specific major version, so we'll get bothered if we delay the webpack upgrade too long.

.github/dependabot.yml
```diff
@@ -0,0 +1,56 @@
...
+      # UNUSED PLUGINS
+      - dependency-name: workbox-webpack-plugin
```
anthony-kopczyk, line 55: This is technically used, but it looks like we immediately unregister any service workers. So I labelled it unused for now as we might be removing this dependency.

johanna-adhoc:
I think we should change the interval to weekly given that we will be doing maintenance updates every two weeks. So the developer that does the update has the PRs available at that point.

johanna-adhoc (approved):
LGTM

TS000 (approved):

PR #636, opened 2021-02-18
[API-4730] Consumer Docs Toggle Flag and placeholders

gileswells:
### Description

https://vajira.max.gov/browse/API-4730

Created a feature flag for the consumer docs and provided a placeholder nav link and route. Set dev and staging to use this flag.

### Process

- [ ] Docs updated
  

src/components/navBar/NavBar.tsx
```diff
@@ -225,6 +226,23 @@ const NavBar = (props: NavBarProps): JSX.Element => {
               </SubNav>
             </li>
           </Flag>
+          <Flag name={[FLAG_API_CONSUMER]}>
```
mjlumetta, line 229: this might be a bit pre-emptive? but it doesn't do any harm either, so I'm fine with committing it

mjlumetta (commented):
looks good to me! have suggested the fix for the 404 issue, plus a name suggestion which is less important. will approve once the 404 thing is fixed.

src/Routes.tsx
```diff
@@ -61,6 +61,9 @@ export const SiteRoutes: React.FunctionComponent = () => (
     <Flag name={[FLAG_API_PUBLISHING]}>
       <Route path={PUBLISHING_PATH} component={Publishing} />
     </Flag>
+    <Flag name={[FLAG_API_CONSUMER]}>
```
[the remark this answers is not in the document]
gileswells: This did fix the 404 page

mjlumetta (dismissed):
lgtm ✨

src/Routes.tsx
```diff
@@ -19,52 +19,58 @@ import Support from './containers/support/Support';
...
+        render={(): JSX.Element => <Redirect to="/explore/verification/docs/authorization" />}
+      />
+      <Route path="/release-notes/:apiCategoryKey?" component={ReleaseNotes} />
+      <Route path="/news" component={News} />
+      <Route path="/support" component={Support} />
+      <Route
+        path="/providers/integration-guide"
+        render={(): JSX.Element => MarkdownPage(ProviderIntegrationGuide)}
+      />
+      {flags.api_publishing && <Route path={PUBLISHING_PATH} component={Publishing} />}
```
mjlumetta, line 66: so the thought here is that comparing against the individual flags from `getFlags` means that we don't actually need changes to the sitemap? makes sense

src/Routes.tsx
```diff
@@ -19,52 +19,58 @@ import Support from './containers/support/Support';
...
 import { Publishing } from './containers/publishing';
 import { CONSUMER_PATH, PUBLISHING_PATH } from './types/constants/paths';
-import { FLAG_CONSUMER_DOCS, FLAG_API_PUBLISHING } from './types/constants';
 
-export const SiteRoutes: React.FunctionComponent = () => (
-  <Switch>
-    <Route exact path="/" component={Home} />
-    <Route exact path="/index.html" component={Home} />
+export const SiteRoutes: React.FunctionComponent = (): JSX.Element => {
+  const flags = getFlags();
```
mjlumetta, line 27: what about using the `useFlags` hook instead of `getFlags`? not like `getFlags` is some ridiculous performance hit or anything, but then we wouldn't have to recompute them.

mjlumetta (commented):
one suggestion on the sitemap generation (`getFlags` -> `useFlags`) but that's it. lmk your thoughts on that and we can get this one moving.

src/Routes.tsx
gileswells: Basically. If we remove the route based on the flag then it will stop us from needing to make any new blocking rules from them being included in the sitemap. Also this does the better job of preventing urls from going to production (albeit not linked from anywhere) before we're ready for them to go live.

src/Routes.tsx
gileswells: Using that will result with this error for improperly using hooks. This was one of the first things I tried. https://reactjs.org/docs/error-decoder.html/?invariant=321

mjlumetta: really? isn't it already a function component with your changes? I tried it locally just now, and I'm not seeing it. maybe I didn't test the same thing you did, though.

gileswells: You have to run `npm run build` to get the sitemap to generate. That's where the error is. The page loads fine and the updates to the routes being present or missing based on the flag work perfectly but it's the sitemap that breaks and throws that error.

mjlumetta: ahh, i see. i hadn't tried the build. yeah, so, i think it's okay to use `getFlags` then if we have to. if so, i'd say it might be worth a comment clarifying why we're using `getFlags` instead of `useFlags`.

the other thing that might be worth trying is going back to creating an anonymous function component in the sitemap config for `topLevelRoutes` that wraps `Routes` inside a `FlagsProvider`. i tried to run a build with that config and `useFlags`, and i _think_ it might work. if you feel that's worthwhile, you can try it out and see if it works; otherwise, let's go with what you've got now.

mjlumetta: i take that back, the sitemap is basically empty. approving now.

mjlumetta (approved):

PR #640, opened 2021-02-22
Fixed immer vulnerability with force resolutions for transitive dependencies

johanna-adhoc:
### Added force resolution packages to enforce version in transitory dependency

Immer high vulnerability in react-dev-utils(dev dependency) is blocking PR builds. This PR add the force resolution package to force the immer version that does not contain the vulnerability

### Process

### Requested feedback
All feedback is welcome.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,22 +4,22 @@ FROM node:12
 
 # Install chromium dependencies
 RUN apt-get update && apt-get install -y libxss1 libxtst6 wget --no-install-recommends \
-    && wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
-    && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
-    && apt-get update \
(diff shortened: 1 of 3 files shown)


gileswells (approved):
🚀 LGTM

Literally my only question is the difference on number of spaces indenting the `Dockerfile` but that's a non-issue

johanna-adhoc:
> 🚀 LGTM
> 
> Literally my only question is the difference on number of spaces indenting the `Dockerfile` but that's a non-issue

My auto formatter is doing that, it takes the eslint configuration and adds the spaces :/

Dockerfile
```diff
@@ -30,6 +30,11 @@ RUN npm install -g @sentry/cli
 
 COPY package.json package-lock.json ./
 
+# Add write permission to package-lock.json so npm resolutions can set the correct dependency version
+USER root
```
mjlumetta, line 34: I think this whole file could run as root anyway now if we wanted... but not super important for this PR

mjlumetta (approved):
LGTM. thank you for fixing this @johanna-adhoc!

TS000 (approved):
Looks great! Take that immer!

PR #647, opened 2021-02-25
Api 5360 add dropdown and revise color ratio

charleystran:
### Description
This PR adds the API dropdown selector to the top of the auth docs page. Additionally the PR changes the contrast ratio of the dropdown selector text and removed the rounded radius for the new API selector at the top of the page.

This PR related to https://vajira.max.gov/browse/API-5360

- [x] Docs updated
  - [x] If you have made structural changes to the site, the [Developer Portal Content Types](https://community.max.gov/x/qUwOg) docs have been updated
  - [x] If you have made/confirmed any decisions about user interactions and accessibility, the [Accessible Component Design](https://community.max.gov/x/xS8mgg) docs have been updated
  

TS000 (approved):
All tests pass and the manual check looked great!

src/components/oauthDocs/BuildingOIDCContent.tsx
```diff
@@ -20,7 +20,7 @@ const BuildingOIDCContent = (): JSX.Element => {
   };
 
   return (
-    <section aria-labelledby="building-oidc-apps">
+    <section aria-labelledby="building-oidc-apps" id="building-oidc-apps">
```
mjlumetta, line 23: this id is also used by the section heading below. if we're using the section as the parent of all the code blocks, we need to choose a unique id here. we could also conceivably use a class, which is slightly preferable imo if we're using it only as a CSS selector (Scopes has a similar thing).

src/containers/documentation/AuthorizationDocs.scss
```diff
@@ -58,15 +58,17 @@
...
+  }
+  .api-selector {
+    background-color: $color-gray;
 
     select {
-      color: $color-white;
+      color: #80f8a2;
       option {
-          color: $color-base;
+          color: #80f8a2;
```
mjlumetta, line 71: this should remain `$color-base`, i believe. depends on your browser, but the option color can be what shows up when you open the select options. mine looks fine on Chrome, but here's what it looks like in Firefox:

Screen Shot 2021-02-26 at 9 42 19 AM

cc @ediiotero if you have any other explanation or context to add on the styling patterns here

mjlumetta (changes requested):
couple of small but important changes to suggest here. lmk if you have thoughts on those considerations

src/containers/documentation/AuthorizationDocs.scss
charleystran: This has been been fixed

src/components/oauthDocs/BuildingOIDCContent.tsx
charleystran: I changed to use a class instead.

mjlumetta (approved):

gileswells (approved):
LGTM

PR #648, opened 2021-02-25
Add checks in CI job to make sure the branch is not "master" before building review instance

mjlumetta:
### Description

on Tuesday 2/24/21, the dev and staging deploys broke. while testing [a potential CodeBuild vulnerability](https://sprocketfox.io/xssfox/2021/01/18/pipeline/uncovered), @duganth-va opened [a PR](https://github.com/department-of-veterans-affairs/developer-portal/pull/641) from a forked repository to the main repo. that PR caused builds of `department-of-veterans-affairs:master` to fail because the `prnum.sh` script identified the `duganth-va:master` PR as the `PR_NUM` for its build job. since `PR_NUM` was set, the CI job treated the build as a PR branch, built the app for a review instance including the wrong `PUBLIC_URL`, and deployed a review instance. having the build assets that were archived built for a review instance with the wrong `PUBLIC_URL` get released and deployed caused it to break in the environments it was deployed in (not prod, fortunately).

this PR does 3 things to mitigate the risk of this issue happening:

- adds a check before the build to make sure the current branch isn't the main branch before creating the `.env.{env}.local` files for review instances. this prevents changes on the main branch from building with the wrong `PUBLIC_URL`.
- adds a check before deploying to review instances to make sure the current branch isn't the main branch. this is less of an issue than archiving broken build assets, but it goes with the previous check and prevents unnecessary files from going in S3.
- adds an `exit 1` condition to the `prnum.sh` call so that if that script fails, the build fails. this catch applies to errors in `prnum.sh`, but it specifically helps prevent this error since Tom is adding a check to `prnum.sh` to make sure the PR is originating from `department-of-veterans-affairs/developer-portal` and not a forked repo.

### Process

- [n/a] ~Tests added or updated for change~
- [n/a] ~Docs updated~
  - [n/a] ~If you have made structural changes to the site, the [Developer Portal Content Types](https://community.max.gov/x/qUwOg) docs have been updated~
  - [n/a] ~If you have made/confirmed any decisions about user interactions and accessibility, the [Accessible Component Design](https://community.max.gov/x/xS8mgg) docs have been updated~
- [n/a] ~Accessibility concerns have been considered and addressed~
  
  - [n/a] ~If you've made any UI changes, you've run an axe check using the axe browser extension that reported no new issues~


cicd/buildspec-ci.yml
```diff
@@ -132,7 +133,8 @@ phases:
       # Check if we deploy to the review bucket
       - phase=deploy
       - |
-        if [[ ${PR_NUM} ]]; then
+        # check BRANCH as well as PR_NUM to ensure we don't treat the main branch like a PR branch
+        if [[ ${PR_NUM} && ${BRANCH} != "master" ]]; then
```
mjlumetta, line 137: this check is less important because the cost of deploying to a review instance unnecessarily is way lower than the cost of building + archiving bad assets. open to removing if people feel inclined to do so.

cicd/buildspec-ci.yml
```diff
@@ -80,10 +80,11 @@ phases:
       - COMMIT_HASH=${CODEBUILD_RESOLVED_SOURCE_VERSION:0:7}
       - git lfs pull
       # Resolve PR_NUM
-      - PR_NUM=$(prnum.sh ${REPO} ${BRANCH})
+      - PR_NUM=$(prnum.sh ${REPO} ${BRANCH}) || exit 1
```
mjlumetta, line 83: https://github.com/department-of-veterans-affairs/lighthouse-codebuild-containers/blob/main/resources/scripts/prnum.sh 

if checking for a PR fails, we want to fail the build job. this includes the failure where the originating repo of the PR isn't the main VA repo.

cicd/buildspec-ci.yml
```diff
@@ -80,10 +80,11 @@ phases:
...
       # Create the required dot files prior to build for proper deployment to review bucket
       - |
-        if [[ ${PR_NUM} ]]; then
+        # check BRANCH as well as PR_NUM to ensure we don't treat the main branch like a PR branch
+        if [[ ${PR_NUM} && ${BRANCH} != "master" ]]; then
```
gileswells, line 87: I don't know if there's a desire on the horizon to migrate the primary branch from `master` to `main` like GitHub is now defaulting. Maybe add an additional condition here that will handle that gracefully if we make that change in the future?

gileswells (approved):
LGTM overall. One question on master/main but shippable as is.

cicd/buildspec-ci.yml
mjlumetta: yeah, that's a good callout. it would be good to migrate over at some point, imo, but it's also not something that is necessarily imminent. i'd prefer not to future-proof in this case because this is already a safeguard for a pretty rare case.

one way to split the difference could be making "master" a constant at the top so we only have to change it in one place if we change the main branch, if you'd like to see that. let me know if you think that's a good idea and i can push another commit.

anthony-kopczyk: I like the idea of the variable. It might be good to call the variable something generic like `release_branch` or something.

cicd/buildspec-ci.yml
anthony-kopczyk: I'm not sure how all of the ci is set up. Will this prevent us from being able to run builds of master without a PR? I'm not sure if there's ever a scenario where we'd want to, but curious.

anthony-kopczyk (commented):
It looks good to me. Just a couple questions because I haven't dealt with the ci files much.

cicd/buildspec-ci.yml
duganth-va: If the PR script were to fail we should already get a failed build job. I still need to add the repo check to the PRNUM script with and `exit 1` on match. Currently if there is no PR it simply returns empty string with a 0 exit code. 

As general note on CodeBuild behavior. CodeBuild will honor the script exit code. The ` || exit 1` syntax is usually only need in multiline yaml scenarios. In these cases `errexit` isn't really thing.

cicd/buildspec-ci.yml
duganth-va: I would support a constant variable as well. 

As an aside we are hoping long term we remove this functionality from ci job anyway so that evaluating the branch would not be an issue in this case. Worth noting though renaming the `master` branch will trigger some needed configuration changes in regards to CodeBuild jobs when we start to isolate these triggers.

duganth-va (dismissed):
💯

cicd/buildspec-ci.yml
mjlumetta: ahh okay, thanks Tom. I was missing that nuance of single-line vs multiline commands. and Anthony, to your question, we do build master when we merge to it; the difference is that we don't build and deploy it as a review instance or try to comment on a PR (since there is none).

mjlumetta:
sorry for the delay in updating this PR, everyone. changes are made and pushed.

gileswells (approved):
LGTM. Nice change on the master/main branch thing.

PR #651, opened 2021-02-26
[API-5452] Upgrade swagger-ui to 3.44.0 and remove temporary CSS fix

ndsprinkle:
### Description
Swagger-ui accepted my accessibility changes, so we are upgrading again. This upgrade also lets us remove the temporary CSS fix in swagger.scss. After this upgrade, we should be good to start on the rest of the accessibility improvements.

### Original issue(s)
https://vajira.max.gov/browse/API-5275


mjlumetta (commented):
looks good to me on paper, but for some reason i'm still seeing that rogue link in the DOM. definitely installed 3.44.0, so i'm wondering if i'm having some kind of weird caching issue or something? so trying to figure out if that's real or not. if not, wondering why the axe tests didn't catch it 🤔

ndsprinkle:
@mjlumetta Are you talking about this `<div className="info__contact">` ? Or are you actually still seeing an empty `<a>` ?

ndsprinkle:
I still see the empty `info__contact` div being rendered, but it's empty inside. Before it had a blank link being rendered inside of class `info__extdocs`.

ndsprinkle:
Actually, ignore that, looking at the wrong link. The link that got fixed would've been under `<div className="info">` next to `<div className="description">`, but it is not there for me, as expected. Which API are you looking at?

anthony-kopczyk (dismissed):
This looks good to me. Going to remove the approval because it looks like visual regression is failing for some reason.

johanna-adhoc (dismissed):
LGTM

mjlumetta:
@ndsprinkle here's what I'm seeing in the DOM:

Screen Shot 2021-02-26 at 12 03 09 PM

I honestly have no idea why this is. if everybody else says it's rendering correctly, I'm good with the changes because they look right.

ndsprinkle:
The failing regression tests are confusing me because I get no failing regression tests locally and the ones that are failing in Jenkins seem to just be some sort of wrapping issue. @gileswells is going to pull my branch down and see if he can get the tests to fail locally. I pointed my .env.local file to all the production URLs and still didn't get failing tests, so I'm not sure what's going on.

anthony-kopczyk (approved):
This looks good to me. Might want to get together with Mike and figure out why his is showing the link still. Maybe there's some kind of issue still.

The failing regression test is odd that it randomly started moving the ">" character up. Maybe a new version of Chrome is the cause. 🤷

PR #655, opened 2021-03-01
API-5390: Fix postcss-loader blocked version in dependabot version config

johanna-adhoc:
### [API-5471: Fix postcss-loader blocked version in dependabot version config](https://vajira.max.gov/browse/API-5471)
Current dependabot configuration is set to ignore version updates from postcss-loader from 4.2.0 to 5.0.0. Correct versions are 3.0.0 to 3.0.0+.

Currently we are getting dependabot version alerts that are blocked.

### Process


anthony-kopczyk (approved):
Looks good!

PR #666, opened 2021-03-08
API-3611 - Test SwaggerUI redux related components (VersionActions, VersionReducers, and VersionSelector)

alexandervalencia:
### Description

This pr supports [API-3611](https://vajira.max.gov/browse/API-3611) by adding tests for the VersionActions, VersionSelector, and VersionReducers.

This is to follow up on an effort to increase the amount of coverage on critical components of the developer-portal site.

### Process

- [ ] Docs updated
  

anthony-kopczyk (commented):
I didn't realize we could test the redux functionality this way. That's cool!

It doesn't look like there's testing of the rendering of the components. I'd like some of that to verify the component has its 'key functions' and that they're accessible to the user. With the react testing library we can find the elements the way a user would. I know this is probably going to a massive add-on. Maybe we could split the ticket even.

anthony-kopczyk:
Adding on more. I do really like this method of testing the redux functions, separate from the code. Redux is essentially our frontend business logic for these components, so testing that logic by itself is going to make sure odd errors aren't happening.

alexandervalencia:
> It doesn't look like there's testing of the rendering of the components. I'd like some of that to verify the component has its 'key functions' and that they're accessible to the user. With the react testing library we can find the elements the way a user would. I know this is probably going to a massive add-on. Maybe we could split the ticket even.

I think this would essentially be a new test file for SwaggerDocs, as SwaggerUI is the main component consuming the redux logic. I think the only other component within the SwaggerPlugins using these would be VersionSelect. VersionSelect is currently mocking the store/functions. I think we can make a new ticket for testing SwaggerDocs to test the rendered implementation.

anthony-kopczyk:
> I think this would essentially be a new test file for SwaggerDocs, as SwaggerUI is the main component consuming the redux logic. I think the only other component within the SwaggerPlugins using these would be VersionSelect. VersionSelect is currently mocking the store/functions. I think we can make a new ticket for testing SwaggerDocs to test the rendered implementation.

Yeah. That makes sense to me - do an integration test for Swagger Docs to test key rendering pieces and test the business logic in the individual component tests.

TS000 (approved):
LGTM 🎸

PR #668, opened 2021-03-08
API-5474: Dependency Update Sprint 03/09

TS000:
### Description
Maintenance involving updating packages and documenting issues for the front end developer portal.
Jira Ticket [here](https://vajira.max.gov/browse/API-5474)

**24 packages updated**.

### Process

1. Run npm outdated command to list the libraries that need to be updated.
2. Address the dependatbot version PRs that are waiting in the queue. Use the list generated in previous step to identify if the PR could potentially introduce breaking changes (Version is listed as latest but not wanted). For PRs that are addressing mayor version updates perform regression testing in your local (unit, e2e, visual) and manually check the API key submission form. If all looks good proceed to approve and merge the PR.
3. Run npm update to update all the libraries that can be upgraded without introducing breaking changes.
4. Run npm outdated to check what libraries still need to be updated to the latest. Update to latest if possible.
5. Run npm audit fix.

### Blocks

Block for package 'node-fetch' that is used by Swagger,  [documented here](https://github.com/swagger-api/swagger-ui/issues/6437#issuecomment-790142395) this leaves us open to Denial of Service which is a low severity issue.

### Note

There is an issue with package 'uswds' introduced in version *2.0.0* that deals with png files missing. The current version *1.6.13* passed `npm outdated` as being the "wanted" version, however it is not the current version which is *2.10.2*. This leaves us open to Prototype Pollution which is a low severity issue.
### Requested feedback
All feedback welcome.


package.json
```diff
@@ -68,93 +68,94 @@
...
     "@types/react-router-hash-link": "^1.2.1",
     "@types/react-tabs": "^2.3.2",
     "@types/uuid": "^8.3.0",
     "@types/webpack": "^4.41.26",
-    "@typescript-eslint/eslint-plugin": "^4.15.1",
-    "@typescript-eslint/parser": "^4.15.2",
-    "enzyme-adapter-react-16": "^1.15.6",
+    "@typescript-eslint/eslint-plugin": "^4.17.0",
+    "@typescript-eslint/parser": "^4.17.0",
+    "@wojtekmaj/enzyme-adapter-react-17": "^0.4.1",
```
johanna-adhoc, line 117: we are currently running react 16 so I am not sure if the enzyme adapter got updated by mistake but I see many unit test failing in my local and I think this is the root cause.

johanna-adhoc:
This dependency update is for sprint 03/09. We will have another ticket to update on spring 03/21. Would be a good idea to change the name of the PR to avoid confusion

johanna-adhoc (approved):
LGTM

package.json
```diff
@@ -106,41 +106,42 @@
...
     "@types/react-router-dom": "^5.1.7",
     "@types/react-router-hash-link": "^1.2.1",
     "@types/react-tabs": "^2.3.2",
     "@types/uuid": "^8.3.0",
     "@types/webpack": "^4.41.26",
     "@typescript-eslint/eslint-plugin": "^4.15.1",
-    "@typescript-eslint/parser": "^4.15.2",
-    "enzyme-adapter-react-16": "^1.15.6",
+    "@typescript-eslint/parser": "^4.15.1",
+    "@wojtekmaj/enzyme-adapter-react-17": "^0.4.1",
```
anthony-kopczyk, line 117: We also have the enzyme adapter for react 16 down below on the dependencies list. Until we re-upgrade to react 17, I think we should take this dependency out.

package.json
```diff
@@ -24,18 +24,18 @@
   },
```
anthony-kopczyk, line 24: It won't let me comment up further.

With the latest react-dev-utils, it's immer version is now set to 8.0.1, so we can take out the "resolutions" in this package.json, since all it's doing is setting immer to 8.0.1.

anthony-kopczyk (changes requested):
This looks good, except with react-dev-utils being upgraded we can remove the resolution for immer to be set to 8.0.1. It looks like they bumped another patch of react-dev-utils this last weekend. There might be other packages that have been updated over the weekend too.

package.json
johanna-adhoc: I agree with this. If the immer version that is installed by default is 8.0.1 then we should remove the resolutions

package.json
johanna-adhoc: This is true too. We can add this adapter back if the story to bump to react 17 is possible to implement, in the meantime we should remove unused dependencies.

TS000: enzyme adapter for React 17 has been removed! Thanks Anthony!

package.json
TS000: Immer resolution has been removed! Thanks you two!

johanna-adhoc (approved):

anthony-kopczyk (approved):
Looks good to go!

PR #670, opened 2021-03-08
Review Deploy Buildspec

duganth-va:
### Description
This PR will isolate pull request review deployments. It will paired with a new Codebuild job https://github.com/department-of-veterans-affairs/devops/pull/8638.

- It includes verifies the PR is coming from the main repository and then retrieves the PR number for comment with the gh cli tool.
-  It can be run manually without a PR as the PR scripts will exit 0 if a PR is not found. 
- It removes the branch name as that is only sent with the webhook and wasn't reliable for manual builds and retry builds. 

Note: This does not yet include the removal of the PR functionality on the CI job. 

### Requested feedback
Request feedback from developers on the new addition.


anthony-kopczyk (commented):
I'm not sure if I like calling it review deployments - thought I can see how they're deployed on S3 since they're static web files. 

While this is an independent chunk of work I'm wondering what the end result is when the work is all done.

Does this sound right?
- We have a CI pipeline that runs a build and checks test
- We have this pipeline (review deploy) that builds the app for PRs and puts in the S3 buckets but does not tag the version
- We have a release pipeline that triggers on PR merges to master, bumps version, and tags.
- We have the deploy pipeline that deploys a tagged docker image/or version to the app to production

My only question is, will the release pipeline still do it's own build when it tags or will it be borrowing the build from the PR review. I'd think it's own so that we make sure everything from master is there, but we enforce PRs to be up to date before merging. To get the version/tag into the React version the release pipeline would need to do its own build with the new tag.

duganth-va:
> I'm not sure if I like calling it review deployments - thought I can see how they're deployed on S3 since they're static web files.

I'm ok with changing the name to whatever feels intuitive for the dev team. 😄  I'll give the rationale behind the name.

This buildspec file's purpose is to deploy review sites of a particular commit hash to the review bucket. This requirement is carried over from the Jenkins pipeline.  It mimics the regular deployment job but there a few differences:

* This file has to create the dot files for the build step to consume.  
* This file builds the app prior to deploy.  

As and alternate approach we could theoretically look to combine these deployment mechanisms into one deployment file. This file would be referenced from at least two different CodeBuild projects. The different projects would be responsible for isolating the IAM permissions. This inherently creates a new set of problems. 

The goal of this file is to separate the deployments of the review sites from CI to solve a complexity problem and a security problem. Hence why I settled on the `review-deployments` name. I have no opinion on the name other than that it makes sense the dev teams. Hopefully that context helps!

duganth-va:
> While this is an independent chunk of work I'm wondering what the end result is when the work is all done.
> 
> Does this sound right?
> 
> We have a CI pipeline that runs a build and checks test
> We have this pipeline (review deploy) that builds the app for PRs and puts in the S3 buckets but does not tag the version
> We have a release pipeline that triggers on PR merges to master, bumps version, and tags.
> We have the deploy pipeline that deploys a tagged docker image/or version to the app to production

Sounds about right would make the following modifications:
* I would add one more step to the release bullet. Right now, I _think_ the release job will build the release artifacts and upload them the archive bucket. This would remove that functionality from CI.  This also removeslogic in CI revolving around the `master` PR. That would be a separate PR. 
* Deploy pipeline would not deploy a docker image. It would (and does now) deploy the static website files to the proper bucket. The change in the future would revolve around what triggers the auto deploys to `dev` and `staging`.

duganth-va:
> My only question is, will the release pipeline still do it's own build when it tags or will it be borrowing the build from the PR review. I'd think it's own so that we make sure everything from master is there, but we enforce PRs to be up to date before merging. To get the version/tag into the React version the release pipeline would need to do its own build with the new tag.

Yes thats the intent and this would happen only after CI Passes. We need the `gh-status.sh` script for that to work properly.

anthony-kopczyk:
I'm good with the name review-deployment. I can't think of another name to call it.

mjlumetta (commented):
overall looks good to me. main thought here would be to abort this build job (without erroring) earlier if there is no PR so that we're not using up S3 resources we don't strictly need. I can't think of a case where that'd really be necessary, although open to feedback from other devs on that.

cicd/buildspec-review-deploy.yml
```diff
@@ -0,0 +1,90 @@
...
+      # Make a url compliant name for S3
+      # Short Ref to commit hash
+      - COMMIT_HASH=${CODEBUILD_RESOLVED_SOURCE_VERSION:0:7}
+      # Verify PR is coming from source repository
+      - pr-verify.sh -r ${REPO} -c ${COMMIT_HASH}
+      # Get PR Num
+      - PRNUM=$(pr-number.sh -r ${REPO}  -c ${COMMIT_HASH})
+      # Log all environment varbiables to cloudwatch in case we need to troubleshoot. AWS is nice enough to hide any secrets from parameter-store.
+      - printenv
+      - git lfs pull
```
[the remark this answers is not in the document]
duganth-va: ah ok cool will remove!

duganth-va:
> main thought here would be to abort this build job (without erroring) earlier if there is no PR so that we're not using up S3 resources we don't strictly need.

I was going to ask if it would be ok if a put a 30 day lifecyle policy on the s3 bucket. Thoughts?

cicd/buildspec-review-deploy.yml
```diff
@@ -0,0 +1,90 @@
...
+          make build_app ENVIRONMENT=${env} || exit 1
+          links="${links} - [${env}](https://s3-us-gov-west-1.amazonaws.com/${s3_review_bucket}/${s3_branch}/${commit_hash}/${env}/index.html) <br>"
+        done
+      # Check CI status of other checks
+      - gh-status.sh -r ${REPO} -c ${COMMIT_HASH} -x ${XCHECKS}
+      # Deploy To s3 review bucket
+      - aws s3 sync --no-progress --acl public-read build/ s3://${S3_REVIEW_BUCKET}/${S3_BRANCH}/${COMMIT_HASH}/
+      # If Review deploy is assocaited with PR, comment on PR.
+      - |
+        if [[ ! -z ${PRNUM} ]]; then
```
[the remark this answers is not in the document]
mjlumetta: I doubt that it's necessary to deploy a review instance prior to a PR opening up. the primary purpose of these deployments is so that PR reviewers can test out the site without having to pull the branch and run the site locally, particularly non-engineers.

I would say this job shouldn't fail if there's no PR, but it doesn't need to deploy the assets without one, either. so if it can exit early and succeed, that would be ideal. but, that said, it's not a huge deal if there are keys in the S3 buckets that we're not using if there's a good retention policy.

mjlumetta:
> I was going to ask if it would be ok if a put a 30 day lifecyle policy on the s3 bucket.

yeah, I'd say that is very safe. there are very few instances where we'd need a single review instance for even a week, so 30 days is absolutely fine.

cicd/buildspec-review-deploy.yml
```diff
@@ -0,0 +1,90 @@
...
+        done
+      - links=""
+        # Run build and generate links per environment
+      - |
+        for env in ${ENVIRONMENTS}; do
+          make build_app ENVIRONMENT=${env} || exit 1
+          links="${links} - [${env}](https://s3-us-gov-west-1.amazonaws.com/${s3_review_bucket}/${s3_branch}/${commit_hash}/${env}/index.html) <br>"
+        done
+      # Check CI status of other checks
+      - gh-status.sh -r ${REPO} -c ${COMMIT_HASH} -x ${XCHECKS}
```
mjlumetta, line 82: thinking about this a bit further, i'm not 100% sure we even need this check. would be curious to get other devs' feedback on this. 

but, to explain why i think we can take it out, there's no real reason all the CI checks need to pass in order to get the review instance. for example, if a unit test fails, that doesn't mean that it would be inappropriate to have a deployed instance of the portal for reviewers to check out. and in fact, it could be beneficial to make them independent, because that means that devs don't need to get all the checks passing in order for a reviewer to test with the review instance. the only dependency is on the build for the site completing successfully, and that's already in this job anyway.

cicd/buildspec-review-deploy.yml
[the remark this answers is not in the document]
duganth-va: No worries having it fail early works for me!

cicd/buildspec-review-deploy.yml
duganth-va: 💯  that makes sense! 💡  I'll wait for feedback prior to edit.

anthony-kopczyk: Mike's logic makes sense to me. The only negative I can think of is a build with failed CI takes up more space in S3, but I don't believe that's a blocker for us - especially if we add a 30 day lifetime on S3 for the review instances.

duganth-va:
I ran into an issue while running this CodeBuild job that was related to the build env specs. I had set it to SMALL which has 3 GB memory, 2 vCPUs. It looks like `fork-ts-checker-webpack-plugin` requires at least 4 GB mem. When I upgraded the build environment to MEDIUM it worked (minus a permissions issue).  Just an FYI for the team.

cicd/buildspec-review-deploy.yml
```diff
@@ -0,0 +1,88 @@
...
+      # Short Ref to commit hash
+      - COMMIT_HASH=${CODEBUILD_RESOLVED_SOURCE_VERSION:0:7}
+      # Verify PR is coming from source repository
+      - pr-verify.sh -r ${REPO} -c ${COMMIT_HASH}
+      # Get PR Num
+      - PRNUM=$(pr-number.sh -r ${REPO}  -c ${COMMIT_HASH})
+      - |
+        if [[ -z ${PRNUM} ]]; then
+          echo "No PR Found aborting job"
+          exit 1
```
mjlumetta, line 60: just `exit`?

cicd/buildspec-ci.yml
```diff
@@ -176,20 +142,15 @@ phases:
...
-          slackpost.sh "Pipeline failure for the developer portal" 
+          slackpost.sh "Pipeline failure for the developer portal"
           ;;
         esac
       # Output build times
       - printf '%b\n' "${time}"
-      # Handle comments.
-      - |
-        if [[ ${PR_NUM} ]]; then
-          prcomment.sh ${REPO} ${PR_NUM} "${comment:-"CI Job has finished successfully."}"
```
mjlumetta, line 194: do we still want a success notification here if there's a PR? might be useful

mjlumetta (commented):
couple of comments, nothing major, i'm pretty okay with this version as-is.

cicd/buildspec-review-deploy.yml
duganth-va: CodeBuild will abort on non zero exit codes hence the 1.

cicd/buildspec-ci.yml
duganth-va: The status check would be in GitHub not sure if you'd want a pr comment as well? Let me know if that would be helpful.

duganth-va:
> These changes have been deployed to an S3 bucket. A build for each environment is available:
> 
> - [dev](https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/td-API-5504-PR-Isolation/53e95d1/dev/index.html)
> - [staging](https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/td-API-5504-PR-Isolation/53e95d1/staging/index.html)
> - [production](https://s3-us-gov-west-1.amazonaws.com/review-developer-va-gov/td-API-5504-PR-Isolation/53e95d1/production/index.html)
> 
> Due to S3 website hosting limitations in govcloud you need to first navigate to index.html explicitly.

This is deployed with this file after compute type changes

cicd/buildspec-review-deploy.yml
mjlumetta: will that mark the job as failed? didn't find guidance for that in the docs, but i didn't go super deep on that.

cicd/buildspec-ci.yml
mjlumetta: eh, doesn't matter that much? it could be, but i am fine not having it. will leave it up to others

mjlumetta (approved):
this is good from my end. i have one question about the exit for the no-PR case, but that's not blocking.

PR #673, opened 2021-03-10
Update auth docs copy, e2e helper, and visual regression test

ediiotero:
### Description

Update auth docs copy, e2e helper, and visual regression test


mjlumetta (approved):
thanks Eddie, and sorry again for not catching the original PR. one small cleanup suggestion: can delete the old, now-unused snapshots too, although that's not absolutely necessary to commit these changes. (think that you'd have to do that manually, `jest-image-snapshot` isn't smart enough to do that by itself.)

mjlumetta (dismissed):

gileswells:
@ediiotero I completely didn't see that you had the changes for the visual regressions in a different PR. While I don't have the text change in my PR I do have the urls and updated screenshots in #679. I suspect when that merges on Monday your branch might end up with a fair number of conflicts.

gileswells:
@ediiotero #679 is now merged so you have some conflicts on this and some redundant changes I think.

src/e2eHelpers.ts
```diff
@@ -16,6 +16,7 @@ export const testPaths = [
   '/explore/authorization',
   '/explore/health',
   '/explore/benefits',
+  '/explore/authorization?api=claims',
```
gileswells, line 19: Is this route needed? I had added `/explore/authorization` a few lines above in a different PR.

gileswells (changes requested):
I think we might have inadvertently doubled up on `/explore/authorization` tests

src/e2eHelpers.ts
ediiotero: yeah you're right, removed it

gileswells (approved):

PR #674, opened 2021-03-11
API-5513 - CurlForm parameter inputs 

ediiotero:
Story: [API-5513](https://vajira.max.gov/browse/API-5513#main)

### Description
Duplicate input ids in the curlforms preventing assistive technologies from accurately associating the inputs with labels.

- added aria-label to curlForm input
- tested with NVDA and Narrator
- passes aXe check


mjlumetta (approved):
lgtm! ✨ 

think you noted this elsewhere (in the ticket?), but VoiceOver, like NVDA, was able to associate the labels correctly before, i think. that said, good to make it work correctly anyway so there's no conflict.

PR #678, opened 2021-03-15
API-5390: Include Build Version In the Developer Portal UI

TS000:
Jira [ticket here](https://vajira.max.gov/browse/API-5390)
### Description
Quickly check the code version that is currently deployed in PROD so that we can easily track the changes that are part of this version plus get the version build to fill out MR request for future deployments.

### Process
1. In order to pass the CodeBuild hash to the production ready Developer Portal build process required including the CodeBuild hash in the React build process within `cicd/buildpsec-ci.yml`. 
2. The environment variable was then picked up by the React build process (React requires all custom environment variables to follow the naming style of `REACT_APP_(your cool var name)`). 
3. A component then picks up the environment variable and hides it in the DOM. Currently the component us being used in `components/Banner`.
4. The hash is placed in the `Makefile` in order to pass it to the Docker build for production.

- [ x] Tests added or updated for change
- [ ] Docs updated
- [x ] Accessibility concerns have been considered and addressed
  
### Requested feedback
I apologies for the amount of builds done in this PR but in order to test the other side of the build processes it had to be done.

Feedback welcome


src/components/appVersion/AppVersion.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import './AppVersion.scss';
+
+const currentAppVersion = process.env.REACT_APP_COMMIT_HASH ?? 'undefined';
+
+const AppVersion = (): JSX.Element => (
+  <p className="app-version">{currentAppVersion}</p>
```
johanna-adhoc, line 7: I would rename the class app-commit so when the build version gets integrated there is no confusion

package.json
```diff
@@ -17,7 +17,8 @@
     "test:accessibility": "node scripts/test.js --projects=e2e.jest.config.js --testMatch='<rootDir>/src/**/?(*.)(accessibility).(j|t)s?(x)' --runInBand",
     "test:accessibility:ci": "node scripts/test.js --projects=e2e.jest.config.js --testMatch='<rootDir>/src/**/?(*.)(accessibility).(j|t)s?(x)' --ci --testResultsProcessor='./node_modules/jest-junit-reporter' --forceExit --runInBand",
     "test:visual": "node scripts/test.js --projects=visual.jest.config.js --runInBand --ci",
-    "test:coverage": "node scripts/test.js --coverage "
+    "test:coverage": "node scripts/test.js --coverage ",
+    "git-info": "git log -1 --oneline > src/components/appVersion/appInfo.txt"
```
johanna-adhoc, line 21: I think the command git-info should not be part of the scripts

cicd/buildspec-ci.yml
```diff
@@ -91,6 +91,7 @@ phases:
         cat <<EOF > .env.${env}.local
         PUBLIC_URL=/${S3_REVIEW_BUCKET}/${S3_BRANCH}/${COMMIT_HASH}/${env}
         REACT_APP_SENTRY_DSN=${PREVIEW_SENTRY_DSN}
+        REACT_APP_COMMIT_HASH=${CODEBUILD_RESOLVED_SOURCE_VERSION}
```
johanna-adhoc, line 94: Should we only add the first 6 characters? I have seen that as a practice

johanna-adhoc (changes requested):

package.json
TS000: oops that was left in there from a test i was doing, ill remove. Thanks!

cicd/buildspec-ci.yml
TS000: I'm not aware of any best method to serve the codebuild hash but if you say 6 char i can make that happen.

src/components/appVersion/AppVersion.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+describe('AppVersion', () => {
+  let wrapper: ReactWrapper<unknown, unknown, React.Component>;
+  beforeEach(() => {
+    wrapper = mount(<AppVersion />);
+  });
+
+  it('should render the hash or undefined', () => {
+    expect(wrapper.find('.app-commit').length).toBe(1);
+  });
+});
```
anthony-kopczyk, line 17: I think we're trying to move away from enzyme. The react testing library should work here.

anthony-kopczyk (commented):
Functionally everything looks good. I just have a couple questions/suggestions.

cicd/buildspec-ci.yml
```diff
@@ -91,6 +91,7 @@ phases:
         cat <<EOF > .env.${env}.local
         PUBLIC_URL=/${S3_REVIEW_BUCKET}/${S3_BRANCH}/${COMMIT_HASH}/${env}
         REACT_APP_SENTRY_DSN=${PREVIEW_SENTRY_DSN}
+        REACT_APP_COMMIT_HASH=${CODEBUILD_RESOLVED_SOURCE_VERSION:0:6}
```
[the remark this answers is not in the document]
TS000: React requires REACT_APP in order for variables to be passed to the available ENV variables. All variables without REACT_APP are ignored in the build process. [See docs here](https://create-react-app.dev/docs/adding-custom-environment-variables/).

Makefile
```diff
@@ -15,6 +15,8 @@ BRANCH ?= notmaster
 ENVIRONMENT ?= dev
 # Sets image name - useful locally if you want an image with a unique name
 IMAGE_NAME ?= developer-portal
+# Sets current application hash - used to see version from the front end
+REACT_APP_COMMIT_HASH ?= REACT_APP_COMMIT_HASH
```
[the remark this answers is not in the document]
TS000: This was a left over from the myriad of things I tried 😅

src/components/appVersion/AppVersion.test.tsx
TS000: Moved over to testing-library 👍

src/components/banner/Banner.tsx
```diff
@@ -81,6 +82,7 @@ const Banner = (): JSX.Element => {
                   aria-hidden="true"
                 />
               </div>
+              <AppVersion />
```
johanna-adhoc, line 85: Can you move the AppVersion Element to the root of the footer?

johanna-adhoc (changes requested):

cicd/buildspec-ci.yml
```diff
@@ -119,6 +119,7 @@ phases:
       - phase=build_app
       - |
         for env in ${ENVIRONMENTS}; do
+          REACT_APP_COMMIT_HASH=${CODEBUILD_RESOLVED_SOURCE_VERSION:0:7}
           time="${time}\n${phase}-${env} - $(date +%r) - started"
           make build_app ENVIRONMENT=${env} || exit 1
         done
```
anthony-kopczyk, line 125: In the Makefile, REACT_APP_COMMIT_HASH is being set to COMMIT_HASH.
This makes me think that the REACT_APP_COMMIT_HASH in buildspec-ci.yml isn't actually being used, since Makefile is using COMMIT_HASH.

.env.test
```diff
@@ -3,3 +3,4 @@ REACT_APP_VETSGOV_SECONDARY_SWAGGER_API=https://api.va.gov
 REACT_APP_ARGONAUT_API_ENABLED=false
 REACT_APP_DEVELOPER_PORTAL_SELF_SERVICE_URL=http://fake.va.gov
 REACT_APP_AUTH_DOCS_V2=true
+REACT_APP_COMMIT_HASH=123456
```
anthony-kopczyk, line 6: Could add a new line to this if you make another commit.

Makefile
```diff
@@ -105,4 +105,5 @@ build_app:
 		--volume "/application/node_modules" \
 		--env NODE_ENV=production \
 		--env BUILD_ENV=${ENVIRONMENT} \
+		--env REACT_APP_COMMIT_HASH=${COMMIT_HASH} \
```
anthony-kopczyk, line 108: _Just me rambling below_

This is more in line with how I was thinking of passing the environment variable. It looks like maybe it's immediately accessible to the Makefile. Thought it would need passed in, but I guess it makes sense. Oooh! I just saw the for loop in buildspec-ci.yml. I see now. Yeah.

Okay, so I was originally think (in the spike ticket), we needed to (in the buildspec-ci.yml) do `make build_app ENVIRONMENT=${env} COMMIT_HASH=${COMMIT_HASH}`.. but since the make command happens on the same machine, that environment variable is already defined. The only reason ENVIRONMENT is being set is because of the for loop.

🤯 

Well, that makes this a little less complicated. Was thinking we had to pass it through 4 or so places.

src/components/appVersion/AppVersion.scss
```diff
@@ -0,0 +1,3 @@
+#app-commit {
+    display: none;
+}
```
anthony-kopczyk, line 3: If you make another commit, can add a new line on the end of this file.

anthony-kopczyk (commented):
One question about maybe a line not being needed - I'm not sure.

And then there are two files that don't end in new lines. If there's a new commit, you can throw those in with it.

.env.test
anthony-kopczyk: Seeing it's not clear in the comments of the review. This file doesn't end in a new line.

src/components/appVersion/AppVersion.scss
anthony-kopczyk: Seeing it's not clear in the comments of the review. This file doesn't end in a new line.

cicd/buildspec-ci.yml
johanna-adhoc: that is a good point... if we are not using REACT_APP_COMMIT_HASH we should remove the declaration.

johanna-adhoc (changes requested):

src/components/banner/Banner.tsx
mjlumetta: yeah, i'd second this suggestion. this location feels kind of arbitrary, and i'd prefer it to be closer to the top level so it's easily accessible.

mjlumetta (dismissed):
changes look good to me overall, with the caveat of the comment from Johanna that i seconded. i've got a couple additional questions that maybe can't be resolved right now but are worth considering.

@duganth-va is refactoring the dev portal pipeline based on the concerns raised in [this thread](https://lighthouseva.slack.com/archives/CAZF880HW/p1614290399049100). the short version as relates to this PR is that the build in the CI job is not going to be used for either review instances or releases (and therefore not the deployed site). PR deploys are going to be part of the new job in #670, and the release job is going to be refactored later. i _think_ it'll probably be fine in both, but we need to make sure that `COMMIT_HASH` variable doesn't get lost in the process in a way that's going to break this functionality.

also, i think it'd be cool to have the tagged version in here too for the release. that won't be that relevant until the build for the release happens in the release job itself, but might be useful to think about now.

Makefile
TS000: This is a grey area and the only way to test is to remove it, build it and see if the variable is there. Just pushed a commit to test this out.

cicd/buildspec-ci.yml
TS000: Just pushed a commit to see if we need REACT_APP_COMMIT_HASH in the build loop or if having in the Docker build process within Makefile is enough. All i know for sure is REACT_APP_COMMIT_HASH in the build loop works.

johanna-adhoc (approved):
LGTM

src/components/banner/Banner.tsx
TS000: Moved to Footer. Thanks you two!

.env.test
TS000: done, thanks!

anthony-kopczyk:
> also, i think it'd be cool to have the tagged version in here too for the release. that won't be that relevant until the build for the release happens in the release job itself, but might be useful to think about now.

This was the intended functionality but after looking through the pipelines we realized the best we could do in the current state is the commit hash. This is absolutely on the table once the pipelines are updated. I thought we made a jira ticket for this, but I guess we didn't. Put in https://vajira.max.gov/browse/API-5877 for it.

src/components/banner/Banner.tsx
```diff
@@ -103,7 +106,7 @@ const Banner = (): JSX.Element => {
                     'small-desktop-screen:vads-u-margin-left--1',
                     'small-desktop-screen:vads-u-margin-top--0',
                   )}
-                  aria-label="Here&apos;s how you know this is an official website"
+                  aria-label="Here's how you know this is an official website"
```
anthony-kopczyk, line 109: I've seen this in another few files. Not sure if other teams will be on board changing this over. Might want to check with someone on Okapi.

anthony-kopczyk (commented):
It all looks good to me. There's one thing I think we should check up on, unrelated to the build version. But other than that it looks all good to go!

src/components/banner/Banner.tsx
TS000: That came from running a format with the lint file. I talked to Quokka and they said to go with what the linter does. I also checked the place where it's used with Wave and it was fine.

anthony-kopczyk (approved):
Looks good to go

PR #680, opened 2021-03-16
Add copy to category and API documentation pages to redirect veterans to VA.gov

mjlumetta:
### Description

makes copy changes to redirect veterans to VA.gov for [API-5736](https://vajira.max.gov/browse/API-5736).

### Process

- [ ] ~Tests added or updated for change~ n/a
- [ ] ~Docs updated~ n/a
  - [ ] ~If you have made structural changes to the site, the [Developer Portal Content Types](https://community.max.gov/x/qUwOg) docs have been updated~ n/a
  - [ ] ~If you have made/confirmed any decisions about user interactions and accessibility, the [Accessible Component Design](https://community.max.gov/x/xS8mgg) docs have been updated~ n/a
- [X] Accessibility concerns have been considered and addressed
  
  - [X] If you've made any UI changes, you've run an axe check using the axe browser extension that reported no new issues

### Requested feedback

the ticket didn't specify too much about the location of this copy, so I used the existing structure of the portal to put them in the slots offered by the site structure (see the Developer Portal Content Types doc linked in the Process section). is there any more UX we should add, or should they go in different locations?

PR #682, opened 2021-03-17
Add content to redirect veterans to VA.gov

mjlumetta:
### Description

adds content to redirect veterans to VA.gov if they are looking for info on benefits, facilities, or forms. completes [API-5736](https://vajira.max.gov/browse/API-5736). alternate approach to #680; this PR adds the info box styles and uses a generic solution for APIs and categories.

### Process

- [ ] Docs updated
  
  - [X] If you've made any UI changes, you've run an axe check using the axe browser extension that reported no new issues


darduda:
Looks good to me!

src/apiDefs/query.test.ts
```diff
@@ -36,6 +36,11 @@ describe('query module', () => {
         trustedPartnerOnly: false,
         urlFragment: 'facilities',
         vaInternalOnly: false,
+        veteranRedirect: {
+          linkText: "Find the facility that's right for you",
+          linkUrl: 'https://www.va.gov/find-locations/',
+          message: 'Are you a Veteran? ',
```
gileswells, line 42: Personally I'd have preferred to enforce the trailing space on the message field in the display portion. I feel like taking out an accidental trailing space with a quote is something that's not an "if" it happens but more of a "when" it happens. Someone thinking they're helping (I'm looking squarely at myself here) might remove it and then miss where it impacts the site. 

I can be ok with it as is though if you have a differing opinion on it.

src/containers/documentation/ApiPage.tsx
```diff
@@ -70,6 +71,18 @@ const ApiPage = (): JSX.Element => {
...
+        <AlertBox
+          status="info"
+          key={api.urlFragment}
+          className={classNames('vads-u-margin-bottom--2', 'vads-u-padding-y--1')}
+        >
+          {api.veteranRedirect.message}
+          <a href={api.veteranRedirect.linkUrl}>
+            {api.veteranRedirect.linkText}
+          </a>
+          .
```
gileswells, line 84: I would prefer, if we make future alterations to this, that there to be a field in the config for a `linkSuffix` rather than hard code the punctuation here. Default it to a `.` in the template if the property is missing works for me but not having a way to change this out seems less than ideal to me.

gileswells (approved):
I have some thoughts but nothing that blocks this from merging to me.

src/containers/documentation/ApiPage.tsx
mjlumetta: yeah... i don't necessarily disagree, because i can easily imagine some more complicated text like "<link>. You may also want to contact your VSO, blah blah blah," whatever. on the other hand, do you feel like this conflicts with your other recommendation to move the trailing space into the display part? feels a little weird to be moving one thing from there to here and another from here to there. i'm okay with whichever implementation, though, because this is small enough that we don't need a perfect abstraction or anything.

gileswells (approved):
LGTM

PR #689, opened 2021-03-24
API-5944 - Fix table css on pages with Swagger-UI documentation

gileswells:
### Description

https://vajira.max.gov/browse/API-5944

Pages with html tables that are generated through their swagger docs were having their css collide with the tables within each endpoint's details accordion.

### Process

Confirm that the tables on these pages are displaying properly while those on the [auth docs](https://developer.va.gov/explore/authorization?api=claims) aren't affected.

https://developer.va.gov/explore/appeals/docs/decision_reviews?version=current
https://developer.va.gov/explore/benefits/docs/claims?version=current

- [ ] Docs updated
  
### Requested feedback

There shouldn't be any visual regressions failures because these 2 urls are not included in our limited testing set of swagger based pages. Do we want to add one of these to make sure they don't get adversely affected in a future push?

### DIFF
--- a/src/containers/documentation/swaggerPlugins/StyleOverride.scss
+++ b/src/containers/documentation/swaggerPlugins/StyleOverride.scss
@@ -1,3 +1,6 @@
+@import '../../../styles/variables';
+@import '~@department-of-veterans-affairs/formation/sass/full';
+
 @mixin table-style {
   padding: 0;
   margin: 0;
@@ -7,12 +10,12 @@
   tr {
     td {
       padding: 1rem;
-      background-color: #EEEEEE;
+      background-color: #eeeeee;
       border: 1px solid #5b616b;
       border-top-style: none;
       border-bottom-style: none;
     }
-    &:last-child{
+    &:last-child {
       td {
         border-bottom: 1px solid #5b616b;
       }
@@ -20,7 +23,8 @@
   }
 }
 
-.opblock-section, .responses-wrapper {
+.opblock-section,
+.responses-wrapper {
   margin: 2em 2em 0;
   padding-bottom: 1em;
 }
@@ -33,7 +37,8 @@
   display: none;
 }
 
-.table-container, .responses-inner {
+.table-container,
+.responses-inner {
   padding: 1em 0 2em !important;
 }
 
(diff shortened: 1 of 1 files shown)


ndsprinkle (approved):

src/containers/documentation/swaggerPlugins/StyleOverride.scss
```diff
@@ -43,30 +48,55 @@
...
+    thead tr {
+      th {
+        color: $color-gray-dark;
+        font-family: $font-sans;
+        font-size: $base-font-size;
+        font-weight: $font-bold;
+      }
+
+      th,
+      td {
```
mjlumetta, line 93: are there any `td`s in the `thead`? wouldn't expect there to be

src/containers/documentation/swaggerPlugins/StyleOverride.scss
```diff
@@ -43,30 +48,55 @@
...
+        background-color: $color-gray-lightest;
```
mjlumetta, line 94: in the browser tools, i'm seeing this and the `font-weight` from line 89 repeated twice. is that from the USWDS stylesheet?

mjlumetta (approved):
looks good, have a couple of questions but no blockers

src/containers/documentation/swaggerPlugins/StyleOverride.scss
gileswells: I didn't see any instances of this however I did this to keep 1:1 pairity with the uswds package in case anyone chose to use it in the future so there wouldn't need to be a second patch down the road.

src/containers/documentation/swaggerPlugins/StyleOverride.scss
gileswells: Yes. The `@import` a few lines up imports the whole of that file into the `.swagger-ui .renderedMarkdown` css namespace so it does bring that in a second time. Unfortunately side effect but it makes the code such that there's less duplication in the styles override file. 

I did have a small update to make due to how `th, td {` worked with the `@import` so that `td` is now namespaced under `.swagger-ui .renderedMarkdown thead tr` as it should be.

ndsprinkle (approved):

mjlumetta (approved):
works for me. as a general rule, the only thing i'd suggest here is keeping an eye on bundle sizes, but looking good.

PR #705, opened 2021-04-07
[API-3130] Reroute users from invalid pages under /explore and /release-notes to the 404 page

ndsprinkle:
### Description
Currently, when attempting to access a fake category under /explore or /release-notes, the user is unexpectedly rerouted to the root of /explore or /release-notes without any indication that the page they were trying to visit is invalid. This change reroutes users to the 404 page instead, which should make it more clear to the user that they tried to hit an unavailable page. 

This reroute is being handled by the default route that handles any non-defined route. In this situation, I am sending users to /404, which is not a defined route, so therefore it renders the 404 page. If this /404 route needs to be hardcoded to render the 404 page, then I can make that change.


src/containers/releaseNotes/CategoryReleaseNotes.test.tsx
```diff
@@ -158,9 +159,10 @@ describe('ReleaseNotesCollection', () => {
               render={(): JSX.Element => <div>/release-notes</div>}
             />
             <Route path="/release-notes/fakeCategory" exact component={CategoryReleaseNotes} />
+            <Route component={NotFound} />
           </Router>,
         );
-        expect(container.innerHTML).toEqual(expect.stringContaining('/release-notes'));
+        expect(container.innerHTML).toEqual(expect.stringContaining('404'));
```
gileswells, line 165: `404` or `/404`? It doesn't look like it should cause any issues but just for uniformity sake.

gileswells (changes requested):
Also I found a missing 404 wildcard under `/support/banana`. If you want to tackle that in this PR or via another ticket I'm ok with either. I'm hoping the fix for that particular issue would be to fix the routes file so we get proper 404's all the way down without redirects.

src/containers/releaseNotes/CategoryReleaseNotes.test.tsx
ndsprinkle: I don't know the reason, but `/404` doesn't appear in the `innerHTML`. Actually, I'm more confused as to why `/release-notes` appears in it originally. The only reason `404` appears is because there is actual text that says "404" on that page.

ndsprinkle: Just let me know if this is good or not, I can change it to check something else if checking another string makes more sense, but I wasn't sure what.

ndsprinkle: Oh and I guess my answer is that we can try to handle the 404 wildcard you mentioned in another PR since I theoretically am moving on to Apigee things

gileswells (approved):
LGTM with the above answers

PR #706, opened 2021-04-07
API-6080 Error Boundary Page

gileswells:
### Description

https://vajira.max.gov/browse/API-6080

This just setups the bulk of the implementation code for an error catching page template so that if an error were to happen on production they wouldn't just get a blank white page. This is not the final design (hence why it's not a PR to master) but is a functional prototype that can get the design layer added to it once https://vajira.max.gov/browse/API-2751 is completed.

### Process

Visit the 404 page and you'll see what looks like the 404 page but it is a different template and the error presented is taken directly from the javascript error that's thrown from `< DummyErrorThrower />`.

- [ ] Docs updated
  
### Requested feedback


alexandervalencia (approved):
It looks good to me. My only thought is that, if intention of the error body is to replicate the navigation, it's missing the API Publishing links.

Up to you if you want to add them, otherwise, approved. 🚀

gileswells:
> It looks good to me. My only thought is that, if intention of the error body is to replicate the navigation, it's missing the API Publishing links.

Overall this is a dupe of the 404 page's template and it might be a good idea to offer those up here as well. There's some tidy up of links that can be done and also a very real potential to merge the base of these page templates depending on how design chooses to make the error page look.

PR #712, opened 2021-04-20
API-5574: upgrade react to 17

anthony-kopczyk:
### Description

This upgrades the version of react to 17. A change in SwaggerDocs was needed so that the swagger docs would render properly. This was due to the change in react 17 that makes teardown effects run asynchronously like other effects.

To get swagger docs working right after this change, we moved a setup function into a `useEffect` hook and moved a function into a `useCallback` hook.

Note that there aren't any updates to the dependabot configuration file because we forgot to add react as a 'problem upgrade' when we had to downgrade from 17 to 16. We'd be removing react in that dependabot config, but since we forgot to add it, there's nothing to remove. Sharing just in case this reminds someone else making changes to the dependencies to update the dependabot config if it's needed.

### Process

- [N/A] Docs updated
  
### Requested feedback

Nothing specific. We need to verify the full app is functioning with this change.

I tried organizing some of the code into logical sections in my last commit, so the files changed is a little intense. Sorry about that!


alexandervalencia (approved):
LGTM! 🚀  Great work determining the fix for Swagger @anthony-kopczyk !

gileswells (approved):
LGTM

PR #714, opened 2021-04-22
API-176 Fix response table overflow

gileswells:
### Description

https://vajira.max.gov/browse/API-176

The response table has been overflowing as too wide on both mobile and desktop but was overly egregious when zooming in further. While the ticket was specifically targeted at fixing zoomed in browsers this resolves issues regarding overflows to all browser widths.

### Process
- [ ] Docs updated
  

anthony-kopczyk (commented):
For the most part this looks good. At extreme small sizes, the tables stay wide, but the rest of our content shrinks. This causes the nav bar to be smaller than the max width of the screen. I'd like someone else to take a look before approving.

In production the nav bar stays the same width as the screen, but the tables have the overflow problem.

image

gileswells:
@anthony-kopczyk I cannot recreate the issue you're showing on the left of your screenshot. I see that the table still goes off screen but I don't see the nav bar bring narrowed. What browser, width and zoom setting are you at to get that to show? 
Personally I don't see that as being a blocker as I believe overall things are better with the proposed changes.

anthony-kopczyk:
> @anthony-kopczyk I cannot recreate the issue you're showing on the left of your screenshot. I see that the table still goes off screen but I don't see the nav bar bring narrowed. What browser, width and zoom setting are you at to get that to show?
> Personally I don't see that as being a blocker as I believe overall things are better with the proposed changes.

@gileswells It's only at extremes when you go down to the mobile menu sizes (around 360 width and lower), and on closer inspection, it's oddly the div at the root of the page that's narrow even though it doesn't have any styling on it. I'm running Chrome Version 90.0.4430.93 (Official Build) (64-bit) - my Chrome is saying that's the latest.

With Changes:
image

With Changes (hovering the div id root):
image

Without Changes:
image

anthony-kopczyk:
@gileswells If I toggle flex-direction on and off on the body tag in the Chrome dev console, it goes back and forth between the super narrow and overflowing the screen. So, it appears to be something with flex box and how it's interacting with the tables when the page makes the width calculations. It doesn't make any sense to me because I don't think flexbox is supposed to behave this way. There's only one child element so it should theoretically take up all the space regardless of flex-direction.

wwanamaker:
The Jira ticket specifies a width of 1280px. @anthony-kopczyk would you be cool with us moving forward as-is (unless Giles happens to find the fix quickly), and then I'll go ahead and create a new ticket for your discovery above?

anthony-kopczyk (approved):
Thanks @wwanamaker . Looking at the ticket acceptance criteria. Yes, this is working at 1280px with 400%.

I think the issue I'm seeing might be with the Chrome mobile mode simulator, or it could be something that specifically happens at 100% on small screens - but seeing the 1280px requirement, I did the math below and my screenshots definitely don't fall under wcag 2.1.

1280 / 4 [since 400%] = 320 - so the screenshots I took were 'smaller' than that size and are definitely above and beyond WCAG anyway. This should be good to go!

charleystran (approved):

PR #724, opened 2021-05-05
API-6080: error boundary feature

gileswells:
### Description
https://vajira.max.gov/browse/API-6080

Adds a designed error page when there's a javascript error within the React application. Previously the page would simply show a blank white page and users would have to know to look in the DevTools Console to see what went wrong. This provides a first level of information on the page and provides links elsewhere.

Because of their massive similarities I also merged the NotFound page and the ErrorBoundaryPage into a singular ErrorPage template with props to differentiate between them.

### Process

To test this you'll need to trigger an error within a component. Something like `throw new Error('This is an error');` will cause this error boundary page to show up. On development environments the React error page will show up first and you need to close that to see the stylized error page.

- [ ] Docs updated
  

christineyost (approved):
Looks good to me! 🎉

PR #733, opened 2021-05-14
API-6710: Update support to help guide veterans

anthony-kopczyk:
### Description

Jira Ticket: https://vajira.max.gov/browse/API-6710

This set of changes is intended to help guide them to va.gov or their local VSO to find data, since the portal contact us page is intended for developers and API questions

### Process

- [N/A] Docs updated
  
### Requested feedback

Specifically double checking to make sure this is consistent with the content in the jira ticket.


gileswells:
@anthony-kopczyk does this pass visual regressions? I know the npm audit currently fails but with on page changes I'm expecting those to fail.

anthony-kopczyk:
> @anthony-kopczyk does this pass visual regressions? I know the npm audit currently fails but with on page changes I'm expecting those to fail.

Oh! I didn't think of running them/forgot about them. They should fail with these changes. I'll get this updated with updated files. Thanks!

src/containers/support/ContactUs.tsx
```diff
@@ -32,16 +33,14 @@ const ContactUs = (): JSX.Element => {
...
-          </p>
-          <p>
-            <strong>Note:</strong> If you are a Veteran seeking help with claims, health records, or
-            other information, visit <a href="https://www.va.gov/">VA.gov</a> or contact your local
-            VSO for assistance.
-          </p>
+          <p>If you have questions about APIs, development, or related topics, use this form to send us a message.</p>
+          <AlertBox
+            status="info"
+            key="contactus"
```
gileswells, line 39: Is there a reason you're using a key here? From my testing you don't need it and if that's the case you can move the entirety of the `<AlertBox />` component into the `ContactUsAlertComment.tsx` file.

gileswells (changes requested):
Unless there's a reason I'm missing for using the key I'd much prefer to see it bundled up a little bit more since it's all the same markup wrapping the actual text.

src/containers/support/ContactUs.tsx
anthony-kopczyk: That comes from copy/pasta when using this component. I pulled it from the APIDocs or something like that (haven't done the alert box before, and now it's fairly apparent to me I didn't spend enough time making sure I knew the component).

I'll look at that component and try to tidy it up. Being able to move the entire list box into that other file will make it much cleaner too.

Thanks!

gileswells (approved):
LGTM

PR #743, opened 2021-05-20
Update instructions for adding an API to apply page

christineyost:
### Description

Update readme instructions for adding a new API to the apply page. Since formik was added, we no longer have the reducers and actions that were mentioned previously. Also called out that the backend will also need to be updated and updated references to the old `developer-portal-lambda-backend`

### Process

- [ ] Docs updated
  

gileswells (approved):
LGTM

PR #746, opened 2021-05-24
[API-5509] Reconcile operation tag swagger UI

gileswells:
### Description

https://vajira.max.gov/browse/API-5509

Because we were able to get a [PR merge](https://github.com/swagger-api/swagger-ui/pull/7224) upstream to swagger-ui to address the accessibility fixes we sidestepped here we can now remove that plugin and go with the standard implementation from the Swagger release.

### Process

- [ ] Docs updated
  

src/styles/swagger.scss
```diff
@@ -146,4 +146,28 @@
...
   }
+
+  button.opblock-summary-control {
+    display: contents;
+    text-align: left;
+    font-weight: normal;
+  }
+  .opblock .opblock-summary-method {
+    color: #fff;
+    line-height: 21px;
```
mjlumetta, line 157: for this selector...

1. is the color necessary? toggling that off didn't change anything.
2. can we use a relative line height?

src/styles/swagger.scss
```diff
@@ -146,4 +146,28 @@
...
+  }
+  .models-control {
```
mjlumetta, line 159: > why is this block necessary? why does it look so bad if you toggle these off?

same question

src/styles/swagger.scss
```diff
@@ -146,4 +146,28 @@
...
+
+    & > span {
+      line-height: 20.8px;
+    }
+  }
+  .model-box-control {
+    display: contents;
+  }
+  .model.model-title {
+    line-height: 24px;
```
mjlumetta, line 171: same suggestion about relative line height

mjlumetta (commented):
looks great to me, and happy to get that plugin out of here! left a bunch of questions. should. be able to approve after a round of answering those.

also, can you catch me up on the fixes we got in to the main package so i can try out some of those pieces of functionality?

src/styles/swagger.scss
```diff
@@ -146,4 +146,28 @@
   button.authorization__btn {
     display: none;
   }
+
+  button.opblock-summary-control {
+    display: contents;
```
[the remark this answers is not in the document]
mjlumetta: tried with VoiceOver in 3 browsers

1. Chrome: looks fine
2. Firefox: reads as a button but doesn't respond to clicks properly (?)
3. Safari: responds to clicks properly but the button isn't announced properly (not good behavior but makes way more sense than the Firefox behavior)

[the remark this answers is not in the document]
gileswells: I hadn't caught that. I'll rework it because taking it out of the DOM tree like that is a non-starter.

src/styles/swagger.scss
```diff
@@ -146,4 +146,28 @@
   button.authorization__btn {
     display: none;
   }
+
+  button.opblock-summary-control {
```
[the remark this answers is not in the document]
gileswells: Yes this is to fight back against the USWDS defaults so that it can be an invisible button to sighted users.

src/styles/swagger.scss
gileswells: More USWDS overriding to get the style to remain

src/styles/swagger.scss
gileswells: updated

src/styles/swagger.scss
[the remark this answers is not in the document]
gileswells: this is also resolved as I moved to flex

src/styles/swagger.scss
```diff
@@ -146,4 +146,32 @@
...
+    display: flex;
+    font-family: Titillium Web, sans-serif;
+    line-height: 1.3;
+    margin: 0;
+    padding: 0;
+    text-align: left;
+    width: 100%;
+  }
+  .model-box-control {
+    display: contents;
```
mjlumetta, line 172: same concern about `display: contents` and screen readers on this one. only tested in Firefox but I don't think it's working here either

src/styles/swagger.scss
```diff
@@ -146,4 +146,32 @@
   button.authorization__btn {
     display: none;
   }
+
+  button.opblock-summary-control {
+    align-items: inherit;
```
mjlumetta, line 151: slight preference for declaring this `align-items: center` explicitly in case something changes higher up the tree in the future. but not worth another commit on its own

mjlumetta (commented):
looks good, mainly have one comment on a `display: contents` again

mjlumetta (approved):

PR #755, opened 2021-06-03
[API-157] Left nav aria-current fixes

gileswells:
### Description

https://vajira.max.gov/browse/API-157

This PR addresses primarily addresses the left sidebar where every parent item to the current page (ex. /explore/benefits/docs/claims?version=current would have /explore/benefits as a parent) would also include an `aria-current="page"` attribute which is inaccurate and confusing to screen readers.

Additionally (self imposed scope creep) I updated the header and footer nav bars to address this as well so that there is no longer a discrepancy on different parts of the page. Previously when on /explore/benefits/docs/claims?version=current the Documentation link in the header nav would also include `aria-current="page"` despite it not being the same url. 

### Process

When checking different pages there should only be, at most, a single element within a navigation section that has `aria-current="page"` on it. The parent element, specifically in the documentation section will now show `aria-current="false"`.

Of note, the `router-react-dom` package does not, at it's currently installed version, support `"false"` as a valid value so this will cause a console warning to be shown on a local environment. This is suppressed on higher level environments. There is an [issue that has been open since October 2020](https://github.com/ReactTraining/react-router/issues/7647) asking for a new minor release as the fix to the allowed values had been merged to master since July 2020.

- [ ] Docs updated
  
### Requested feedback

If someone wants to take a stab at getting `react-router-dom` updated to using master or a specific commit hash instead of being behind on 5.2.0 I would happily do what I can to assist on that but I haven't been able to get that working myself.


ediiotero (approved):
LGTM 👍

PR #759, opened 2021-06-07
Fix overflow bug in the curl generator

hinzed1127:
### Description

The bearer tokens for OAuth are very long, resulting in an overflow on the curl generator. This fixes an issue where the token/copy-paste button were pushed very far to the right, requiring a lot of scrolling.

**Before**
Screen Shot 2021-06-07 at 6 20 28 PM

**After**
Screen Shot 2021-06-07 at 6 19 13 PM

### DIFF
--- a/src/components/codeWrapper/CodeWrapper.scss
+++ b/src/components/codeWrapper/CodeWrapper.scss
@@ -5,4 +5,8 @@
     margin: 0;
     overflow: visible;
   }
+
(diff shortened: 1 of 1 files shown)


jw81 (approved):

SlakrHakr (approved):

gileswells (approved):
LGTM

PR #762, opened 2021-06-09
API-7647 change appeals-status docs url to docserver S3

michaelkimball:
### Description

[API-7647](https://vajira.max.gov/browse/API-7647)
Added appeals status to docserver and migrated the dev portal url to it

### Process

- [ ] Docs updated
  

gileswells (dismissed):
LGTM

mjlumetta (commented):
i've got a couple of comments on this change.

1. this URL won't take effect in the Swagger UI docs. if both are present, the `metadataUrl` takes precedence over the `openApiUrl` and the `openApiUrl` is basically a backup. you can see this if you look at the Appeals Status page; it still uses the URL returned by that metadata endpoint.
2. more of a curiosity question than a feedback question: why are the path prefixes for the Docserver under `/internal` instead of `/services`? not that it matters very much.

but yeah, point 1 may be a bit gnarly to untangle. we want the Docserver to take work away from providers, but the versioning framework (to date) has been maintained by providers. should the Docserver take on the versions part of the metadata? should we rewrite the versioning framework? some combination of the above? no clear answers as of now, but worth considering.

michaelkimball:
> 
> 
> i've got a couple of comments on this change.
> 
>     1. this URL won't take effect in the Swagger UI docs. if both are present, the `metadataUrl` takes precedence over the `openApiUrl` and the `openApiUrl` is basically a backup. you can see this if you look at the Appeals Status page; it still uses the URL returned by that metadata endpoint.
> 
>     2. more of a curiosity question than a feedback question: why are the path prefixes for the Docserver under `/internal` instead of `/services`? not that it matters very much.
> 
> 
> but yeah, point 1 may be a bit gnarly to untangle. we want the Docserver to take work away from providers, but the versioning framework (to date) has been maintained by providers. should the Docserver take on the versions part of the metadata? should we rewrite the versioning framework? some combination of the above? no clear answers as of now, but worth considering.

Thanks Mike for looking at this PR! The metadata endpoint makes sense. I think I'll have to take this back to the team to figure out the best solution. We can probably end up delivering something similar.

To answer the question about `/services` vs `/internal`, the change was more to indicate that this OAS is not being served from a live API, rather an S3 bucket so there should be no worries about the downtime.

christineyost (commented):
Looks good to me! Confused by the test failure though. Maybe it just needs to be re-run?

dsteeve (dismissed):
lgtm

ediiotero (approved):
LGTM

PR #765, opened 2021-06-09
API-7512/Dependency update - 2

ediiotero:
### Description

Story [API7512](https://vajira.max.gov/browse/API-7512)

Second pr to update dependencies


gileswells (approved):
LGTM

PR #766, opened 2021-06-09
Switch to validating apply form only on submit

RockslideJr:
### Description

[API-7521](https://vajira.max.gov/browse/API-7521)
This switches the apply form to only validate when the submit button is pressed. It also updates the submit button to always be enabled, and focuses the first invalid input after the button is pressed.

### Process

- [ ] Docs updated
  

christineyost (approved):
Looks good to me! ✨  Played with the S3 deployed version a bit and the form seems to be behaving as it's supposed to

src/containers/apply/validateForm.ts
```diff
@@ -34,12 +34,17 @@ export const validateForm = (values: Values): FormikErrors<Values> => {
...
+  let count = 0;
   /*
    * This removes any fields that have an 'undefined' error (as returned by validatePresence)
    * This is needed, otherwise formik thinks there is still an error
    */
   return Object.entries(errors).reduce((reducedErrors, [key, value]) => {
     if (value) {
+      if (count === 0) {
+        document.getElementsByName(key)[0].focus();
+        count = count + 1;
```
mjlumetta, line 46: I've got two questions about this bit of logic.

1. we're setting focus of a DOM element inside this `validateForm` function which is used to update some Formik state. why does that happen here, as opposed to somewhere more directly tied to the DOM? seems like it could be coupling things that aren't super tightly related (or maybe someone wouldn't know to look here for focus management things, etc).
2. how does the `count` variable work? seems like it can only be 0 or 1, based on the fact that this is in an if block based on `count === 0`. if it's a 0/1 thing, should it be a boolean? also, more importantly, what does it do – is it just finding the first element with an error?

mjlumetta (commented):
overall looks good, liking the change. I have some questions about how the validations show up – it doesn't look like the error shows up in the DOM in that span where it does currently, which seems like something we'd want, and the `aria-invalid` attribute doesn't get set to true. also, when I entered a wrong email ("notanemail"), it gave me what I think is the Chrome browser error ("Please include an '@' in the email address. 'notanemail' is missing an '@'.") but nothing else. do you know how that works?

anyway, exciting that we're making this change!

src/containers/apply/validateForm.ts
RockslideJr: Oh yeah this whole bit of logic was garbage. I'm in the process of changing it all

RockslideJr: I will say i agree on your thoughts about the dom manipulation belonging somewhere else, I just couldn't think of a great place to do it. I tried putting it into the form component itself, but there's no way to tell if you actually need to do the focussing, as it can either be rendering due to the validation completing, or due to any other change to the form state, and both look the same to the component.

src/containers/apply/ApplyForm.tsx
```diff
@@ -130,7 +130,7 @@ const ApplyForm: FC<ApplyFormProps> = ({ onSuccess }) => {
                 className="form-checkbox"
               />
 
-              <button type="submit" className="vads-u-width--auto">
+              <button onClick={submitForm} type="button" className="vads-u-width--auto">
```
mjlumetta, line 133: why the change in the button type? shouldn't it still be `submit`?

src/containers/apply/validateForm.ts
```diff
@@ -34,17 +34,18 @@ export const validateForm = (values: Values): FormikErrors<Values> => {
     errors.oAuthRedirectURI = validateOAuthRedirectURI(values.oAuthRedirectURI);
   }
 
-  let count = 0;
+  setTimeout(() => {
```
mjlumetta, line 37: clever, liking this improvement, especially the `querySelectorAll` usage. I still wish we could disentangle this from the validation logic itself if possible... maybe if we used `useEffect` or a similar hook? does Formik have some kind of hook like that? (that would be super cool if they did.) I think this can work if there's not an alternative, but still feel like it'd be great if we could massage it into something different.

src/containers/apply/validateForm.ts
```diff
@@ -34,17 +34,18 @@ export const validateForm = (values: Values): FormikErrors<Values> => {
...
+    const firstError = document.querySelectorAll<HTMLElement>('[aria-invalid=true]')[0];
+
+    firstError.focus();
```
mjlumetta, line 40: looks like we need to check for the presence of an error before calling `focus()`

Screen Shot 2021-06-17 at 5 40 28 PM

mjlumetta (commented):
there's one error we definitely need to fix, and I still have questions about alternatives on the focus logic, although I think this works if we can't find a better option. I also wanted to ask if there's a problem with merging this without having validation messages for the API checkboxes and the terms of service checkbox. I know @ediiotero is adding validation for the checkboxes in #772, but I also said this PR was going to fix a potential issue with his, so I have no idea where we should land on that 😅 anyway, hope this is helpful

src/containers/apply/ApplyForm.tsx
RockslideJr: This is per @brentschneider 's request. If we leave it as a submit button, it runs all the semantic html validations before it submits the form, which means our validation never happens, and the only thing you see is the semantic html validation.

src/containers/apply/validateForm.ts
RockslideJr: ah, great catch, thanks!

src/containers/apply/validateForm.ts
RockslideJr: your persistence has paid off; I've thought of another place to put this

src/containers/apply/ApplyForm.tsx
alexandervalencia: Is there a way to add `novalidate` to the `<form>` tag with Formik?

RockslideJr: TIL `novalidate`. Looks like it works!

src/containers/apply/validateForm.ts
mjlumetta: love that solution!

mjlumetta (approved):
looks good!! I would still ask about adding the Terms of Service error message – is there a ticket for that? but the changes look good ✨

mjlumetta (approved):
re-approving

PR #768, opened 2021-06-11
Add consumer onboarding overview page behind feature flag

mjlumetta:
### Description

adds the overview page to the consumer onboarding docs, plus the structure like main nav changes and the consumer onboarding side nav. completes [API-7687](https://vajira.max.gov/browse/API-7687).

key links:

- [Figma design](https://www.figma.com/file/9sOD2GDAsyNTayvOqKlcJV/API-3854-Consumer?node-id=1397%3A0)
- [content](https://docs.google.com/document/d/1zjJIHPUrwug8fAGc1QEYGmb_8HCAMZ5Ymcq6UwN06vs/edit#)

### Process

- [x] Tests added or updated for change
- [ ] Docs updated
  - [ ] If you have made structural changes to the site, the [Developer Portal Content Types](https://community.max.gov/x/qUwOg) docs have been updated
  - [n/a] If you have made/confirmed any decisions about user interactions and accessibility, the [Accessible Component Design](https://community.max.gov/x/xS8mgg) docs have been updated
- [x] Accessibility concerns have been considered and addressed
  
  - [x] If you've made any UI changes, you've run an axe check using the axe browser extension that reported no new issues

### Requested feedback

wondering if anyone has opinions on whether or not to use markdown for some part of this functionality. not sure where we are with that at the moment.

PR #769, opened 2021-06-14
[API-7594] Benefits API/CategoryPage/PageHeader Updates for Consumer Docs

ndsprinkle:
### Description

As part of the consumer docs epic, this PR updates the Benefits API category page. Because of how the CategoryPage and PageHeader components are designed, this technically updates all the API category pages. The remaining consumer docs tickets for the other API category pages will not need to update the hierarchy as that is handled by this PR. However, text/content updates may still be necessary. 

The part of this ticket that adds the new card component is not currently implemented in this PR. Depending on the timing of the separate PR that creates this new card component, it may be possible to add it to this PR before merging, but is not a priority. 

This closes API-7594.

### Screenshots

image


src/components/pageHeader/PageHeader.tsx
```diff
@@ -9,6 +10,12 @@ interface PageHeaderProps {
   description?: string;
   halo?: string;
   header: string;
+  veteranRedirect?: {
```
mjlumetta, line 13: think we're going to want all this stuff to be done directly in the `CategoryPage` component still. the `PageHeader` component is used on pretty much every page, so this feels overly specific to put in here. were there any particular benefits you see to putting it here that i might be missing?

src/containers/documentation/CategoryPage.tsx
```diff
@@ -61,17 +61,23 @@ const CategoryPage = (): JSX.Element => {
...
+        header={categoryName}
+        veteranRedirect={veteranRedirect}
+        apiCategoryKey={apiCategoryKey}
+      />
       {intro({})}
+      <div className="vads-u-width--full">
+        {overview({})}
+        <Flag name={[FLAG_CONSUMER_DOCS]}>
+          <p>
+            <Link to="/consumer-docs">
```
mjlumetta, line 74: technically this link should be to `/onboarding`, but Giles made a constant `CONSUMER_PATH` in `src/types/constants/paths.ts` that i've used and updated in #768. so probably can use that constant as the `to` value here

src/content/apiDocs/benefits/benefitsOverview.mdx
```diff
@@ -1,3 +1 @@
-## Overview
-
-Our goal is to make easier and more secure for VSOs and other organizations to interact with the VA on behalf of Veterans. Individual API documentation is available above.
\ No newline at end of file
+Our benefits APIs allow you to submit and track electronic benefits claims, submit PDF claims and supplemental documentation, and more. Using these APIs allows you to speed up claims processing when compared to traditional mail or fax submissions. Read the API documentation to find out which claims these APIs handle. [Review the latest release notes](/release-notes/benefits).
```
mjlumetta, line 1: really wish the release notes link could be built in structurally, which we can't do with a markdown component... probably not ideal in any current situation, but it would be good if we could make it a structural feature of the category page. might have to write this overview part differently though, and doing some version of a React component could be a net negative. would be curious for your thoughts there.

mjlumetta (commented):
yeah, overall looks good. have a couple of comments and a question but shouldn't take much to move this forward

mjlumetta:
also, wondering if we should at least get rid of the "Overview" h2 on the Forms API overview to avoid this double h2

Screen Shot 2021-06-14 at 11 49 10 AM

i'm also not sure those should be h2's in the first place, but that is a sort of separate question i'm asking Brent.

ndsprinkle:
Met with Mike, going to make a few changes and also address the content updates for all the other category pages and then re-request review.

ndsprinkle:
Latest commits now update all API category pages and have some tweaks from conversations with Mike and Brent. 

This closes API-7595
This closes API-7598
This closes API-7599
This closes API-7600
This closes API-7601

src/__mocks__/fakeCategories.tsx
```diff
@@ -66,9 +66,6 @@ export const fakeCategories: APICategories = {
       },
     ],
     content: {
```
mjlumetta, line 68: would probably add in the `consumerDocsLinkText` here?

mjlumetta (approved):
I've got a few suggestions here, but in general, think this is looking great. like the consolidation of `intro` and `overview` a lot. lmk what you think about these potential changes.

src/__mocks__/fakeCategories.tsx
mjlumetta: also might make what was the `intro` the `overview` now, since it has a dummy element there, but not necessary by any means

src/apiDefs/schema.ts
```diff
@@ -24,7 +24,7 @@ export interface APICategoryContent {
 }
 
 export const ApiCategoryContentPropType = PropTypes.shape({
-  intro: PropTypes.any.isRequired,
+  consumerDocsLinkText: PropTypes.string,
```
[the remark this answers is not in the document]
ndsprinkle: I don't have a huge opinion on which way to take this. I like my approach as it gives the ability to make the consumer docs link whatever text for whatever category, but obviously there are currently only two different messages. And like you say, we'll need that boolean flag eventually, so for now it would be easy to just use the boolean to show one message or the other.

src/apiDefs/schema.ts
```diff
@@ -15,7 +15,7 @@ export interface VeteranRedirectMessage {
 }
 
 export interface APICategoryContent {
-  readonly intro: React.FunctionComponent;
+  readonly consumerDocsLinkText?: string;
```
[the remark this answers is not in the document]
ndsprinkle: Yeah, I'll make it required, I think that makes more sense. I'll change the CategoryPage too so that it always shows instead of conditionally showing.

src/__mocks__/fakeCategories.tsx
ndsprinkle: Will do. Since I'm changing it to be required, I think I'll have to add it here now anyway.

src/apiDefs/schema.ts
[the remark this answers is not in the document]
mjlumetta: yeah... if Darcy and Brent are reasonably confident that this is a consistent pattern, I'd lean toward using the boolean and refactoring later if we need to, but I think either is fine. gonna leave it up to your call.

mjlumetta (approved):
re-approving ✨

PR #772, opened 2021-06-14
Api 7522/api requirment

ediiotero:
### Description

Story: [API 7522](https://vajira.max.gov/browse/API-7522)

PR to group apply page API checkbox validation


RockslideJr (dismissed):

src/containers/apply/SelectedApis.tsx
```diff
@@ -67,26 +69,42 @@ const apiInfo = [
...
+        <legend className={classNames(selectAPIClass, 'vads-u-font-size--base', labelClass)}>
+          Select the APIs you want to access. <span className="apply-required">&#40;*Required&#41;</span>
+        </legend>
+        <span id="api-checkbox-error" className={validationClass}>
+          <ErrorMessage name="apis" />
+        </span>
+        <p>You can always request access to more APIs later.</p>
+        <fieldset
+          className="vads-u-margin-top--2"
+          aria-label="Please select all of the Standard APIs you'd like access to:"
```
mjlumetta, line 93: can this be an `aria-labelledby` referencing the legend here? kind of a judgment call, but wondering whether it adds anything in particular. same for the OAuth APIs

src/containers/apply/validateForm.ts
```diff
@@ -26,7 +26,7 @@ export const validateForm = (values: Values): FormikErrors<Values> => {
   }
 
   if (!anyApiSelected(values)) {
-    errors.apis = 'At least one API is required for submission.';
+    errors.apis = 'Choose at least one API.';
```
mjlumetta, line 29: was the old error ever used here? it wasn't, right? I don't remember why this didn't prevent submission

src/containers/apply/SelectedApis.tsx
```diff
@@ -67,26 +69,42 @@ const apiInfo = [
...
-    <fieldset
-      className="vads-u-margin-top--2"
-      aria-label="Please select all of the Standard APIs you'd like access to:"
-    >
-      <legend className="vads-u-font-size--lg">Standard APIs:</legend>
-      <ApiCheckboxList apiCheckboxes={apiInfo} />
+const SelectedAPIs = (): JSX.Element => {
+  const { errors, touched } = useFormikContext();
+  const checkboxName = 'apis';
+  const shouldDisplayErrors = !!errors[checkboxName] && !!touched[checkboxName];
```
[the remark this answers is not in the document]
mjlumetta: actually - I have a question about this now versus after John's changes. right now, since validation occurs once _any checkbox_ is touched, if you tab pass Claims Attributes it immediately shows the error. John's changes should solve that, but is it a problem now?

src/containers/apply/SelectedApis.tsx
```diff
@@ -67,26 +69,42 @@ const apiInfo = [
...
+  const containerClass = shouldDisplayErrors ? 'usa-input-error' : '';
+  const labelClass = shouldDisplayErrors ? 'usa-input-error-label' : '';
+  const validationClass = shouldDisplayErrors ? 'usa-input-error-message' : '';
+  const selectAPIClass = shouldDisplayErrors ? 'vads-u-font-weight--bold' : 'vads-u-font-weight--normal';
+  
+  return (
+    <fieldset className={classNames(containerClass, 'apply-api-select')}>
```
mjlumetta, line 82: should we add an accessible name to the top-level fieldset in addition to the nested fieldsets? testing with VoiceOver, Safari announced all 3 fieldsets as a group, using the legend as the name for the top level one, but Chrome only announced the nested fieldsets. may have to test that out again to confirm.

src/containers/apply/SelectedApis.tsx
```diff
@@ -67,26 +69,42 @@ const apiInfo = [
...
+      <span className="vads-u-margin-top--3 apply-checkbox-labels">
+        <legend className={classNames(selectAPIClass, 'vads-u-font-size--base', labelClass)}>
+          Select the APIs you want to access. <span className="apply-required">&#40;*Required&#41;</span>
+        </legend>
+        <span id="api-checkbox-error" className={validationClass}>
```
mjlumetta, line 87: should we add an `aria-live="polite"` here? we have not been particularly good about including that in form validation, but I believe it's something we should be doing.

mjlumetta (commented):
leaving some comments here – overall looks very good. a lot of these comments are accessibility ones. the big question for me is how the validation occurs when the blur event fires on the first checkbox touched; this seems like a usability problem, particularly for keyboard users. that said, I think John's work on #766 might fix that if that PR goes first.

src/containers/apply/SelectedApis.tsx
mjlumetta: actually I got this answer from Trevor Pierce in the Ad Hoc Slack:

> If you're dynamically adding error messages after an event like blur() fires, we usually use role="alert" on the element being added, which has the same effect as aria-live="assertive" to announce the error immediately. That's definitely abrupt, and I don't know there's been any testing with users to ask their preference or if they have better examples for us to riff on.

that said, either would be an improvement over what we do now.

src/containers/apply/ApplyForm.tsx
```diff
@@ -104,7 +104,7 @@ const ApplyForm: FC<ApplyFormProps> = ({ onSuccess }) => {
       <div className={classNames('vads-l-col--12', 'vads-u-padding-x--2p5')}>
         <Formik initialValues={initialValues} onSubmit={submitForm} validate={validateForm}>
           {({ dirty, isValid, isSubmitting, values }): React.ReactNode => (
-            <Form className="usa-form">
+            <Form className="vads-l-col--12 medium-screen:vads-l-col--5">
```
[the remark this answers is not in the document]
ediiotero: the 5 column with came from @brentschneider
In the design the gray box is slightly wider than the input boxes above

src/containers/apply/DeveloperInfo.tsx
```diff
@@ -8,29 +8,29 @@ const DeveloperInfo = (): JSX.Element => (
...
-      className="vads-u-margin-top--4"
+      className="vads-u-margin-top--4 medium-screen:vads-l-col--10"
     />
 
     <TextField
       label="Organization"
       name="organization"
       required
-      className="vads-u-margin-top--4"
+      className="vads-u-margin-top--4 medium-screen:vads-l-col--10"
```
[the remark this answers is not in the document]
ediiotero: good call

src/containers/apply/SelectedApis.tsx
```diff
@@ -67,26 +69,42 @@ const apiInfo = [
...
+  const checkboxName = 'apis';
+  const shouldDisplayErrors = !!errors[checkboxName] && !!touched[checkboxName];
+  const containerClass = shouldDisplayErrors ? 'usa-input-error' : '';
+  const labelClass = shouldDisplayErrors ? 'usa-input-error-label' : '';
+  const validationClass = shouldDisplayErrors ? 'usa-input-error-message' : '';
+  const selectAPIClass = shouldDisplayErrors ? 'vads-u-font-weight--bold' : 'vads-u-font-weight--normal';
+  
+  return (
+    <fieldset className={classNames(containerClass, 'apply-api-select')}>
+      <span className="vads-u-margin-top--3 apply-checkbox-labels">
```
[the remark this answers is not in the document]
ediiotero: agreed, the span is a bit weird there. I think I had an issue with a div and a problem with conflicting classes. Using a div now and it works. 
Adding padding to the fieldset was pushing content down while the text from the legend was above the gray box. Wrapping the content in a div fixed that.

src/containers/apply/SelectedApis.tsx
ediiotero: done 👍

src/containers/apply/validateForm.ts
ediiotero: No, it wasn't.  My guess is it was due to the submit button being disabled till the form was valid

src/containers/apply/SelectedApis.tsx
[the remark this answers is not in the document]
ediiotero: it is not a problem now, the errors for the checkboxes never fire

src/containers/apply/SelectedApis.tsx
ediiotero: added the aria-live="polite" attribute but it doesn't seem to announce the error using nvda, it did announce when using role="alert". 
@mjlumetta should we add the role="alert"?

src/containers/apply/SelectedApis.tsx
ediiotero: added an aria-label="API Select Checkboxes" open to suggestions

src/containers/apply/validateForm.ts
mjlumetta: makes sense. so submit button is disabled, there is no displayed error, but they must correct the error to enable submission

src/containers/apply/SelectedApis.tsx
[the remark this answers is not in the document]
mjlumetta: can you explain further? should I clarify the scenario I'm describing?

src/containers/apply/SelectedApis.tsx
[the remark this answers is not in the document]
mjlumetta: okay cool! not a huge deal but this makes somewhat more sense semantically. thanks for updating

src/containers/apply/SelectedApis.tsx
mjlumetta: I'd suggest maybe adding an `aria-labelledby` referring to the "Select the APIs you want to access" legend, to conform with what visual users see? that might be overkill with the use of that content to label the nested fieldsets, but you could also remove it as a label from those nested fieldsets. thoughts?

src/containers/apply/SelectedApis.tsx
mjlumetta: it works for me in VoiceOver on Chrome and Safari, but if NVDA isn't working, then yeah, maybe try either `role="alert"` or `aria-live="assertive"`? if you go with `role="alert"`, I would say don't include `aria-live` so there aren't potential conflicts across browsers and screen readers.

mjlumetta (commented):
left a couple of comments on the existing threads. I think the screen reader improvements look good – suggested a couple tweaks but those are up to your judgment. the bigger issue is the early validation for the API checkboxes. is it clear what I'm talking about there? can sync up or try to explain more clearly if you can clarify what your understanding of it is.

src/containers/apply/SelectedApis.tsx
```diff
@@ -70,34 +70,34 @@ const apiInfo = [
 ];
 
 const SelectedAPIs = (): JSX.Element => {
-  const { errors, touched } = useFormikContext();
+  const { errors } = useFormikContext();
   const checkboxName = 'apis';
-  const shouldDisplayErrors = !!errors[checkboxName] && !!touched[checkboxName];
+  const shouldDisplayErrors = !!errors[checkboxName];
```
mjlumetta, line 75: this doesn't really change the behavior currently (since the validation is triggered once the fields are touched), but since this won't go forward without John's changes, it should be good ✅

mjlumetta (dismissed):
looks good!! ✨

christineyost (approved):
Looks good to me! 🎉

PR #778, opened 2021-06-21
API-5785: Add Sentry to production deploy job

gileswells:
### Description

https://vajira.max.gov/browse/API-5785

This adds Sentry releases to the CICD process so that we can see Sentry events bundled into the different releases. Also, as an added bonus, I added the Sentry logging of errors back to the project as the ErrorBoundary page caught the errors before it got to the app level logging code.

### Process

Kinda hard to really test the CICD process on this until it actually goes to production.

- [ ] Docs updated
  
### Requested feedback

If anyone knows how to post the last error Sentry in the build process that would be great but as far as I can see there's only an environment variable for build passing / failing.


michaelkimball:
It looks like there is a way to [ship logs](https://docs.sentry.io/product/cli/send-event/#sending-breadcrumbs) with the Sentry message
Maybe you could pick up the log in ` /root/.npm/_logs/` and if it exists attach it to the message.

gileswells:
> Maybe you could pick up the log in ` /root/.npm/_logs/` and if it exists attach it to the message.

I like the idea of sending the logs to Sentry but the downside is that this file doesn't use any npm steps. The only solution I can think of (just pushed the commit) is dual outputting the commands to a `build-output.txt` file and also to the console so you get the output in both a file and in the CodeBuild logs. 

Unfortunately, it's ugly so I'm not the biggest fan of it. Personally, while having the logs in Sentry would be good, we can get the full logs through the CI job in CodeBuild already so I'm not sure if the ugliness is worth the data being duplicated into another system. 

@michaelkimball thoughts?

michaelkimball:
It could be helpful when onboarding new members that aren't able to get Codebuild access for awhile, but I agree that that the benefit is limited. Other than that, it's nice to be able to view things in one place. Also, I think Sentry allows for email notifications. I'm not sure what setting that up on Codebuild looks like. Either way, I don't have a huge preference.

gileswells:
@michaelkimball I reverted the change. I don't think it adds more value than the complexity of its implementation and usage.

michaelkimball (approved):
Into the breach! 🚀

PR #780, opened 2021-06-22
API-8096 API-8097: Open redirect fixes for Fortify

gileswells:
### Description

https://vajira.max.gov/browse/API-8096
https://vajira.max.gov/browse/API-8097

Updates no longer take the url param and directly push that into a link tag but instead grabs it from a data store item where it's static text.

### Process

- [ ] Docs updated
  

src/containers/documentation/ApiNotFoundPage.tsx
```diff
@@ -14,6 +14,8 @@ import { PAGE_HEADER_ID } from '../../types/constants';
 const ApiNotFoundPage = (): JSX.Element => {
   const { apiCategoryKey } = useParams<APINameParam>();
   const category = lookupApiCategory(apiCategoryKey);
+  // You can't get to this page without a valid API so this ?? is just for TypeScript's benefit
+  const api = lookupApiByFragment(apiCategoryKey) ?? { urlFragment: '' };
```
mjlumetta, line 18: this doesn't appear to work (see comment below) because it doesn't follow the contract for `lookupApiByFragment` because the parameter is supposed to be an API URL fragment/key, not the category key. it coincidentally works for 4 of 6 APIs because there's an API that has the same key/identifier as the category.

can we redirect to the 404 page if the category is null? does that solve the scan issue?

src/containers/documentation/QuickstartPage.tsx
```diff
@@ -21,7 +21,11 @@ const QuickstartPage = (): JSX.Element => {
   if (quickstartContent) {
     return <QuickstartWrapper categoryName={name} quickstartContent={quickstartContent} />;
   } else {
-    return <Redirect to={`/explore/${apiCategoryKey}`} />;
+    const apiLookup = lookupApiByFragment(apiCategoryKey);
+    if (apiLookup) {
+      return <Redirect to={`/explore/${apiLookup.urlFragment}`} />;
```
mjlumetta, line 26: same thing here – works for the first 4 APIs but not the last 2. would suggest checking that the category is present (however we look that up, with `getApiDefinitions` or `lookupCategory`) rather than looking up an API (which will work accidentally in many cases but not all).

mjlumetta (changes requested):
some mixup between use of categories and APIs here, i think. works for 4 categories (Appeals, Benefits, Facilities, Forms) by coincidence, in that each has an API with the same key/identifier as the category's key, but not the other 2.

src/containers/documentation/ApiNotFoundPage.tsx
gileswells: What we should be doing is showing the 404 on the url it happens on and in fact [there's a ticket](https://vajira.max.gov/browse/API-6348) I'm about to pull in that does exactly that. The main issue is pulling the url path parameter directly into a `<Link` or `<Redirect` and Fortify isn't able to dig into the code to determine that even getting to the offending code means that it's been validated by way of everything else happening.

gileswells:
@mjlumetta I've had to rework how it functions but this satisfies the Fortify scan.

gileswells:
Going to drop this to draft status while we wait to see if these open redirects can be explained away in the Fortify scans. Depending on that result we can close / revisit this.

gileswells:
Closing since this is way out of date and there hasn't been an update saying these updates are required.

PR #787, opened 2021-06-25
API-7851 - Remove state from token requests

beaugrantham-va:
Remove `state` from calls to `/token` as it's not required nor used.

https://vajira.max.gov/browse/API-7851


charleystran (approved):
LGTM

lukemarkwordtlibertyits (approved):
Looks Good

PR #791, opened 2021-06-29
Dynamically create apply page checkbox list

ediiotero:
Story: [API-7515](https://vajira.max.gov/browse/API-7515)

### Description

Currently the Apply page is using it's own custom JSON representations for APIs that are displayed as checkboxes. 
This PR is using the api query to get api definitions to populate the checkbox list on the Apply page.


src/apiDefs/data/benefits.ts
```diff
@@ -29,6 +29,7 @@ const benefitsApis: APIDescription[] = [
     urlFragment: 'claims',
     vaInternalOnly: false,
     veteranRedirect: benefitsContent.veteranRedirect,
+    altID: 'claims',
```
ediiotero, line 32: added altIDs to match the kong keys on the backend

src/apiDefs/data/facilities.ts
```diff
@@ -1,9 +1,10 @@
-import { facilitiesContent, FacilitiesReleaseNotes } from '../../content/apiDocs/facilities';
+ import { facilitiesContent, FacilitiesReleaseNotes } from '../../content/apiDocs/facilities';
```
gileswells, line 1: Extra space added inadvertently

src/apiDefs/schema.ts
```diff
@@ -75,6 +75,7 @@ export interface APIDescription {
   readonly deactivationInfo?: APIDeactivationInfo;
   readonly multiOpenAPIIntro?: React.FunctionComponent;
   readonly veteranRedirect?: VeteranRedirectMessage;
+  readonly altID?: string;
```
gileswells, line 78: Is this optional because it's not required for previously deactivated APIs?

src/containers/apply/Apply.test.tsx
```diff
@@ -58,23 +65,8 @@ describe('Apply', () => {
...
-      expect(screen.getByRole('checkbox', { name: /VA Claims API/ })).toBeInTheDocument();
-      expect(screen.getByRole('checkbox', { name: /VA Health API/ })).toBeInTheDocument();
-      expect(
-        screen.getByRole('checkbox', { name: /Community Care Eligibility API/ }),
-      ).toBeInTheDocument();
-      expect(
-        screen.getByRole('checkbox', { name: /VA Veteran Verification API/ }),
-      ).toBeInTheDocument();
+    it.each(allCheckboxApis)('displays the expected checkboxes', name => {
+      expect(screen.getByRole('checkbox', { name })).toBeInTheDocument();
```
gileswells, line 69: This test is an additive test where it's checking to make sure they've all been added. I'd optionally like to see the reverse where those that shouldn't be added are confirmed to not be on the page. I can see this being either in this file or in the `ApplyForm.test.ts` file. Either one works for me and I certainly feel it doesn't need to be in both.

src/containers/apply/SelectedApis.tsx
```diff
@@ -25,47 +29,19 @@ const ApiCheckboxList = ({ apiCheckboxes }: APICheckboxListProps): JSX.Element =
...
-  {
-    id: 'communityCare',
-    label: 'Community Care Eligibility API',
-  },
-  {
-    id: 'verification',
-    label: 'VA Veteran Verification API',
-  },
-];
+const oauthInfo = getAllOauthApis()
```
gileswells, line 32: I feel like this could be better placed in the `src/apiDefs/query.ts` file where we do more of these queries and then can also layer in the associated unit tests to make sure that these query tests are working. Adding a test to this query to make sure it's not grabbing more than it should would in theory make my previous test request unnecessary.

src/containers/apply/SelectedApis.tsx
```diff
@@ -25,47 +29,19 @@ const ApiCheckboxList = ({ apiCheckboxes }: APICheckboxListProps): JSX.Element =
...
-  {
-    id: 'vaForms',
-    label: 'VA Forms API',
-  },
-  {
-    id: 'confirmation',
-    label: 'VA Veteran Confirmation API',
-  },
-];
+const standardApi = getAllApis()
```
gileswells, line 39: Same as above for `oauthInfo`. Could also bundle it into something like `getAllOpenRegistrationApis(oauth: boolean): APIDescription[] => {}` to kill 2 birds with one stone. Open to ideas here.

gileswells (changes requested):
Looks great but I had some ideas on possible improvements.

src/containers/apply/SelectedApis.tsx
ediiotero: moved it over

src/apiDefs/schema.ts
ediiotero: Yeah that's one reason. I had to add this to keep the value of the checkboxes the same as they need to match the keys on the backend.

src/containers/apply/Apply.test.tsx
ediiotero: moved it to the query file and added tests

gileswells (approved):
I can leave the variable name change suggestions but if you agree let me know when done and it'll be a quick re-approve.

src/apiDefs/query.ts
```diff
@@ -50,6 +51,12 @@ const apisFor = (apiList: string[]): APIDescription[] => {
 
 const includesOAuthAPI = (apiList: string[]): boolean => apisFor(apiList).some(api => !!api.oAuth);
 
+const getAllCurrentOauthApis = (): APIDescription[] =>
+  getAllOauthApis().filter((api: APIDescription) => api.enabledByDefault && api.altID && !isApiDeactivated(api));
+
+const getAllCurrentStandardApis = (): APIDescription[] =>
+  getAllApis().filter((api: APIDescription) => api.enabledByDefault && api.altID && !api.oAuth && !isApiDeactivated(api));
```
mjlumetta, line 58: i have some questions about these new functions. first, what does "current" here mean? from context, it seems like "enabled and not deactivated"; it's unclear where the `altID` factors in. to date, this data access layer kind of has the structure where `query` is the gateway to the `data` directory, and the `deprecated` and `env` files depend on `query` to dictate logic about what APIs are actually available in the UI (but `query` does not handle logic about enabled/disabled and deactivation). these functions would break that pattern; it's fine if we want to do that, but i'd be curious to hear your thoughts on why. my preference would be to use `Flag` on the apply page around these checkboxes rather than adding in additional logic here.

couple other related thoughts:
1. would still benefit from a `getAllStandardApis` function here (or "key auth" – i have a slight preference for that since it is more specific, regardless of what the UI says).
2. `enabledByDefault` does not take into account the environment variables, which are a big part of how enabled/disabled status is managed. using `Flag` would solve that, but we could also include that logic here.

src/containers/apply/SelectedApis.tsx
```diff
@@ -25,47 +28,17 @@ const ApiCheckboxList = ({ apiCheckboxes }: APICheckboxListProps): JSX.Element =
...
-  {
-    id: 'vaForms',
-    label: 'VA Forms API',
-  },
-  {
-    id: 'confirmation',
-    label: 'VA Veteran Confirmation API',
-  },
-];
+const standardApi = getAllCurrentStandardApis()
```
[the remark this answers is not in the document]
ediiotero: agreed, made the change

src/apiDefs/query.test.ts
```diff
@@ -110,4 +113,22 @@ describe('query module', () => {
       expect(getAllQuickstartCategorySlugs()).toStrictEqual(['health']);
     });
   });
+
+  describe('getAllCurrentOauthApis', () => {
+    it.each(getAllCurrentOauthApis())('get all current and active oauth apis', api => {
+      expect(api).toHaveProperty('altID');
+      expect(api).toHaveProperty('oAuth');
+      expect(api).not.toHaveProperty('deactivationInfo');
```
mjlumetta, line 121: i don't know how accurate this test is? non-deactivated APIs can still have `deactivationInfo` set. in general, i'm curious what the thought behind these tests are.

also, when these tests are run, do they have unique names? i would suggest writing it in such a way that they can be distinguished, if we're going to use `it.each` here.

src/apiDefs/schema.ts
mjlumetta, line 78: can you add a comment for this field to describe why it exists (vs `urlFragment`)? also, why optional vs required?

src/containers/apply/SelectedApis.tsx
```diff
@@ -25,47 +28,17 @@ const ApiCheckboxList = ({ apiCheckboxes }: APICheckboxListProps): JSX.Element =
...
+  .map((api: APIDescription) => ({
+    id: api.altID ?? '',
```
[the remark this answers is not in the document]
gileswells: There wouldn't be an empty name here though. In order to get in the list of standard APIs you have to have an `altID`. This is just TypeScript being TypeScript.

src/containers/apply/Apply.test.tsx
```diff
@@ -55,25 +58,4 @@ describe('Apply', () => {
...
-      expect(screen.getByRole('checkbox', { name: /VA Claims API/ })).toBeInTheDocument();
-      expect(screen.getByRole('checkbox', { name: /VA Health API/ })).toBeInTheDocument();
-      expect(
-        screen.getByRole('checkbox', { name: /Community Care Eligibility API/ }),
-      ).toBeInTheDocument();
-      expect(
-        screen.getByRole('checkbox', { name: /VA Veteran Verification API/ }),
-      ).toBeInTheDocument();
-    });
-  });
```
alexandervalencia, line 78: I'm curious if we're still doing the same testing elsewhere now?

alexandervalencia (approved):
Looks good!

src/containers/apply/Apply.test.tsx
ediiotero: I feel like the SelectedApis test in ApplyForm.test.tsx covers this

PR #804, opened 2021-07-07
Fix visual bug and update image snapshots 

ediiotero:
PR to fix a visual bug on the Apply page, removes the ; at the end of the checkbox list.
Also updates the visual regression snapshots for the Apply page and the Side Nav (Claims Attributes was removed).


gileswells (dismissed):
LGTM

gileswells (approved):

PR #805, opened 2021-07-07
Update Community Care API OAS url

christineyost:
### Description

This PR is for [API-7646](https://vajira.max.gov/browse/API-7646). It updates the Community Care API OAS url so that it points to the OAS that is transformed by Docserver. It also disables the Community Care API in staging since this API is not in the staging environment. Disabling the Community Care API in the staging environment led to an issue building the sitemap because there are no health APIs in the staging environment. To overcome that error I added a filter statement to the sitemapConfig in the `Routes.tsx` file

### Process

- [ ] Docs updated
  

gileswells:
Personally I'd prefer we also block the sitemap all together on dev and staging environments if there's going to be discrepancies like this unless there's a good use case I'm missing for them to be active on the lower environments.

christineyost:
Confirmed the sitemap is already blocked for dev and staging environments. It will still be built for those environments though.

alexandervalencia (approved):
🚀

PR #806, opened 2021-07-12
API-6348 - Provide better URL validation and 404 handling in Routes.tsx

gileswells:
### Description

https://vajira.max.gov/browse/API-6348

One of the overall goals here is to ensure that the proper 404 page shows up in place (without a redirect) within the dev portal. There are many paths that currently will render a black content block but not show the 404 page. [example with redirect](https://developer.va.gov/release-notes/banana) [example of blank content](https://developer.va.gov/explore/vaForms/banana)

### Process

Test all page urls to ensure that all pages are rendering the proper content and also check pages, especially wildcard routes, to ensure the proper 404 page shows up as expected.

- [ ] Docs updated
  
### Requested feedback

This is a pretty substantial change as it aims to consolidate a lot of the routing logic into this single Routes.tsx file (which as an added benefit helps make the sitemap setup easier). Very much looking for thoughts and comments or places to improve.


Cgfor3 (dismissed):
Updates all look good - I'd wait for someone more knowledgeable to approve though :D

SlakrHakr (dismissed):
i havent looked at a ton of typescript previously but everything looks alright to my novice eyes.
i did at least pull down both master and this branch to compare locally and i cant reproduce the redirects in this new branch at all 👍🏻

mjlumetta:
getting blank content for the following routes (including no side nav):

- /explore/fake
- /explore/benefits/fake
- /explore/benefits/docs
- /explore/fake/docs/fake
- /release-notes/fake
- /support/fake

src/containers/documentation/DocumentationRoot.tsx
```diff
@@ -103,33 +102,27 @@ const oldRouteToNew = [
...
-  return (
-    <ContentWithNav
-      nav={<ExploreSideNav />}
-      content={
-        <Switch>
-          {oldRouteToNew.map(routes => (
-            <Redirect key={routes.from} exact from={routes.from} to={routes.to} />
-          ))}
-          <Route path="/explore/authorization" component={AuthorizationDocs} exact />
-          {!shouldRouteCategory && <Redirect from="/explore/:apiCategoryKey" to="/404" />}
```
mjlumetta, line 119: do we still want this safeguard to protect against wholes in the logic higher up? `/explore/fake` is currently not working.

mjlumetta (changes requested):
some holes in the logic - see [this comment](https://github.com/department-of-veterans-affairs/developer-portal/pull/806#issuecomment-882814239). think this is a good idea but needs a little tightening up, probably?

i think the thing we really need here is some integration tests for this `Routes` file. all this handling of routes and wildcard routes in particular is getting confusing, so there's no basis for confidence when you're making changes. having a test suite both to catch regressions and provide a specification of expected behavior would solve those problems; i think that would ideally be part of this change.

gileswells:
> I think the thing we really need here is some integration tests for this `Routes` file.

@mjlumetta That was always the plan. Part of the reason I didn't fix the currently broken tests / write the new ones to make sure we were covering all our bases was I wanted to get some feedback to see if people would be on board with this style of declaring the routes directly like this before passing off to secondary routes per section (like in `DocumentationRoot`).

I'm working on those tests now and some of the holes that I apparently missed.

src/Routes.tsx
```diff
@@ -65,39 +68,87 @@ export const SiteRoutes: React.FunctionComponent = (): JSX.Element => {
...
-        render={(): JSX.Element => <Redirect to="/explore/verification/docs/authorization" />}
-      />
-      <Route path="/release-notes/:apiCategoryKey?" component={ReleaseNotes} />
+        />
+      )}
+
+      {/* API Documentation */}
+      <Route exact path="/explore" component={DocumentationRoot} />
+      <Route exact path="/explore/authorization" component={DocumentationRoot} />
+      {Object.entries(apiDefinitions)
```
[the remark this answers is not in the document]
gileswells: I tried that a few different times BUT always with the shorthand `<></>` instead of the longhand `<React.Fragment></React.Fragment>` and now I'm feeling abundantly silly. I'll give that a go.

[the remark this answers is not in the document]
gileswells: Ended up reworking this to use a route builder method that loops over the APIs to create a simple array of paths. Much simplier in the end over this and avoids the issue of using a [`<React.Fragment>` within a `<Switch>`](https://stackoverflow.com/questions/58169397/redirect-doesnt-work-inside-switch-when-fragment-has-been-used).

src/containers/documentation/DocumentationRoot.tsx
gileswells: I don't think it's necessary with the updates to the routes changes made today/yesterday.

src/content/explorePage.mdx
```diff
@@ -1,10 +1,12 @@
-<h1 style={{ 'margin-top': '0' }}>
+<h1 style={{ 'marginTop': '0' }}>
```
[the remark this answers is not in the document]
gileswells: I believe that this page in it's entirety was only ever visible in the very weirdest of circumstances and with the routing changes this page should, in theory at least, never be routed. Will confirm before killing off.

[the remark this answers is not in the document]
gileswells: Well it would be routed to if an API was disabled in the config. While I can see how we might want to use that in the event of discontinuing an API without a replacement the `ApiNotFoundPage` component is a much better way to do that. I'd suspect that any true sunset of an API would be accompanied by a redirect to it's replacement or at least to the main category page.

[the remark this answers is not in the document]
gileswells: I've updated this to use the `ApiNotFoundPage` component and killed this file (and it's associated unit / e2e testing) off.

gileswells:
> getting blank content for the following routes (including no side nav):
> 
> * /explore/fake
> * /explore/benefits/fake
> * /explore/benefits/docs
> * /explore/fake/docs/fake
> * /release-notes/fake
> * /support/fake

All of these should now show the 404 page without a redirect. Also there's now e2e coverage to enforce this.

src/containers/consumerOnboarding/ConsumerOnboardingRoot.e2e.ts
```diff
@@ -0,0 +1,13 @@
+import 'jest';
+import { puppeteerHost } from '../../e2eHelpers';
+
+describe('Onboarding Page', () => {
+  it('should show the 404 page on /onboarding/invalid', async () => {
```
mjlumetta, line 5: my preference for these tests would probably be to have them all in a single `Routes.e2e.ts` file rather than in the test suite for each top-level page. if the motivation for this PR is to consolidate the routing logic into a single place (`Routes.tsx`), it makes sense to also test that logic in the same place, imo. it also makes looking up routing behavior easier; if i want to know what should happen when i go to `/explore/fake/docs/appeals`, i know i can look at that file and see the tests. thoughts?

src/containers/documentation/ApiNotFoundPage.tsx
```diff
@@ -26,7 +28,7 @@ const ApiNotFoundPage = (): JSX.Element => {
       {category?.name && <PageHeader header={category.name} />}
       <ul>
         {category?.apis
-          .filter((item: APIDescription) => !isApiDeactivated(item))
+          .filter((item: APIDescription) => !isApiDeactivated(item) && flags[item.urlFragment])
```
mjlumetta, line 31: why aren't we just using `Flag` here?

src/containers/documentation/ApiPage.tsx
```diff
@@ -58,19 +57,17 @@ const getApi = (apiName?: string): APIDescription | null => {
 const ApiPage = (): JSX.Element => {
   const location = useLocation();
   const params = useParams<APINameParam>();
+  const flags = useFlag([FLAG_API_ENABLED_PROPERTY]);
```
mjlumetta, line 60: why not

```suggestion
  const flags = useFlag([FLAG_API_ENABLED_PROPERTY, api.urlFragment]);
```

(if you move it after line 62)

i would also suggest clarifying what type of flag it is in the variable name so it's easier to remember exactly which value we're using here.

src/utils/routesHelper.ts
```diff
@@ -0,0 +1,20 @@
...
+  Object.entries(apiDefinitions).flatMap(([key, value]: [string, APICategory]) => {
+    const apiRoutes = [];
+    apiRoutes.push(`/explore/${key}`);
+    if (value.content.quickstart) {
+      apiRoutes.push(`/explore/${key}/docs/quickstart`);
+    }
+    if (value.apis.length > 0) {
+      value.apis.forEach(api => {
+        // Not required but this outputs the urls to generate the sitemap
+        apiRoutes.push(`/explore/${key}/docs/${api.urlFragment}`);
```
mjlumetta, line 13: how does this work with disabled APIs (or categories)?

mjlumetta (commented):
looking good. my main question is about potentially incorporating all these tests into one `Routes.e2e.ts` file – i think that could be beneficial to test all the routing in one place if we're trying to encapsulate the main logic inside `Routes.tsx`. open to discussion on that. everything else is minor comments.

src/Routes.tsx
```diff
@@ -131,24 +101,18 @@ export const SiteRoutes: React.FunctionComponent = (): JSX.Element => {
...
-      <Route exact path={PUBLISHING_PATH} component={Publishing} />
-      <Route exact path={PUBLISHING_EXPECTATIONS_PATH} component={Publishing} />
-      <Route exact path={PUBLISHING_ONBOARDING_PATH} component={Publishing} />
+      {PUBLISHING_ROUTER_PATHS.map((path: string) => (
+        <Route exact path={path} component={Publishing} key={path} />
+      ))}
 
       {/* Consumer Docs */}
-      {flags.consumer_docs && (
-        <>
```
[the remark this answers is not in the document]
gileswells: The main issue was that the sitemap generator wasn't able to go into sub-routes files. The whole section at the bottom of the Routes.tsx file was specifically for the sitemap. With the changes to how we do routes we don't need this last section (notice how there's no listed component on these routes).

src/containers/consumerOnboarding/ConsumerOnboardingRoot.e2e.ts
gileswells: I like that. I'll do that today.

src/utils/routesHelper.ts
gileswells: This helper doesn't care if something is disabled or depreciated. Because it takes in the categories (and their apis) as a parameter it just builds from there based on what it's given. If there was ever a need to build all the routes, even including deactivated routes, this helper would be already good to go. It assumes that you've made the decisions about deactivated routes before calling this helper.

src/containers/documentation/ApiNotFoundPage.tsx
gileswells: Personally I prefer a single place for the conditions with in a filter/map situation. Having a filter and then following it up with a flag later when it can all be handled in one filter method seems less straight forward. Also by waiting to use `<Flag` later you're passing more items to the `.map()` loop which caused more processing time. Milliseconds at worst but still why loop if it might return empty anyways?

src/Routes.tsx
```diff
@@ -171,34 +135,9 @@ interface SitemapConfig {
...
   return {
-    paramsConfig: {
-      '/explore/:apiCategoryKey/docs/:apiName': apiCategoryOrder
-        .filter(apiCategory => getApiRouteParams(apiCategory).length > 0)
-        .map(apiCategory => ({
-          apiCategoryKey: apiCategory,
-          apiName: getApiRouteParams(apiCategory),
-        })),
-    },
+    paramsConfig: {},
```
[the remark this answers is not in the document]
gileswells: That is a concern that I have too but part of the primary drive behind this was the issue that wildcard routes within React Router aren't really that stellar in terms of what to do with a 404. There's no standard way of loading up a wildcard route but allow it to skip past if there's no matching value on the passed route parameter. 

Like `/explore/invalid` would, on a route of `/explore/:apiCategory`, always load the `DocumentationRoot` component and once you're in there it's impossible to kick back up to the parent and continue on through the routes till you hit the 404 catchall at the end. You can do that with a redirect, as we previously did / as many tutorials recommend, but the better user experience is to show the 404 in the place it happens.

src/utils/routesHelper.ts
gileswells: I now see how this is a mistake with the current routes and it returning all paths to the sitemap rather than just the active ones / the ones enabled via .env configs. Sorting it out now.

src/containers/documentation/ApiPage.tsx
gileswells: That suggestion as written wouldn't work. The available `api.urlFragment` values aren't flag names. Also if I tried to use it at the end of the `useFlag()` method then the api is possibly undefined (impossible in this routing but TypeScript doesn't know that) causing it to fallback using an inline conditional which just makes this turn ugly IMO.

```ts
const apiEnabledFlag = useFlag([FLAG_API_ENABLED_PROPERTY])[api?.urlFragment ?? ''];
```

I did end up changing the variable name to clarify things though which I think helps nicely.

src/components/Routes.e2e.ts
```diff
@@ -0,0 +1,20 @@
...
+  it.each([
+    '/api-publishing/invalid',
+    '/explore/invalid',
+    '/explore/invalid/docs/quickstart',
+    '/onboarding/invalid',
+    '/release-notes/invalid',
+    '/support/invalid',
+  ])('should show the 404 page on %s', async (path: string) => {
+    await page.goto(`${puppeteerHost}${path}`, { waitUntil: 'networkidle0' });
+    const pageNotFound = await page.evaluate(() => document.querySelector('h1')?.innerHTML);
```
mjlumetta, line 14: i'd have a slight preference for going with the [`pptr-testing-library`](https://testing-library.com/docs/pptr-testing-library/intro#usage) query here. don't think it's a huge issue with such a small query, but i think it's good to be in the habit of using the library since it expresses things in a user-centered way.

if you're interested in current examples:

```
mikelumetta@Mike-Lumettas-MacBook-Pro alt-developer-portal$ grep -rns pptr-testing-library src/
src//containers/documentation/AuthorizationDocs.e2e.ts:1:import { getDocument, queries } from 'pptr-testing-library';
src//containers/News.e2e.ts:1:import { getDocument, queries } from 'pptr-testing-library';
src//containers/releaseNotes/ReleaseNotes.e2e.ts:1:import { getDocument, queries } from 'pptr-testing-library';
src//containers/releaseNotes/CategoryReleaseNotes.e2e.ts:1:import { getDocument, queries } from 'pptr-testing-library';
src//App.e2e.ts:9:import { getDocument, queries } from 'pptr-testing-library';
```

this one is also up to you, no strong preference

src/Routes.tsx
```diff
@@ -1,15 +1,9 @@
+/* eslint-disable no-console */
```
mjlumetta, line 1: delete

src/Routes.tsx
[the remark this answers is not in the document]
mjlumetta: thanks, that makes sense. we can cross that bridge when we come to it

src/containers/documentation/ApiNotFoundPage.tsx
mjlumetta: oh, i agree we shouldn't use `Flag` and `filter()` together. but there's already a single flag for it, so i think we can just get rid of the filter function if we want? i'd prefer to use the flag because it's sort of a canonical way of figuring out what is and isn't available in React components, but this is okay if you prefer it. (also, i think using the flag would have been better before when it was just deactivated, too.)

[example](https://github.com/department-of-veterans-affairs/developer-portal/blob/master/src/containers/documentation/CategoryPage.tsx#L30)

src/containers/documentation/ApiPage.tsx
mjlumetta: you could do this if you wanted:

```
const api = getApi(params.apiName);
const category = lookupApiCategory(params.apiCategoryKey);
// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
const apiEnabledFlag = useFlag([FLAG_API_ENABLED_PROPERTY, api!.urlFragment]);
```

but given your point, i'm fine with the current approach if that's your preference. 

also, did you push the variable change commit? don't see it right now

src/Routes.tsx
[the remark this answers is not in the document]
mjlumetta: oh i was referring to the fragment use, not the rendering in the live site. that totally makes sense because of what you said, but i'm surprised that the sitemap generator worked for these routes given the trouble we ran into trying to use fragments earlier. maybe `react-router-sitemap` is doing its own thing there

mjlumetta (approved):
bunch of comments but they're all stylistic. i think this looks good overall and can be merged if you want to. there are some comments that i'd like to see resolved a bit more than others, but none of them are blocking.

src/containers/documentation/ApiPage.tsx
gileswells: Forgot to commit / push. Will do with other tweaks.

src/Routes.tsx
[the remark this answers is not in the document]
gileswells: I think the fragment issue was related to redirect or something else I recall. This shouldn't have been affected in my testing.

src/apiDefs/query.ts
```diff
@@ -14,11 +14,32 @@
  */
 
 import apiDefs, { apiCategoryOrder } from './data/categories';
+import { isApiDeactivated } from './deprecated';
+import { isHostedApiEnabled } from './env';
```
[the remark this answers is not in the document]
gileswells: I think @wwanamaker could create us a backlog ticket to do the refactor of this and personally I'd vote for option 3. Having a single condition to trigger inclusion or not is ideal to me.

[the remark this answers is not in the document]
wwanamaker: API-8933 in Jira!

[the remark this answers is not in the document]
mjlumetta: thanks to both of you! sounds good

src/containers/documentation/ApiNotFoundPage.tsx
```diff
@@ -28,7 +27,7 @@ const ApiNotFoundPage = (): JSX.Element => {
       {category?.name && <PageHeader header={category.name} />}
       <ul>
         {category?.apis
-          .filter((item: APIDescription) => !isApiDeactivated(item) && flags[item.urlFragment])
+          .filter((item: APIDescription) => flags[item.urlFragment])
```
mjlumetta, line 30: wait – we don't want to show deactivated APIs here, do we? here's what the Health category looks like, including Urgent Care:

Screen Shot 2021-07-28 at 2 47 52 PM

mjlumetta (commented):
one functional question about the `ApiNotFoundPage` component

src/containers/documentation/ApiNotFoundPage.tsx
gileswells: We don't. I got your previous comment backwards in my head thinking that just the flags would be enough to turn off the deactivated APIs for us. I've reverted this.

mjlumetta (approved):

src/containers/documentation/ApiNotFoundPage.tsx
mjlumetta: ohhhh. yes, the flag could be enough, but it's a different flag

PR #816, opened 2021-08-03
Proof of concept for headless CMS with Strapi

mjlumetta:
### Description

This is a proof of concept of using a headless CMS for developer portal content for [API-2344](https://vajira.max.gov/browse/API-2344). Specifically, it uses [Strapi](https://strapi.io/documentation/), an open source headless CMS. It could be any headless CMS, though; if we were actually going to implement a headless CMS for the developer portal, we would likely want a spike to explore a few other options.

#### Trying it out

1. Go to the [Strapi repo](https://github.com/mjlumetta/portal-content-strapi) and check it out locally.
2. In the Strapi repo, run `npm run develop` or `npm start`. Either command will allow you to edit content; the difference is `develop` will allow you to create new content types, whereas `start` won't allow you to edit content types. The server will start on `http://localhost:1337`. (1337 is the default port, and you shouldn't change that because it's hardcoded in this branch.)
3. Go to `http://localhost:1337` and create an admin account so you can edit content.
3. Make sure that the Strapi content is publicly accessible for read and write using the Users & Permissions Plugin, because I didn't figure out how to do authenticated API writes yet. In the Strapi admin portal, go to Settings (left nav) -> Users & Permissions Plugin (page nav) -> Roles. Select the Public role. Go to the Permissions section and expand the Application accordion. Under this accordion will be a list of content types (APIs, categories, FAQ page, Contact Us page, etc); click the "Select all" checkbox for every content type.
3. Check out this branch.
4. Run `node scripts/populateCMS.js` on this branch. This will put in a bunch of test data for all the content types. It mostly reflects the content of the current developer portal but includes some silly and/or nonsensical changes to demonstrate that the content is editable from the CMS. You can fix the weird things or make more stuff weird, whatever you prefer. _Note: doesn't include FAQ questions (the questions themselves) because I didn't figure out how to do relations in the content API yet._
5. Run this branch locally with `npm start`. You should see the same content that you can see in the Strapi admin portal.
6. Explore the Strapi admin portal and make whatever edits you want! Try it out and see what it's like. You can see the available content types in the site nav on the left under Collection Types and Single Types at the top.

#### Notes

- For the proof of concept, I converted some individual pages and some models to use Strapi as the content source of truth. Specifically, the pages are all of the Support pages – the overview, FAQ, and Contact Us – and the models are APIs + categories, so most stuff under Documentation and Release Notes. I probably didn't get everything but this should communicate the gist.
- When content is not loaded, a loading indicator (from VADS) displays.
- The POC includes examples of both using local React state and the Redux store for managing content as state. The single pages under Support use React state (because it's a single page, it is truly local), and the API and category models use Redux because they need to be available across several pages. We could use either or both if we were to actually implement this (although I think it's probably a good idea to use Redux for things that are shared across pages).
- APIs and categories have a `friendlyId` field in the Strapi schema to be correlated with the category key or API identifier (URL fragment). If we were doing this for real we'd find a slicker way of doing it (and also hopefully do so in a way that's generalizable to other Lighthouse platform apps, like the doc server or the platform backend). As is, that's why several APIs are commented out, because I didn't want to make fake data and it breaks without test data.
- A note on how Strapi works: as an open source solution, it's intended to be developed. You start a project with an app that is generated by Strapi and add content types either by using the content types plugin in the admin UI locally or by editing JSON + JS files in the Strapi project. The Strapi project is just an app with controllers and models, and when files in the source code are edited, if the server is running using `npm run develop`, it will restart with the new models and controllers. The content types plugin is just a UI that writes these source files for you when in development mode. I imagine that some of the paid solutions would have processes available on top of this to make changes easier. I'm going to try and give this topic and related questions – deployment, development/change management process, brief overview of paid solutions, general UX – a deeper look in the write-up because I think it deserves consideration. Also a good reason to look at other options for compare/contrast.

More coming in the write-up in Confluence, especially about deployment, development, UX, and other such considerations.

### Process

**all n/a8**

- [ ] Docs updated
  

brentschneider (commented):
I think I'm running into some config or install issues.
Can we pair on walking through the process next week?

PR #820, opened 2021-08-05
[API-6384] Update CollapsiblePanel to be va-accordion

gileswells:
### Description

https://vajira.max.gov/browse/API-6384

As CollapsiblePanel is deprecated this PR updates all instances of it to instead use the va-accordion that's coming from the web-components package that is being built.

### Process

- [ ] Docs updated
  
### Requested feedback

One question I have would be around the usage of `<va-accordion>` in the `RequestProductionAccess` and `WorkingWithOurAPIs` components. Should that be done as it since it is a singular accordion item per grouping or should we use the `GroupedAccordions` component like we do on the FAQ page? 
One consideration of using `GroupedAccordions` would be that we'd need to add an optional prop to disable the Expand/Collapse All link per accordion group (or remove it when the length of accordion items is just 1).


src/components/groupedAccordions/GroupedAccordions.test.tsx
```diff
@@ -26,24 +27,6 @@ describe('GroupedAccordions', () => {
     expect(heading).toBeInTheDocument();
   });
 
-  it('renders each accordion (measured by button triggers)', () => {
```
gileswells, line 29: The mindset with removing this set of tests is based on the mindset that `<va-accordion>` is tested within that package already so this feels redundant.

src/containers/consumerOnboarding/WorkingWithOurAPIs.test.tsx
```diff
@@ -27,10 +32,4 @@ describe('WorkingWithOurAPIs', () => {
     const heading = screen.getByRole('heading', { level: 2, name: headingText });
     expect(heading).toBeInTheDocument();
   });
-  it('renders all accordions in a closed state', () => {
```
gileswells, line 30: [Same as above](https://github.com/department-of-veterans-affairs/developer-portal/pull/820/files#diff-f17f932e7a51be3704434cf29388281fa6899345800952c0338bd686f5eb3df2L29), feels redundant.

src/containers/publishing/components/publishingExpecations/PublishingExpectations.test.tsx
```diff
@@ -21,13 +21,6 @@ describe('PublishingExpectations', () => {
     ).toBeInTheDocument();
   });
 
-  it('renders all accordions in a closed state', () => {
```
gileswells, line 24: [Same as above](https://github.com/department-of-veterans-affairs/developer-portal/pull/820/files#diff-f17f932e7a51be3704434cf29388281fa6899345800952c0338bd686f5eb3df2L29), feels redundant.

src/containers/publishing/components/publishingExpecations/PublishingExpectations.tsx
```diff
@@ -27,39 +26,53 @@ const PublishingExpectations = (): JSX.Element => (
     </p>
 
     <h2>Security</h2>
-    <CollapsiblePanel panelName="We follow industry best practices regarding security and privacy to ensure the data is safe and secure.">
-      <SecurityMarkdown authenticationUrl="/explore/authorization" />
-    </CollapsiblePanel>
+    <va-accordion>
```
gileswells, line 29: I am curious about the usage of `<va-accordion>` here. Do we want to put instances of `<va-accordion>` all throughout the codebase where needed or do we want to use `<CollapsiblePanel>` throughout and limit the accordion's tags in that component that we manage within this repo?

Also, as an asside, because of the predictable pattern of the content in this page I could absolutely see this being refactored into a loop with an object that holds the content.

alexandervalencia: I'm not sure about what you're asking in your first question? Isn't the point of the PR to replace the deprecated `CollapsiblePanel` everywhere?

alexandervalencia (commented):
Looks good, I just had two questions for you @gileswells

src/components/groupedAccordions/GroupedAccordions.test.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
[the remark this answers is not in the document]
gileswells: Yes...

gileswells:
@alexandervalencia The CollapsiblePanel is what's going away BUT we have a component that we use to wrap some instances of CollapsiblePanel called GroupedAccordions. The question is does anyone feel like it's a good idea to wrap all instances of `<va-accordion>` into the `GroupedAccordions` component so we have a single place to modify them all if need be. Personally I think I'm against the idea but it's a thought I had.

alexandervalencia:
I see what you were asking. No, I agree with your final assessment, I think it's a good idea to have more granular control over each instance than just using them in the `GroupedAccordions` component.

One final thing I noticed that we lost from using the `CollapsiblePanel` component is that previously the screen would auto-scroll to the top of the opened panel. Are we okay losing that functionality or does it need to be added back?

gileswells:
@alexandervalencia I didn't see any e2e tests for that so I figured it wasn't a critical feature. Also I didn't see where we were doing that on our end so I figured it was removed from the va-accordion when it was migrated to from CollapsiblePanel during the refactoring. Best to not re-add it over top in my opinion as the accordions now don't natively scroll to the top by design.

alexandervalencia (approved):
LGTM! 🚀

PR #821, opened 2021-08-09
API-8441 Internal API checkbox

ediiotero:
Story: [API-8441](https://vajira.max.gov/browse/API-8441)

### Description
This PR add an internal API (Address Validation) to the apply page.

### Process
The form requires a VA email, either at the top of the form or in the inputs below the internal API checkbox


src/containers/consumerOnboarding/components/sandbox/validateForm.ts
```diff
@@ -37,6 +43,23 @@ export const validateForm = (values: Values): FormikErrors<Values> => {
...
   }
 
+  if (anyInternalApisSelected(values.apis)) {
+    const vaEmailPattern = /^[A-Za-z0-9._%+-]+@va.gov$/;
+
+    errors.internalApiInfo = {
+      programName: validatePresence('program name', values.internalApiInfo.programName),
+      sponsorEmail: validateEmail(values.internalApiInfo.sponsorEmail),
+      // eslint-disable-next-line no-negated-condition
+      vaEmail: !vaEmailPattern.test(values.email) ? validateVAEmail(values.internalApiInfo.vaEmail) : '',
```
alexandervalencia, line 53: I actually prefer it as you have it, but I think this is a valid option to remove the eslint comment
```suggestion
      vaEmail: vaEmailPattern.test(values.email) ? '' : validateVAEmail(values.internalApiInfo.vaEmail),
```

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.test.tsx
```diff
@@ -173,6 +173,65 @@ describe('SandboxAccessForm', () => {
       expect(await screen.findByText('Enter an http or https URI.')).toBeInTheDocument();
     });
 
+    it('validates internal api fields when clicked', async () => {
```
alexandervalencia, line 176:
```suggestion
    it('validates internal api fields when selected', async () => {
```

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.test.tsx
```diff
@@ -173,6 +173,65 @@ describe('SandboxAccessForm', () => {
...
+        userEvent.click(screen.getByRole('checkbox', { name: /Terms of Service/ }));
+      });
+      userEvent.click(screen.getByRole('button', { name: 'Submit' }));
+
+      expect(await screen.findByText('Enter your program name.')).toBeInTheDocument();
+      expect(await screen.findByText('Enter a valid email address.')).toBeInTheDocument();
+      expect(await screen.findByText('Enter a valid VA-issued email address.')).toBeInTheDocument();
+    });
+
+    it('validates internal api fields when clicked and does not ask for VA email', async () => {
```
alexandervalencia, line 206: I think the description here is a little unclear what we're actually checking. Are we checking that if you put in a VA email address it validates the other fields but won't ask for a valid VA-issued address?

Could we somehow describe that a little more clearly, if that's the case?

alexandervalencia (commented):
A few questions for you, should be pretty minor to fix

src/containers/consumerOnboarding/components/sandbox/OAuthAppInfo.scss
```diff
@@ -1,5 +1,5 @@
-.oauth-uri-input {
-    #oAuthRedirectURIFormField {
-        max-width: 100%;
-    }
-}
+// .oauth-uri-input {
+//     #oAuthRedirectURIFormField {
+//         max-width: 100%;
+//     }
+// }
```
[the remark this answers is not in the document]
ediiotero: Oh! I totally meant to remove that

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
```diff
@@ -8,27 +8,45 @@ import { Flag } from '../../../../flags';
...
-          <CheckboxRadioField
-            type="checkbox"
-            name="apis"
-            label={api.name}
-            value={api.altID ?? api.urlFragment}
-          />
-        </Flag>
-      ))}
+      {hostedApis.map(api => {
+        const apiCheckboxName = api.altID ?? api.urlFragment;
```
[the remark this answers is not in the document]
ediiotero: I agree, This was just a TS issue I had to work around to make it happy.
Started [here](https://github.com/department-of-veterans-affairs/developer-portal/pull/791#pullrequestreview-697350591)

src/containers/consumerOnboarding/components/sandbox/validateForm.ts
ediiotero: tried it the other was and it prevents the form from submitting

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.test.tsx
ediiotero: done

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.test.tsx
ediiotero: done

alexandervalencia (dismissed):
Great work @ediiotero ! 🚀

alexandervalencia (approved):
♻️  🚀

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
```diff
@@ -8,38 +8,54 @@ import { Flag } from '../../../../flags';
...
+                        trustedPartnerOnly={api.trustedPartnerOnly}
+                        vaInternalOnly={api.vaInternalOnly}
+                      />
+                    </span>
+                  </>
+                }
+                value={apiCheckboxName}
+                className="vads-u-padding-left--1p5"
+              />
+              {internalApiSelected && <InternalOnlyInfo />}
```
mjlumetta, line 54: why is this part of the API checkbox list? and even if it is part of the list, does each internal only API need its own internal only info?

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
```diff
@@ -8,38 +8,54 @@ import { Flag } from '../../../../flags';
...
-        </Flag>
-      ))}
+      {hostedApis.map(api => {
+        const apiCheckboxName = api.altID ?? api.urlFragment;
+        const internalApiSelected =
+          checkedApis?.apis.includes(apiCheckboxName) && api.vaInternalOnly;
+        const internalApisBorderClass = internalApiSelected ? 'vads-u-border-left--4px' : '';
+        const internalApisBorderColorClass = internalApiSelected
+          ? 'vads-u-border-color--primary-alt-light'
+          : '';
```
mjlumetta, line 32: we've talked a bit about including these kinds of conditionals directly in `classNames` rather than ternaries recently (@gileswells and @alexandervalencia, i think?) for slightly more clarity. optional of course

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
```diff
@@ -64,6 +80,7 @@ const SelectedAPIs = ({ selectedApis }: SelectedApisProps): JSX.Element => {
     : 'vads-u-font-weight--normal';
 
   const oauthApisSelected = anyOAuthApisSelected(selectedApis);
+  const apisSelected = values as Values;
```
mjlumetta, line 83: can you get rid of this line by doing `useFormikContext<Values>()` on line 71?

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
```diff
@@ -8,38 +8,54 @@ import { Flag } from '../../../../flags';
...
 interface APICheckboxListProps {
   apiCheckboxes: APIDescription[];
+  checkedApis?: Values;
 }
 
-const ApiCheckboxList = ({ apiCheckboxes }: APICheckboxListProps): JSX.Element => {
-  // we will need to change this filter when we allow internal apis on the apply page
-  const hostedApis = apiCheckboxes.filter(api => !api.vaInternalOnly && !api.trustedPartnerOnly);
+const ApiCheckboxList = ({ apiCheckboxes, checkedApis }: APICheckboxListProps): JSX.Element => {
+  const hostedApis = apiCheckboxes.filter(api => api.altID);
```
mjlumetta, line 21: hey @ediiotero, I know this filter function isn't from this PR, but i wanted to check in about it because it's related. i don't think that `altID` should be treated as a meaningful value for determining what APIs are or are not on the sandbox form. i think we should treat it _only_ as an override to send to the server when the server and client identifiers for an API differ. this current logic treats it as a value that determines whether or not an API can actually be sent to the server. there are a few reasons why i think that's tricky and we should find another way to express that logic:

1. what happens when we need different values on different forms? the available APIs for the sandbox and production forms diverge because the production form has all internal only APIs, not just Address Validation. also, @keithosayande ran into problems updating these tests after adding `altID` to Address Validation without your changes from this branch, which is not the end of the world but i think also makes the tests a bit unnecessarily brittle.
2. i don't think it would be obvious to newcomers to the portal what the meaning of `altID` and why it's being filtered on here; it feels pretty opaque to me. it would require heavy comments to explain since it's a hack and not an intuitive pattern, and even then, it would be another thing to keep track of. i'm not sure that i would know what it meant if i hadn't seen the PR where you added the `altID`.
3. since having these two redundant identifiers is a hack, we're going to need to refactor at some point. the less that depends on `altID` at that point, the less work that refactor will be, so it could be beneficial to restrict the scope of that.

do you have thoughts about what the logic should be? i would think that for an API to be on this page, it would either have to be not internal only or Address Validation.

---

i also think that this kind of business logic should go in `SelectedApis`. `SelectedApis` is already responsible for determining where OAuth APIs and key auth APIs go, and i think it makes sense to keep the rest of the business logic (external APIs + Address Validation) there as well and keep `ApiCheckboxList` dumb (so that it can be used easily for any other future scenario).

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
```diff
@@ -8,38 +8,54 @@ import { Flag } from '../../../../flags';
 import { FLAG_HOSTED_APIS } from '../../../../types/constants';
 import { anyOAuthApisSelected } from './validateForm';
 import { OAuthAppInfo } from './OAuthAppInfo';
+import { InternalOnlyInfo } from './InternalOnlyInfo';
+import { Values } from './SandboxAccessForm';
 import './SelectedApis.scss';
 
 interface APICheckboxListProps {
   apiCheckboxes: APIDescription[];
+  checkedApis?: Values;
```
mjlumetta, line 17: can this be mandatory across the board? even though it won't apply to OAuth APIs right now, there could be OAuth APIs in the future that are internal only and need to be on this page. it would make the `APICheckboxList` more generic to make this required, imo.

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.test.tsx
```diff
@@ -23,9 +23,7 @@ const allOauthApis = getAllOauthApis()
 
 const allKeyAuthApis = getAllKeyAuthApis()
   .filter(api => api.altID)
```
mjlumetta, line 25: same comment as in the source code. @keithosayande has a version on his branch for #831 (not sure if he's pushed yet)

mjlumetta (commented):
some comments here, mainly about the use of `altID`, how that overlaps with #831, and how that will work out in the future

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
ediiotero: The designs [(figma)](https://www.figma.com/file/HuMj9rdAX9pIXSD2VLSu2o/Apply---Address-Validation?node-id=761%3A2285) show the input boxes below the internal api selected. At this point I'm not sure if each internal api would need its own set of input boxes. I was told we don't  expect a consumer to apply for more than one internal api at a time. I'm open to suggestions if you think this should go some place else.

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
ediiotero: ok I'll work on moving that

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.test.tsx
```diff
@@ -300,8 +353,9 @@ describe('SandboxAccessForm', () => {
   describe('SelectedApis', () => {
     describe('Standard APIs', () => {
       it.each(allKeyAuthApis)('toggles the %s checkbox on click', name => {
+        const checkboxName = RegExp(name, 'g');
```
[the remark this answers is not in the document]
ediiotero: I believe that is correct. Trying to run the tests again to confirm but running them seems to be pushing my machine to its limits lol

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
ediiotero: are you saying we should look for "not internal only" APIs or a list of approved internal APIs? That should work. I will make the changes

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.test.tsx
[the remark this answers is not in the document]
mjlumetta: can we use the same check for Open Data + Internal Only? whether that's the regex or the complete string. either is fine imo but would be good to use consistent checks

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
mjlumetta: I think this probably needs to be answered to go forward? that's a pretty significant discrepancy in behavior and functionality. if each internal only API has its own unique set of inputs, then I would expect this as you've designed it, but if not, this section shouldn't be in this component and definitely not in the loop. i don't know what the requirements say for that specifically, but intuitively internal only info seems like it would apply to the whole form, not an individual API. @brentschneider can you give a ruling here?

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
mjlumetta: yeah. the set of external/"not internal only" APIs + Address Validation (which can be hardcoded since it's an exception, or if you want to add another config value, that's okay too)

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
ediiotero: done

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
ediiotero: done. thanks

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
ediiotero: okay, I'll add the comment. We do have a future ticket to allow a user to select more than one internal api when the next internal api is added to the page.

mjlumetta (commented):
my biggest issue is that i can't submit the form – not sure if that's supposed to be part of this ticket or not, but it seems like this would be the place. other issues that are functional:

1. does spacing/layout work on both versions of the page?
2. mixing `altID` and `APPLY_FIELDS_TO_URL_FRAGMENTS` – this seems like a problem waiting to happen. i don't want to impose more changes on this PR, so i wouldn't say you have to go back and fix `anyOAuthApisSelected`, but `anyInternalApisSelected` should rely on `altID`. and then a follow-up ticket needs to be created to refactor `anyOAuthApisSelected` (or refactor it here if you feel up to it, that's fine too).

everything else is code cleanliness that i think could be beneficial but can be ignored or questions.

mjlumetta:
Screen Shot 2021-08-31 at 1 05 43 PM

bit wonky spacing, not sure if that needs to be fixed

src/containers/consumerOnboarding/components/sandbox/DeveloperInfo.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 import { TextField } from '../../../../components';
 
-const classNames = 'vads-u-margin-top--4 medium-screen:vads-l-col--10';
+const classNames = 'vads-u-margin-top--4 medium-screen:vads-l-col--10 large-screen:vads-l-col--8';
```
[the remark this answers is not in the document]
ediiotero: yes, this is limiting the size of the input boxes so they are more inline with the figma designs

src/containers/consumerOnboarding/components/sandbox/InternalOnlyInfo.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { TextField } from '../../../../components';
+import { Values } from './SandboxAccessForm';
+
+const InternalOnlyInfo = (): JSX.Element => {
+  const { errors, values } = useFormikContext();
+  const formValues = values as Values;
+  const vaEmailPattern = /^[A-Za-z0-9._%+-]+@va.gov$/;
+  const internalApiInfoName = 'internalApiInfo';
+  const shouldDisplayInputError = !!errors[internalApiInfoName];
+  const internalApiInfoClass = shouldDisplayInputError ? 'vads-u-margin-left--3' : '';
```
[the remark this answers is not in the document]
mjlumetta: so the inputs without errors are supposed to be left aligned? if that's what the design says, then all good, but want to confirm with @brentschneider

src/containers/consumerOnboarding/components/sandbox/validateForm.ts
```diff
@@ -2,18 +2,24 @@ import { FormikErrors } from 'formik';
...
+import { includesOAuthAPI, includesInternalOnlyAPI } from '../../../../apiDefs/query';
 import { Values } from './SandboxAccessForm';
 
 export const anyOAuthApisSelected = (apis: string[]): boolean => {
   const apiIdsByField = apis.flatMap(formField => APPLY_FIELDS_TO_URL_FRAGMENTS[formField]);
   return includesOAuthAPI(apiIdsByField);
 };
 
+export const anyInternalApisSelected = (apis: string[]): boolean => {
+  const apiIdsByField = apis.flatMap(formField => APPLY_FIELDS_TO_URL_FRAGMENTS[formField]);
```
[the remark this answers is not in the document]
ediiotero: sorry I'm a bit confused, I thought you wanted to move away from using altID? 
We should find time to talk about this

src/containers/consumerOnboarding/components/sandbox/InternalOnlyInfo.tsx
```diff
@@ -5,13 +5,11 @@ import { TextField } from '../../../../components';
...
 const InternalOnlyInfo = (): JSX.Element => {
-  const { errors, values } = useFormikContext();
-  const formValues = values as Values;
+  const { errors, values } = useFormikContext<Values>();
   const vaEmailPattern = /^[A-Za-z0-9._%+-]+@va.gov$/;
   const internalApiInfoName = 'internalApiInfo';
   const shouldDisplayInputError = !!errors[internalApiInfoName];
-  const internalApiInfoClass = shouldDisplayInputError ? 'vads-u-margin-left--3' : '';
-  const internalInputClassNames = 'vads-u-margin-top--2 large-screen:vads-l-col--8';
+  const internalApiInfoClass = shouldDisplayInputError ? 'vads-u-margin-left--0p5' : '';
```
mjlumetta, line 12: preference for using `classNames` for this logic, especially since this variable is consumed by `classNames`

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
```diff
@@ -2,22 +2,21 @@ import classNames from 'classnames';
...
 import './SelectedApis.scss';
 
 interface APICheckboxListProps {
   apiCheckboxes: APIDescription[];
-  checkedApis: Values;
 }
 
-const ApiCheckboxList = ({ apiCheckboxes, checkedApis }: APICheckboxListProps): JSX.Element => {
+const ApiCheckboxList = ({ apiCheckboxes }: APICheckboxListProps): JSX.Element => {
+  const formValues = useFormikContext<Values>().values;
```
mjlumetta, line 19: nice, very slick. love using context to cut out props like that (where appropriate)

mjlumetta (approved):
one style thing but that's not necessary to change. looks good ✨ ✅

src/containers/consumerOnboarding/components/sandbox/InternalOnlyInfo.tsx
ediiotero: going to make this change in [#835](https://github.com/department-of-veterans-affairs/developer-portal/pull/835)

PR #822, opened 2021-08-10
Fixed text in dialog buttons.

johanna-adhoc:
### Description
Fixes [API-9047](https://vajira.max.gov/browse/API-9047)

When you click the cancel button, can we have the “Yes Leave” option be “Yes, leave” and the other one be “No, stay on form” (with commas)

Screen Shot 2021-08-10 at 1 11 20 PM


gileswells (approved):
LGTM

PR #825, opened 2021-08-11
API-7391 Veteran Verification V1

dbrown-va:
### Description

https://vajira.max.gov/browse/API-7391

Addition of V1 Veteran Verification as well as the previous V0 reference by way of the Metadata Controller addition under https://github.com/department-of-veterans-affairs/vets-api/pull/7618.

V0 will be removed after consumers have been migrated to the V1 endpoint. 

Blocked until https://github.com/department-of-veterans-affairs/lighthouse-platform-tools/pull/136 gateway routes are deployed.

### Requested feedback

We are not familiar with the Docserver components, this may require the V1 docs endpoint first be uploaded(?) to Docserver and then the url PR'd here updated.


beaugrantham-va (approved):

dbrown-va:
Closing as a different PR addressed this by adding docserver metadata.

PR #826, opened 2021-08-11
API-9049/Fixes to the APICheckboxList

johanna-adhoc:
Fixes [API-9049](https://vajira.max.gov/browse/API-9049)

### Description
Adjust spacing between the list of APIs. - might be the tag messing up the spacing
Naming convention--should all have API after them

This ticket is also:
Creating the APICheckboxList as a reusable component.
Adding the API tags to the current API list in the apply page.

As part of the spacing fix margins and paddings were removed from the api tag components. Margins were added to improve positioning on CardLink component but introducing issues in the API CheckboxList. These margins were migrated to the CardLink Component instead.

Visual and Unit testing was updated.

### Process

### Requested feedback
any

Consumer Flag ON
Screen Shot 2021-08-13 at 9 14 03 AM
Screen Shot 2021-08-13 at 9 13 55 AM

Consumer Flag OFF
Screen Shot 2021-08-13 at 9 14 22 AM


brentschneider:
Looks good to me! @johanna-adhoc

src/components/apiCheckboxList/APICheckboxList.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import { APIDescription } from '../../apiDefs/schema';
+import { CheckboxRadioField, ApiTags } from '../index';
+import { Flag } from '../../flags';
+import { FLAG_HOSTED_APIS } from '../../types/constants';
+
+interface APICheckboxListProps {
+  apiCheckboxes: APIDescription[];
```
mjlumetta, line 8: slight preference for calling this `apis` (since the component takes APIs as input and generates the checkboxes as output)

src/components/apiTags/VAInternalOnlyTag.tsx
```diff
@@ -4,7 +4,7 @@ import * as React from 'react';
 import './ApiTags.scss';
 
 const VAInternalOnlyTag = (): JSX.Element => (
-  <div className={classNames('api-tags', 'vads-u-margin-bottom--1p5', 'vads-u-font-size--sm')}>
```
mjlumetta, line 7: since this change has been made basically three times, should we consider refactoring these components to share the same code? like make an `APITags` file and have it export all 3. but that is (a) not important in light of this deliverable and (b) super easy, so i have no inclination to say it should happen here, necessarily

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
```diff
@@ -24,7 +24,18 @@ const ApiCheckboxList = ({ apiCheckboxes }: APICheckboxListProps): JSX.Element =
...
-            label={api.name}
+            label={
+              <>
+                <span>{api.name}</span>
+                <span className="vads-u-display--inline-block vads-u-margin-left--1">
+                  <ApiTags
+                    openData={api.openData}
+                    trustedPartnerOnly={api.trustedPartnerOnly}
+                    vaInternalOnly={api.vaInternalOnly}
+                  />
```
mjlumetta, line 35: should the new `ApiCheckboxList` component replace this one entirely, or is that a later cleanup thing?

mjlumetta (dismissed):
couple of comments about what's in scope, but i would imagine we'll leave those for later. this is good to go ✨

src/components/apiCheckboxList/APICheckboxList.tsx
johanna-adhoc: Agree, I will make this change

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
johanna-adhoc: I didn't replace this code because there is some filtering happening to hide the internal APIs and because some rework of the entire API list is coming anyway I decided to make as little changes as possible.

src/components/apiTags/VAInternalOnlyTag.tsx
johanna-adhoc: I created a reusable component!

src/components/apiTags/ApiTag.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+        case tagTypes.VAInternalOnly:
+            label = 'Internal VA use only';
+            background = 'vads-u-background-color--gold';
+            break;
+        default:
+            label = 'Internal VA use only';
+            background = 'vads-u-background-color--gold';
+            break;
+    }
```
mjlumetta, line 39: i think my inclination would've been to create the label + background as props and replace the `type` prop with `label` and `background` in `ApiTags`, or possibly export each tag as its own component from here using the generic component (so each tag still has its own dedicated component with no props as before, but they're all in this file and all based on the same base component). but this is a trivial question and unlikely to matter so it should in no way block this PR.

src/components/apiTags/ApiTag.tsx
```diff
@@ -0,0 +1,57 @@
+import classNames from 'classnames';
```
mjlumetta, line 1: spacing in this file looks a little wonky but non-blocking obviously

mjlumetta (commented):
lgtm, good to go. it appears [this Github change](https://betterprogramming.pub/dealing-with-github-password-authentication-deprecation-1b59ced90065) might be breaking the review instance job, though? maybe we can override the checks in this case since (a) not a functional issue and (b) not the CI job itself

mjlumetta (approved):
oops meant to approve 🤦

src/components/apiTags/ApiTag.tsx
johanna-adhoc: Made a small improvement. I took the setup logic of the label and background to a "configuration" component so these contents get applied within the ApiTag component in a more clean way. 

I don't want to create extra wrappers for a configuration as simple as a label and a background color so I will pass up on keeping the other components (OpenDataTag, VAInternalTag, etc). Instead this comment reminded me to remove them!

PR #830, opened 2021-08-17
Temporarily disable list and loop component

mjlumetta:
### Description

completes [API-9263](https://vajira.max.gov/browse/API-9263).

this PR removes the list and loop component from the production access form until we're able to work out the tricky design details. it feature flags the component and replaces it with a single text input, since each field that uses list and loop can have only one value. since that's the case, we are treating the list and loop component as an enhancement that we don't need for the MVP and can iterate on later. the feature flag enables us to continue developing the list and loop component without having to do some crazy version control tracking. 

the affected fields are

- notification emails (2)
- sign up links (2)
- support/FAQ links (2)
- privacy/TOS documents (4)

changing the inputs for these fields means that their type in the form schema has to change as well. since we do have a feature flag here, it requires some manipulation to maintain both options at once. the schema changes apply to both the Typescript interface and the Yup schema validator. 

- the Yup schema is relatively simple to change; each field in the schema becomes a ternary depending on the environment variable. 
- for the Typescript side, because Typescript is a compiler and not part of the executable runtime, each of these fields now has to be typed as `string | string[]`. since the only time we have to directly set the values is for Formik's `initialValues` prop, we can calculate the initial values using the flag and then Formik will take care of everything.
 
this also includes some style fixes to the `TextField` component and the form as a whole as directed by @brentschneider.

### Process

- [ ] Docs updated
  

src/components/formField/TextField/TextField.tsx
```diff
@@ -40,7 +40,10 @@ const TextField: FC<TextFieldProps> = ({
   const containerClass = shouldDisplayErrors ? 'usa-input-error' : '';
   const labelClass = shouldDisplayErrors ? 'usa-input-error-label' : '';
   const validationClass = shouldDisplayErrors ? 'usa-input-error-message' : '';
-  const fieldClass = props.as === 'textarea' ? classNames('vads-u-margin-top--2p5') : '';
+  const fieldClass = classNames({
+    'vads-u-margin-top--1': props.as !== 'textarea',
+    'vads-u-margin-top--2p5': props.as === 'textarea',
+  });
```
alexandervalencia, line 46: nice update using the `classNames` more properly 👍

brentschneider (approved):
I'll leave the code for other engineers to review, but the description of the change sounds spot on.

alexandervalencia (commented):
Everything looks okay, I think I would prefer that the env variables were added to the various `.env` files as part of the PR.

src/components/formField/TextField/TextField.tsx
gileswells: We need to do this more often.

gileswells (dismissed):
LGTM

src/components/formField/TextField/TextField.tsx
mjlumetta: thanks yeah i agree... would rather us do this than the ternaries that are pretty common right now

gileswells (approved):
Seems that there's some vertical spacing that's changed from this last set of snapshot updates but so long as that's intentional or acceptable to you I'm good with it.

mjlumetta:
yeah @gileswells those changes were requested by @brentschneider

PR #833, opened 2021-08-17
[API-5748]: Scaling font-size

Cgfor3:
### Description

Add css to override USWDS font-size styling to the html element to allow content to change dynamically according to the browser font-size
Tested changes on Safari/Firefox/Chrome

### Process
- [ ] Docs updated
  
### Requested feedback
N/A


package-lock.json
```diff
@@ -9397,9 +9397,9 @@
           }
         },
         "tar": {
-          "version": "6.1.6",
-          "resolved": "https://registry.npmjs.org/tar/-/tar-6.1.6.tgz",
-          "integrity": "sha512-oaWyu5dQbHaYcyZCTfyPpC+VmI62/OM2RTUYavTk1MDr1cwW5Boi3baeYQKiZbY2uSQJGr+iMOzb/JFxLrft+g==",
+          "version": "6.1.8",
+          "resolved": "https://registry.npmjs.org/tar/-/tar-6.1.8.tgz",
+          "integrity": "sha512-sb9b0cp855NbkMJcskdSYA7b11Q8JsX4qe4pyUAfHp+Y6jBjJeek2ZVlwEfWayshEIwlIzXx0Fain3QG9JPm2A==",
```
TS000, line 9402: Is there a reason these upgrades were included in this PR?

TS000 (changes requested):
I think we accidentally included a package upgrade in this PR.

mjlumetta (commented):
looks great! way better now. i have two requested changes:

1. the one issue i can see is the side nav does not scale across the site. could we please put those sizes in rem? the site will be perfect on this front if that's done (unless there's some line height discrepancies i missed, maybe? but it'll be great).
2. unless the `package-lock.json` changes are meaningfully, can we not commit them please? not the end of the world if we do but usually ideal to avoid that churn

Cgfor3:
> 1. the one issue i can see is the side nav does not scale across the site. could we please put those sizes in rem? the site will be perfect on this front if that's done (unless there's some line height discrepancies i missed, maybe? but it'll be great).

What did you use to test this issue? I want to be sure I'm able to reproduce this

src/components/sideNav/SideNav.scss
```diff
@@ -38,3 +38,7 @@ a.va-api-secondary-skip-nav {
     background: $va-api-color-blue-dark;
   }
 }
+
+ul.usa-sidenav-list a {
+  font-size: 1.5rem;
+}
```
Cgfor3, line 44: Using `rem` here to scale based on the value of the `html` element's `font-size` which we have now set to scale depending on the browser's `font-size` setting.

The browser's default `font-size` is `16px` which will result in our `html` element to have a `font-size` of `10px`, which is where the value of `1.5rem` comes from since we want a default `font-size` of `15px` for the list item anchor elements.

mjlumetta (approved):
thanks @Cgfor3! great to have this done and support adjustable font sizing ✨

TS000 (approved):
LGTM 🍡

PR #837, opened 2021-08-18
Remove claims attributes from dev and staging

ediiotero:
### Description

Changing this to remove claims attributes from the apply page api list in staging and dev


gileswells (approved):
LGTM

PR #839, opened 2021-08-19
Focus first input on each prod access form step

mjlumetta:
### Description

completes [API-9119](https://vajira.max.gov/browse/API-9119).

sets focus on the first input of step 1 when the user navigates to step 1, whether by loading the page or by navigating back from step 2. this is an accessibility requirement for keyboard users and screen reader users. steps 2, 3, and 4 already had this behavior implemented.

I wanted to write integration tests but I'm having this obscure issue with Puppeteer. hoping to add later, tbd whether that will happen.

### Process

- [ ] Docs updated
  

johanna-adhoc (approved):
LGTM

gileswells (commented):
I'm not so sure about this concept overall. Your PR mentions it's an accessibility requirement yet `jsx-a11y/no-autofocus` from eslint specifically calls out not using autofocus. (I tried to change this out to use the `autoFocus` prop to clean up the code and got an accessibility warning). I get that you're not using the autofocus tag specifically but the effect is the same. 

Also this is only functional for me when I load the page directly. If I'm clicking from the previous page (/onboarding/request-prod-access) I don't see the focus being set.

mjlumetta:
@gileswells those are both great points. i will test out navigating from the previous page as well and try to fix that (if that's even the right call).

as for the a11y issue, you are definitely right in general, and i hadn't thought super carefully about the screen reader case. i think there's a possibility that this is an "it depends" kind of scenario. i pulled this from [the article](https://brucelawson.co.uk/2009/the-accessibility-of-html-5-autofocus/) linked from the `jsx-a11y/no-autofocus` page:

> I believe that autofocussing into a form is a usability win for most people if the purpose of the page is that form—for example, the Google hompage. So, I wouldn’t autofocus into the search box or comment field on this page because the primary purpose of this page is discussing this article which requires prior reading. Most people are lurkers, not commenters, so it’s against usability to autofocus into the comment form. But a contact form is the primary purpose of a contact page, and it saves users a keystroke to put the caret into the first field—particularly usable for mobile phone users.
> 
> However, screenreader users find themselves dumped in the middle of a form, potentially without context (athough it’s always possible to tell the screenreader to repeat the title of the page, and we all give our pages useful titles, don’t we?).

i think the description in the first paragraph does describe this form, actually. the entire content of this page is the form, so i don't think it's necessarily wrong to jump to the first input on the first input. the only front matter we'd skip is the h1 (not super meaningful if they knew where they were going) and the h2, which is more meaningful but perhaps not necessary, and both can be discovered by navigating by headings. however, since there isn't that much front matter to go past, then why bother focusing the first input on step 1? why not err on the side of not autofocusing for step 1? so i am sold on defaulting to what you said since that's going to be right in the majority of cases and safer.

if the ruling for step 1 is not to focus on that input, then that could make sense because you're coming to the _page_ first. for steps 2-4, because we haven't changed pages, we would still manage focus the same way by putting it on the first input, and same for navigating back from step 2 to step 1. we could also, alternatively, put the focus on the h2 that describes the step in those cases, if we don't want to go past that.

mjlumetta:
@gileswells confirmed that you're correct, in all cases. reworking so that it only does this focus handling when navigating back to step 1, not on page load.

gileswells (approved):
LGTM

PR #840, opened 2021-08-19
Set focus on first input with an error after validation for each step in the prod access form

mjlumetta:
### Description

completes [API-9175](https://vajira.max.gov/browse/API-9175).

sets focus to the first input with an error after the form validation runs for each step in the prod access form. the mechanism for doing this is a click handler for the submit button; once the button is clicked to submit the form, the handler runs. in the handler, we set a timeout that will find all elements with `aria-invalid="true"` and set the focus on the first one. that way, once the validation on submission happens, it will set focus to the right input if there are any errors.

it's probably worth noting that, with a multi-step form, it's possible that it could set focus inappropriately if any inputs on the next step were to have `aria-invalid="true"` already for some reason. but i think our setup for only validating the current step does prevent that, so we should be fine.

### Process

- [ ] Docs updated
  

dsteeve (approved):

dsteeve (approved):

johanna-adhoc (approved):
💯

johanna-adhoc (approved):

PR #843, opened 2021-08-24
Remove trustedPartnerOnly field from API definitions

mjlumetta:
### Description

completes [API-9465](https://vajira.max.gov/browse/API-9465).

Trusted partner only APIs were added as a concept to the developer portal in early 2020 (?) for the Loan Guaranty API, but it was never developed into something user-facing that was different from being an internal only API. Given that it had been in limbo so long and that it was causing hard-to-catch bugs in development, particularly for the prod access form, we want to refactor it out of the code until there is a clear business case so that developers, especially newcomers to this app, don't get confused.

### Process

- [ ] Docs updated
  

johanna-adhoc (approved):
LGTM

PR #851, opened 2021-09-01
API-7163 Non-existing APIs on signup

gileswells:
### Description

https://vajira.max.gov/browse/API-7163

This PR handles the front-end change to give better debugging information when trying to signup for an API that doesn't exist within the backend.

### Process

On `src/apiDefs/data/benefits.ts:14` change the `altID` to be something else like `claims-bad` and then attempt to submit the signup form on the apply page. You should see a clear list of errors presented while on dev.

- [ ] Docs updated
  
### Requested feedback

I have this set right now to only show up on development by tying in the `NODE_ENV` variable. Personally I don't think that this should be visible on production unless we wanted to tie it into the contact form by pre-populating the error message or something like that. (being able to pass a body field to the contact us page that pre-populates an error message might be a good idea for more streamlined support)


src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.tsx
```diff
@@ -47,12 +47,22 @@ interface SandboxAccessFormProps {
   onSuccess: (results: ApplySuccessResult) => void;
 }
 
+interface SandboxAccessFormError {
```
mjlumetta, line 50: should this relate to the `HttpErrorResponse` interface in `makeRequest` at all? they share a similar structure. wouldn't force it if not but worth considering if this is something we might be looking to standardize across most forms in the future.

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.tsx
```diff
@@ -169,13 +183,22 @@ const SandboxAccessForm: FC<SandboxAccessFormProps> = ({ onSuccess }) => {
...
                 Need assistance? Create an issue through our <Link to="/support">Support page</Link>
+                {submissionErrors.length > 0 ? (
+                  <ul>
+                    {submissionErrors.map((item: string) => (
+                      <li key={item}>{item}</li>
+                    ))}
+                  </ul>
+                ) : (
+                  ''
+                )}
```
mjlumetta, line 201: would generally prefer to avoid a JSX ternary if the falsy expression isn't rendering anything

```suggestion
                {submissionErrors.length > 0 && (
                  <ul>
                    {submissionErrors.map((item: string) => (
                      <li key={item}>{item}</li>
                    ))}
                  </ul>
                )}
```

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.tsx
```diff
@@ -86,7 +96,11 @@ const SandboxAccessForm: FC<SandboxAccessFormProps> = ({ onSuccess }) => {
         email: values.email,
       });
     } catch (error: unknown) {
-      setSubmissionError(true);
+      setSubmissionHasError(true);
+      if (process.env.NODE_ENV === 'development') {
+        const errors = (error as SandboxAccessFormError).body?.errors ?? [];
```
mjlumetta, line 101: i might add a comment reminding readers that this errors array should be present for 400 (or 4xx) error responses but not for 5xx responses, and the 5xx case is when the default array will be used in practice (and this code won't have any effect as a result)

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.tsx
```diff
@@ -86,7 +96,11 @@ const SandboxAccessForm: FC<SandboxAccessFormProps> = ({ onSuccess }) => {
         email: values.email,
       });
     } catch (error: unknown) {
-      setSubmissionError(true);
+      setSubmissionHasError(true);
+      if (process.env.NODE_ENV === 'development') {
```
mjlumetta, line 100: i have a no right or wrong answer question about this conditional, but wanted to ask – do you think it makes more sense to add the conditional where the state is updated or where the errors are rendered? i think there might be something to be said for keeping component state management consistent across environments, but that is also a totally negligible concern. could also use `Flag` for that if it were part of the rendering logic. but yeah, this is an incredibly minor concern, doesn't need to be addressed

mjlumetta (approved):
looks good, does what it says on the tin. tested locally and on the review instance. i left a few comments about clarity and implementation, but they're all very minor, so address them as you think best

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.tsx
gileswells: I agree.

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.tsx
gileswells: Changed this up to extend `HttpErrorResponse` to allow me access through to `body.errors`.

mjlumetta (dismissed):
looks good to go, will re-approve when linting errors are fixed

mjlumetta (approved):
good to go ✨

gileswells:
Found a way to trigger an uncaught exception. I'll reopen once resolved.

gileswells:
I was wrong and it was an issue on my local machine.

PR #852, opened 2021-09-01
Refactor of the step calculation in the prod access form by defining steps in the API data model

mjlumetta:
### Description

completes [API-9449](https://vajira.max.gov/browse/API-9449) and [API-9442](https://vajira.max.gov/browse/API-9449). i took the approach of refactoring the step calculation not to use hardcoded strings because i think that reduces risk of repeating these errors we ran into in development in the future.

- The `APIDescription` interface now includes an enum property called `prodAccessSteps` that defines what the last step should be for that API.
- `ProductionAccess` now checks `prodAccessSteps` rather than a hardcoded array of server-side API identifiers (`altID` or `urlFragment`, in other words) to determine the number of steps.
- The major benefit of this change is that developers are required to provide this value when adding a new API. If the step calculation is based on a hardcoded array of string ids, a developer adding an API needs to (a) know the prod access form exists in the first place, (b) know that the API's number of steps in the form needs to be accounted for as part of adding it, and (c) know the correct string id to add to that array. Using an enum in the API data model circumvents most of this risk.

also, exports `apisFor` from `query.ts` so that it can be consumed by components and not used as an internal utility for that module only.

### Process

- [ ] Docs updated
  

keithosayande (dismissed):
Looks good to go

johanna-adhoc (dismissed):
🏅

keithosayande (approved):
LGTM!

PR #863, opened 2021-09-10
Commented out the 508 Compliance Modal

keithosayande:
### Description
This ticket comments out the 508 Compliance Modal. UX and Marmoset are going to do more research on the 508 compliance question and modal. Until then, that code is commented out, so that whenever that is done, we can quickly make the changes. 

### Process

- [ ] Docs updated
  

mjlumetta (approved):
looks good, works as expected. so, since we're still asking the question, guessing the intent is to send it to the server in the post-prod changes and add it to the email to support?

PR #864, opened 2021-09-10
Remove code for list and loop component since it's unlikely to be reintroduced

mjlumetta:
### Description

the list and loop component is unlikely to be reintroduced per @brentschneider. as such, we are removing this code to reduce complexity. the ticket for reintroducing this component, should that happen, is [API-9122](https://vajira.max.gov/browse/API-9122); i will link the last commit with this feature-flagged code in the ticket so the person implementing it can have it as a reference.

### Process

- [ ] Docs updated
  

johanna-adhoc (approved):
Thanks for cleaning old flags 💯

PR #868, opened 2021-09-13
Updated content

keithosayande:
### Description
Updated the content of the 508 bullet point on the Request production access page.

### Process

- [ ] Docs updated
  

src/containers/consumerOnboarding/RequestProductionAccess.tsx
```diff
@@ -26,7 +26,7 @@ const RequestProductionAccess = (): JSX.Element => (
         API.
       </li>
       <li>
-        All apps and websites must be{' '}
+        We encourage all apps and websites to comply with{' '}
         <a href="https://section508.gov/" target="_blank" rel="noreferrer">
           Section 508
         </a>{' '}
```
mjlumetta, line 32: gotta delete the "compliant" on the line below, but that's it

mjlumetta (changes requested):
currently reads "We encourage all apps and websites to comply with Section 508 compliant", so need to delete that last word. also, did you want to add the env variable for consumer docs? i added it to my PR, but i imagine yours will get there faster

src/containers/consumerOnboarding/RequestProductionAccess.tsx
keithosayande: Done.

mjlumetta (approved):
lgtm ✨ 

do you know why there are only 2 screenshots that got updated?

keithosayande:
> lgtm ✨
> 
> do you know why there are only 2 screenshots that got updated?

This PR on changed 1 thing?

mjlumetta:
the screenshot thing is because the screenshot is getting taken twice right now (old URL + new URL), and each screenshot is taken at 3 viewports (mobile, tablet, and desktop, roughly). so i would expect 6 files to change rather than 2, but there could be some reason for that that i'm not aware of.

mjlumetta (approved):

PR #877, opened 2021-09-20
Api 9710

stiehlrod:
### Description
[APi-9710](https://vajira.max.gov/browse/API-9710)
After researching what causes the screen reader to pause, I decided to add a dash before each label that blends in with the background.

If you're making non-content updates make sure you have read the development guide: https://github.com/department-of-veterans-affairs/developer-portal/blob/master/docs/development.md

### Process
- [ ] Tests added or updated for change  n/a
- [ ] Docs updated -  n/a
- [ ] If you have made structural changes to the site, the [Developer Portal Content Types] -  n/a (https://community.max.gov/x/qUwOg) docs have been updated
- [x] If you have made/confirmed any decisions about user interactions and accessibility, the [Accessible Component Design](https://community.max.gov/x/xS8mgg) docs have been updated

### Requested feedback


src/components/apiTags/ApiTag.tsx
```diff
@@ -28,18 +29,27 @@ const apiTagConfig: APITagConfiguration[] = [];
...
   label: 'Internal VA use only',
+  screenReaderLabel: ' - ',
 };
 
 /**
  *  APITag Component
  */
 const ApiTag = ({ type }: ApiTagProps): JSX.Element => (
   <div className={classNames('api-tags', 'vads-u-font-size--sm')}>
+    <span
```
mjlumetta, line 46: i think there may be a slightly simpler way to solve this. instead of creating a new span, we can prepend this string to the text of `apiTagConfig[type].label`. 

i think it might also be possible to just use a space instead of a dash? that would probably work, although maybe if some screen reader/browser combos are smushing them together across elements, it might collapse that space as well. can try it out if you think that's worthwhile.

src/components/apiTags/ApiTag.tsx
```diff
@@ -28,18 +29,27 @@ const apiTagConfig: APITagConfiguration[] = [];
 apiTagConfig[tagTypes.OpenData] = {
   background: 'vads-u-background-color--primary-alt-light',
   label: 'Open Data',
+  screenReaderLabel: ' - ',
 };
 
 apiTagConfig[tagTypes.VAInternalOnly] = {
   background: 'vads-u-background-color--gold',
   label: 'Internal VA use only',
+  screenReaderLabel: ' - ',
```
mjlumetta, line 38: i don't think this necessarily needs to be part of the config since it's the same for each tag type, and i would anticipate being the same for any future tags

mjlumetta (commented):
this does a good job solving the functional problem for VoiceOver/Chrome, so all my comments are style related. the main theme is that i think it might be possible to simplify a bit further.

in addition to those thoughts, i wanted to ask what are your thoughts on solving this in `ApiTag` versus where it's used in the sandbox and prod access forms? i could see merits to either side, so curious what your perspective would be. the files for each are

- sandbox: `src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx`
- prod: `src/components/apiCheckboxList/APICheckboxList.tsx`. basically this one is newer and ideally we'd let this generic component be shared across both, but that refactor hasn't been done yet (plus depends on the future of the portal)

anyway, would love to hear your thoughts!

src/components/apiTags/ApiTag.tsx
```diff
@@ -37,9 +37,14 @@ apiTagConfig[tagTypes.VAInternalOnly] = {
 
 /**
  *  APITag Component
+ *  added a span with a dash to slow down the screen reader so the checkbox text does not run into the label
```
mjlumetta, line 40: i would suggest a couple things about this comment:

1. move to right above the span on line 44 (see below for syntax)
2. add more detail to the comment. this issue only occurs on VoiceOver/Chrome that we've seen, but could be on other platforms, so that's worth noting. also sharing the reasoning for a dash over a space, why that's more reliable, and if there are any links you found when you were looking it up that explain it really clearly, that would be helpful too. and maybe explicitly saying it's a hack.

syntax of JSX comment:

```
{/* ... */}
{/*
    ...
*/}
<span>...</span>
```

src/components/apiTags/ApiTag.tsx
```diff
@@ -37,9 +37,14 @@ apiTagConfig[tagTypes.VAInternalOnly] = {
 
 /**
  *  APITag Component
+ *  added a span with a dash to slow down the screen reader so the checkbox text does not run into the label 
  */
 const ApiTag = ({ type }: ApiTagProps): JSX.Element => (
   <div className={classNames('api-tags', 'vads-u-font-size--sm')}>
+    <span style={{ left: '-10000px', position: 'absolute'}}
```
mjlumetta, line 44: it's usually a good rule of thumb not to apply styles directly on the element using the `style` attribute. i think the way we approached trying it out in the browser might have been misleading – that's my bad. that's good for trying out style changes without reloading the whole app and starting over, but for writing the code to commit, we'd want to use CSS (SCSS technically). the `ApiTags.scss` file in this directory would be a good place for that code.

mjlumetta (commented):
couple of comments to go, but this is looking good. also looks like the CI job is failing at the lint job, mostly about aligning spacing in the JSX, so will need to fix those. you can run `npm run lint` to see the errors, and if you have questions about fixing them, lmk.

mjlumetta (approved):
great job Jennica ✨

PR #881, opened 2021-09-23
Update CI failure slackwebhook

duganth-va:
### Description
When CI jobs fail outside of the test a notification is sent to Slack to see if there is a problem with CodeBuild. These notifications were using the wrong webhook and being sent to DSVA workspace. 
This PR also updates the message with codebuild ID for easier identification.

### DIFF
--- a/cicd/buildspec-ci.yml
+++ b/cicd/buildspec-ci.yml
@@ -151,7 +151,7 @@ phases:
           ;;
         *)
           comment="An unhandled error occured with the build process. This is most likely a failure with the build pipeline. A post has been made in Slack"
-          slackpost.sh "CI Pipeline failure for the developer portal. CodeBuild URL: ${CODEBUILD_BUILD_URL}"
+          slackpost.sh "CI Pipeline failure for the developer portal. <${CODEBUILD_BUILD_URL}|CodeBuild URL>"
           ;;
         esac
       # Output build times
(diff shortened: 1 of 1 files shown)


cicd/buildspec-ci.yml
```diff
@@ -149,7 +149,7 @@ phases:
           ;;
         *)
           comment="An unhandled error occured with the build process. This is most likely a failure with the build pipeline. A post has been made in Slack"
-          slackpost.sh "Pipeline failure for the developer portal"
+          slackpost.sh "CI Pipeline failure for the developer portal. CodeBuild Job: ${CODEBUILD_BUILD_ID}"
```
gileswells, line 152: Can we also provide a link to the build logs? If we have the job ID and job name it looks like this should be doable. This is the current build's url: `https://us-gov-west-1.console.amazonaws-us-gov.com/codesuite/codebuild/008577686731/projects/developer-portal-ci/build/developer-portal-ci%3Aefeaa8b5-6e2f-4948-aaa8-7f8198c4a19d?region=us-gov-west-1`

gileswells (changes requested):
I'd like to see this include a link rather than just show a build ID. Makes it more usable for AWS console users as well as AWS CLI users.

cicd/buildspec-ci.yml
```diff
@@ -41,7 +41,7 @@ env:
     DOCKER_USERNAME: '/dvp/devops/DOCKER_USERNAME'
     DOCKER_PASSWORD: '/dvp/devops/DOCKER_PASSWORD'
     # This webhook currently targets DSVA workspace #lighthous-deploys channel
```
gileswells, line 43: This comment should also be updated to reflect where it's being sent to now.

gileswells (changes requested):

duganth-va:
https://lighthouseva.slack.com/archives/C01CP09F04Q/p1632409507049300

gileswells (approved):
LGTM

PR #884, opened 2021-09-27
API-10193 Veteran Verification API v1.0.0 release note

dbrown-va:
### Description
https://vajira.max.gov/browse/API-10193

Adds Veteran Verification API release note for v1.0.0 release and GRAS history fix.

### Process

v1.0.0 is already enabled through the API Gateway kong and documentation set to the default version on the portal.

The v0.0.0 needs to be confirmed with the Support team so that proper notification is given to current Sandbox users.  The current Production consumer is actively testing v1.0.0.


gileswells (approved):
LGTM

SlakrHakr (approved):

PR #888, opened 2021-10-01
API-10382 - Implement auth doc updates

mheusgit:
### Description
Implements: API-10382 Implement auth doc updates for Clinical Health API

PR is three content changes located on the authorization page:
Add "launch" row with text as noted, including link to Scopes section
Add "launch" row with text as noted, including link to Requesting Authorization section
Add "fhirUser" row with text as noted, including link to fhirUserScope section


src/components/oauthDocs/AuthCodeFlowContent.tsx
```diff
@@ -179,6 +179,37 @@ https://sandbox-api.va.gov${baseAuthPath}/authorization?
               </p>
             </td>
           </tr>
+          <tr>
+            <td>
+              <code>launch</code>
+            </td>
```
christineyost, line 185: From Darcy's document it sounds like this row should only show when Clinical Health API is the selected

christineyost (commented):
Looks good to me! The only thing is I think the launch parameter in the first table is only supposed to show if the Clinical Health API is selected

src/components/oauthDocs/AuthCodeFlowContent.tsx
mheusgit: Its been corrected

christineyost (approved):
Looks good to me!

PR #889, opened 2021-10-04
Removed consumer docs flags. Replaced /apply and /go-live urls with their new urls

keithosayande:
### Description
This PR cleans us the codebase now that consumer docs are in production and are stable. The consumer docs feature flags have been removed from the codebase. /apply and /go-live links have been replaced in the codebase with their consumer docs equivalents. All relevant tests have been updated.

### Process

- [ ] Docs updated
  
### DIFF
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -5,17 +5,14 @@ import { Redirect, Route } from 'react-router-dom';
 import { getActiveApiDefinitions, getApiCategoryOrder } from './apiDefs/query';
 import { MarkdownPage } from './components';
 import ConsumerOnboardingRoot from './containers/consumerOnboarding/ConsumerOnboardingRoot';
-import DisabledApplyForm from './containers/DisabledApplyForm';
 import DocumentationRoot from './containers/documentation/DocumentationRoot';
 import Home from './containers/Home';
 import News from './containers/News';
 import ErrorPage from './containers/ErrorPage';
 import ReleaseNotes from './containers/releaseNotes/ReleaseNotes';
 import Support, { sections as supportSections, SupportSection } from './containers/support/Support';
-import PathToProduction from './content/goLive.mdx';
 import TermsOfService from './content/termsOfService.mdx';
 import ProviderIntegrationGuide from './content/providers/integrationGuide.mdx';
-import { Flag, getFlags } from './flags';
 import { Publishing } from './containers/publishing';
 import {
   CONSUMER_APPLICATION_PATH,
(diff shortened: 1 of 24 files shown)


src/Routes.tsx
```diff
@@ -92,32 +86,16 @@ export const SiteRoutes: React.FunctionComponent = (): JSX.Element => {
...
-              fallbackComponent={DisabledApplyForm}
-            />
-          )}
-        />
-      )}
 
-      {flags.consumer_docs ? (
-        <Redirect from="/go-live" to={CONSUMER_PROD_PATH} />
-      ) : <Route path="/go-live" render={(): JSX.Element => MarkdownPage(PathToProduction)} />}
+      <Route path={CONSUMER_APPLICATION_PATH} component={ProductionAccess} />
```
gileswells, line 90: Is there a specific reason this route isn't included in the `CONSUMER_ROUTER_PATHS` variable and then listed in the map below?

src/components/header/Header.tsx
```diff
@@ -113,11 +114,14 @@ const Header = (): JSX.Element => {
...
               <section aria-label="Network issue alert">
                 {/* message written for specific issue on 8/25/21, update next time it needs to be used */}
-                There were recent network issues affecting all VA sites and usage of VA Lighthouse APIs in
-                sandbox and production environments from 9:56 am to 10:44 am EDT. Please review the&nbsp;
-                <a href="https://valighthouse.statuspage.io/" target="_blank" rel="noreferrer">Status Page</a>
-                &nbsp;for details and reach out to <Link to="/support/contact-us">Support</Link> if you
-                have any questions.
+                There were recent network issues affecting all VA sites and usage of VA Lighthouse
+                APIs in sandbox and production environments from 9:56 am to 10:44 am EDT. Please
+                review the&nbsp;
```
gileswells, line 119: This text won't likely ever get seen but the pattern we've used elsewhere was to end the line with `{' '}` to enforce the space between the text and the following inline tag.

src/e2eHelpers.ts
```diff
@@ -17,9 +17,7 @@ import { mockSwagger as mocks } from './__mocks__/mockSwagger';
 // Paths to test in visual regression and accessibility tests
 export const testPaths = [
   '/',
-  '/apply',
```
gileswells, line 20: The related visual regressions need to be deleted as well. The redirect makes it so that the screenshot for this and for `/go-live` will live on in the repo till someone manually resolves it. In my review the full list of screenshots to be removed are:

```
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-apply-properly-1-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-apply-properly-2-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-apply-properly-3-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-go-live-properly-1-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-go-live-properly-2-snap.png
test/image_snapshots/visual-regression-test-ts-visual-regression-test-renders-go-live-properly-3-snap.png
```

gileswells (changes requested):
One question and then a couple smaller bits.

For the visual regression deletions you could, if you choose, delete them all and get a fresh set since there are currently about 20 with very minor modifications that aren't large enough to trigger a failing visual regression.

src/Routes.tsx
keithosayande: The prod access form isn't a child of the `ConsumerOnboardingRoot` so if I remove this line, a 404 error is thrown. Adding the prod access form to the `ConsumerOnboardingRoot` would make the form render the sidebar nav and we don't want that.

gileswells (approved):
My only question is whether or not deleting the .keep file would cause any issues but if you've seen no adverse side effects from that then I'm good with this.

PR #900, opened 2021-11-01
API-9848 - GitHub Action PR builds on main fork

gileswells:
### Description
https://vajira.max.gov/browse/API-9848

This duplicates the PR review build jobs from a chained job (`buildspec-review-deploy.yml / developer-portal-review-deploy`) that fires after CodeBuild finishes the tests in `buildspec-ci.yml / developer-portal-ci` into it's own process that runs parallel to the npm tests through GitHub Actions.

### Process

- [ ] Docs updated
  
### Requested feedback

Currently this will run in addition to the existing CodeBuild jobs so that we can observe a side-by-side comparison to make sure any differences are spotted.


FonzMP (dismissed):
I'm good with all of this. :shipit:

imcgonigle (dismissed):

FonzMP (approved):
LGTM 🚀

src/styles/swagger.scss
```diff
@@ -1,5 +1,63 @@
...
 
+/* cyrillic-ext */
+@font-face {
+  font-family: 'Source Code Pro';
+  font-style: normal;
+  font-weight: 400;
+  src: url(https://fonts.gstatic.com/s/sourcecodepro/v14/HI_SiYsKILxRpg3hIP6sJ7fM7PqlMOvWnsUnxlC9.woff2)
+    format('woff2');
+  unicode-range: U+0460-052F, U+1C80-1C88, U+20B4, U+2DE0-2DFF, U+A640-A69F, U+FE2E-FE2F;
+}
```
SlakrHakr, line 11: im assuming these swagger stylesheet changes were required to get visual regression tests working in github actions? i guess im confused about why they were needed. How were these tests passing before?

SlakrHakr (approved):

src/styles/swagger.scss
gileswells: Before they were passing because across the board Source Code Pro wasn't installed across the different systems that run the tests and generate the visual regressions. This made it harder to debug as well because visually on developer's systems the images wouldn't match what our browsers were rendering either as we'd get it properly if we locally had Source Code Pro installed which, without any evidence, I think is a +50% bet for a developer's machine. For some reason even just installing the font into the docker box or onto the GitHub Actions runner machine wasn't doing the trick so this was the solution. (also +1 for leveraging the CDN version of the font for efficiency)

PR #901, opened 2021-11-02
Removing claims attributes from portal

ediiotero:
### Description
Story: [API-10674](https://vajira.max.gov/browse/API-10674)

Removing what is left of Claims Attributes from the portal for clean up


gileswells (changes requested):
Overall looks good to go but there are a couple items remaining in the `.env` files named `REACT_APP_CLAIMS_ATTRIBUTES_API_ENABLED` that can be removed.

gileswells (approved):
LGTM

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.test.tsx
```diff
@@ -395,7 +395,7 @@ describe('SandboxAccessForm', () => {
 
   describe('SelectedApis', () => {
     describe('Standard APIs', () => {
-      const filteredKeyAuthApis = allKeyAuthApis.filter(api => api !== /Claims Attributes API/);
+      const filteredKeyAuthApis = allKeyAuthApis.filter(api => api !== /Benefits Intake API/);
```
dsteeve, line 398: not sure if this is what we want to do.  Was Claims Attributes being filtered out before because it was there in the allKeyAuthApis (and not really active) and now we don't need to filter out anything and can just use "allKeyAuthApis" instead?   not sure though, so am asking the question.

mjlumetta: @dsteeve i'm 99% sure you're correct here. and if not, why would Benefits Intake be filtered out? it's in the form

src/types/constants/index.ts
```diff
@@ -15,7 +15,6 @@ export const PRODUCTION_ACCESS_URL = `${BACKEND_BASE_URL}/production_request`;
 export const CONTACT_US_URL = `${BACKEND_BASE_URL}/contact-us`;
 
 export const APPLY_STANDARD_APIS = [
```
mjlumetta, line 17: seems like this array could probably be removed and computed instead, although up to you whether that's a refactor you think is worth doing now

src/containers/consumerOnboarding/components/sandbox/SandboxAccessForm.test.tsx
ediiotero: good catch, I removed it

dsteeve (approved):
lgtm!

PR #916, opened 2021-12-02
API-11422: Modal: Are you US Based - Fixes -

SlakrHakr:
### Description
https://vajira.max.gov/browse/API-11422

Made small changes to the "non-US based company" modal:
image (3)

### Process


src/containers/consumerOnboarding/ProductionAccess.e2e.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const doc = await getDocument(page);
+    const dialogHidden = await queries.queryByRole(doc, 'dialog');
+    expect(dialogHidden).toEqual(null);
+
+    await page.click('#isUSBasedCompanyFormFieldno');
+    await page.click('#is508CompliantFormFieldyes');
+    await page.click('#apisFormFieldappeals');
+    await page.click('#termsOfServiceFormField');
+    await page.click('button.usa-button[type=submit]');
```
imcgonigle, line 18: Is this the same for all three tests? If so it might make sense to add it to a before each block instead of repeating it in every test.

imcgonigle (approved):
I added one question/suggestion but overall it looks good to me

src/containers/consumerOnboarding/ProductionAccess.e2e.ts
SlakrHakr: the beginning isnt exactly the same in the third test.
this middle section could be extracted into a method i suppose, i just prefer is a bit more this way (considering a before each block cant work cleanly here), more explicit this way.
that being said, if you (or anyone else) feels strongly about consolidating i can make that happen

imcgonigle: 👍 I figured there was a reason it was done this way

gileswells (approved):

PR #917, opened 2021-12-07
Add PGD to CCG Auth Docs

christineyost:
### Description

Add PGD to CCG Auth Docs as described in [API-10868](https://vajira.max.gov/browse/API-10868). To accomplish this I broke up `AuthInfo` into `acgInfo` and `ccgInfo`. APIs like the Veterans' Health API can use both ACG and CCG for authentication. When that is the case, the base URLs and scopes will be different, so it seems best to have two separate objects describing these values.

### Process

- [ ] Docs updated
  

mheusgit (approved):
Looks good.  From reviewing the changes it appears a ticket to move API configurations and associated security settings away from Dev Portal into a DB is past due.

christineyost:
> Looks good. From reviewing the changes it appears a ticket to move API configurations and associated security settings away from Dev Portal into a DB is past due.

Yes, definitely agree it would be nice to store this data elsewhere! Since we don't have anything set up for that currently though, just adding it in here

dsteeve (approved):
looks good!

PR #923, opened 2021-12-15
API-11749: [COLOR CONTRAST]: Fix color contrast for deprecated endpoints on dev portal - P4

SlakrHakr:
### Description
https://vajira.max.gov/browse/API-11749

This change darkens the endpoint method text for deprecated endpoints for more contrast against the light gray container.

image


brentschneider:
Hi Lee, 
I rechecked the a11y contrast applied to the staging link ...  we are just short (3.13) to passing AA contrast 4.5:1 
_It's that muting that's applied to the page which is throwing me off from knowing exactly which value to use._   

On the button text color value, can we try the next darker color,  `$color-gray` to see if it passes?

SlakrHakr:
@brentschneider
$color-gray seems to have improved its score but it wasn't enough: 3.27 is the new one with $color-gray
$color-gray-dark wasn't good enough either: 4

$color-black was enough to appease axe, this is what that looks like:
image

brentschneider:
Sweet deal, thanks for iterating forward on the color spectrum.
@SlakrHakr approved from a UX perspective.

gileswells (approved):

PR #929, opened 2022-01-05
API-12187  Developer Portal visual regression tests - enforced isolation

mheusgit:
### Description
PR adds a call to jestPuppeteer.resetBrowser() to isolate each visual regression test.  Previously, bad browser state was breaking later test for some developer machines.

### Process

- [x] Tests added or updated for change:  Now that the browser is not reused between test, visual regression testing should run longer but be more consistent for everyone.


christineyost (approved):
Looks good to me!

PR #943, opened 2022-01-17
API-000 Removed invalid scopes from Claims CCG docs

dbrown-va:
### Description

Removed invalid scopes from Claims CCG docs.  Since Client Credentials Grant flow does not have an attended user, `openid`, `profile`, and `offline_access` are not applicable.


dsteeve:
This can be closed.  we did this work under our ticket number and is currently in dev and staging. 
https://github.com/department-of-veterans-affairs/developer-portal/pull/945

PR #945, opened 2022-01-18
Api 12694 fix scopes for benefits claims ccg flow

stevelong00:
### Description

Removed invalid scopes from Claims CCG docs. Since Client Credentials Grant flow does not have an attended user, openid, profile, and offline_access are not applicable.

### Process

- [ ] Updated unit tests and visual regression tests accordingly


dsteeve (approved):
looks good to me!

mheusgit (approved):
Looks good

PR #946, opened 2022-01-19
sorts all apis list on the production apply page

SlakrHakr:
### Description
This change ensures the list of APIs on the production apply page is sorted alphabetically (as it is on the sandbox apply page).


dsteeve (approved):
looks good

PR #965, opened 2022-02-02
API-12252: Add Benefits Reference Data to DeveloperPortal in PROD

SlakrHakr:
### Description
Removes feature flag to enable new benefits reference data api in production.


gileswells (approved):
#thatwaseasy
Approving for now but I wonder if visuals will overlap on this one and flag it.

PR #973, opened 2022-02-11
API-12233: Update LGY Loan Guaranty on Portal to use OAS from Docserver - DEV STAGING PROD

SlakrHakr:
### Description
https://vajira.max.gov/browse/API-12233

LGY API is now on docserver. This just updates the portal to point at the docserver url instead of directly at the api.

### DIFF
--- a/src/apiDefs/data/loanGuaranty.ts
+++ b/src/apiDefs/data/loanGuaranty.ts
@@ -26,7 +26,7 @@ const loanGuarantyApis: APIDescription[] = [
     description: 'Use the Loan Guaranty API to Manage VA Home Loans.',
     docSources: [
       {
-        openApiUrl: `${OPEN_API_SPEC_HOST}/services/loan_guaranty/docs/v1/api`,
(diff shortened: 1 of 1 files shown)


gileswells (approved):

src/apiDefs/data/loanGuaranty.ts
```diff
@@ -26,7 +26,7 @@ const loanGuarantyApis: APIDescription[] = [
     description: 'Use the Loan Guaranty API to Manage VA Home Loans.',
     docSources: [
       {
-        openApiUrl: `${OPEN_API_SPEC_HOST}/services/loan_guaranty/docs/v1/api`,
+        openApiUrl: `${OPEN_API_SPEC_HOST}/internal/docs/loan_guaranty_property/v1/openapi.json`,
```
christineyost, line 29: Looks good to me!

Just wondering if you want to add the metadata url in here too. That won't change the behavior for this API on the dev portal now, but if another version of this API is added another PR will have to be made if the metadata URL is not included now. If it is included now no additional PR would be needed to show a new version on the portal

PR #981, opened 2022-02-15
[API-13220] Facilities Versioning Banner

Cgfor3:
### Description

- Add va-alert banner to header and display only on facilities docs page
- Add props interface for va-alert component

### Process

- N/A

### Requested feedback

- N/A

### DIFF
--- a/src/components/header/Header.tsx
+++ b/src/components/header/Header.tsx
@@ -131,7 +131,8 @@ const Header = (): JSX.Element => {
           <va-alert status="info" visible>
             <h3 slot="headline">
               A new version of the VA Facilities API (v1) will launch in the first quarter of 2022.
-              We will add release notes when it&apos;s live.
+              We will add <Link to="/release-notes/facilities">release notes</Link> when it&apos;s
+              live.
             </h3>
           </va-alert>
         )}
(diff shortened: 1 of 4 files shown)


src/components/header/Header.tsx
```diff
@@ -127,6 +127,14 @@ const Header = (): JSX.Element => {
             className="vads-u-margin-top--0"
           />
         </Flag>
+        {location.pathname === '/explore/facilities/docs/facilities' && (
+          <va-alert status="info" visible>
+            <h3 slot="headline">
+              A new version of the VA Facilities API (v1) will launch in the first quarter of 2022.
+              We will add release notes when it&apos;s live.
```
gileswells, line 134: I'd like to see a link to the release notes page here. Without linking to where that would live I feel like developers might expect it to be included on this page.

gileswells (changes requested):
Small addition to the alert. Also please add the `impact: users` label to the PR so the MR automation will pick up this PR properly. (this is new and I hadn't yet made it more publicly known so you didn't miss an announcement)

gileswells (approved):
LGTM

gileswells (approved):

PR #998, opened 2022-03-08
API-13910: Add LGY Guaranty Remittance API to Sandbox Apply Page

SlakrHakr:
### Description
LGY Guaranty Admittance is not internal only. Configuring them to be external will add them to the sandbox apply page on the frontend.


gileswells (approved):
LGTM

PR #1016, opened 2022-04-07
API-15027: Delete Video from About Us Page

SlakrHakr:
### Description
Video is old and outdated and needs to be removed.


FonzMP (approved):
:shipit:

PR #1023, opened 2022-04-14
fix support "contact us" path

SlakrHakr:
### Description
Fixes issue with contact us path to LCMS.

Related PR regarding the same change to the rev-proxy:
https://github.com/department-of-veterans-affairs/devops/pull/11068


FonzMP (approved):
:shipit:

PR #1027, opened 2022-04-19
set cookie SameSite value to Strict

SlakrHakr:
### Description
I think this will fix the current issue in dev with the developer-portal talking to LCMS. (LCMS not being able to access the cookie generated and stored by the developer-portal)
It seems that the default "SameSite" value is `None` and if that is set and "Secure" isn't set, the cookie is "blocked"?

Strict should be what we want anyways. Should restrict it's usage to the specific dev-developer.va.gov domain.


FonzMP (approved):
:shipit:

PR #1030, opened 2022-04-20
Api 9122/update notification email field

imcgonigle:
### Description
[API-9122](https://vajira.max.gov/browse/API-9122)

This PR adds a multi-entry input so that teams can add multiple notification email addresses on the production access form.

### Process

- [ ] Docs updated
  

SlakrHakr:
Mobile view for the list of emails seems off? I would expect to still be able to see the email i entered I would think?
Screenshot 2022-04-28 at 15-42-44 VA API Platform Production access form

Why does "removing" an email not actually remove it from the view?
image
It seems to just gray out the box maybe? I tried it in both firefox and safari, got the same result.

SlakrHakr (changes requested):
Code looks good and makes sense.
Couple of concerns around the UI functionality is all.

SlakrHakr (approved):
Functionality works great!

PR #1051, opened 2022-05-06
set post_to_lpb flag to true in production

SlakrHakr:
### Description

LPB submissions from developer-portal have been thoroughly tested in dev and staging.
Moving forward with adding LPB to the flow of portal form submissions in a shadow capacity.


FonzMP (approved):
:shipit:

PR #1057, opened 2022-05-16
[SPIKE] API Filtering page rough POC

gileswells:
### Description

https://vajira.max.gov/browse/API-15779

This PR is simply a visual proof of concept on what an API filtering page might roughly look / function like coming out of the information architecture meetings with Okapi and UX about the future state of the developer portal as more and more APIs are added.

### Process

- [ ] Docs updated
  
### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8202,12 +8202,6 @@
       "integrity": "sha512-1APHAyr3+PCamwNw3bXCPp4HFLONZt/yIH0sZp0/469KWNTEy+qN5jQ3GVX6DMZ1UXAi34yVwtTeaG/HpBuuzw==",
       "dev": true
     },
-    "ansi-regex": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
-      "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-      "dev": true
-    },
     "ansi-styles": {
       "version": "3.2.1",
       "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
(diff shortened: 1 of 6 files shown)


FonzMP:
so just perusing the documentation for react-isotope, I can see the rigid structure, but can you expand on limitations you found? 
Also, is this the only package we're considering? 

I will say on my initial peruse, I'm not impressed by this one and there doesn't seem to be an extensive list, so I'm leaning on the side of custom implementation as well, but just curious to hear a bit more of your thoughts.

gileswells:
@FonzMP That's the only package that I've found to even come close to being able to do what we need. This is more of a rough POC so that UX can click through and see it functioning in this limited state but my suggestion is also heavily leaning toward custom implementation.

FonzMP:
@gileswells sounds good. i wasn't all that impressed by it and there didn't seem to be an extensive list of packages for it, so I'm on board with this. 👍

gileswells:
> wouldnt be crazy to implement i wouldnt think?

No and that's where I'm sitting. The functional logic of this package is about 60 lines of code at most. We can easily create our own logic (or if we wanted to go overboard our own releasable package) and use cypress for the testing and achieve full coverage quite easily.

brentschneider:
Great convo! 

@gileswells I appreciate the demo work and exploring what limited options are available as packages.

Aside from some accessibility improvements like showing the number of items, displaying the selected filters, and providing pagination, the SPIKE shows the filter concept works. ✅

Spot on, @SlakrHakr ...
> end of the day, this is a list of tags matched with filters

Along with the a11y considerations, I'm also thinking about at the [ability to switch between rows or cards](https://www.figma.com/proto/qFYotEst3ofQxANRDmAkXk/Developer-portal-2.0?page-id=1464%3A2526&node-id=1464%3A2740&viewport=820%2C742%2C0.81&scaling=min-zoom&starting-point-node-id=1464%3A2527).

Don't know if this idea changes your approach to the problem?

PR #1063, opened 2022-05-17
Add additional trigger for is_labeled check

gileswells:
### Description

This should resolve the issue where the `is_labeled` check will sometimes run before `actions/labeler@v3` in the [auto-label-content-prs.yml](https://github.com/department-of-veterans-affairs/developer-portal/blob/master/.github/workflows/auto-label-content-prs.yml) workflow. That would cause the `is_labeled` check to fail because it ran too early. This adds an additional check after the auto labeler has it's fair chance to run.

### Process

- [ ] Docs updated
  

FonzMP (approved):
:shipit:

PR #1066, opened 2022-05-25
[API-16379] - Add VA Letter Generator API to staging

gileswells:
### Description

https://vajira.max.gov/browse/API-16379

Add the VA Letter Generator API to staging for docs, release notes, apply, and prod access.

### Process

- [ ] Docs updated
  

src/types/constants/index.ts
```diff
@@ -23,7 +23,7 @@ export const LPB_CONTACT_US_URL = `${LPB_BACKEND_BASE_URL}/v0/support/contact-us
 export const APPLY_OAUTH_APIS = ['claims', 'communityCare', 'health', 'verification'];
 export const PAGE_HEADER_ID = 'page-header';
 export const PAGE_HEADER_AND_HALO_ID = 'header-halo';
-export const APPLY_INTERNAL_APIS = ['address_validation'];
+export const APPLY_INTERNAL_APIS = ['address_validation', 'va_letter_generator'];
```
SlakrHakr, line 26: I don't believe we want show the internal only fields for this api
image


if we do want to show them, the backend api only allows for submitting one set of these values, so it would need to be updated to handled submitting different values if a users signs up for this and address validation in the same form

SlakrHakr (changes requested):

src/containers/consumerOnboarding/components/sandbox/SelectedApis.tsx
```diff
@@ -19,16 +20,16 @@ const ApiCheckboxList = ({ apiCheckboxes }: APICheckboxListProps): JSX.Element =
   const formValues = useFormikContext<Values>().values;
   const hostedApis = apiCheckboxes.filter(
     api =>
-      !api.vaInternalOnly  ||
-      APPLY_INTERNAL_APIS.includes(api.urlFragment),
+      !api.vaInternalOnly ||
+      (APPLY_INTERNAL_APIS.includes(api.urlFragment) &&
+        isHostedApiEnabled(api.urlFragment, api.enabledByDefault)),
```
gileswells, line 25: Previously this wouldn't check environments to see if an API should be listed. This allow us to put APIs that are internal only onto the apply page while still respecting if they're not yet enabled by default.

src/types/constants/index.ts
SlakrHakr: ok, this is that api, wanted to make sure

it is supposed to be on the apply page, but without these "internal to VA only" fields
https://vajira.max.gov/browse/API-15342?focusedCommentId=1409543&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1409543

but its only ccg auth, so it couldnt go onto the apply page right now anyways
https://vajira.max.gov/browse/API-15342?focusedCommentId=1409218&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1409218

there is a spun-off ticket for the work of adding this api to the sandbox apply form, once ccg on the backend is setup
https://vajira.max.gov/browse/API-15567

gileswells: Gotcha. I think I still want to leave the fix to allow internal APIs on lower environments but overall this can't be on just the apply page yet without CCG being ready to go.

SlakrHakr (approved):

PR #1079, opened 2022-06-17
API-16011 Adds deprecation notice for HLR V1

stiehlrod:
### Description
Adds deprecation notice for HLR V1.
[Ticket: API-16011](https://vajira.max.gov/browse/API-16011)
[Vets API PR](https://github.com/department-of-veterans-affairs/vets-api/pull/10063)
### Process
- [x] Docs updated


gileswells (approved):
LGTM. There is an extra space before the `---` but I'm ok with it either way as it still renders properly.

PR #1088, opened 2022-07-01
July 6 _lastUpdated Update

kbratti-va:
### Process

- [ ] Docs updated
  

src/content/apiDocs/health/fhirAPIReleaseNotes.mdx
```diff
@@ -1,3 +1,15 @@
+### July 6, 2022
+We have enabled an additional FHIR _sort parameter, "_lastUpdated", for the following R4 resources:
```
gileswells, line 2: The "_lastUpdated" here is the only instance in this file where it would be enclosed with double quotes. Other usages of this don't have any quotes at all and further still others use backticks which will make it show up as monospaced character when rendered on the portal. I feel like the choice for no quotes, backticks, or double quotes would be one for UX to decide.

gileswells (changes requested):
I am happy to be overruled by UX on the quoting item I've raised in a comment. Functionally everything else i good to go.

PR #1089, opened 2022-07-05
API-17064: Update existing version and API select dropdown to match new design from UX

SlakrHakr:
### Description

https://vajira.max.gov/browse/API-17064

Updates "Update page" functionality on both OAuth pages (Selecting the API) and API documentation pages (selecting the API version).

image
image

Updates the text on the button to make it clear the entire page is being updated by this action.

Moves the button below the dropdown selector on smaller resolutions.

Button is now disabled initially and enabled when a selection next to it is made.

Tooltip was added when button is pressed on the OAuth pages. This tooltip was not added on the API documentation pages as it seems like swagger-ui is re-rendering the entire page at that point? Hoping that is ok as the entire page shows a loading indicator but wanted to mention it here.


src/components/apiSelector/APISelector.tsx
```diff
@@ -54,10 +70,30 @@ const APISelector = (props: APISelectorProps): JSX.Element => {
...
-            <button onClick={onButtonClick} type="button">
-              Select
+          <div
+            className={classNames(
+              'vads-l-col--12',
+              'medium-screen:vads-l-col--3',
+            )}
+          >
+            <button
+              disabled={apiSelectionButtonDisabled}
```
gileswells, line 80: This should start as disabled so that there isn't a flash of color when initially loading the page.

src/components/apiSelector/APISelector.tsx
```diff
@@ -54,10 +70,30 @@ const APISelector = (props: APISelectorProps): JSX.Element => {
...
+              type="button"
+              className="page-updater"
+              data-for="update-page-button"
+              data-tip="Page updated!"
+              data-iscapture="true"
+              data-effect="solid"
+              data-place="top"
+              data-event="click"
+              data-event-off="mouseout"
+              data-delay-hide="5000"
```
gileswells, line 91: The tooltip doesn't go away on it's own, you have to scroll away. Additionally, even if there is no scroll or the timer hasn't expired, the tooltip should definitely self close if the user opens the dropdown again to start selecting a new option regardless of how long it's been visible.

Currently, if I never scroll the ACG page I can continue to keep changing APIs and it will leave the "Page updated" tooltip visible the entire time.

gileswells (changes requested):
There are some user flow things that feel off to me.

src/components/apiSelector/APISelector.tsx
SlakrHakr: I went this route because I couldnt figure out a way to make the tooltip only appear on click of the button but disappear on click of anything else. I might have to sync up with you on how youd suggest doing something like this.

src/components/apiSelector/APISelector.tsx
SlakrHakr: ok, i thought `React.useEffect(() => {` would do it immediately on load, without a flash of anything.
i'll look into this one.

SlakrHakr: ok, this should be fixed now:
https://github.com/department-of-veterans-affairs/developer-portal/pull/1089/commits/d122a4008e62ad6f0945b2615b0c605dc1eda16a

src/components/apiSelector/APISelector.tsx
SlakrHakr: ok, i think everything here should be addressed now.

> the tooltip should definitely self close if the user opens the dropdown again to start selecting a new option regardless of how long it's been visible

it doesn't quite do exactly this, but the tooltip hides when a new selection has been made in this scenario. The tooltip should hide on any click within the window after it has been shown.

SlakrHakr:
Looking into why the unit tests are failing...

> SyntaxError: Cannot use import statement outside a module

i also get the error if i switch back to master and run unit tests locally interestingly

gileswells (approved):
LGTM

PR #1090, opened 2022-07-06
API-16650: Secondary Text Not Using Correct Styling Rules in Prod Access Form

SlakrHakr:
### Description

https://vajira.max.gov/browse/API-16650

Change formatting of secondary text "If yes, we will ask for more details" in production request form to match other similar elements in the form.

image


gileswells (approved):
Personally I would have preferred to find a solution that continued to use a `<p>` tag but I see that other secondary texts across this form are using divs so this makes sense rather than needing to update all the other instances.

PR #1096, opened 2022-07-19
[API-17501] - Future Proofing MR Auto Creation workflow

gileswells:
### Description

When a PR's merge commit message is too long it will skip adding the PR number to the end of the commit. The MR creation uses that PR number from the release title (which is derived from the commit message) to properly format the MR. When the message was too long this would cause the MRs to fail to generate.

This will, if the PR number is missing, append it to the end of the Release Title when it's created.

### Process

- [ ] Docs updated
  

SlakrHakr (approved):
love it

PR #158, opened 2019-07-30
health docs copy update

jasalisbury:
https://github.com/department-of-veterans-affairs/vets-contrib/issues/2815

Update health API copy per feedback from the health team. One thing I can't figure out is how to get the URLs on the health quickstart page to show up as links. I've tried using `[]()` syntax but the link target doesn't work properly when using the environment variable. Using `< link >` doesn't work either. If anyone has ideas let me know.

See [HealthApiDocumentationUpdated2.docx](https://github.com/department-of-veterans-affairs/developer-portal/files/3447535/HealthApiDocumentationUpdated2.docx) for the list of changes.

PR #1105, opened 2022-08-03
point health doc urls at sandbox instead of production

SlakrHakr:
### Description

Health Quick Start documentation mentions several urls that currently point to production. The documentation around them talks about developing in sandbox. This code change changes the urls [here](https://developer.va.gov/explore/health/docs/quickstart) to point to sandbox.

### Note

I scoured the project to make sure this environment variable isn't being used anywhere else. I only found it used in this one document, so I think we should be safe to change it.

Additional info: this is the PR where this environment variable was introduced:
https://github.com/department-of-veterans-affairs/developer-portal/pull/158


FonzMP (approved):
This actually really surprises me that this value is not used elsewhere... but I did a quick check as well and seemingly good? 

:shipit:

PR #1106, opened 2022-08-04
API-18310: Add Internal Only Label to LGY API in Sandbox Access Form

SlakrHakr:
### Description

Flag Guaranty Remittance as "internal only" due to it having restricted access for now.
https://vajira.max.gov/browse/API-18310

### Note

All LPB environments have been updated as well.


FonzMP (approved):
:shipit:

FonzMP (approved):
:shipit:

PR #1110, opened 2022-08-26
[API-18903] - Fix links on 404 page and add Sentry

gileswells:
### Description

https://vajira.max.gov/browse/API-18903

This updates the 404 page to pull from LPB for the list of APIs and also removes the "Authorization" link from the Release Notes section as that doesn't exist and would itself 404.

Also adds Sentry logging to 404 page loads.

### Process

- [ ] Docs updated
  
### DIFF
--- a/src/components/apisLoader/ApisLoader.tsx
+++ b/src/components/apisLoader/ApisLoader.tsx
@@ -23,7 +23,7 @@ const ApisLoader: React.FunctionComponent<ApisLoaderProps> = (props): JSX.Elemen
         <div />
       ) : (
         <va-alert status="error" visible>
-          <h2 className="headline">Loading Error:</h2>
+          <h2 slot="headline">Loading Error:</h2>
           <p className="vads-u-margin-y--0">
             API details failed to load from Lighthouse Platform Backend. Please reload or try again
             later if the issue persists.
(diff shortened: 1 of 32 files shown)


SlakrHakr:
This is what it looks like if it fails to make the call to LPB, just spins forever

image

This might not be the end of the world, but it would probably make sense to have some sort of fall back, even if it just didnt have those api links in that case?

src/components/apisLoader/ApisLoader.tsx
```diff
@@ -0,0 +1,41 @@
...
+      return props.hideSpinner ? <div /> : <LoadingIndicator {...defaultLoadingProps()} />;
+    case apiLoadingState.ERROR:
+      return props.hideError ? (
+        <div />
+      ) : (
+        <va-alert status="error" visible>
+          <h2 slot="headline">Loading Error:</h2>
+          <p className="vads-u-margin-y--0">
+            API details failed to load from Lighthouse Platform Backend. Please reload or try again
+            later if the issue persists.
```
SlakrHakr, line 30: functionality of presenting the error message is awesome!
i think we should remove "Lighthouse Platform Backend" from the verbiage though. maybe as simple as:
```
API details failed to load.
Please reload or try again later if issue persists.
```
?

gileswells: Sounds good to me. I'll do that and unit test the component tomorrow.

PR #1122, opened 2022-10-13
[API-20646] - Change out Math.random() for static forgery token

gileswells:
### Description

https://vajira.max.gov/browse/API-20646

The initial WASA scan for the developer portal flagged the use of `Math.random()` as an issue because it is not cryptographically secure. Our use of it is not necessary and can be omitted entirely because we're using `SameSite=String` which coupled with sending the value as both a header and cookie means it cannot be sent maliciously via csrf regardless of the actual contents of the token.

### Process

- [ ] Docs updated
  
### DIFF
--- a/src/types/constants/index.ts
+++ b/src/types/constants/index.ts
@@ -10,6 +10,7 @@ export const OPEN_API_SPEC_HOST: string = process.env.REACT_APP_VETSGOV_SWAGGER_
 
 const LPB_PREFIX = process.env.PUBLIC_URL ?? '';
 const LPB_BASE_URL = `${LPB_PREFIX}/platform-backend`;
+export const LPB_FORGERY_TOKEN = 'CsrfBlocker';
 export const LPB_APPLY_URL = `${LPB_BASE_URL}/v0/consumers/applications`;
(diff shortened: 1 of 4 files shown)


jw81 (approved):
LGTM 👍

src/types/constants/index.ts
```diff
@@ -10,6 +10,7 @@ export const OPEN_API_SPEC_HOST: string = process.env.REACT_APP_VETSGOV_SWAGGER_
 
 const LPB_PREFIX = process.env.PUBLIC_URL ?? '';
 const LPB_BASE_URL = `${LPB_PREFIX}/platform-backend`;
+export const LPB_FORGERY_TOKEN = 'CsrfBlocker';
```
SlakrHakr, line 13: this isnt quite the same as proper csrf protection, but might be good enough for now
ideally each form render would put a unique value into the session cookie, there by ensuring the form was viewed and then submitted to the backend endpoint

rails has csrf protection built in between its frontend and backend, i attempted to model this implementation off of it: https://medium.com/rubyinside/a-deep-dive-into-csrf-protection-in-rails-19fa0a42c0ef

PR #1132, opened 2022-10-28
[API-20045] - Use category veteran redirect message as fallback

andresriveratoro:
### Description

This PR updates API pages to display the category veteran redirect message as a fallback when an API specific veteran redirect message doesn't exist.


gileswells (approved):
LGTM

PR #1155, opened 2022-12-15
[API-22495] - Conditionally hide prod aud from ccg auth

gileswells:
### Description

https://vajira.max.gov/browse/API-22495

So we don't show the `sandboxAud` value on the production okta endpoint, we're hiding the production endpoint when the `sandboxAud` and `productionAud` are matching values.

### Process

- [ ] Docs updated
  
### DIFF
--- a/src/components/oauthDocs/ccg/GeneratingJWTContent.tsx
+++ b/src/components/oauthDocs/ccg/GeneratingJWTContent.tsx
@@ -76,12 +76,14 @@ const GeneratingJWTContent: FC<GeneratingJWTProps> = ({ apiName, productionAud, 
                 {`https://deptva-eval.okta.com/oauth2/${sandboxAud}/v1/token`}
               </a>
             </p>
-            <p>
-              For the {apiName} in production, this is:{' '}
-              <a href={`https://va.okta.com/oauth2/${productionAud}/v1/token`}>
-                {`https://va.okta.com/oauth2/${productionAud}/v1/token`}
-              </a>
-            </p>
+            {sandboxAud !== productionAud && (
+              <p>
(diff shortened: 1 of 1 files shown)


src/components/oauthDocs/ccg/GeneratingJWTContent.tsx
```diff
@@ -76,12 +76,14 @@ const GeneratingJWTContent: FC<GeneratingJWTProps> = ({ apiName, productionAud,
                 {`https://deptva-eval.okta.com/oauth2/${sandboxAud}/v1/token`}
               </a>
             </p>
-            <p>
-              For the {apiName} in production, this is:{' '}
-              <a href={`https://va.okta.com/oauth2/${productionAud}/v1/token`}>
-                {`https://va.okta.com/oauth2/${productionAud}/v1/token`}
-              </a>
-            </p>
+            {sandboxAud !== productionAud && (
```
imcgonigle, line 79: This change makes sense with the context of the Jira ticket but it's not clear why this is done without looking at the ticket. Do you think it makes sense to add a test to make sure this doesn't get removed in the future by someone who doesn't have that context?

I think we should at least have a comment here explaining the reason for the check.

gileswells: A `git blame` would bring the dev back to this PR that gives additional context. So while I don't disagree that adding context is a good thing I think that we've got it fairly covered, albeit with a quick search, with it as is.

imcgonigle: You're 100% correct. Most editors will even display the commit message when you focus the line

imcgonigle (approved):

PR #1161, opened 2023-01-25
Log violation to console

imcgonigle:
### Description

I had to fix a failing accessibility test and it didn't have any information in the logs. This PR adds information about any violations encountered during the accessibility tests.

**Before**
Screenshot 2023-01-25 at 3 32 56 PM

**After**
Screenshot 2023-01-25 at 3 57 52 PM


gileswells (approved):
Love it. I think these were technically visible in the screenshot BUT this is a much better solution.

PR #1166, opened 2023-02-01
[API-23464] - Replace Swagger with Stoplight Elements

gileswells:
### Description

https://vajira.max.gov/browse/API-23464

The original ticket directed to try both Stoplight Elements and RapiDoc however RapiDoc does not load the r4 Patient Health API spec file so the only current option I can see as feasible is now Stoplight Elements. This is a ROUGH demo and will need work to be formed into a working final, deployable version.

### Process

- [ ] Docs updated
  

gileswells:
Plan going forward will be to wait for Swagger UI's update this quarter.

PR #1169, opened 2023-02-08
[API-22006] - Attempt to add source maps to Sentry Releases

jw81:
### Description

- Adds logic to try and upload source maps with Sentry Releases
- Removes a script (`scripts/upload_source_map.sh`) that supposedly use to do this through integration with a Jenkinsfile.  But that Jenkinsfile no longer exists and it doesn't appear that anything is referencing the script anymore.
  - The original commit where that script was created and referenced by the Jenkinsfile can be found [here](https://github.com/department-of-veterans-affairs/developer-portal/commit/b8c46813dc7bae86710d8de5bd16ce03563674c7).


gileswells (approved):
Looks like as good a solution as any

PR #1173, opened 2023-02-13
[API-22993] Docs page styling updates

Cgfor3:
### Description

- Update spacing around drop-down selector on docs pages
- Update paragraph font-size and margin
- Update Heading font-family and margin where applicable
- Update link font-size and font-family where applicable

### Process

- [ ] Docs updated
  

imcgonigle (approved):
It would probably be good to add at least one fallback font but not 100% necessary.

Cgfor3:
> It would probably be good to add at least one fallback font but not 100% necessary.

I will check with Maddie for what the fallback should be if we include one at all. Good thinking!

gileswells (approved):

PR #1184, opened 2023-02-23
Add leading '/' to script and style paths

imcgonigle:
### Description
This PR fixes the script and link src attributes so they point to the correct location.


jw81 (approved):

gileswells (approved):
Looks good to me too

PR #1209, opened 2023-03-30
Add .nvmrc file to project root

Cgfor3:
### Description

- Add .nvmrc file to help standardize dev environments during local development


.nvmrc
```diff
@@ -0,0 +1 @@
+18
```
gileswells, line 1: While there shouldn't be any issues with just targeting 18, do we want to enforce an exact version rather than just the major version? Open for thoughts.

Cgfor3: I would only be concerned with being less prescriptive if we learned there were minor versions of node 18 that were incompatible with our project. Generally, if a user runs `nvm install`, it will pull down the latest version available in the specified major version.

gileswells (approved):

PR #1218, opened 2023-04-27
Line number span color is consistent

andresriveratoro:


src/containers/documentation/swaggerPlugins/StyleOverride.scss
```diff
@@ -43,6 +43,9 @@
 .opblock {
   .opblock-body {
     background-color: white;
+    pre span.linenumber {
+      color: #AEB0B5 !important;
+    }
```
andresriveratoro, line 48: This `!important` is required to override a SwaggerUI rule that has less specificity but uses `!important`.

gileswells: I dislike how much Swagger uses `!important`

gileswells (approved):

PR #1226, opened 2023-05-15
[API-24008] Explore skeleton

Cgfor3:
### Description

- Add page header content
- Add placeholder for filter section
- Add list of apis linking to corresponding docs pages

### Process

- [ ] Docs updated
  
### DIFF
--- a/cypress/e2e/containers/documentation/ExploreRoot.cy.js
+++ b/cypress/e2e/containers/documentation/ExploreRoot.cy.js
@@ -1,0 +1,25 @@
+/// <reference types="cypress" />
+
+describe('Explore Root', () => {
+  beforeEach(() => {
+    cy.intercept('GET', '/platform-backend/v0/providers/transformations/legacy.json*', {
+      fixture: 'legacy.json',
+    }).as('LPB datastore');
+
+    cy.visit('/explore');
+  });
+
+  it('has core page elements', () => {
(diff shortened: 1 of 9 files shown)


cypress/e2e/containers/documentation/ExploreRoot.cy.js
```diff
@@ -0,0 +1,25 @@
...
+  it('has core page elements', () => {
+    cy.get('[data-cy="page-header"]').should('have.text', 'Explore our APIs');
+    cy.get('[data-cy="explore-filters"]').should('be.visible');
+    cy.get('[data-cy="api-list"]').should('be.visible');
+  });
+
+  describe('HAPPY PATH', () => {
+    it('navigates to first api page', () => {
+      cy.get('[data-cy="api-list"] a').first().click();
+      cy.url().should('include', '/explore/api/');
```
gileswells, line 21: I feel like this would be a better test, knowing that we have essentially a static legacy.json file to test against, if we were to test against the url including `/explore/api/address-validation` but that update would come better once we've updated to using the `urlSlug` field for url fragment as right now we're using `urlSegment` which has `address_validation` as it's value.

This is a long comment to say, I think this can be improved but let's do that after the query.ts updates are completed.

gileswells (approved):
I think this is quite solid and something that we could likely merge (with UX's blessing) and allow more hands to start tackling more nuanced parts of the explore page filtering.

imcgonigle (approved):
Looks good

PR #1231, opened 2023-05-24
[API-24006] - Header

andresriveratoro:
https://vajira.max.gov/browse/API-24006


src/components/header/Header.tsx
```diff
@@ -1,140 +1,138 @@
...
+          <p className="vads-u-margin-y--0">
+            Version 0 of the Veteran Confirmation API is deprecated and scheduled for deactivation
+            on April 4, 2024. Version 1 of the Veteran Confirmation API is now active.
+          </p>
+        </va-alert>
+      );
+    default:
+      return null;
+  }
+};
```
gileswells, line 49: I would like to see this `HeaderAlerts` element updated to use the correct url structures. Currently, this will never be seen on page.

src/components/navBar/NavBar.tsx
```diff
@@ -1,36 +1,25 @@
...
-import { getApiCategoryOrder, getApiDefinitions } from '../../apiDefs/query';
-
 interface NavBarProps {
   isMobileMenuVisible: boolean;
   onMobileNavClose: () => void;
+  isSearchBarVisible?: boolean;
+  toggleSearchBar?: () => void;
 }
 
 const navItemStyles = (isFirstChild = false): string =>
```
gileswells, line 25: This can be updated to just a string vs being a function now that no items include `isFirstChild` as true. Should help with code coverage.

src/components/header/Header.tsx
gileswells: This also could be split out to it's own component file entirely if desired which would likely make it much easier to unit test each alert on it's own vs having to test the header as a whole for each alert.

gileswells (changes requested):
Overall looks good. The code coverage specifically on `Header.tsx` did drop a bit across the different coverage metrics so I'd like to see that back up to at least where it was.

src/components/headerAlerts/HeaderAlerts.tsx
```diff
@@ -0,0 +1,39 @@
...
+            A new version of Appeals Status API (v1) will launch later this year.
+          </p>
+        </va-alert>
+      );
+    case '/explore/api/va-facilities/docs':
+      return (
+        <va-alert background-only show-icon status="info" visible>
+          <p className="vads-u-margin-y--0">
+            Version 1 of the VA Facilities API is launching soon. We will add{' '}
+            <Link to="/release-notes/facilities">release notes</Link> when it&apos;s live.
```
gileswells, line 23: Need to update to the new release notes page url

```suggestion
            <Link to="/explore/api/va-facilities/release-notes">release notes</Link> when it&apos;s live.
```

gileswells (changes requested):
Literally one last change needed then good to go

gileswells (approved):
:shipit:

PR #1233, opened 2023-05-25
API-23996 Update query.ts to use new urlSlug field 

imcgonigle:
### Description

This PR adds two new functions to `query.ts` called `lookupApiBySlug` and `lookupApiCategoryBySlug`  that both use  the new `urlSlug` to find an API or Category instead of `urlFragment`. We also update the rest of the codebase to use the new  functions.

[Jira Ticket](https://vajira.max.gov/browse/API-23996)


gileswells (approved):
Overall looks good to me

PR #1244, opened 2023-06-15
Cgfor3/api 27237 veteran history deprecation banner

Cgfor3:
### Description

- Dynamically check which banner to apply from an array of options
- Add docs explaining how to add banners going forward

### Process

- [x] Docs updated
  
### DIFF
--- a/src/components/header/Header.tsx
+++ b/src/components/header/Header.tsx
@@ -130,6 +130,14 @@ const Header = (): JSX.Element => {
             </p>
           </va-alert>
         )}
+        {location.pathname === '/explore/verification/docs/veteran_verification' && (
+          <va-alert background-only show-icon status="info" visible>
+            <p className="vads-u-margin-y--0">
+              Veteran Service History and Eligibility versions 0 and 1 are deprecated and scheduled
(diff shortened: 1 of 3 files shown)


imcgonigle (dismissed):
Looks good

src/components/header/Header.tsx
```diff
@@ -107,37 +108,19 @@ const Header = (): JSX.Element => {
...
-              Veteran Service History and Eligibility versions 0 and 1 are deprecated and scheduled
-              for deactivation on May 30, 2024. Version 2 is now live.
-            </p>
-          </va-alert>
-        )}
+        {deprecationBannerTargets
+          .filter(target => target.path === location.pathname)
+          .map(target => (
+            <va-alert
+              key={`${target.path} ${target.content}`}
```
gileswells, line 115: I think we don't need to have `${target.content}` included as part of the key. We shouldn't ever have 2 banners with the same path. Maybe provide a field to set a custom key if you think there will be collisions?

gileswells (commented):
Overall I like it, just a question on the key. Do we need any changes with this to account for unit testing?

PR #1245, opened 2023-06-16
[API-24014] explore row layout

Cgfor3:
### Description

- Add logic to iterate through each category and pass categoryName to each API within the category
- Add faLock to appropriate tags
- Add restricted access to a static list of APIs
- Updated APIDescription schema and confirmed existing instances to new schema

### Process

- [ ] Docs updated
  

src/containers/documentation/ExploreRoot.tsx
```diff
@@ -1,11 +1,75 @@
...
+  'Clinical Health API (FHIR)',
+  'Community Care Eligibility API',
+  'Contact Information',
+  'Decision Reviews API',
+  'Direct Deposit',
+  'Guaranty Remittance API',
+  'Loan Guaranty API',
+  'Loan Review',
+  'VA Letter Generator API',
+];
```
gileswells, line 26: I'm not a huge fan of this BUT provided that we create a ticket to account for removing this by adding support to LPB on a per API basis after IA launch I'm ok with it.

src/containers/documentation/ExploreRoot.tsx
```diff
@@ -1,11 +1,75 @@
...
+
+    return tags;
+  };
+
+  const getApiList = (): JSX.Element[] => {
+    let apiList: JSX.Element[] = [];
+    const apiListKeys = Object.keys(apiDefs);
+
+    apiListKeys.forEach(apiCategory => {
+      const categoryName = apiDefs[apiCategory].urlSlug;
```
gileswells, line 58: I feel like the logic for getting the category name, because it's specifically needed for the ExploreApiCard component, should be handled within the component. Passing in just the api variable through a single parameter (with a second parameter being the key) should be more than enough. Basically let the component handle its own logic rather than depend on an external process.

If this component ever needed to get added elsewhere then we'd have to duplicate this logic to that location as well.

src/containers/documentation/ExploreRoot.tsx
```diff
@@ -39,15 +103,7 @@ export const ExploreRoot = (): JSX.Element => {
...
-          {apis.map(api => (
-            <ExploreApiCard
-              description={api.description}
-              filterTags={['PLACEHOLDER TAG']}
-              key={api.urlSlug}
-              name={api.name}
-              urlSlug={api.urlSlug}
-            />
-          ))}
+          {getApiList()}
```
gileswells, line 106: I get wanting to bundle up all the logic in a method but if the suggestion to move the category logic into the ExploreApiCard component is adopted then I would rather the loop just be here rather than as a method being called as it is here.

gileswells (changes requested):
Some thoughts on how to handle the loop and where to do category tag logic.

src/containers/documentation/ExploreRoot.tsx
Cgfor3: Agreed - LPB currently exposes "restrictedAccessDetails" (string) and "restrictedAccessToggle" (boolean) with each API, so could we leverage that? AFAIK these are all set to false for the moment. If that attribute is intended for something else, then we can create another attribute

src/containers/documentation/ExploreRoot.tsx
Cgfor3: On it

src/components/exploreApiCard/ExploreApiCard.tsx
```diff
@@ -14,21 +19,53 @@ export const ExploreApiCard = ({
...
+
+  const getApiCategory = (): string => {
+    const categoryNames = Object.keys(apiDefs);
+    const apisByName = (api: APIDescription): boolean => api.name === name;
+    const categoryByApiName = (category: string): boolean =>
+      apiDefs[category].apis.filter(apisByName).length === 1;
+
+    return categoryNames.find(categoryByApiName) as string;
+  };
+
```
Cgfor3, line 33: @gileswells let me know if this is more inline with what you were thinking

gileswells: Are there instances where we won't have access to api.categoryUrlFragment? If we have that then you could utilize lookupApiCategory(api.categoryUrlFragment) as a much simpler option, right?

gileswells: Yes moving the logic into the component is the goal so it's self contained but overall I think we already have this logic within the query.ts file so we should leverage it.

Cgfor3: The mapping does not seem to work for every API on my end. I'm using `lookupApiCategoryBySlug` since I'm not seeing `categoryUrlFragment` passed from LPB and I'm still only seeing a select few of the categories returned 
Screen Shot 2023-06-21 at 11 14 30 AM

src/actions/index.ts
```diff
@@ -78,6 +78,16 @@ export const setApis: ActionCreator<SetAPIs> = (apis: APICategories) => {
...
+  ['facilities', 'vaForms', 'health', 'loanGuaranty', 'va-benefits', 'verification'].forEach(
+    (category: string) => {
+      apis[category].apis = apis[category].apis.map((item: APIDescription) => ({
+        ...item,
+        categoryUrlFragment: category,
+      }));
+    },
+  );
+  // eslint-disable-next-line no-console
+  console.log(apis);
```
gileswells, line 90: Remove

```suggestion
```

src/apiDefs/query.ts
```diff
@@ -118,9 +118,9 @@ const lookupApiCategoryBySlug = (urlSlug: string): APICategory | null => {
   return Object.values(categories).find(hasMatchingIdentifier) ?? null;
 };
 
-const lookupApiCategory = (categoryKey: string): APICategory | null =>
+const lookupApiCategory = (categoryKey: string): APICategory =>
   // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
gileswells, line 122: This line can likely be removed since there's no more `?? null` conditional.

src/components/exploreApiCard/ExploreApiCard.tsx
```diff
@@ -1,34 +1,26 @@
...
-  urlSlug: string;
-}
-
-export const ExploreApiCard = ({
-  description,
-  filterTags = [],
-  name,
-  urlSlug,
-}: ExploreApiCardProps): JSX.Element => (
+export const ExploreApiCard = ({ api }: { api: APIDescription }): JSX.Element => (
```
gileswells, line 9: Absolutely love how much simpler this component is now. 🥰

src/components/exploreApiCard/ExploreApiTags.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <FontAwesomeIcon
+        className={classNames('explore-fa-lock', 'vads-u-color--gray-medium')}
+        icon={faLock}
+      />
+    )}
+    {tag.replace(/-/g, ' ').toUpperCase()}
+  </span>
+);
+
+export const ExploreApiTags = ({ api }: { api: APIDescription }): JSX.Element => {
```
gileswells, line 26: I am slightly conflicted here. Because we only use this under the `/explore` top level folder the name of `ExploreApiTags` makes sense but the naming of this as just `ApiTags` seems like another viable option because we will have these utilized on an API's detail page that isn't directly part of the "Explore" page.

This is just a thought and I'm ok with it either way. Change if you think this rationale makes sense enough to rename it.

src/components/exploreApiCard/ExploreApiCard.scss
```diff
@@ -2,7 +2,18 @@
...
+    padding: 0 0.4rem;
+    margin-right: 0.8rem;
+    font-variant: all-small-caps;
+  }
+
+  .explore-fa-lock {
+    height: 14px;
+    margin-right: 0.6rem;
+    width: 12px;
   }
```
gileswells, line 18: It might be best to move these CSS rules out to their own file / not have then within the `.explore-api-card-container` so that they can be dropped into the [API Overview page](https://www.figma.com/file/ufpSfFIB56bvfwAHjrkM8H/Portal-ReVAMP?type=design&node-id=1452-47183&mode=design&t=UWLX5WbTolcZjbBL-0) without issue.

gileswells (changes requested):
Minor things but no big functional changes from here.

src/components/exploreApiCard/ExploreApiTags.tsx
Cgfor3: My only concern is that ApiTags already exists and I'm not sure what the differences are or where the existing component is used

Cgfor3: IF this work has already been done two years ago I'm gonna be real sad

gileswells: I knew the name sounded familiar. It was kinda done previously but not with this styling and definitely not with this granularity. https://developer.va.gov/onboarding/production-access-application This is the last place that I can see it exists. Likely a future consideration to revise how we do the visible tags on the prod access form.

gileswells (approved):
Just fixed the unit test since last approval of Calvin's code.

PR #1255, opened 2023-06-24
[API-24010] Filters

andresriveratoro:
https://jira.devops.va.gov/browse/API-24010


src/components/apiFilters/ApiFilters.tsx
```diff
@@ -0,0 +1,274 @@
...
+    if (window.innerWidth < 768) {
+      window.addEventListener('scroll', checkStickiness);
+    }
+
+    return () => {
+      if (window.innerWidth < 768) {
+        window.removeEventListener('scroll', checkStickiness);
+      }
+    };
+  }, []);
```
gileswells, line 208: Is there anything we can tie this to so it's not firing on every single change? Is something like `window.scrollY` supported in this so we've got less process calculating for differences when the vertical position hasn't changed?

gileswells (changes requested):
Looks great, just one question about the sticky implementation.

src/components/apiFilters/ApiFilters.tsx
gileswells: Guess it's not possible. 🤷‍♂️

gileswells (approved):
LGTM

PR #1270, opened 2023-07-04
[API-27821] Update tag restricted access rules

Cgfor3:
### Description

- update generateFilterTags to abide by new guidelines
- Add manual checks for outlier API conditions
- Change tags schema to include 'showLock' property

### Process

- [ ] Docs updated
  
### DIFF
--- a/src/utils/generateFilterTags.ts
+++ b/src/utils/generateFilterTags.ts
@@ -1,53 +1,78 @@
+export interface TagDataProps {
+  showLock: boolean;
+  tagName: string;
+}
+
+const AUTHORIZATION_CODE_GRANT = 'AUTHORIZATION CODE GRANT';
+const CLIENT_CREDENTIALS_GRANT = 'CLIENT CREDENTIALS GRANT';
+
 const OAUTHTYPES = {
-  AuthorizationCodeGrant: 'AUTHORIZATION CODE GRANT',
-  ClientCredentialsGrant: 'CLIENT CREDENTIALS GRANT',
+  AuthorizationCodeGrant: AUTHORIZATION_CODE_GRANT,
+  ClientCredentialsGrant: CLIENT_CREDENTIALS_GRANT,
 };
 
-const RESTRICTED_ACCESS_APIS = [
+const ACG_RESTRICTED_APIS = ['Clinical Health API (FHIR)', 'Community Care Eligibility API'];
+const NON_OAUTH_RESTRICTED_APIS = [
   'Address Validation API',
-  'Benefits Documents',
-  'Clinical Health API (FHIR)',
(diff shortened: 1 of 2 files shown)


src/utils/generateFilterTags.ts
```diff
@@ -1,53 +1,78 @@
...
+      tags = [...tags, { showLock: false, tagName: 'loan-guaranty' }];
       break;
     case 'vaForms':
-      tags = [...tags, 'forms'];
+      tags = [...tags, { showLock: false, tagName: 'forms' }];
       break;
     default:
-      tags = [...tags, categoryUrlFragment];
+      tags = [...tags, { showLock: false, tagName: categoryUrlFragment }];
   }
```
gileswells, line 59: I'd like for us to have a ticket to cover making this entirely driven by LPB data in the future. Obviously this will work for now but for the future we should definitely drive this with LPB.

gileswells (approved):
Would like to see this reworked to be entirely driven by LPB data but this looks about as good as we can get with the layers of rules on when what should be displayed and how.

src/utils/generateFilterTags.ts
Cgfor3: That was my hope going forward. https://lighthouseva.slack.com/archives/C042CQB8KF0/p1688418441485259

I'll bring this up during standup tomorrow morning to make sure we're tracking this enhancement

src/utils/generateFilterTags.ts
```diff
@@ -1,53 +1,78 @@
...
   name: string,
   oAuthTypes: string[] | null,
   openData: boolean,
-): string[] => {
-  let tags: string[] = [];
+): TagDataProps[] => {
+  let tags: TagDataProps[] = [];
+
+  if (hasRestrictedAccess(name, oAuthTypes)) {
+    tags = [...tags, { showLock: true, tagName: 'RESTRICTED ACCESS' }];
```
andresriveratoro, line 47: This is minor, but we can omit spreading an empty array here.

andresriveratoro (approved):

src/utils/generateFilterTags.ts
Cgfor3: How would you approach this? If tags is empty on this line, we're spreading an empty array. But if it skips this conditional, and enters the switch statement as an empty array I would need to account for it there

andresriveratoro, line 47:
```suggestion
    tags = [{ showLock: true, tagName: 'RESTRICTED ACCESS' }];
```

This is what I meant. You are correct that you will have to account for the possibility that tags isn't empty at the switch statement.

gileswells (approved):

PR #1271, opened 2023-07-05
Update topic filter names to use abbreviated form

Cgfor3:
### Process

- [ ] Docs updated
  
### Requested feedback
An alternative solution to using an object to map names was to chain `.replace('APIs', '').replace('Veteran', '').trim()`. Let me know which of these two solutions is preferred.


gileswells (approved):
LGTM, does it pass unit tests?

gileswells (approved):
LGTM

PR #1277, opened 2023-07-06
[API-24015] - S3 and React based redirects

gileswells:
### Description

This PR will create the redirects necessary for the S3 bucket but due to an issue with how site traffic passes through the application load balancer those redirects are failing until a patch can be created and applied.

This also includes a fallback component that will do the same list of redirects but within Javascript which is not as SEO optimized. This is temporary and will hopefully be removed in the future once the S3 redirects start working properly.

### Process

- [ ] Docs updated
  

src/components/SiteRedirects.tsx
```diff
@@ -0,0 +1,329 @@
...
+    },
+    {
+      new: '/explore',
+      old: '/oauth',
+    },
+  ];
+  redirects.some((item: RedirectData): boolean => {
+    const testFullPath = location.pathname + location.search;
+    if (location.pathname === item.old || testFullPath === item.old) {
+      window.location.href = item.new;
```
gileswells, line 320: I'm using window.location.href to force a page reload as just using the react-router `<redirect />` component was throwing an issue when redirecting to a pre-filtered explore page.

imcgonigle (approved):
Looks good

src/components/SiteRedirects.tsx
andresriveratoro: How about [useHistory](https://v5.reactrouter.com/web/api/Hooks/usehistory)?

andresriveratoro (approved):

PR #1278, opened 2023-07-07
Update forms ACG and CCG links

andresriveratoro:
[27930](https://jira.devops.va.gov/browse/API-27930)


gileswells (approved):
Simple

src/containers/documentation/RequestSandboxAccess.tsx
```diff
@@ -7,32 +7,18 @@ import { Link } from 'react-router-dom';
 import { PageHeader } from '../../components';
 import { LPB_APPLY_URL } from '../../types/constants';
 import { ApplySuccessResult } from '../../types/forms/apply';
-import {
-  AUTHORIZATION_CCG_PATH,
-  AUTHORIZATION_PKCE_PATH,
```
Cgfor3, line 12: We should remove these constants from our paths file as well

gileswells: Very soon, we should be able to use these path constants to to track down all instances of these legacy URLs and update them accordingly.

andresriveratoro: I can remove these in the next PR

Cgfor3: They only exist for the sandbox and prod access forms. The local sandbox form component isn't used as it's been replaced, and Andres is updating the production form in another PR so there shouldn't be other instances of these URLs

Cgfor3: JK Routes.tsx line 36 is double redirecting now

`    <Route path="/oauth" render={(): JSX.Element => <Redirect to="/explore/authorization" />} />
`

Cgfor3 (dismissed):
LGTM

Cgfor3 (approved):
LGTM

PR #1280, opened 2023-07-07
Update production access form ACG and CCG links

andresriveratoro:
Part of [27930](https://jira.devops.va.gov/browse/API-27930)


Cgfor3 (approved):
Looks good. If you don't want to remove the path constants that's completely fine. We should have a larger effort to clean up/remove legacy code

andresriveratoro:
Merged this to https://github.com/department-of-veterans-affairs/developer-portal/pull/1278

PR #1287, opened 2023-07-13
fix: apply initial auth filters

andresriveratoro:


src/components/apiFilters/ApiFilters.tsx
```diff
@@ -45,7 +45,7 @@ export const ApiFilters = ({ apis, setApis }: ApiFiltersProps): JSX.Element => {
     params.categoryUrlSlugs?.split('+') ?? [],
   );
   const [authFilter, setAuthFilter] = useState<string[]>(
-    new URLSearchParams(location.search).get('auth')?.split('+') ?? [],
+    new URLSearchParams(location.search).get('auth')?.split(' ') ?? [],
```
gileswells, line 48: Um... What's the rationale behind this?

gileswells (changes requested):

src/components/apiFilters/ApiFilters.tsx
andresriveratoro: If you go to https://developer.va.gov/explore?auth=acg+ccg you will see auth filters are not being applied properly - no APIs are displayed and the filter pill is broken. Without this change, the initial value of authFilter is `["acg ccg"]` instead of `["acg", "ccg"]`.

Cgfor3: Browsers translate the `+` shorthand encoding to a space. It's the equivalent to `?auth=acg%20ccg`

PR #1299, opened 2023-07-25
[API-28250] Validate OAuth public key

andresriveratoro:
[API-28250](https://jira.devops.va.gov/browse/API-28250)


cypress/e2e/containers/consumerOnboarding/ProductionAccess.cy.js
```diff
@@ -207,7 +207,10 @@ describe('Production Access Form', () => {
 
     verificationFields(cy);
     cy.get('#apisFormFieldccgclaims').click();
-    cy.get('#oAuthPublicKeyFormField').type('{ }');
+    cy.get('#oAuthPublicKeyFormField').type(
+      '{"kid": null,"kty": "RSA","e": "AQAB","use": null,"n": "2Fb4_D4-RSjvl11txu-0s9bThk8hTo2SJauTRrS9N7piFlpGi6PBql3KzLmEu_T36YMbmTjDRPyybEEBD_XkEDuNdWSQph5Da7atfFM04IW5WH3MGPuvmaH6WpZB4Li5qESTFaMk0677uCDvOLcJmfa8bzunvbtlB4U-1WLjtDBODWiVpLlGEUofNQdX2MvTF9shtm-QqPk7K-a2Z36LrZpgcQBB1U8QtqexdaLrMgaoxmEbSgXGAc-uDkmQx1VOAsREozYZ9f1tASmOKGlxfVyBHcf6dePxq1cewpmrUfRTezky5A4K6v17uBYSpEols4ritWDRDymb7rFlUwxBjqdCjmtV18HiLIrgBNPQ2-5Jlnt-BCJg3lP_UG0r6cMO2DEtTkAkDcy4HzNuMQCrXn5ZL4kSUITrf9Mixny3vFn3aVcSNsCqLUSAfnpfRIz9oUUz5xI-FD9QsJJ1vneC8mfo-1lNaVRLNhn2t9VWY0kqhNNzS2HIktkZGzGv7gsB"}',
```
jw81, line 211: Would it be possible to change this example data to be something more obviously "fake"?
I'm fairly sure that repo security crawler bots might flag this as a potential security risk.
So if we can make the data look less real, then it'll hopefully not trigger the bots.  But at the very least, if it's obvious fake data and still somehow triggers the bots, then it'll be immediately obvious to future-us when we go to investigate if the security alert is real.

jw81: FWIW, I tinkered with the PR app and was able to give the field the following input and it worked.  So maybe we could do something similar with the fake data in this test?

```
{
  "kty": "RSA",
  "n": "blah",
  "e": "blah"
}
```

gileswells (approved):

PR #1301, opened 2023-07-26
[API-28383] Display schema docs at full width

andresriveratoro:
[API-28383](https://jira.devops.va.gov/browse/API-28383)


Cgfor3 (approved):
👍

PR #1302, opened 2023-07-27
[API-27857] Fix Swagger UI table overflow

andresriveratoro:
[API-27857](https://jira.devops.va.gov/browse/API-27857)


Cgfor3 (approved):
👍

PR #1308, opened 2023-07-31
[API-28413] Update breadcrumb navigation to explore APIs with filters

andresriveratoro:
[API-28413](https://jira.devops.va.gov/browse/API-28413)

### DIFF
--- a/src/components/apiFilters/ApiFilters.test.tsx
+++ b/src/components/apiFilters/ApiFilters.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { fireEvent, render, screen } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { BrowserRouter as Router, Router as HistoryRouter } from 'react-router-dom';
 import { createMemoryHistory } from 'history';
 import { Provider } from 'react-redux';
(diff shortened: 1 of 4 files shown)


src/components/apiFilters/ApiFilters.test.tsx
```diff
@@ -83,4 +83,22 @@ describe('ApiFilters', () => {
...
+    const history = createMemoryHistory();
+    history.push('/explore/va-benefits');
+    render(
+      <Provider store={store}>
+        <HistoryRouter history={history}>
+          <ApiFilters apis={apis} setApis={setApis} />
+        </HistoryRouter>
+      </Provider>,
+    );
+    expect(localStorage.getItem('exploreApisPath')).toBe('/explore/va-benefits');
```
gileswells, line 97: Is this being set by the `history.push()` on like 87 or is this a carry over from the explicit `localStorage.setItem()` that's being called in [ApiBreadbrumbs.test.tsx](https://github.com/department-of-veterans-affairs/developer-portal/pull/1308/commits/9ea1b20fcc43ca4cbf21de9214760f622959bfdc#diff-8caecf31103c18a4980df1d3b1560c985c305bf346935e2a62c9654751c37476R49)?

Just want to make sure that these tests are working in isolation.

gileswells (commented):
One question about test isolation but overall looks good. Much prefer that you were able to work these in with the unit tests vs cypress as it'll help with code coverage metrics.

gileswells (approved):
The only other question I have is if we want to do anything in terms of when people navigate away from the explore page or leave the site entirely. How long do we want to keep the `localStorage` around for. That's a complex question for another day if it even needs to be answered at all, which it likely does not. ✅

PR #1311, opened 2023-07-31
[API-27988] - Add GHA workflow for monitoring security alerts

jw81:
### Description

Original Issue :: [API-27988](https://jira.devops.va.gov/browse/API-27988)

- Adds a GitHub Action workflow that checks for any security vulnerability alerts in the repo and sends a Slack notification to the `#team-okapi-alerts` channel if any are found.

### DIFF
--- a/.github/workflows/notify-security-alerts.yml
+++ b/.github/workflows/notify-security-alerts.yml
@@ -1,0 +1,78 @@
+name: Notify Security Alerts
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 0 15 * * 1-5
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    name: Check for security alerts and notify team via Slack
+    steps:
+      - id: check_for_code_scanning_alerts
+        name: Check for code-scanning alerts
+        run: |
+          echo ${{ secrets.DEVPORTAL_SECURITY_SCAN_WORKFLOW_TOKEN }} | gh auth login --with-token
+          gh api '/repos/department-of-veterans-affairs/developer-portal/code-scanning/alerts?state=open' \
+            > code-scanning.json
+          cat code-scanning.json
+          if [ "[]" == "$(cat code-scanning.json)" ]; then
+            echo "security_alert_exists='false'" >> $GITHUB_OUTPUT
+          else
+            echo "security_alert_exists='true'" >> $GITHUB_OUTPUT
+          fi
+      - id: check_for_dependabot_alerts
+        name: Check for dependabot alerts   
+        run: |
+          echo ${{ secrets.DEVPORTAL_SECURITY_SCAN_WORKFLOW_TOKEN }} | gh auth login --with-token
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-security-alerts.yml
```diff
@@ -0,0 +1,78 @@
...
+          cat code-scanning.json
+          if [ "[]" == "$(cat code-scanning.json)" ]; then
+            echo "security_alert_exists='false'" >> $GITHUB_OUTPUT
+          else
+            echo "security_alert_exists='true'" >> $GITHUB_OUTPUT
+          fi
+      - id: check_for_dependabot_alerts
+        name: Check for dependabot alerts   
+        run: |
+          echo ${{ secrets.DEVPORTAL_SECURITY_SCAN_WORKFLOW_TOKEN }} | gh auth login --with-token
```
gileswells, line 26: Shouldn't need to be repeated

.github/workflows/notify-security-alerts.yml
```diff
@@ -0,0 +1,78 @@
...
+          cat dependabot.json
+          if [ "[]" == "$(cat dependabot.json)" ]; then
+            echo "security_alert_exists='false'" >> $GITHUB_OUTPUT
+          else
+            echo "security_alert_exists='true'" >> $GITHUB_OUTPUT
+          fi
+      - id: check_for_secret_scanning_alerts
+        name: Check for secret scanning alerts   
+        run: |
+          echo ${{ secrets.DEVPORTAL_SECURITY_SCAN_WORKFLOW_TOKEN }} | gh auth login --with-token
```
gileswells, line 38: Shouldn't need to be repeated

gileswells (changes requested):
Overall looks good, just a couple lines that can be removed to tidy it up.

.github/workflows/notify-security-alerts.yml
jw81: Ah, ok.
For some reason, I was thinking that `step` order wasn't guaranteed.  So I was thinking I needed to give each `step` some logic to make sure it was logged in.  But, it looks like that's [not the case](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions#jobs).
> Steps are executed in order and are dependent on each other.

Removed in [this commit](https://github.com/department-of-veterans-affairs/developer-portal/pull/1311/commits/86e44fba51d89964ee78d47b52d808886c4cc390).

.github/workflows/notify-security-alerts.yml
jw81: Ah, ok.
For some reason, I was thinking that `step` order wasn't guaranteed.  So I was thinking I needed to give each `step` some logic to make sure it was logged in.  But, it looks like that's [not the case](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions#jobs).
> Steps are executed in order and are dependent on each other.

Removed in [this commit](https://github.com/department-of-veterans-affairs/developer-portal/pull/1311/commits/86e44fba51d89964ee78d47b52d808886c4cc390).

gileswells (approved):

PR #1316, opened 2023-08-02
Replace production access form API labels

andresriveratoro:
### Description
[API-28416](https://jira.devops.va.gov/browse/API-28416)

This PR replaces the API labels on the production access form with the same labels used for the Explore API page


src/components/apiTags/ApiTags.test.tsx
```diff
@@ -5,22 +5,23 @@ import { VaInternalOnly } from '../../apiDefs/schema';
...
 describe('ApiTags', () => {
-  it('renders the VA internal only tag and not the open data tag', () => {
+  it('renders the restricted access tag and not the open data tag', () => {
     render(
       <FlagsProvider flags={getFlags()}>
         <ApiTags openData={false} vaInternalOnly={VaInternalOnly.StrictlyInternal} />
       </FlagsProvider>,
     );
 
+    screen.debug();
```
gileswells, line 19: Should this be removed?

src/components/apiTags/ApiTags.tsx
```diff
@@ -16,8 +16,8 @@ const ApiTagsPropTypes = {
 type ApiTagsProps = PropTypes.InferProps<typeof ApiTagsPropTypes>;
 const ApiTags: React.FunctionComponent<ApiTagsProps> = (props: ApiTagsProps): JSX.Element => (
   <>
-    {props.vaInternalOnly && <ApiTag type={tagTypes.VAInternalOnly} />}
-    {props.openData && <ApiTag type={tagTypes.OpenData} />}
+    {props.vaInternalOnly && <ExploreApiTag showLock tagName="Restricted Access" />}
+    {props.openData && <ExploreApiTag tagName="Open Data" />}
```
gileswells, line 20: Does it make more sense to include a full component (`ApiTags` in this case) or would it make more sense to allow for `<ExploreApiTags />` to allow for a property where someone can block certain tag types from being shown. For instance, what if `ApiTags` was removed entirely and we just had `<ExploreApiTags api={api} skip={['topic']} />` where this would allow for all the same tags and layout to be used but would skip the topic tag from rendering? I put this as an array as we could potentially want to have skips setup for all of the types (`'auth'|'lock'|'restricted'|'topic'`) and potentially multiple at once.

Not convinced either way but curious for your thoughts.

gileswells (changes requested):
Curious about the `<ApiTags />` component remaining vs expanding the scope of `<ExploreApiTags />` as an alternative option.

src/components/apiTags/ApiTags.tsx
andresriveratoro: I ended up removing the old ApiTags component. I don't think we need an ApiTags component or to expand the ExploreApiTags component for a component this simple. I extracted the ExploreApiTag component out of ExploreApiTags to further signal that ApiTag is a generic component that can be used elsewhere.

gileswells: I like it. Solve the CI tests and I'll re-review the PR.

gileswells (dismissed):
LGTM

gileswells (dismissed):
LGTM

Cgfor3 (approved):
:shipit:

PR #1330, opened 2023-08-10
[API-24651] - Remove chalk package

gileswells:
### Description

https://jira.devops.va.gov/browse/API-24651

Package isn't necessary on it's own so changing to use the one under `react-dev-tools` instead.


jw81 (approved):
👍

PR #1333, opened 2023-08-11
[API-28910] - Fix Notify Security Alerts GHA workflow

jw81:
### Description
Original Issue :: [API-28910](https://jira.devops.va.gov/browse/API-28910)
- Tweaks the structure of the "Notify Security Alerts" GHA workflow so that it properly notifies us when an alert has been detected.

### DIFF
--- a/.github/workflows/notify-security-alerts.yml
+++ b/.github/workflows/notify-security-alerts.yml
@@ -4,51 +4,60 @@ on:
   schedule:
     - cron: 0 15 * * 1-5
 jobs:
-  notify:
+  check_code_scanning_alerts:
     runs-on: ubuntu-latest
-    name: Check for security alerts and notify team via Slack
+    outputs:
+      alert_detected: ${{ steps.check.outputs.alert_detected }}
     steps:
-      - id: check_for_code_scanning_alerts
-        name: Check for code-scanning alerts
+      - id: check
         run: |
           echo ${{ secrets.DEVPORTAL_SECURITY_SCAN_WORKFLOW_TOKEN }} | gh auth login --with-token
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-security-alerts.yml
```diff
@@ -4,51 +4,60 @@ on:
...
-        name: Check for dependabot alerts   
+
+  check_dependabot_alerts:
+    runs-on: ubuntu-latest
+    outputs:
+      alert_detected: ${{ steps.check.outputs.alert_detected }}
+    steps:
+      - id: check
         run: |
+          echo ${{ secrets.DEVPORTAL_SECURITY_SCAN_WORKFLOW_TOKEN }} | gh auth login --with-token
```
jw81, line 29: Since I broke this out into its own `job` rather than a nested `step`, then we have to authenticate with GH again.

.github/workflows/notify-security-alerts.yml
```diff
@@ -4,51 +4,60 @@ on:
...
-        name: Check for secret scanning alerts   
+
+  check_secret_scanning_alerts:
+    runs-on: ubuntu-latest
+    outputs:
+      alert_detected: ${{ steps.check.outputs.alert_detected }}
+    steps:
+      - id: check
         run: |
+          echo ${{ secrets.DEVPORTAL_SECURITY_SCAN_WORKFLOW_TOKEN }} | gh auth login --with-token
```
jw81, line 44: Since I broke this out into its own `job` rather than a nested `step`, then we have to authenticate with GH again.

gileswells (approved):

PR #1337, opened 2023-08-16
[API-15112] Update component-library dependency to 18.0.1

andresriveratoro:


andresriveratoro:
I tried updating the import for `defineCustomElements` as found in the [documentation](https://design.va.gov/about/developers/install#load-the-web-component-library), but I was unable to get it to work in Jest. It results in the following error: `TypeError: (0 , component_library_1.defineCustomElements) is not a function`. Updating the `transformIgnorePatterns` property in [unit.jest.config.js](https://github.com/department-of-veterans-affairs/developer-portal/blob/master/unit.jest.config.js#L43) to exclude the component-library from being transformed results in the same error.

gileswells (approved):
Shocked I tell you at how small of an update this is.

PR #1343, opened 2023-08-22
[API-28921] - implement sandbox form blocking

gileswells:
### Description

https://jira.devops.va.gov/browse/API-28921

Using the value for `vaInternalOnly` on an `APIDescription` this PR will block access to the sandbox access form to APIs that should only be available through contacting support for sandbox access.


jw81 (approved):
LGTM

PR #1349, opened 2023-08-31
[API-29516] - Patch failure scenario on smoketest

gileswells:
### Description

https://jira.devops.va.gov/browse/API-29516

Now failing tests will not stop the cleanup steps from running leaving lots of orphaned apps in Okta.


jw81 (approved):
LGTM

PR #1350, opened 2023-09-01
[API-29153] - Update sandbox-access-form package

gileswells:
### Description

https://jira.devops.va.gov/browse/API-29153

Updates the sandbox access form to use the latest version which returns the organization field and makes it required again.


jw81 (approved):
LGTM

PR #1352, opened 2023-09-12
Make auth filter classname unique

Cgfor3:
### Description

Both filter types share a classname that should be unique
- Added new classname for the auth filter and duplicated the styling to apply to both


jw81 (approved):
LGTM 👍

gileswells (approved):

gileswells (approved):

PR #1354, opened 2023-09-14
[API-29775] - Block sandbox-access route by LPB property

gileswells:
### Description

https://jira.devops.va.gov/browse/API-29775

This will use the `blockSandboxForm` property as the source of truth for whether or not to show the sandbox access form.

Additionally I updated some of the 404 conditionals to use the `throw new Error('API not found')` as a way to limit the number of different places we were calling for the 404 page to have to load in. It will show an error on local but when deployed that error should disappear and users will see the 404 page right away.

### DIFF
--- a/src/containers/documentation/ApiPage.test.tsx
+++ b/src/containers/documentation/ApiPage.test.tsx
@@ -8,7 +8,6 @@ import { AppFlags, FlagsProvider, getFlags } from '../../flags';
 import { fakeCategories, unmetDeactivationInfo } from '../../__mocks__/fakeCategories';
 import * as apiDefs from '../../apiDefs/query';
 import store from '../../store';
-// import { apiLoadingState } from '../../types/constants';
 import ApiPage from './ApiPage';
 
 // Convenience variables to try and keep the index values out of the test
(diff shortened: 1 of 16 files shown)


imcgonigle (dismissed):

src/containers/documentation/ApiPage.test.tsx
```diff
@@ -8,7 +8,7 @@ import { AppFlags, FlagsProvider, getFlags } from '../../flags';
 import { fakeCategories, unmetDeactivationInfo } from '../../__mocks__/fakeCategories';
 import * as apiDefs from '../../apiDefs/query';
 import store from '../../store';
-import { apiLoadingState } from '../../types/constants';
+// import { apiLoadingState } from '../../types/constants';
```
Cgfor3, line 11: Remove comment

src/containers/documentation/ApiPage.test.tsx
```diff
@@ -59,9 +59,9 @@ describe('ApiPage', () => {
 
   const lookupApiBySlugMock = jest.spyOn(apiDefs, 'lookupApiBySlug');
   const lookupApiCategoryMock = jest.spyOn(apiDefs, 'lookupApiCategory');
-  const apisLoadedSpy = jest
-    .spyOn(apiDefs, 'getApisLoadedState')
-    .mockReturnValue(apiLoadingState.LOADED);
+  // const apisLoadedSpy = jest
+  //   .spyOn(apiDefs, 'getApisLoadedState')
+  //   .mockReturnValue(apiLoadingState.LOADED);
```
Cgfor3, line 64: same as above

Cgfor3 (dismissed):
LGTM

Cgfor3 (approved):
:shipit:

PR #1360, opened 2023-09-21
Update navigation aria labels to avoid SR repetition

andresriveratoro:
### Description
https://jira.devops.va.gov/browse/API-29807


Cgfor3 (approved):
👍

PR #1365, opened 2023-09-27
API-29550/Add missing scope descriptions

imcgonigle:
### Description
This PR adds the descriptions of the scopes being added in API-29550


gileswells (dismissed):
We really need to work out a solution to load in scopes from LPB...

gileswells (approved):
LGTM

PR #1379, opened 2023-10-12
Fix broken import from major version update

gileswells:
### Description

Closes #1351 

This fixes an import path that no longer exists with the version update from 5.17.0 to 6.1.3
No tests were affected by this, just the import path.


Cgfor3 (approved):
LGTM

PR #1384, opened 2023-10-16
[API-29927] Remove ICN section

Cgfor3:
### Description

- Remove section for Guaranty Remittance and Loan Review APIs


gileswells (approved):
LGTM

I'd much prefer that we have a more standard way to have these types of things be handled through an LPB property on the API itself but this works for what's needed.

PR #1387, opened 2023-10-18
[API-30849] - Support stealth launched apis

gileswells:
### Description
https://jira.devops.va.gov/browse/API-30849

This adds a filter into the root `getAllApis()` method to allow stealth launched APIs to be ignored in all cases except when viewed directly by their URL. This is conditioned based on an `isStealthLaunched` property on the APIDescription object.


Cgfor3 (approved):
LGTM

PR #1388, opened 2023-10-18
Remove jest-puppeteer package

gileswells:
### Description

This package was previously used to support accessibility tests that were being run under jest. These have long since been replaced with Cypress tests that cover the same requirements.


Cgfor3 (approved):
:shipit:

PR #1389, opened 2023-10-20
[API-30849] - Remove Flag around SwaggerDocs

gileswells:
### Description

https://jira.devops.va.gov/browse/API-30849

Removes this check because it's blocking specifically the swagger docs on a stealth launched API from being viewable on the portal.


Cgfor3 (approved):
🔥

PR #1393, opened 2023-10-23
[API-31131] Update CCG Launch Description

Cgfor3:
### Description

- Update CCG launch context description for Guaranty Remittance and Loan Review APIs

### DIFF
--- a/src/components/oauthDocs/ccg/RetrievingTokenContent.tsx
+++ b/src/components/oauthDocs/ccg/RetrievingTokenContent.tsx
@@ -1,57 +1,16 @@
 import React, { FC } from 'react';
 import { useParams } from 'react-router-dom';
 import { CodeBlock } from '../../../components';
+import { getLaunchDescription } from '../../../utils/ccgLaunchContextHelper';
 
 interface RetrievingTokenProps {
   scopes: string[];
   baseAuthPath: string;
 }
 
-const getLaunchDescription = (urlSlug: string): JSX.Element => {
-  switch (urlSlug) {
-    case 'guaranty-remittance':
-    case 'loan-review':
-      return (
-        <td>
-          <p>
-            The launch scope and parameter limit the scope of an access token by indicating the
-            token is for a specific lender ID.
-          </p>
-          <p>
-            It must be a base64-encoded JSON object, the value of which is the lender portal ID that
-            is associated to the specific lender ID. The format of the object will be:{' '}
-            <code>{'{ "portal_id": "TEST1234567890SERVICE"}'}</code>
-          </p>
-          <p>
-            When encoded using base64, the object will look like this:{' '}
-            <code>LWIgeyJwYXRpZW50IjoiMTAwMDcyMDEwMFYyNzEzODcifQo==</code>
-          </p>
-        </td>
-      );
-    default:
-      return (
-        <td>
-          <p>
-            The launch scope and parameter limit the scope of an access token by indicating the
-            token is for a specific patient or encounter.
-          </p>
-          <p>
-            It must be a base64-encoded JSON object, the value of which is the patient&apos;s ICN.
-            The format of the object will be: <code>{'{ "patient": "1000720100V271387"}'}</code>
-          </p>
-          <p>
-            When encoded using base64, the object will look like this:{' '}
-            <code>LWIgeyJwYXRpZW50IjoiMTAwMDcyMDEwMFYyNzEzODcifQo==</code>
-          </p>
-        </td>
-      );
-  }
-};
-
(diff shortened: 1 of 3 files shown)


src/components/oauthDocs/ccg/RetrievingTokenContent.tsx
```diff
@@ -1,13 +1,58 @@
...
+            token is for a specific lender ID.
+          </p>
+          <p>
+            It must be a base64-encoded JSON object, the value of which is the lender portal ID that
+            is associated to the specific lender ID. The format of the object will be:{' '}
+            <code>{'{ "portal_id": "TEST1234567890SERVICE"}'}</code>
+          </p>
+          <p>
+            When encoded using base64, the object will look like this:{' '}
+            <code>LWIgeyJwYXRpZW50IjoiMTAwMDcyMDEwMFYyNzEzODcifQo==</code>
```
gileswells, line 27: This is not the base64 encoded string of `{ "portal_id": "TEST1234567890SERVICE"}`.

src/components/oauthDocs/ccg/RetrievingTokenContent.tsx
```diff
@@ -1,13 +1,58 @@
...
+    default:
+      return (
+        <td>
+          <p>
+            The launch scope and parameter limit the scope of an access token by indicating the
+            token is for a specific patient or encounter.
+          </p>
+          <p>
+            It must be a base64-encoded JSON object, the value of which is the patient&apos;s ICN.
+            The format of the object will be: <code>{'{ "patient": "1000720100V271387"}'}</code>
```
gileswells, line 40: Same thing here with the space before the "patient" property. Would also likely need to update the base64 string below to match.

src/components/oauthDocs/ccg/RetrievingTokenContent.tsx
```diff
@@ -1,13 +1,58 @@
...
+      return (
+        <td>
+          <p>
+            The launch scope and parameter limit the scope of an access token by indicating the
+            token is for a specific lender ID.
+          </p>
+          <p>
+            It must be a base64-encoded JSON object, the value of which is the lender portal ID that
+            is associated to the specific lender ID. The format of the object will be:{' '}
+            <code>{'{ "portal_id": "TEST1234567890SERVICE"}'}</code>
```
gileswells, line 23: I'd ideally like to see us format this JSON a little better. Specifically looking at that arbitrary space before the "portal_id" property name.

gileswells (changes requested):
Some changes. I wouldn't be opposed to setting the JSON object earlier in the switch statement and then using `JSON.stringify(object)` and `btoa(JSON.stringify(object))`. This would be a good first step toward the likely inevitable switch to this object coming down from LPB's `oauth_info` field.

src/components/oauthDocs/ccg/RetrievingTokenContent.tsx
Cgfor3: Good catch they didn't provide the encoded string, and I didn't think to generate it. 

Is there a preferred method of getting this? Or is using any online base64 encoder fine? I got `eyAicG9ydGFsX2lkIjogIlRFU1QxMjM0NTY3ODkwU0VSVklDRSJ9` as the encoded string

src/utils/ccgLaunchContextHelper.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const encodedData = btoa(stringData);
+  return [stringData, encodedData];
+};
+
+export const getLaunchDescription = (urlSlug: string): JSX.Element => {
+  switch (urlSlug) {
+    case 'guaranty-remittance':
+    case 'loan-review':
+      const portalData = { portal_id: 'TEST1234567890SERVICE' };
+      const [portalStringData, portalEncodedData] = createContextData(portalData);
```
Cgfor3, line 14: @gileswells is this closer to what you had in mind? It's using the stringify method you described and starting with the object outside the template string resolves the issue of having extra random spaces

gileswells: Yes. The similarities of the blocks of text give me a desire to add more things as variables vs large blocks of html code so that we can keep the content uniform with just swapping in like `patient's ICN` or `lender ID`. The pure simplicity of just swapping out a name / field here and there does kinda run afoul with the middle paragraph so maybe not so much a possibility as my desires might have.

gileswells (approved):
I like the idea of making this content more variable driven but might not be overly possible.

jw81 (approved):
LGTM 👍 
The code makes sense to me and when I look at Guaranty Remittance and Loan Guaranty CCG docs in the review instance, they look like they should.  
Also looked at the Patient Health CCG docs in the review instance and they too looked as they should.

PR #1396, opened 2023-10-25
[API-29794] - Update Veteran Resources modal aria labels for SR

andresriveratoro:
### Description
https://jira.devops.va.gov/browse/API-29794


src/components/veteranResources/VeteranResources.test.tsx
```diff
@@ -12,57 +12,14 @@ describe('VeteranResources', () => {
...
-    expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
-
-    const modalButton = screen.getByRole('button');
-    fireEvent.click(modalButton);
-    await waitFor(() => expect(screen.queryByRole('dialog')).toBeInTheDocument());
-
-    const dialogCloseButton = screen.getByText('Close');
-    fireEvent.click(dialogCloseButton);
-    await waitFor(() => expect(screen.queryByRole('dialog')).not.toBeInTheDocument());
-  });
```
jw81, line 41: In the PR build, when I click the "close" button on the modal, it doesn't appear to close.
It looks like the "x" icon and clicking off of the modal work just fine.  But something seems to be wonky with clicking the "close" button.

Also, is there maybe a reason we shouldn't be keeping these 3 specs?

andresriveratoro: Good catch. I've updated the component to use the correct event prop.

The tests were removed because we replaced the React modal component with the web component Modal from VADS and we're unable to access shadow DOM elements in our tests. This functionality is also already [tested](https://github.com/department-of-veterans-affairs/component-library/blob/main/packages/web-components/src/components/va-modal/test/va-modal.e2e.ts) as part of the VADS.

jw81 (approved):
LGTM 👍

PR #1423, opened 2023-11-29
[API-31520] - Implement Touchpoint Survey

jw81:
### Description

Original Issue :: [API-31520](https://jira.devops.va.gov/browse/API-31520)
- Adds the UX-designed Touchpoint survey to the DevPortal.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -15,6 +15,8 @@
       })(window,document,'script','dataLayer','GTM-P39RXZWL');</script>
     <!-- End Google Tag Manager -->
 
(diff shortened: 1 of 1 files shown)


jw81:
**_NOTE_** :: Everything here is working, except we'd need to update the visual regressions.
But, I'm trying a new approach over in [this PR](https://github.com/department-of-veterans-affairs/developer-portal/pull/1425).  If that PR, we're trying the "Custom button & modal on your website" config option that Touchpoints provides.  If that PR doesn't work out and we need to fall back to this PR, then we'll need to go back into the Touchpoints config and switch it back over to using the default button configuration option.

PR #1425, opened 2023-12-13
[API-31520] - Implement Touchpoint Survey (revisited)

jw81:
### Description

Original Issue :: [API-31520](https://jira.devops.va.gov/browse/API-31520)
- Adds the UX-designed Touchpoint survey to the DevPortal.


cypress/e2e/containers/documentation/DocumentationRoot.cy.js
```diff
@@ -33,7 +33,7 @@ describe('Documentation Root', () => {
   });
 
   it('Side nav does not link to Sandbox Access form on internal APIs', () => {
-    cy.visit('/explore/api/appeals-status');
-    cy.get('[href="/explore/api/appeals-status/sandbox-access').should('not.exist');
+    cy.visit('/explore/api/appeals');
+    cy.get('[href="/explore/api/appeals/sandbox-access').should('not.exist');
```
gileswells, line 37: I think I would rather see this left as `appeals-status` and then we update the legacy.json file to use the url that matches what's in production. Overall it's not necessary but if we're making a change, I'd rather we make it to be in line with what we have on prod.

src/components/pageContent/PageContent.tsx
```diff
@@ -49,6 +70,60 @@ const PageContent = (): JSX.Element => {
...
+    // Debounce function to help limit the firing of another function
+    const debounce = (func: () => void, delay: number): (() => void) => {
+      let inDebounce: ReturnType<typeof setTimeout> | null;
+      return (): void => {
+        if (inDebounce !== null) {
+          clearTimeout(inDebounce);
+        }
+        inDebounce = setTimeout(func, delay);
+      };
+    };
```
gileswells, line 83: So this debounce is waiting 100ms after the event finishes to run just one time. I feel like the experience shown to the user would be better if it was setup to run at most every 100ms instead of wait at least 100ms after the scrolling event stops firing.

gileswells (changes requested):

cypress/e2e/containers/documentation/DocumentationRoot.cy.js
jw81: Updated [here](https://github.com/department-of-veterans-affairs/developer-portal/pull/1425/commits/47b88160ff642285e6184c380efce1a9c231f4a3).

src/components/pageContent/PageContent.tsx
jw81: Switched over to a "throttle" approach [here](https://github.com/department-of-veterans-affairs/developer-portal/pull/1425/commits/3f3107c9c37b6ab968881d0d570cb0cd6f29d5b8).  LMK if it's not quite what you were thinking.

src/components/pageContent/PageContent.tsx
```diff
@@ -49,6 +70,70 @@ const PageContent = (): JSX.Element => {
...
+          }, limit - (Date.now() - lastRan));
+        }
+      };
+    };
+
+    // Adjust vertical positioning of Touchpoint survey button
+    const adjustSurveyPosition = (): void => {
+      const touchpointButton = document.querySelector('#touchpoints-survey');
+      const touchpointElement = touchpointButton instanceof HTMLElement ? touchpointButton : null;
+      const footer = document.querySelector('footer');
```
gileswells, line 99: I think that these can be moved outside of this function. In general, dom queries are more expensive in terms of performance and since these elements don't actually change we shouldn't need to query to find them again each time we adjust the button position.

gileswells (changes requested):
I like the change to throttle(), just concerned about best practice on the dom queries.

src/components/pageContent/PageContent.tsx
jw81: Ah, that's a really good point!  Only need to find them once.
I'll update this.

jw81: Updated that code [here](https://github.com/department-of-veterans-affairs/developer-portal/pull/1425/commits/0a92c683e4b432ed25452bd7cc2024ce704ff09c).

jw81 (commented):
Added a quick note about just removing the logo from Touchpoints vs adding CSS to programmatically hide the logo.

src/components/pageContent/PageContent.tsx
jw81: All that custom Javascript has been removed since UX has decided to go with a sticky experience for the button.

cypress/e2e/containers/documentation/DocumentationRoot.cy.js
jw81: The custom JS was removed since UX decided to go with a more "static" experience.

src/components/pageContent/PageContent.tsx
jw81: The custom JS was removed since UX decided to go with a more "static" experience.

andresriveratoro (approved):
LGTM

gileswells (approved):

PR #1427, opened 2023-12-19
[API-31641] - Test users deeplink page

gileswells:
### Description
https://jira.devops.va.gov/browse/API-31641

Create the test users deeplink page template and populate it with data.

To trick the dev-portal into allowing access to the test users page you must run this command in your browser's [DevTools](https://developer.chrome.com/docs/devtools/open) in the console tab. Do this on the homepage of the dev portal before navigating to any other pages of the site. This will only need to be done once and it will persist for all future visits to PR builds of the dev portal on that browser.

```
localStorage.setItem('persist:root', "{\"id\":\"1\",\"testUserHash\":\"\\\"just-a-testing-hash\\\"\",\"_persist\":\"{\\\"version\\\":-1,\\\"rehydrated\\\":true}\"}");window.location.reload();
```


gileswells:
Blocked by #1429

src/services/lpb.ts
```diff
@@ -0,0 +1,53 @@
...
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          method: 'POST',
+          url: LPB_TEST_USER_ACCESS_PATH,
+        };
+        if (isPrReviewBuild()) {
+          delete requestObject.body;
+          requestObject.method = 'GET';
+        }
```
gileswells, line 45: I am not a fan of this but because of how PR builds work this is a necessary hack to allow UX and others to see the page before it is live.

andresriveratoro (approved):

PR #1429, opened 2023-12-28
[API-32467] - Migrate to @reduxjs/toolkit

gileswells:
### Description

https://jira.devops.va.gov/browse/API-32467

Update Redux to a more modern implementation through the @reduxjs/toolkit package.


src/App.tsx
```diff
@@ -1,20 +1,21 @@
 import * as React from 'react';
 import classNames from 'classnames';
 import { ScrollRestoration } from 'react-router-dom';
-import { useDispatch, connect } from 'react-redux';
+import { connect } from 'react-redux';
```
andresriveratoro, line 4: I believe we can replace `connect` and `mapStateToProps` with hooks where needed as suggested [here](https://redux.js.org/usage/migrating-to-modern-redux#migrating-connect-to-hooks).

andresriveratoro (approved):
LGTM! Awesome work modernizing our Redux! I left a comment about something minor we can remove before merging.

src/App.tsx
gileswells: I can actually rip out the whole `connect()` and `mapStateToProps()` entirely as this component isn't directly accessing the state at all. The saving of the api list is now handled in the api service file so this component doesn't need to connect at all.

Because it's such a small change and with limited devs available to review today, so that I can get to the work depending on this PR I'm going to merge this as is and make this change in the follow up PR.

PR #1430, opened 2024-01-03
[API-32575] - Update facilities API deactivation date

gileswells:
### Description

Update Facilities API deactivation date in the page banner.


jw81 (approved):
LGTM

PR #1434, opened 2024-01-16
[API-31641] - Fix missing csrf header

gileswells:
### Description

https://jira.devops.va.gov/browse/API-31641

Missing the header in the request causing 403 errors.


jw81 (approved):
LGTM

PR #1435, opened 2024-01-19
[API-27141] - Add address fields to production access request form

andresriveratoro:
### Description

https://jira.devops.va.gov/browse/API-27141


src/containers/consumerOnboarding/components/productionAccessForm/BasicInformation.tsx
```diff
@@ -87,6 +91,43 @@ const BasicInformation: FC = () => {
           required
           className="vads-u-margin-top--4 medium-screen:vads-l-col--10"
         />
+        <va-select label="Country" name="country" value="USA" required>
+          <option value="USA">United States</option>
+        </va-select>
```
gileswells, line 96: The concept of this field seems weird. Like why have the field at all if you've only got 1 option to select from.

gileswells (approved):
LGTM. Just weird to have a single selection dropdown when there really isn't a choice to be made. 🤷‍♂️

PR #1442, opened 2024-02-07
[API-25709] Disallow emails with IP in domain

andresriveratoro:
### Description

https://jira.devops.va.gov/browse/API-25709


gileswells (dismissed):
LGTM

PR #1443, opened 2024-02-09
fix: use newer cypress image

andresriveratoro:
### Description

This fixes the GPG errors currently occurring in our GitHub actions that use the cypress/included image.


gileswells (approved):
LGTM

PR #1452, opened 2024-03-04
[API-34529] - Add GHA for Detecting API oAuth Property Changes

jw81:
### Description
Original Issue :: [API-34529](https://jira.devops.va.gov/browse/API-34529)
- Adds a GitHub Action (GHA) workflow to routinely check and see if the oAuth values have changed for the Address Validation API.
- Right now, this script just checks on Address Validation, but we can iterate on this script over time to include more/all APIs if necessary.
- Once this is in `Master`, then I'll come back and modify the GHA workflow so that it sends alerts to the `#team-okapi-alerts` Slack channel when necessary.


gileswells (approved):
A good start. Would like to see us expand this into a more robust test set in the future.

jw81:
> A good start. Would like to see us expand this into a more robust test set in the future.

Agreed.

PR #1455, opened 2024-03-06
[API-34671] - Conditionally disable test users page

gileswells:
### Description

https://jira.devops.va.gov/browse/API-34671

Specifically requested for the Clinical Health API but built to be configurable, this will allow us to add `disableTestUsersPage: true` to the `acgInfo` property within LPB to disable an API's Test Users page when they have ACG OAuth available.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8669,9 +8669,9 @@
       }
     },
     "node_modules/caniuse-lite": {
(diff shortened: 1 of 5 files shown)


jw81 (approved):
LGTM 👍

PR #1456, opened 2024-03-07
[API-34711] - Update Auth Doc CCG Pages Text

jw81:
### Description
Original Issue :: [API-34711](https://jira.devops.va.gov/browse/API-34711)
- changes the `expires_in` value from `900` to `300` as requested on CCG docs
- removes the trailing `\` from the "Retrieving an access token" curl examples as requested on CCG docs
  - takes into account that API's that use the `launchScope` will have an extra line in the curl example.


gileswells (approved):
LGTM but likely will need new visual regressions.

Edit: or not... Nice.

PR #1478, opened 2024-03-29
[API-29738] Fix top level swagger accordion accessible names

andresriveratoro:
### Description
https://jira.devops.va.gov/browse/API-29738


gileswells:
I have had LOADS of problems with turning on deeplinking within Swagger in the past. This is going to need a LOT of testing to make sure that all is well.

PR #1482, opened 2024-04-02
[API-35262] - Fix release titles and mr creation

gileswells:
### Description

https://jira.devops.va.gov/browse/API-35262

This resolves a bad regex that caused extra API calls that would sometimes get rate limited causing down the line errors. It also is more fault tolerant in cases where the extra API call might still be necessary so that even if it does get rate limited it will still function properly.


jw81 (approved):
LGTM

PR #1490, opened 2024-04-08
[API-35191] Request production access page text and style updates

andresriveratoro:
### Description
https://jira.devops.va.gov/browse/API-35191


src/styles/base.scss
```diff
@@ -165,3 +165,22 @@ va-omb-info,
 .dev-portal-omb-info {
   margin-top: 32px;
 }
+
+a.vads-c-action-link--green {
```
gileswells, line 169: So long as this broadly scoped CSS class doesn't cause issues with our other current usage of this class on the ApiOverviewPage component, I have no issues.

gileswells (approved):
Just one comment to acknowledge before merging but I suspect it's good to go as is.

PR #1493, opened 2024-04-11
[API-35183] Replace breadcrumbs with vads-breadcrumbs

andresriveratoro:
### Description
https://jira.devops.va.gov/browse/API-35183


gileswells (approved):
LGTM

The only thing I've spotted is we're missing the breadcrumb on the Test Users page. I've made a bug ticket for that as it's unrelated to the change over to the USWDS component.

PR #1495, opened 2024-04-15
[API-35184] Replace gov banner with vads

andresriveratoro:
### Description
https://jira.devops.va.gov/browse/API-35184


gileswells (approved):
LGTM

PR #1498, opened 2024-04-18
[API-35427] - update restricted access handling

gileswells:
### Description
https://jira.devops.va.gov/browse/API-35428
also
https://jira.devops.va.gov/browse/API-35273

This PR handles the updates to respond to how LPB now presents the restricted access data through the legacy.json file. That change just so happens to also complete the requirements of API-35273.


scripts/fake_lpb.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 cd ./public/data/
+cp ../../cypress/fixtures/legacy.json ./platform-backend/v0/providers/transformations/legacy.json
```
gileswells, line 4: The goal here, while not a ticket AC, is to remove the duplication of this file. This change, along with the .gitkeep to maintain the folder and the new .gitignore, allows us to just update the singular legacy.json file in the fixtures folder to support both Cypress tests and also PR builds.

PR #1506, opened 2024-04-30
[API-35351] - Support apikey and oauth on the same API

gileswells:
### Description
https://jira.devops.va.gov/browse/API-35351

This PR allows for a single API to have both oAuth and apikey authentication methods. This carries down to the sandbox access form and implements the restricted access field for blocking the sandbox access form on a per auth type basis.


andresriveratoro (approved):
LGTM

PR #1511, opened 2024-05-16
API-36422 Replace placeholder with real ACG well-known config URI

caseywilliams:
### Description

[API-36422](https://jira.devops.va.gov/browse/API-36422): This ticket is to update the ACG auth page to replace the existing well-known OAuth config URI, which is a placeholder, with the actual well-known config URI for the current API.

Please note that the ticket mentions both ACG and CCG, but only the ACG page includes the placeholder URI, so only the ACG page has been changed here (so far, anyway - I'm waiting for confirmation that the CCG page requires no updates).

### Requested feedback

I could see at least two ways to do this - I'll leave comments with details.


src/components/oauthDocs/acg/IdToken.tsx
```diff
@@ -1,58 +1,61 @@
...
-        <code>exp</code> - the expiration time of the JWT. The token cannot be accepted by VA
-        Developer after this time, and your application should not use an expired token to identify
-        a user.
-      </li>
-    </ul>
-  </>
-);
+const IdToken = ({ api }: ApiRequiredProps): JSX.Element => {
+  const baseAuthPath = api.oAuthInfo?.acgInfo?.baseAuthPath ?? '/oauth2/{api}/v1';
+  const wellKnownConfigUri = `https://sandbox-api.va.gov${baseAuthPath}/.well-known/openid-configuration`;
```
caseywilliams, line 7: Initially I had hoped to just read this URI from the `api` object, and it did look like it had some relevant information - here's an example of an `api` object I was seeing locally:
```jsonc
{
    // ...other content omitted for brevity:
    "apikey": false,
    "oAuth": true,
    "oAuthInfo": {
        "acgInfo": {
            "baseAuthPath": "/oauth2/veteran-verification/v1",
            "gitHubTestUsers": "https://github.com/department-of-veterans-affairs/vets-api-clients/blob/master/test_accounts/verification_test_accounts.md",
            "productionAud": "aus7y0sefudDrg2HI2p7",
            "productionWellKnownConfig": "https://dev-api.va.gov/oauth2/veteran-verification/v1/.well-known/openid-configuration",
            "sandboxAud": "aus7y0sefudDrg2HI2p7",
            "sandboxWellKnownConfig": "https://dev-api.va.gov/oauth2/veteran-verification/v1/.well-known/openid-configuration",
            "scopes": [
                "profile",
                "openid",
                "offline_access",
                "service_history.read",
                "disability_rating.read",
                "veteran_status.read"
            ],
            "lighthouseAuthServiceCategory": "veteran-verification"
        }
    },
    "oAuthTypes": [
        "AuthorizationCodeGrant"
    ],
    // ...etc.
}
```
My first thought was to just use the `api.oauthInfo.acgInfo.sandboxWellKnownConfig` to get this URI, but I didn't go with that approach for a few reasons:
1. Both the `sandboxWellKnownConfig` and the `productionWellKnownConfig` include `dev-api.va.gov`, which seemed wrong. This may just be  quirk of the seeded data though? I wasn't sure.
2. Neither the `sandboxWellKnownConfig` nor the `productionWellKnownConfig` properties are documented as part of [the ACGInfo schema in schema.ts](https://github.com/department-of-veterans-affairs/developer-portal/blob/e7e8269a940f06c38da7379f870c2562849e7ded/src/apiDefs/schema.ts#L102-L109).
3. I noticed that [this component, `AuthCodeGrantContent`, which is used on this same page](https://github.com/department-of-veterans-affairs/developer-portal/blob/e7e8269a940f06c38da7379f870c2562849e7ded/src/components/oauthDocs/acg/AuthCodeGrantContent.tsx#L9), already builds a sandbox URI based on the `api.oAuthInfo?.acgInfo?.baseAuthPath` value - I ultimately copied that approach.

Any thoughts/requests for changes?

